PR #17, opened 2022-08-17
Header in isolation validation

someone235:


consensus/core/src/header.rs
```diff
@@ -18,11 +19,6 @@ impl Header {
             hasher.write(parent);
         }
         hasher.update(nonce.to_le_bytes());
```
michaelsutton, line 21: add `time_in_ms` to hasher (in the same order as done in the go ref)

consensus/src/pipeline/header_processor/validate_in_isolation.rs
```diff
@@ -0,0 +1,53 @@
...
+    time::{SystemTime, UNIX_EPOCH},
+};
+
+impl HeaderProcessor {
+    pub(super) fn validate_header_in_isolation(self: &Arc<HeaderProcessor>, header: &Header) -> ConsensusResult<()> {
+        if header.hash == self.genesis_hash {
+            return Ok(());
+        }
+
+        self.check_header_version(header)?;
```
michaelsutton, line 16: I was expecting calls to `check_block_timestamp_in_isolation` and `check_parents_limit` here

PR #26, opened 2022-08-28
Concurrent processing pipeline -- Part 2 

michaelsutton:
Completes the processing pipeline structure with a block body processor and a virtual state processor. Does not implement the actual processing, but is aimed at getting closer to the final consensus API.


consensus/src/pipeline/block_processor/processor.rs
```diff
@@ -0,0 +1,102 @@
...
+            db,
+            statuses_store,
+            reachability_service,
+            task_manager: BlockTaskDependencyManager::new(),
+        }
+    }
+
+    pub fn worker(self: &Arc<BlockBodyProcessor>) {
+        loop {
+            select! {
```
someone235, line 51: Not sure why do we need select for one channel

consensus/src/pipeline/block_processor/processor.rs
```diff
@@ -0,0 +1,102 @@
...
+        // Wait until all workers are idle before exiting
+        self.task_manager.wait_for_idle();
+
+        // Pass the exit signal on to the following processor
+        self.sender.send(BlockTask::Exit).unwrap();
+    }
+
+    fn queue_block(self: &Arc<BlockBodyProcessor>, hash: Hash) {
+        if let Some(block) = self.task_manager.try_begin(hash) {
+            let res = self.process_block_body(&block);
```
someone235, line 84: You're doing nothing with it?

consensus/src/pipeline/block_processor/processor.rs
```diff
@@ -0,0 +1,102 @@
...
+
+            for dep in dependent_tasks {
+                let processor = self.clone();
+                rayon::spawn(move || processor.queue_block(dep));
+            }
+        }
+    }
+
+    fn process_block_body(self: &Arc<BlockBodyProcessor>, block: &Block) -> BlockProcessResult<()> {
+        Ok(())
```
someone235, line 106: Add TODO

someone235 (changes requested):

consensus/src/consensus/mod.rs
```diff
@@ -150,24 +176,29 @@ impl Consensus {
...
-        // Spawn the asynchronous header processor.
+        // Spawn the asynchronous processors.
         let header_processor = self.header_processor.clone();
-        thread::spawn(move || header_processor.worker())
-
-        // TODO: add block body processor and virtual state processor workers and return a vec of join handles.
+        let body_processor = self.body_processor.clone();
+        let virtual_processor = self.virtual_processor.clone();
+
+        vec![
```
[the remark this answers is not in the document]
michaelsutton: For kaspad flow we use consensus as a Service imp which returns a vec of handles

someone235 (approved):

PR #29, opened 2022-08-31
Added address struct + bech32 encoder/decoder

tmrlvi:
Added `addresses` crate with `Address` struct and bech32 encoder / decoder
Usage:
```rust
// Decoding
let address_str: String = "kaspa:qqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqkx9awp4e".to_string();
let address: Address = address_str.try_into()?;

// Encoding
let address = Address { prefix: Prefix::Mainnet, payload: vec![0u8; 32], version: 0u8 };
let address_str: String = address.into();
```


crypto/addresses/src/bech32.rs
```diff
@@ -0,0 +1,195 @@
...
+    100, 100, 15, 100, 10, 17, 21, 20, 26, 30, 7, 5, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100,
+    100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100,
+    100, 100, 100, 29, 100, 24, 13, 25, 9, 8, 23, 100, 18, 22, 31, 27, 19, 100, 1, 0, 3, 16, 11, 28, 12, 14, 6, 4, 2,
+];
+
+// Checksome for bech32
+// https://bch.info/en/specifications
+fn polymod<'data, I>(values: I) -> u64
+where
+    I: Iterator<Item = &'data u8>,
```
elichai, line 16: Can you make this `Iterator<Item=u8>`? a reference is bigger than a u8(and it is a copy type) so there's no need to pass references

crypto/addresses/src/bech32.rs
```diff
@@ -0,0 +1,195 @@
...
+    c ^ 1
+}
+
+fn checksum(payload: &[u8], prefix: &[u8]) -> u64 {
+    polymod(
+        prefix
+            .iter()
+            .chain(&[0u8])
+            .chain(payload)
+            .chain(&vec![0u8, 0u8, 0u8, 0u8, 0u8, 0u8, 0u8, 0u8]),
```
elichai, line 48: You can remove the `vec!` and just pass an array of zeros

elichai: just `[0u8; 8]`

crypto/addresses/src/bech32.rs
```diff
@@ -0,0 +1,195 @@
...
+    )
+}
+
+// Convert 8bit array to 5bit array with right padding
+fn conv8to5(payload: &[u8]) -> Vec<u8> {
```
elichai: Why is this much more complicated than this? https://github.com/kaspanet/kaspad/blob/54fbfadf84ab43cd4eff7df131ebd18b50bdd882/util/bech32/bech32.go#L260
do they do the same thing or something different?

If it's the same it can also return an Iterator instead of collecting to a vec

elichai: Or is it actually https://github.com/kaspanet/kaspad/blob/54fbfadf84ab43cd4eff7df131ebd18b50bdd882/util/bech32/bech32.go#L166 ?

crypto/addresses/src/bech32.rs
```diff
@@ -0,0 +1,161 @@
...
+        prefix
+            .iter()
+            .chain(&[0u8])
+            .chain(payload)
+            .chain(&vec![0u8, 0u8, 0u8, 0u8, 0u8, 0u8, 0u8, 0u8]),
+    )
+}
+
+// Convert 8bit array to 5bit array with right padding
+fn conv8to5 (payload: &[u8]) -> Vec<u8> {
```
elichai, line 53: Is this the equivalent of https://github.com/kaspanet/kaspad/blob/54fbfadf84ab43cd4eff7df131ebd18b50bdd882/util/bech32/bech32.go#L260 or of https://github.com/kaspanet/kaspad/blob/54fbfadf84ab43cd4eff7df131ebd18b50bdd882/util/bech32/bech32.go#L166?

crypto/addresses/src/bech32.rs
tmrlvi: If i need to pass it as cloned inter, should I use `[0u8; 8].iter().cloned()`? It seems that just passing `[0u8; 8]` translate to `Iterator<Item=&u8>`

crypto/addresses/src/bech32.rs
tmrlvi: It is the second one. The prefix is done by
```
     let fivebit_prefix: Vec<u8> = self
            .prefix
            .to_string()
            .as_bytes()
            .iter()
            .map(|c| c & 0x1fu8)
            .collect();
```

crypto/addresses/src/bech32.rs
tmrlvi: It is the second one.

crypto/addresses/src/bech32.rs
elichai: you should add `.copied()` after the first `iter()` and then `[0u8; 8]` should work as is:
https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=3b92190d8ae19cce7b56140e2806f0d4

crypto/addresses/src/bech32.rs
tmrlvi: Done

crypto/addresses/src/bech32.rs
tmrlvi: done

crypto/addresses/src/bech32.rs
```diff
@@ -0,0 +1,161 @@
...
+            current_idx += 1;
+        }
+    }
+    eight_bit
+}
+
+impl Address {
+    pub(crate) fn encode_payload(&self) -> String {
+        // Convert into 5 bits vector
+        let fivebit_payload = conv8to5(&[vec![self.version], self.payload.clone()].concat());
```
elichai, line 102: This also doesn't require vec

crypto/addresses/src/bech32.rs
```diff
@@ -0,0 +1,161 @@
...
+        let fivebit_prefix: Vec<u8> = self
+            .prefix
+            .to_string()
+            .as_bytes()
+            .iter()
+            .map(|c| c & 0x1fu8)
+            .collect();
+
+        let checksum = checksum(fivebit_payload.as_slice(), fivebit_prefix.as_slice());
```
elichai, line 111: If this would accept `fivebit_prefix` as an iterator than we wouldn't need to collect it above

tmrlvi: I have a problem with ownership in this case. Without the collect, there is no reference kept for `prefix.to_string`. 
I can solve it by creating an extra variable `prefix_str`. Is there a better way?

elichai: yeah, you need an extra variable

elichai: Also, maybe add `as_str()` to `Prefix` that returns a `&' static str`, that will solve that and also save yet another allocation :)

crypto/addresses/src/bech32.rs
tmrlvi: Should I use `as_slice` instead? to make them both `&[u8]`?

elichai: Yeah

crypto/addresses/src/bech32.rs
tmrlvi: done (with `as_str` we don't need extra variable :) )

crypto/addresses/src/bech32.rs
tmrlvi: done

crypto/addresses/src/bech32.rs
```diff
@@ -0,0 +1,163 @@
...
+            .copied()
+            .map(|c| c & 0x1fu8);
+
+        let checksum = checksum(fivebit_payload.as_slice(), fivebit_prefix);
+
+        String::from_utf8(
+            [fivebit_payload, conv8to5(&checksum.to_be_bytes()[3..])]
+                .concat()
+                .iter()
+                .map(|c| CHARSET[*c as usize])
```
elichai, line 119: you can do `char::from(CHARSET[*c as usize])` and then collect directly into a String, without going through a Vec and doing utf8 checks

tmrlvi: for some reason, using char is slightly slower (~10% slower)

crypto/addresses/src/lib.rs
```diff
@@ -0,0 +1,144 @@
...
+    fn try_from(value: String) -> Result<Self, Self::Error> {
+        match value.split_once(':') {
+            Some((prefix, payload)) => Self::decode_payload(prefix.try_into()?, payload),
+            None => Err(AddressError::MissingPrefix),
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
elichai, line 89: Can you please add all the tests from here: https://github.com/kaspanet/kaspad/blob/54fbfadf84ab43cd4eff7df131ebd18b50bdd882/util/bech32/bech32_test.go

tmrlvi: I've added the test, but to keep then intact, I had to leave test-only prefixes in the Prefix enum:
```
#[derive(PartialEq, Eq, Clone, Debug)]
pub enum Prefix {
    Mainnet,
    Testnet,
    Devnet,
    #[cfg(test)]
    A,
    #[cfg(test)]
    B,
}
```
The alternative would be to recalculate the checksum, but the point is to check my calculations

Tell me if you want me to change the prefixes

elichai (approved):

PR #35, opened 2022-09-05
Process in concurrent batches

michaelsutton:


someone235 (approved):

consensus/src/consensus/test_consensus.rs
```diff
@@ -130,6 +131,24 @@ impl TestConsensus {
...
         &self.consensus.reachability_store
     }
+
+    pub fn processing_counters(&self) -> &Arc<ProcessingCounters> {
+        &self.consensus.counters
+    }
+}
+
+impl Service for TestConsensus {
+    fn ident(self: Arc<TestConsensus>) -> String {
```
elichai, line 141: Can we make ident `&'static str`

PR #39, opened 2022-09-11
Add Uint implementation

elichai:
This adds a macro for generating Uints that are multiple of u64,
it contains fuzzing code with corpuses at https://github.com/kaspanet/rusty-kaspa-corpus

Currently it uses U320 when doing the DAA calculation because the window is 2641, so if we have lowest difficulty this is 2^256*2641 which is ~268 bits, so this leaves 2^52 timestamp difference before the multiplication could theoretically overflow


consensus/src/processes/difficulty.rs
```diff
@@ -81,50 +81,54 @@ impl<T: HeaderStoreReader> DifficultyManager<T> {
...
+        // We need Uint320 to avoid overflow when summing and multiplying by the window size.
+        // TODO: Try to see if we can use U256 instead, by modifying the algorithm.
         let difficulty_blocks_len = difficulty_blocks.len();
-        let targets_sum: BigInt = difficulty_blocks
+        let targets_sum: Uint320 = difficulty_blocks
             .into_iter()
-            .map(|diff_block| big_from_compact_target(diff_block.bits))
+            .map(|diff_block| Uint320::from(u256_from_compact_target(diff_block.bits)))
             .sum();
         let average_target = targets_sum / (difficulty_blocks_len as u64);
```
tiram88, line 91: The overflow of U256 can be avoided by relying on following general idea (pseudo-code, by no means valid rust code):
```
min_diff_block = min(difficulty_blocks)
max_diff_block = max(difficulty_blocks)
offsets_sum = sum(diff_block - min_diff_block)
average_target = min_diff_block + offsets_sum / difficulty_blocks_len
```
This however requires that following condition be true:
`max_diff_block - min_diff_block < 2^256 / difficulty_blocks_len`

Otherwise, an alternate execution path should rely on the current U320 implementation.

Note that the `offsets_sum` approach does floor exactly the same way as the `targets_sum` does.

AVG floor demonstration

michaelsutton: Seems like we will defer this optimization to a following PR

PR #44, opened 2022-09-17
Make muhash::finalize/serialize 95% faster

elichai:
This PR implements the Extended Euclidean Algorithm for the generic Uint, creates `Uint3072`, and uses that to invert field elements in the muhash field (instead of the current Fermat's little theorem squaring ladder).

While doing that the fuzzer found a bug in the right/left shift behavior of non powers of 2 (like `U320`, `U192`, or `U3720`) so this also fixes that and adds a `Uint192` fuzzer to specifically fuzz non powers of 2 (Uint192 is much faster than Uint3072 so it works better for fuzzing).


elichai:
Rebased on top of https://github.com/kaspanet/rusty-kaspa/pull/45, so that one should be merged first

Issue #60, opened 2022-10-19
Virtual processor - Part 2


someone235:
1. Change Block to have `Arc<Header>` and pass block around w/o an arc
  2. Switch `BlueWorkType` to Uint192
  3. Consider the DAA window average optimization by Tiram 
  4. Remove fee param from Tx output 
  5. Remove many Arcs from within Tx structs 
  6. Use BlockHashMap/BlockHashSet type defs where ever possible 
  7. Add prefix support to logs trace macro and perhaps add debug/info macros
  8. Consider switching from saving the mergeset DAA blocks to saving the complementary set, i.e. the mergeset non-DAA blocks. On average this means moving from saving O(mergeset) to 0
  9. Apply DbWriter as a generalization of Batch writing and use it to cleanup remaining caching.rs code duplications
  10. Add genesis block to config
  11. Take cache sizes from kaspa-go + give it more thought
  12. Rename rust files with a single convention

michaelsutton:
Pending todos for completing the virtual processor implementation: 

- pick virtual parents from body tips according to pruning rules
- check finality violation
- handle disqualified chain loop
- call pruning point move
- implement coinbase validation
- acceptance data format
- virtual utxo-set ownership semantics 
- basic build block template support (only header and coinbase, w/o mempool txs)

PR #62, opened 2022-10-19
Refactor block struct

michaelsutton:
Created a `MutableBlock` struct which test consensus returns from build methods, and changed `Block` to have header through an Arc, making it immutable and easy to pass by cloning (w/o another Arc around). This also saves from a full header clone on header processor commit


someone235 (approved):

someone235 (approved):

michaelsutton:
Addressed some of the items in https://github.com/kaspanet/rusty-kaspa/issues/60

PR #63, opened 2022-10-20
Call maybe_update_pruning_point_and_candidate and let it handle moving two pruning points at once

someone235:


consensus/src/pipeline/virtual_processor/processor.rs
```diff
@@ -304,20 +308,41 @@ impl VirtualStateProcessor {
...
     fn maybe_update_pruning_point_and_candidate(self: &Arc<Self>) {
-        // TODO: call this method from resolve_virtual
+        let virtual_sp = self.virtual_state_store.read().get().unwrap().ghostdag_data.selected_parent;
+        if virtual_sp == self.genesis_hash {
+            return;
+        }
+
+        let ghostdag_data = self.ghostdag_store.get_compact_data(virtual_sp).unwrap();
         let pruning_read_guard = self.pruning_store.upgradable_read();
         let current_pp = pruning_read_guard.pruning_point().unwrap();
```
michaelsutton, line 318: Change to `pruning_read_guard.get()` to obtain the full `PruningPointInfo` and use index and candidate below (in both if/else branches) instead of re-accessing the cache

consensus/src/pipeline/virtual_processor/processor.rs
```diff
@@ -304,20 +308,41 @@ impl VirtualStateProcessor {
...
+        let current_pp_bs = self.ghostdag_store.get_blue_score(current_pp).unwrap();
         let current_pp_candidate = pruning_read_guard.pruning_point_candidate().unwrap();
-        let virtual_gd = self.virtual_state_store.read().get().unwrap().ghostdag_data.to_compact();
         let (new_pruning_point, new_candidate) =
-            self.pruning_manager.next_pruning_point_and_candidate_by_block_hash(virtual_gd, None, current_pp_candidate, current_pp);
+            self.pruning_manager.next_pruning_point_and_candidate_by_block_hash(ghostdag_data, None, current_pp_candidate, current_pp);
 
         if new_pruning_point != current_pp {
+            let mut past_pruning_points_to_add = Vec::new();
+            for current in self.reachability_service.backward_chain_iterator(virtual_sp, current_pp, false) {
```
michaelsutton, line 326: Why not iterate forward from the prev pp to the new one and check for the pruning window condition? Perhaps we can change to this later. It makes more sense because that exact loop is needed anyhow for building the new pp utxo set

consensus/src/pipeline/virtual_processor/processor.rs
```diff
@@ -304,20 +308,41 @@ impl VirtualStateProcessor {
...
+                let current_header = self.headers_store.get_compact_header_data(current).unwrap();
+                if current_header.pruning_point == current_pp || current_header.blue_score < current_pp_bs + self.pruning_depth {
+                    break;
+                }
+
+                if past_pruning_points_to_add.is_empty() || *past_pruning_points_to_add.last().unwrap() != current_header.pruning_point
```
michaelsutton, line 332: Maybe something like `if !past_pruning_points_to_add.last().is_some_and(|h| **h == current_header.pruning_point) {` is more readable (`is_some_and` is from nightly but we can copy it as a util)

consensus/src/pipeline/virtual_processor/processor.rs
```diff
@@ -304,20 +308,41 @@ impl VirtualStateProcessor {
...
             let mut batch = WriteBatch::default();
-            let new_pp_index = pruning_read_guard.pruning_point_index().unwrap() + 1;
             let mut write_guard = RwLockUpgradableReadGuard::upgrade(pruning_read_guard);
+            for (i, past_pp) in past_pruning_points_to_add.iter().copied().rev().enumerate() {
+                self.past_pruning_points_store.insert_batch(&mut batch, current_pp_index + i as u64 + 1, past_pp).unwrap();
+            }
+            let new_pp_index = current_pp_index + past_pruning_points_to_add.len() as u64;
             write_guard.set_batch(&mut batch, new_pruning_point, new_candidate, new_pp_index).unwrap();
-            self.past_pruning_points_store.insert_batch(&mut batch, new_pp_index, new_pruning_point).unwrap();
             self.db.write(batch).unwrap();
```
michaelsutton, line 346: General comment: ideally we extract the past pruning point list from the devnet refs and put them into the jsons and compare them at the end of the json tests

consensus/src/pipeline/virtual_processor/processor.rs
```diff
@@ -290,6 +292,8 @@ impl VirtualStateProcessor {
             }
             _ => panic!("expected utxo valid or disqualified {}", new_selected),
         }
+
+        self.maybe_update_pruning_point_and_candidate()
```
michaelsutton, line 296: Add a TODO to turn this call into a channel msg send to a "pruning processor"

consensus/src/pipeline/virtual_processor/processor.rs
michaelsutton: Or have `next_pruning_point_and_candidate_by_block_hash` return a list

consensus/src/processes/parents_builder.rs
```diff
@@ -183,5 +183,311 @@ impl<T: HeaderStoreReader, U: ReachabilityStoreReader, V: RelationsStoreReader>
...
+    use super::ParentsManager;
+    use consensus_core::{
+        blockhash::{BlockHashes, ORIGIN},
+        header::Header,
+    };
+    use hashes::Hash;
+    use itertools::Itertools;
+    use parking_lot::RwLock;
+
+    struct HeaderStoreMock {
```
michaelsutton, line 213: Seems like a classic place for a mock lib. Should we add one? What do you think @elichai ? The downside of https://docs.rs/mockall/latest/mockall/ is that you need to add `#[automock]` annotation to the original trait.

consensus/src/processes/pruning.rs
```diff
@@ -137,13 +137,18 @@ impl<S: GhostdagStoreReader, T: ReachabilityStoreReader, U: HeaderStoreReader, V
...
             };
-            let (next_or_current_pp, _) = self.next_pruning_point_and_candidate_by_block_hash(
+            let (added_past_pruning_points, _) = self.next_pruning_points_and_candidate_by_ghostdag_data(
                 ghostdag_data,
                 suggested_low_hash,
                 current_candidate,
                 current_pruning_point,
             );
-            next_or_current_pp
+
```
michaelsutton, line 146: nit: 
`added_past_pruning_points.last().copied().unwrap_or(current_pruning_point)`

consensus/src/processes/pruning.rs
```diff
@@ -69,8 +69,8 @@ impl<S: GhostdagStoreReader, T: ReachabilityStoreReader, U: HeaderStoreReader, V
             None => current_candidate,
         };
 
-        let mut new_pruning_point = current_pruning_point;
-        let mut new_pruning_point_bs = self.ghostdag_store.get_blue_score(new_pruning_point).unwrap();
+        let mut added_past_pruning_points = Vec::with_capacity((self.pruning_depth / self.finality_depth) as usize); // If the pruning point is more out of date than that, an IBD with headers proof is needed anyway.
```
michaelsutton, line 72: Maybe put the comment in the line above since it becomes a bit long

michaelsutton, line 72: Maybe just `added_pruning_points` or `new_pruning_points`

consensus/src/processes/pruning.rs
```diff
@@ -137,13 +137,18 @@ impl<S: GhostdagStoreReader, T: ReachabilityStoreReader, U: HeaderStoreReader, V
                 }
                 Err(err) => panic!("Unexpected reachability error: {:?}", err),
             };
-            let (next_or_current_pp, _) = self.next_pruning_point_and_candidate_by_block_hash(
+            let (added_past_pruning_points, _) = self.next_pruning_points_and_candidate_by_ghostdag_data(
```
michaelsutton, line 140: Same

consensus/src/pipeline/virtual_processor/processor.rs
```diff
@@ -304,26 +309,35 @@ impl VirtualStateProcessor {
...
+        if virtual_sp == self.genesis_hash {
+            return;
+        }
 
-        if new_pruning_point != current_pp {
+        let ghostdag_data = self.ghostdag_store.get_compact_data(virtual_sp).unwrap();
+        let pruning_read_guard = self.pruning_store.upgradable_read();
+        let current_pruning_info = pruning_read_guard.get().unwrap();
+        let current_pp_bs = self.ghostdag_store.get_blue_score(current_pruning_info.pruning_point).unwrap();
+        let (past_pruning_points_to_add, new_candidate) = self.pruning_manager.next_pruning_points_and_candidate_by_ghostdag_data(
```
michaelsutton, line 321: Same (`new_pruning_points`)

michaelsutton (approved):

PR #70, opened 2022-10-24
Override Hasher on BlockHashMap and BlockHashSet

elichai:
The BlockHash contains the PoW nonce, which means that trying to DOS the hashmap costs you in PoW so the PoW function actually behaves as DOS resistancy for HashMaps where the BlockHash is the key.
meaning that we don't need to pass it through `SipHash` instead we can just take the first/last 64 bits of the hash and use those as the hash for the hashmap.

### DIFF
--- a/consensus/core/src/lib.rs
+++ b/consensus/core/src/lib.rs
@@ -1,4 +1,5 @@
 use std::collections::{HashMap, HashSet};
+use std::hash::{BuildHasher, Hasher};
 
 use hashes::Hash;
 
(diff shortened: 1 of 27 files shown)


consensus/core/src/lib.rs
```diff
@@ -14,14 +15,95 @@ pub mod utxo;
...
 /// Integer type for accumulated PoW of blue blocks. We expect no more than
 /// 2^128 work in a single block (btc has ~2^80), and no more than 2^64
-/// overall blocks, so 2^192 is definitely a justified upper-bound.  
+/// overall blocks, so 2^192 is definitely a justified upper-bound.
 pub type BlueWorkType = math::Uint192;
 
-/// Defined in order to make it easy to experiment with various hashers
-pub type BlockHashSet = HashSet<Hash>;
+/// This HashMap skips the hashing of the key and uses the key directly as the hash.
+/// Should only be used for block hashes that has correct DAA,
```
michaelsutton, line 22: has -> have

crypto/hashes/src/lib.rs
```diff
@@ -3,14 +3,15 @@ mod pow_hashers;
...
 use serde::{Deserialize, Serialize};
 use std::fmt::{Debug, Display, Formatter};
+use std::hash::{Hash as StdHash, Hasher as StdHasher};
 use std::str::{self, FromStr};
 
 pub const HASH_SIZE: usize = 32;
 
 pub use hashers::*;
 
 // TODO: Check if we use hash more as an array of u64 or of bytes and change the default accordingly
```
michaelsutton, line 13: Is this still a TODO?

crypto/hashes/src/lib.rs
```diff
@@ -56,6 +57,24 @@ impl Hash {
     }
 }
 
+// Override the default Hash implementation, to: A. improve perf a bit (siphash works over u64s), B. allow a hasher to just take the first u64.
+// Don't change this without looking at `consensus/core/src/blockhash/BlockHashMap`.
+impl StdHash for Hash {
+    #[inline(always)]
+    fn hash<H: StdHasher>(&self, state: &mut H) {
+        self.iter_le_u64().for_each(|x| x.hash(state));
```
michaelsutton, line 65: Reverse or change `from_u64_word` so that we hash by the word (for tests)

crypto/hashes/src/lib.rs
elichai: Yes, when the code is "complete" we should check which methods are used the most

michaelsutton (approved):

elichai (commented):
The changes LGTM

PR #72, opened 2022-10-27
Update emulator to use systemtime, and bug fix for `check_block_timestamp_in_isolation` 

D-Stacks:
Changes:
1) [Bug fix] - use `max_block_time` in check `check_block_timestamp_in_isolation`
2) [update emulator] -  use system time in `block.header.timestamp` and remove usage of the delay, in order to make emulator compatible with new validation logic

### DIFF
--- a/kaspad/src/emulator.rs
+++ b/kaspad/src/emulator.rs
@@ -13,7 +13,7 @@ use std::{
         Arc,
     },
     thread::{self, spawn, JoinHandle},
-    time::Duration,
+    time::{Duration, SystemTime},
 };
 
 /// Emits blocks randomly in the round-based model where number of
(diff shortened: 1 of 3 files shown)


kaspad/src/emulator.rs
```diff
@@ -51,14 +51,11 @@ impl RandomBlockEmitter {
 
         let mut tips = vec![self.genesis];
         let mut total = 0;
-        let mut timestamp = 0u64;
 
         while total < self.target_blocks {
             let v = min(self.max_block_parents, poi.sample(&mut thread_rng) as u64);
-            timestamp += (self.delay as u64) * 1000;
-            if v == 0 {
```
michaelsutton, line 59: We need to keep this since otherwise `new_tips` below becomes empty and the next round has no parents

D-Stacks: yes, i thought i already pushed this back in, i removed it cause i thought it was some timestamp check due to placement of the code.

D-Stacks: okay it is back in.

michaelsutton (approved):

PR #76, opened 2022-11-08
Coinbase validation

michaelsutton:
Implements coinbase validation (which is part of UTXO validation for chain-block candidates).

Some changes/design decisions: 
1. Extract merged block miner data externally (while processing the mergeset)
2. Store the non-DAA mergeset (`mergeset \ DAA window`) instead of the DAA mergeset (`mergeset ∩ DAA window`). This still fits the usage of checking if a merged block is in the DAA window, but is expected to be empty most of the time.


consensus/src/processes/coinbase.rs
```diff
@@ -64,62 +105,121 @@ impl CoinbaseManager {
...
+        let script_pub_key_len = miner_data.script_public_key.script().len();
+        if script_pub_key_len > self.coinbase_payload_script_public_key_max_len as usize {
+            return Err(CoinbaseError::PayloadScriptPublicKeyLenAboveMax(
+                script_pub_key_len,
+                self.coinbase_payload_script_public_key_max_len,
+            ));
+        }
+
+        payload.truncate(LENGTH_OF_BLUE_SCORE + LENGTH_OF_SUBSIDY); // Keep only blue score and subsidy
+        payload.extend(
```
someone235, line 178: nit: In go we have an optimization where we create a new vector only if the length was changed (which almost never happens)

consensus/src/processes/coinbase.rs
```diff
@@ -649,4 +341,69 @@ mod tests {
             assert_eq!(cbm.calc_block_subsidy(t.daa_score), t.expected, "test '{}' failed", t.name);
         }
     }
+
+    #[test]
+    fn payload_serialization_test() {
+        let params = &MAINNET_PARAMS;
```
someone235, line 347: Can you also add a hardcoded hex string to check?

someone235 (changes requested):

consensus/src/pipeline/virtual_processor/utxo_validation.rs
```diff
@@ -135,13 +148,30 @@ impl VirtualStateProcessor {
...
+            .expected_coinbase_transaction(daa_score, miner_data, ghostdag_data, mergeset_rewards, mergeset_non_daa)
+            .unwrap()
+            .tx;
+        trace!(
+            "mergeset: {} blues, {} reds, {} non-DAA",
+            ghostdag_data.mergeset_blues.len(),
+            ghostdag_data.mergeset_reds.len(),
+            mergeset_non_daa.len()
+        );
+        if hashing::tx::hash(coinbase) != hashing::tx::hash(&expected_coinbase) {
```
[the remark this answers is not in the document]
michaelsutton: I figured it's a little safer to keep exactly as in go.

consensus/src/processes/coinbase.rs
michaelsutton: vec.truncate keeps capacity as is, so this is equal

consensus/src/processes/coinbase.rs
```diff
@@ -64,62 +105,121 @@ impl CoinbaseManager {
...
+
+    pub fn serialize_coinbase_payload(&self, data: &CoinbaseData) -> CoinbaseResult<Vec<u8>> {
+        let script_pub_key_len = data.miner_data.script_public_key.script().len();
+        if script_pub_key_len > self.coinbase_payload_script_public_key_max_len as usize {
+            return Err(CoinbaseError::PayloadScriptPublicKeyLenAboveMax(
+                script_pub_key_len,
+                self.coinbase_payload_script_public_key_max_len,
+            ));
+        }
+        let payload: Vec<u8> = data.blue_score.to_le_bytes().iter().copied()                    // Blue score                   (u64)
```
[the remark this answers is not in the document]
michaelsutton: Should be exact prealloc since all iterators here are size exact

someone235 (approved):

michaelsutton:
Addressed some of the items in https://github.com/kaspanet/rusty-kaspa/issues/60

PR #77, opened 2022-11-11
Sighash

someone235:
This PR also includes a toy implementation of check_scripts

### DIFF
--- a/consensus/core/src/hashing/sighash.rs
+++ b/consensus/core/src/hashing/sighash.rs
@@ -145,7 +145,7 @@ pub fn calc_schnorr_signature_hash(
     hash_type: SigHashType,
     reused_values: &mut SigHashReusedValues,
 ) -> Hash {
-    let input = tx.populated_inputs().nth(input_index).unwrap();
+    let input = tx.populated_input(input_index);
     let mut hasher = TransactionSigningHash::new();
     hasher
         .write_u16(tx.tx.version)
--- a/consensus/core/src/hashing/sighash.rs
+++ b/consensus/core/src/hashing/sighash.rs
@@ -1,0 +1,254 @@
+use hashes::{Hash, Hasher, HasherBase, TransactionSigningHash};
+
+use crate::{
+    subnets::SUBNETWORK_ID_NATIVE,
+    tx::{PopulatedTransaction, ScriptPublicKey, TransactionOutpoint, TransactionOutput},
+};
+
+use super::{sighash_type::SigHashType, HasherExtensions};
+
+pub struct SigHashReusedValues {
+    previous_outputs_hash: Option<Hash>,
+    sequence_hash: Option<Hash>,
+    sig_op_counts_hash: Option<Hash>,
+    outputs_hash: Option<Hash>,
+}
+
+impl SigHashReusedValues {
+    pub fn new() -> Self {
+        Self { previous_outputs_hash: None, sequence_hash: None, sig_op_counts_hash: None, outputs_hash: None }
+    }
+}
+
+fn previous_output_hash(tx: &PopulatedTransaction, hash_type: SigHashType, reused_values: &mut SigHashReusedValues) -> Hash {
+    if hash_type.is_sighash_anyone_can_pay() {
+        return 0.into();
+    }
+
+    if let Some(previous_outputs_hash) = reused_values.previous_outputs_hash {
+        previous_outputs_hash
+    } else {
+        let mut hasher = TransactionSigningHash::new();
+        for input in tx.tx.inputs.iter() {
+            hasher.update(input.previous_outpoint.transaction_id.as_bytes());
+            hasher.write_u32(input.previous_outpoint.index);
+        }
+        let previous_outputs_hash = hasher.finalize();
+        reused_values.previous_outputs_hash = Some(previous_outputs_hash);
+        previous_outputs_hash
+    }
+}
+
+fn sequence_hash(tx: &PopulatedTransaction, hash_type: SigHashType, reused_values: &mut SigHashReusedValues) -> Hash {
+    if hash_type.is_sighash_single() || hash_type.is_sighash_anyone_can_pay() || hash_type.is_sighash_none() {
+        return 0.into();
+    }
+
+    if let Some(sequence_hash) = reused_values.sequence_hash {
+        sequence_hash
+    } else {
+        let mut hasher = TransactionSigningHash::new();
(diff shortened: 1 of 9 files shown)


consensus/core/src/hashing/sighash.rs
```diff
@@ -0,0 +1,255 @@
...
+    hasher.write_var_bytes(script_public_key.script());
+}
+
+pub fn calc_schnorr_signature_hash(
+    tx: &PopulatedTransaction,
+    input_index: usize,
+    hash_type: SigHashType,
+    reused_values: &mut SigHashReusedValues,
+) -> Hash {
+    let input = tx.populated_inputs().nth(input_index).unwrap();
```
michaelsutton, line 148: Perhaps pass the (input, entry) tuple from the caller as well, since `nth` re-evaluates the iterator linearly until reaching `n`

consensus/core/src/hashing/sighash.rs
```diff
@@ -0,0 +1,255 @@
...
+    let mut hasher = TransactionSigningHash::new();
+    hasher.write_u16(tx.tx.version);
+    hasher.update(previous_output_hash(tx, hash_type, reused_values));
+    hasher.update(sequence_hash(tx, hash_type, reused_values));
+    hasher.update(sig_op_counts_hash(tx, hash_type, reused_values));
+    hash_outpoint(&mut hasher, input.0.previous_outpoint);
+    hash_script_public_key(&mut hasher, &input.1.script_public_key);
+    hasher.write_u64(input.1.amount);
```
michaelsutton, line 156: Hasher operations can be chained here  and above

consensus/core/src/hashing/sighash.rs
someone235: I think it's pretty cheap and doesn't justify complicating the API (you'll still need to pass the input_index because it's used in `outputs_hash`)

michaelsutton: So can we add the following method to `PopulatedTransaction` and use it here:

```
    pub fn populated_input(&self, input_index: usize) -> (&TransactionInput, &UtxoEntry) {
        (&self.tx.inputs[input_index], &self.entries[input_index])
    }
```

consensus/core/src/hashing/sighash.rs
```diff
@@ -0,0 +1,258 @@
...
+    subnets::SUBNETWORK_ID_NATIVE,
+    tx::{PopulatedTransaction, ScriptPublicKey, TransactionOutpoint, TransactionOutput},
+};
+
+use super::{sighash_type::SigHashType, HasherExtensions};
+
+#[derive(Default)]
+pub struct SigHashReusedValues {
+    previous_outputs_hash: Option<Hash>,
+    sequence_hash: Option<Hash>,
```
michaelsutton, line 13: `sequence_hash` -> `sequences_hash`

consensus/core/src/hashing/sighash.rs
```diff
@@ -0,0 +1,258 @@
+use hashes::{Hash, Hasher, HasherBase, TransactionSigningHash};
+
+use crate::{
+    subnets::SUBNETWORK_ID_NATIVE,
+    tx::{PopulatedTransaction, ScriptPublicKey, TransactionOutpoint, TransactionOutput},
+};
+
+use super::{sighash_type::SigHashType, HasherExtensions};
+
```
michaelsutton, line 9: Maybe add the doc from go:

```
/// Holds all fields used in the calculation of a transaction's sig_hash which are
/// the same for all transaction inputs.
/// Reuse of such values prevents the quadratic hashing problem.
```

consensus/core/src/hashing/sighash.rs
```diff
@@ -0,0 +1,258 @@
...
+impl SigHashReusedValues {
+    pub fn new() -> Self {
+        Self { previous_outputs_hash: None, sequence_hash: None, sig_op_counts_hash: None, outputs_hash: None }
+    }
+}
+
+fn previous_output_hash(tx: &PopulatedTransaction, hash_type: SigHashType, reused_values: &mut SigHashReusedValues) -> Hash {
+    if hash_type.is_sighash_anyone_can_pay() {
+        return 0.into();
```
michaelsutton, line 26: I think we should replace all `0.into()` in this file with `ZERO_HASH` (defined in `hashes::`)

consensus/core/src/hashing/sighash.rs
```diff
@@ -0,0 +1,258 @@
...
+    outputs_hash: Option<Hash>,
+}
+
+impl SigHashReusedValues {
+    pub fn new() -> Self {
+        Self { previous_outputs_hash: None, sequence_hash: None, sig_op_counts_hash: None, outputs_hash: None }
+    }
+}
+
+fn previous_output_hash(tx: &PopulatedTransaction, hash_type: SigHashType, reused_values: &mut SigHashReusedValues) -> Hash {
```
michaelsutton, line 24: `previous_output_hash` -> `previous_outputs_hash`

consensus/core/src/hashing/sighash.rs
```diff
@@ -0,0 +1,258 @@
...
+            hasher.update(input.previous_outpoint.transaction_id.as_bytes());
+            hasher.write_u32(input.previous_outpoint.index);
+        }
+        let previous_outputs_hash = hasher.finalize();
+        reused_values.previous_outputs_hash = Some(previous_outputs_hash);
+        previous_outputs_hash
+    }
+}
+
+fn sequence_hash(tx: &PopulatedTransaction, hash_type: SigHashType, reused_values: &mut SigHashReusedValues) -> Hash {
```
michaelsutton, line 43: `sequence_hash` -> `sequences_hash`

consensus/core/src/hashing/sighash.rs
```diff
@@ -0,0 +1,258 @@
...
+}
+
+fn payload_hash(tx: &PopulatedTransaction) -> Hash {
+    if tx.tx.subnetwork_id == SUBNETWORK_ID_NATIVE {
+        return 0.into();
+    }
+
+    // TODO: Right now this branch will never be executed, since payload is disabled
+    // for all non coinbase transactions. Once payload is enabled, the payload hash
+    // should be cached to make it cost O(1) instead of O(tx.inputs.len()).
```
michaelsutton, line 86: Since it only hashes the payload, technically the comment stating `O(tx.inputs.len())` is wrong

michaelsutton (approved):

PR #79, opened 2022-11-16
Optimize multi-level parents builder

michaelsutton:
Improves performance drastically (avoiding a quadratic `O(num_parents^2)` term)

### DIFF
--- a/consensus/src/processes/parents_builder.rs
+++ b/consensus/src/processes/parents_builder.rs
@@ -65,17 +65,40 @@ impl<T: HeaderStoreReader, U: ReachabilityStoreReader, V: RelationsStoreReader> 
                 .iter()
                 .filter(|h| block_level <= h.block_level as usize)
                 .map(|h| (h.header.hash, smallvec![h.header.hash]))
+                // We use smallvec with size 1 in order to optimize for the common case
+                // where the block itself is the only reference block
                 .collect::<BlockHashMap<SmallVec<[Hash; 1]>>>();
 
-            for parent in direct_parent_headers
-                .iter()
-                // We need to iterate parent's parents only if parent is not at block_level
-                .filter(|h| block_level > h.block_level as usize)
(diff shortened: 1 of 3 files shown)


consensus/src/processes/parents_builder.rs
```diff
@@ -119,7 +115,7 @@ impl<T: HeaderStoreReader, U: ReachabilityStoreReader, V: RelationsStoreReader>
                         reference_blocks
                     };
 
-                    if is_empty_level {
+                    if candidates_by_level_to_reference_blocks_map[block_level].is_empty() {
```
someone235: This doesn't look equivalent to me, since the current loop mutates `candidates_by_level_to_reference_blocks_map`

someone235 (changes requested):

consensus/src/processes/parents_builder.rs
michaelsutton: Seems like all we can optimize here is the case of no direct parents at level + first parent has more than one parent at level -- a case which should be very rare (since low levels are expected to have direct parents, and high levels are a chain)

consensus/src/processes/parents_builder.rs
```diff
@@ -63,8 +64,8 @@ impl<T: HeaderStoreReader, U: ReachabilityStoreReader, V: RelationsStoreReader>
             let mut level_candidates_to_reference_blocks = direct_parent_headers
                 .iter()
                 .filter(|h| block_level <= h.block_level as usize)
-                .map(|h| (h.header.hash, vec![h.header.hash]))
-                .collect::<BlockHashMap<_>>();
+                .map(|h| (h.header.hash, smallvec![h.header.hash]))
+                .collect::<BlockHashMap<SmallVec<[Hash; 1]>>>();
```
someone235, line 68: I would explain the use of SmallVec in a comment (why 1 is justified)

someone235 (changes requested):

someone235 (approved):

PR #90, opened 2022-12-21
Consensus API for Mempool

michaelsutton:
Implements consensus API for populating and validating a mempool transaction


consensus/src/pipeline/virtual_processor/processor.rs
```diff
@@ -480,6 +481,21 @@ impl VirtualStateProcessor {
...
 
+    pub fn validate_transaction_and_populate(&self, mutable_tx: &mut MutableTransaction) -> Result<(), RuleError> {
+        let virtual_utxo_view = self.virtual_utxo_store.as_ref(); // TODO: manage lock with virtual state lock
+        let virtual_state = self.virtual_state_store.read().get().unwrap(); // TODO: lock
+        let virtual_daa_score = virtual_state.daa_score;
+        let virtual_past_median_time = self.past_median_time_manager.calc_past_median_time(&virtual_state.ghostdag_data).0; // TODO: save median time in virtual state
+
+        self.transaction_validator.validate_tx_in_isolation(&mutable_tx.tx).unwrap(); // TODO: return rule error
+        self.transaction_validator.utxo_free_tx_validation(&mutable_tx.tx, virtual_daa_score, virtual_past_median_time).unwrap(); // TODO: return rule error
+        let validated_tx = self.validate_transaction_in_utxo_context(&mutable_tx.tx, virtual_utxo_view, virtual_daa_score).unwrap(); // TODO: return rule error
```
tiram88, line 492: This behaviour of `validate_transaction_in_utxo_context` does not seem to match glolang impl and does not entirely fulfill mempool needs.

We want to fill `mutable_tx` with all available entries even if a `ValidatedTransaction` cannot be built because some entries are still missing.

Your intention to introduce  `Vec<Option<UtxoEntry>>` is a perfect fit for this.

We also need to populate `calculated_mass`. Just wondering if this could it be delegated to transaction_validator_populated.rs `validate_populated_transaction_and_get_fee`. Wdyt?

tiram88 (approved):
I only thoroughly reviewed the consensus-core part (excluding sighash) and consensus parts connected directly or indirectly with the `ConsensusAPI` implementation.

Everything looks great.

consensus/core/src/tx.rs
```diff
@@ -267,21 +312,49 @@ impl<'a> ValidatedTransaction<'a> {
...
+    pub tx: Transaction,
+    /// Partially filled UTXO entry data
+    pub entries: Vec<Option<UtxoEntry>>,
+    /// Populated fee
+    pub calculated_fee: Option<u64>,
+    /// Populated mass
+    pub calculated_mass: Option<u64>,
+}
+
+impl MutableTransaction {
```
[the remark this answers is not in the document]
michaelsutton: Added.

tiram88 (approved):

consensus/core/src/tx.rs
```diff
@@ -267,21 +312,66 @@ impl<'a> ValidatedTransaction<'a> {
...
+/// Represents a mutable owned transaction along with partially filled UTXO entry data and optional fee and mass
+pub struct MutableTransaction {
+    /// The inner transaction
+    pub tx: Transaction,
+    /// Partially filled UTXO entry data
+    pub entries: Vec<Option<UtxoEntry>>,
+    /// Populated fee
+    pub calculated_fee: Option<u64>,
+    /// Populated mass
+    pub calculated_mass: Option<u64>,
```
someone235, line 336: Maybe it should be private? Calculated based on entries, and recalculate each time you add/remove an entry?

consensus/src/consensus/test_consensus.rs
```diff
@@ -240,7 +252,7 @@ pub fn create_temp_db() -> (TempDbLifetime, Arc<DB>) {
 /// Callers must keep the `TempDbLifetime` guard for as long as they wish the DB instance to exist.
 pub fn create_permanent_db(db_path: String) -> (TempDbLifetime, Arc<DB>) {
     let db_dir = PathBuf::from(db_path);
-    fs::create_dir(db_dir.as_path()).unwrap();
+    fs::create_dir(db_dir.as_path()).expect("The provided DB directory already exists");
```
someone235, line 255: Maybe change to "it's probably because"? Maybe the disk was removed etc

consensus/core/src/api/error.rs
```diff
@@ -0,0 +1,25 @@
+use hashes::Hash;
+use thiserror::Error;
+
+use crate::tx::{TransactionId, TransactionOutpoint};
+
+#[derive(Error, Debug)]
+pub enum ConsensusError {
```
someone235, line 7: I prefer to expose the underlying rule errors directly rather than converting them

consensus/src/pipeline/virtual_processor/processor.rs
```diff
@@ -480,10 +493,53 @@ impl VirtualStateProcessor {
...
 
-    pub fn build_block_template(self: &Arc<Self>, miner_data: MinerData, mut txs: Vec<Transaction>) -> BlockTemplate {
+    pub fn validate_mempool_transaction_and_populate(&self, mutable_tx: &mut MutableTransaction) -> TxResult<()> {
+        let virtual_read = self.virtual_stores.read();
+        let virtual_state = virtual_read.state.get().unwrap();
+        let virtual_utxo_view = &virtual_read.utxo_set;
+        let virtual_daa_score = virtual_state.daa_score;
+        let virtual_past_median_time = virtual_state.past_median_time;
+
+        self.transaction_validator.validate_tx_in_isolation(&mutable_tx.tx)?;
```
someone235, line 503: The first part can be done outside the lock. Anyway, this part requires more design.

consensus/src/pipeline/virtual_processor/processor.rs
```diff
@@ -480,10 +493,53 @@ impl VirtualStateProcessor {
...
+        self.transaction_validator.utxo_free_tx_validation(&mutable_tx.tx, virtual_daa_score, virtual_past_median_time)?;
+        self.validate_mempool_transaction_in_utxo_context(mutable_tx, virtual_utxo_view, virtual_daa_score)?;
+
+        Ok(())
+    }
+
+    fn validate_block_template_transaction(
```
someone235, line 510: TBD

consensus/src/pipeline/virtual_processor/utxo_validation.rs
```diff
@@ -184,34 +185,64 @@ impl VirtualStateProcessor {
...
+    pub(super) fn validate_mempool_transaction_in_utxo_context(
+        &self,
+        mutable_tx: &mut MutableTransaction,
+        utxo_view: &impl UtxoView,
+        pov_daa_score: u64,
+    ) -> TxResult<()> {
+        let mut has_missing_outpoints = false;
+        for i in 0..mutable_tx.inputs().len() {
+            if mutable_tx.entries[i].is_some() {
+                // We prefer a previously populated entry if such exists
```
someone235, line 230: Something to think about: what if the DAA score changes and it affects coinbase maturity?

consensus/src/model/stores/utxo_set.rs
```diff
@@ -20,7 +20,7 @@ pub trait UtxoSetStoreReader {
 }
 
 pub trait UtxoSetStore: UtxoSetStoreReader {
-    fn write_diff(&self, utxo_diff: &UtxoDiff) -> Result<(), StoreError>;
+    fn write_diff(&mut self, utxo_diff: &UtxoDiff) -> Result<(), StoreError>;
```
someone235, line 23: Write a comment that explains why the `&mut` is needed

consensus/core/src/tx.rs
```diff
@@ -347,6 +347,11 @@ impl MutableTransaction {
         Self { tx, entries: entries.into_iter().map(Some).collect(), calculated_fee: None, calculated_mass: None }
     }
 
+    pub fn as_verifiable(&self) -> impl VerifiableTransaction + '_ {
+        debug_assert!(self.is_verifiable());
```
someone235, line 351: nit: I think it should be a regular `assert`

someone235 (changes requested):

consensus/core/src/tx.rs
tiram88: AFAIU, calculated mass is based on the inner `Transaction` outputs. It does not depend on the entries.

michaelsutton: Fee calc only makes sense once all utxo entries are obtained, and never changes after. Mass requires the mass calculator bcs it's based on consensus params

consensus/src/pipeline/virtual_processor/processor.rs
michaelsutton: ?

consensus/src/pipeline/virtual_processor/utxo_validation.rs
michaelsutton: Added a comment

consensus/core/src/tx.rs
michaelsutton: Done.

someone235 (approved):

PR #98, opened 2023-01-06
Updated README.md regarding installation of clang toolchain

HLXEasy:
Clang toolchain must be installed, so I've added an entry onto the "Getting started" section.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,8 +4,13 @@ Work in progress to implement the Kaspa full-node and related libraries in the R
 
 ## Getting started
 
-- Install the [rust toolchain](https://rustup.rs/).
-
+- Install Protobuf (required for grpc)
+  - Linux: `sudo apt install protobuf-compiler libprotobuf-dev`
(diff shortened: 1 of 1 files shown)


KaffinPX:
"because of rocksdb" could be good

HLXEasy:
> "because of rocksdb" could be good

Thx for the suggestion, done.

elichai:
I think this is only true for linux?

HLXEasy:
> I think this is only true for linux?

Updated regarding required installation steps, separated by Linux and Windows.

michaelsutton (commented):
Looks good. Thank you. I suggest to replace `Linux i. e.` with `Linux:` (x4)

HLXEasy:
> Looks good. Thank you. I suggest to replace `Linux i. e.` with `Linux:` (x4)

Done.

michaelsutton (approved):

PR #103, opened 2023-01-15
MiningManager, Mempool & BockTemplateBuilder

tiram88:
Implements the `MiningManager` and its components: `Mempool`, `BlockTemplateBuilder` and `BlockTemplateCache`.

The `MinigManager` is designed to be synchronous and executes on a single thread. Concurrent accesses to the underlying `Mempool` and `BlockTemplateCache` are protected by two locks, the second protecting both the cache and the builder when appropriate.

The `MinigManager` is essentially a facade for its components.

The design essentially sticks to the original golang version. Though, some dependencies and execution paths are reorganized to avoid passing unnecessary references. For example, the `MempoolUtxoSet` is now owned by the `TransactionsPool` instead of the `Mempool`.

**One notable difference in behavior is introduced**: when a pool, either the `TransactionsPool` or the `OrphanPool`, gets filled with high priority transactions, new inserts are rejected, preventing the pool to grow beyond the limit. 

The implementation avoids smart pointers, except in the cache. Transactions entering the mempool are owned locally and all transactions leaving the mining manager are cloned.

The code is covered by unit tests exclusively. This is made possible by introducing a `ConsensusMock` and `CoinbaseManagerMock`. This differs from the original version which contains integration tests for the `MiningManager` involving a `TestConsensus`.

This PR is still lacking the upcoming script engine to be 100% operational. This will be adapted in a later PR when the missing dependency will be available. At this stage, some integration tests might be added too.


mining/src/cache.rs
```diff
@@ -0,0 +1,45 @@
...
+
+    pub(crate) fn clear(&mut self) {
+        // This differs from golang implementation.
+        // The cache timer is reset to 0 so its lifetime is expired.
+        self.last_update_time = 0;
+        self.block_template = None;
+    }
+
+    pub(crate) fn get_immutable_cached_template(&self) -> Option<Rc<BlockTemplate>> {
+        if SystemTime::now().duration_since(UNIX_EPOCH).unwrap().as_millis() as u64 - self.last_update_time > self.cache_lifetime {
```
michaelsutton, line 33: I suggest using `i64` for `self.last_update_time` since sys time can go back if the system clock is changed so we'd like to avoid the overflow here

mining/src/cache.rs
```diff
@@ -0,0 +1,45 @@
...
+}
+
+impl BlockTemplateCache {
+    pub(crate) fn new(cache_lifetime: Option<u64>) -> Self {
+        let cache_lifetime = cache_lifetime.unwrap_or(DEFAULT_CACHE_LIFETIME);
+        Self { last_update_time: 0, block_template: None, cache_lifetime }
+    }
+
+    pub(crate) fn clear(&mut self) {
+        // This differs from golang implementation.
```
michaelsutton, line 26: Can remove this comment line now

mining/src/cache.rs
```diff
@@ -0,0 +1,45 @@
...
+    time::{SystemTime, UNIX_EPOCH},
+};
+
+/// CACHE_LIFETIME indicates the default duration in milliseconds after which the cached data expires.
+const DEFAULT_CACHE_LIFETIME: u64 = 1_000;
+
+pub(crate) struct BlockTemplateCache {
+    /// Time, in milliseconds, when the cache was last updated
+    last_update_time: u64,
+    block_template: Option<Rc<BlockTemplate>>,
```
michaelsutton, line 13: When `MiningManager` is used and passed between threads this line will not compile bcs of the Rc. It must be an Arc

mining/src/manager.rs
```diff
@@ -0,0 +1,178 @@
...
+        &self.block_template_builder
+    }
+
+    /// validate_and_insert_transaction validates the given transaction, and
+    /// adds it to the set of known transactions that have not yet been
+    /// added to any block.
+    ///
+    /// The returned transactions are clones of objects owned by the mempool.
+    pub fn validate_and_insert_transaction(
+        &mut self,
```
michaelsutton, line 115: Should not be `mut`

mining/src/block_template/builder.rs
```diff
@@ -0,0 +1,137 @@
...
+
+        let block_txs = selector.select_transactions();
+        Ok(self.consensus().build_block_template(miner_data.clone(), block_txs)?)
+    }
+
+    /// modify_block_template clones an existing block template, modifies it to the requested coinbase data and updates the timestamp
+    pub(crate) fn modify_block_template(
+        &self,
+        new_miner_data: &MinerData,
+        block_template_to_modify: Rc<BlockTemplate>,
```
michaelsutton, line 113: set arg type to `&BlockTemplate`

mining/src/mempool/mod.rs
```diff
@@ -0,0 +1,120 @@
...
+/// - An orphan transaction is unorphaned when all its UTXO entries have been
+///   built or found.
+/// - There are transaction priorities: high and low.
+/// - Transactions submitted to the mempool by a RPC call have **high priority**.
+///   They are owned by the node, they never expire in the mempool and the node
+///   rebroadcasts them once in a while.
+/// - Transactions received through P2P have **low-priority**. They expire after
+///   60 seconds and are removed if not inserted in a block for mining.
+pub(crate) struct Mempool {
+    config: Rc<Config>,
```
michaelsutton, line 40: When `MiningManager` is used and passed between threads this line will not compile bcs of the Rc. It must be an Arc (same goes for Config in the pools within)

mining/src/manager.rs
```diff
@@ -0,0 +1,178 @@
...
+        include_orphan_pool: bool,
+    ) -> OwnerSetTransactions {
+        self.mempool.read().get_transactions_by_addresses(script_public_keys, include_transaction_pool, include_orphan_pool)
+    }
+
+    pub fn transaction_count(&self, include_transaction_pool: bool, include_orphan_pool: bool) -> usize {
+        self.mempool.read().transaction_count(include_transaction_pool, include_orphan_pool)
+    }
+
+    pub fn handle_new_block_transactions(&self, block_transactions: &[Transaction]) -> MiningManagerResult<Vec<MutableTransaction>> {
```
michaelsutton, line 160: Add a TODO about using tx acceptance data to verify that new block txs are actually accepted into virtual state. On the broader level this is related to mempool replay feature well

mining/src/mempool/revalidate_high_priority_transactions.rs
```diff
@@ -0,0 +1,53 @@
...
+        // validations calls to consensus.
+        let ids = self.transaction_pool.index(true).topological_index()?;
+        let mut valid_ids = vec![];
+
+        for transaction_id in ids.iter() {
+            // Try to take the transaction out of the storage map so we can mutate it with some self functions.
+            // The redeemers of removed transactions are removed too so the following call may return a None.
+            if let Some(mut transaction) = self.transaction_pool.all_mut().remove(transaction_id) {
+                let is_valid = self.revalidate_transaction(&mut transaction.mtx)?;
+                // Then put the transaction back into the storage map.
```
michaelsutton, line 27: Explain that we modified the tx out of the map, and now return it post modification. Alternative is to use `RefCell` which should still be considered

mining/src/mempool/revalidate_high_priority_transactions.rs
```diff
@@ -0,0 +1,53 @@
...
+                self.transaction_pool.all_mut().insert(*transaction_id, transaction);
+                if is_valid {
+                    // A following transaction should not remove this one from the pool since we process
+                    // in topological order
+                    // TODO: consider the scenario of two high priority txs sandwiching a low one
```
michaelsutton, line 32: Explain that in this case topology order is not guaranteed since we topologically sorted only high-prio

mining/src/model/owner_txs.rs
```diff
@@ -0,0 +1,26 @@
...
+use std::collections::{HashMap, HashSet};
+
+use super::TransactionIdSet;
+
+pub type ScriptPublicKeySet = HashSet<ScriptPublicKey>;
+
+/// Transaction ids involved in either sending to or receiving from an
+/// address or its [`ScriptPublicKey`] equivalent.
+#[derive(Default)]
+pub struct OwnerTransactions {
```
michaelsutton, line 11: The name is not so clear. What is the logical/functional role of this struct?

mining/src/model/owner_txs.rs
```diff
@@ -0,0 +1,26 @@
...
+}
+
+impl OwnerTransactions {
+    pub fn is_empty(&self) -> bool {
+        self.sending_txs.is_empty() && self.receiving_txs.is_empty()
+    }
+}
+
+#[derive(Default)]
+pub struct OwnerSetTransactions {
```
michaelsutton, line 23: The name is not so clear here too

mining/src/manager.rs
```diff
@@ -0,0 +1,178 @@
...
+                    }
+                }
+                Err(err) => {
+                    return Err(err)?;
+                }
+            }
+        }
+    }
+
+    pub(crate) fn block_candidate_transactions(&self) -> Vec<MutableTransaction> {
```
michaelsutton, line 95: Add a TODO suggesting to check the cost of cloning all mempool transactions (if #mempool txs >> #txs which fit in a block, then most of these clones will not be used).

michaelsutton (approved):

PR #118, opened 2023-02-05
Transaction script engine

michaelsutton:
wip


crypto/txscript/src/opcodes/macros.rs
```diff
@@ -0,0 +1,95 @@
+macro_rules! opcode_serde {
+    ($type:ty) => {
```
someone235, line 2: Add some comment that explains that this is used for OP_PUSHDATA etc

crypto/txscript/src/lib.rs
```diff
@@ -0,0 +1,489 @@
...
+            Ok(())
+        }
+    }
+
+    fn execute_script(&mut self, script: &[u8]) -> Result<(), TxScriptError> {
+        let script_result = script
+            .iter()
+            .batching(|it| {
+                // reads the opcode num item here and then match to opcode
+                it.next().map(|code| deserialize(*code, it))
```
someone235, line 168: I think it'll be clearer if `deserialize` will be renamed to `deserialize_opcode_data` and the current lambda will have its own `deserialize_opcode` function

someone235 (changes requested):

someone235 (approved):

PR #120, opened 2023-02-07
Address manager

someone235:


addressmanager/src/lib.rs
```diff
@@ -0,0 +1,200 @@
...
+            if self.addresses.len() <= MAX_ADDRESSES {
+                return;
+            }
+
+            let to_remove = self.addresses.iter().max_by(|a, b| a.1.cmp(b.1)).unwrap();
+            self.db_store.remove(to_remove.0.ip, to_remove.0.port).unwrap();
+        }
+
+        pub fn get(&self, address: NetAddress) -> u64 {
+            self.db_store.get(address.ip, address.port).unwrap()
```
michaelsutton, line 152: Shouldn't this be a read from cache (`self.addresses`)?

addressmanager/src/lib.rs
```diff
@@ -0,0 +1,200 @@
...
+        }
+
+        pub fn remove(&mut self, address: NetAddress) {
+            self.db_store.remove(address.ip, address.port).unwrap()
```
michaelsutton, line 156: Seems like the cache is not updated here

addressmanager/src/lib.rs
```diff
@@ -0,0 +1,200 @@
...
+            self.keep_limit();
+        }
+
+        fn keep_limit(&mut self) {
+            if self.addresses.len() <= MAX_ADDRESSES {
+                return;
+            }
+
+            let to_remove = self.addresses.iter().max_by(|a, b| a.1.cmp(b.1)).unwrap();
+            self.db_store.remove(to_remove.0.ip, to_remove.0.port).unwrap();
```
michaelsutton, line 148: Should call `self.remove`

addressmanager/src/lib.rs
```diff
@@ -0,0 +1,200 @@
...
+    use rand::{distributions::WeightedIndex, prelude::Distribution};
+
+    use crate::{
+        stores::not_banned_address_store::{DbNotBannedAddressesStore, NotBannedAddressesStore, NotBannedAddressesStoreReader},
+        NetAddress, MAX_ADDRESSES, MAX_CONNECTION_FAILED_COUNT,
+    };
+
+    pub struct Store {
+        db_store: DbNotBannedAddressesStore,
+        addresses: HashMap<NetAddress, u64>,
```
michaelsutton, line 122: I think we should have a newtype around the u64 to avoid errors with the banned u64 which represents time: 
```pub struct ConnectionFailureCount(pub u64)``` 

and of course use it in the DB store as well

File: addressmanager/src/stores/banned_address_store.rs [hunk 5]
```diff
@@ -0,0 +1,77 @@
...
+impl From<AddressKey> for Ipv6Addr {
+    fn from(k: AddressKey) -> Self {
+        k.0.into()
+    }
+}
+
+#[derive(Clone)]
+pub struct DbBannedAddressesStore {
+    db: Arc<DB>,
+    access: CachedDbAccess<AddressKey, u64>,
```
michaelsutton, line 54: I think we should have a newtype around the u64 to avoid errors with the non-banned u64 which represents fail counts: 
```pub struct ConnectionBanTimestamp(pub u64)```

File: addressmanager/src/stores/not_banned_address_store.rs [hunk 6]
```diff
@@ -0,0 +1,102 @@
+use database::{
+    db::DB,
+    errors::{StoreError, StoreResult},
+    prelude::{CachedDbAccess, DirectDbWriter},
+};
+use std::net::Ipv6Addr;
+use std::{error::Error, fmt::Display, sync::Arc};
+
+const STORE_PREFIX_CONNECTION_FAILED_COUNT: &[u8] = b"not-banned-addresses-connection-failed-count";
```
michaelsutton, line 9: Pretty long name, maybe simply `not-banned-addresses` like in go ?

addressmanager/src/stores/not_banned_address_store.rs
```diff
@@ -0,0 +1,102 @@
...
+    }
+}
+
+impl Display for AddressKey {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        let ip_port: (Ipv6Addr, u16) = (*self).into();
+        write!(f, "{}:{}", ip_port.0, ip_port.1)
+    }
+}
+
```
michaelsutton, line 39: Maybe add a non-urgent TODO to generalize this key serialization scheme (looks very much like utxo index key)

addressmanager/src/lib.rs
```diff
@@ -0,0 +1,200 @@
...
+        Arc::new(Mutex::new(Self {
+            banned_address_store: DbBannedAddressesStore::new(db.clone(), MAX_ADDRESSES as u64),
+            not_banned_address_store: not_banned_address_store_with_cache::new(db),
+        }))
+    }
+
+    pub fn add_address(&mut self, address: NetAddress) {
+        // TODO: Don't add non routable addresses
+
+        // We mark `connectionFailedCount` as 0 only after first success
```
michaelsutton, line 49: -> `connection_failed_count`

addressmanager/src/lib.rs
```diff
@@ -0,0 +1,200 @@
...
+        pub fn remove(&mut self, address: NetAddress) {
+            self.db_store.remove(address.ip, address.port).unwrap()
+        }
+
+        pub fn get_all_addresses(&self) -> impl Iterator<Item = NetAddress> + '_ {
+            self.addresses.keys().copied()
+        }
+
+        pub fn get_random_addresses(&self, count: usize, exceptions: HashSet<NetAddress>) -> Vec<NetAddress> {
+            let count = min(count, self.addresses.len());
```
michaelsutton, line 164: I think this is safer:

```
            let addresses = self.addresses.iter().filter(|(a, _)| !exceptions.contains(a)).collect_vec();
            let mut weights =
                addresses.iter().map(|(_, &count)| 64f64.powf((MAX_CONNECTION_FAILED_COUNT + 1 - count) as f64)).collect_vec();
            let count = min(count, addresses.len());
```

having zeroed entries might make the count too large and you might try sampling from a vec of zeros which probably is an err

addressmanager/src/lib.rs
```diff
@@ -0,0 +1,200 @@
...
+            }
+            None => false,
+        }
+    }
+}
+
+fn unix_time() -> u64 {
+    SystemTime::now().duration_since(UNIX_EPOCH).unwrap().as_millis() as u64
+}
+
```
michaelsutton, line 102: Add a comment explaining that here we want the full store to be in memory and we don't mind assuming single-thread access since this is not an intensive operation

michaelsutton (approved):

PR #125, opened 2023-02-16
Implement check_sig_op_counts

someone235:


crypto/txscript/src/lib.rs
```diff
@@ -72,6 +72,66 @@ pub struct TxScriptEngine<'a, T: VerifiableTransaction> {
...
+    // TODO: Check for overflows
+    let mut num_sigs: u64 = 0;
+    for (i, op) in opcodes.iter().enumerate() {
+        match op {
+            Ok(op) => {
+                match op.value() {
+                    codes::OpCheckSig | codes::OpCheckSigVerify | codes::OpCheckSigECDSA => num_sigs += 1,
+                    codes::OpCheckMultiSig | codes::OpCheckMultiSigVerify | codes::OpCheckMultiSigECDSA => {
+                        if i > 0
+                            && opcodes[i - 1].as_ref().expect("they were checked before").value() >= codes::OpTrue
```
tiram88, line 111: Might perhaps be simplified to
```
                        let prev_code = opcodes[i - 1].as_ref().expect("they were checked before");
                        if i > 0 && (codes::OpTrue..codes::Op16).contains(&prev_code.value()) {
                            num_sigs += to_small_int(prev_code) as u64;
                        } else {
                            num_sigs += MAX_PUB_KEYS_PER_MUTLTISIG as u64;
                        }
```

crypto/txscript/src/lib.rs
```diff
@@ -72,6 +72,66 @@ pub struct TxScriptEngine<'a, T: VerifiableTransaction> {
...
+    if signature_script_ops.is_empty() || signature_script_ops.iter().any(|op| op.is_err() || !op.as_ref().unwrap().is_push_opcode()) {
+        return 0;
+    }
+
+    let p2sh_script = signature_script_ops.last().expect("checked if empty above").as_ref().expect("checked if err above").get_data();
+    let p2sh_ops = parse_script::<T>(p2sh_script).collect_vec();
+    get_sig_op_count_by_opcodes(&p2sh_ops)
+}
+
+fn get_sig_op_count_by_opcodes<T: VerifiableTransaction>(opcodes: &[Result<Box<dyn OpCodeImplementation<T>>, TxScriptError>]) -> u64 {
```
tiram88, line 101: I guess the golang "not precise" case is no longer needed.

tiram88 (changes requested):

crypto/txscript/src/lib.rs
someone235: Right. It was also not used in golang

tiram88 (approved):

PR #129, opened 2023-02-19
Utxoindex integration test + some fixes

someone235:


indexes/utxoindex/src/stores/indexed_utxos.rs
```diff
@@ -96,6 +109,22 @@ impl UtxoEntryFullAccessKey {
         bytes.extend_from_slice(transaction_outpoint_key.as_ref());
         Self(Arc::new(bytes))
     }
+
+    /// Extracts a [`ScriptPublicKey`] and [`TransactionOutpoint`] of the  [`UtxoEntryFullAccessKey`]
+    pub fn extract_script_public_key_and_outpoint(&self) -> (ScriptPublicKey, TransactionOutpoint) {
+        let version = ScriptPublicKeyVersion::from_le_bytes(self.0[..VERSION_TYPE_SIZE].try_into().unwrap());
+        let script_size =
```
D-Stacks, line 116: Also not sure why we require script_size here. script could also just be decoded via: `self.0[VERSION_TYPE_SIZE..TRANSACTION_OUTPOINT_KEY_SIZE]`.

indexes/utxoindex/src/stores/indexed_utxos.rs
```diff
@@ -145,22 +175,24 @@ impl UtxoSetByScriptPublicKeyStoreReader for DbUtxoSetByScriptPublicKeyStore {
...
-                        usize::MAX,
-                        false,
-                    )
-                    .map(move |value| {
-                        let (k, v) = value.expect("expected `key: TransactionOutpoint`, `value: CompactUtxoEntry`");
-                        (k, v)
-                    }),
+                self.access.seek_iterator(Some(script_public_key_bucket.as_ref()), None, usize::MAX, false).map(|res| {
+                    let (key, entry) = res.unwrap();
+                    let (_, outpoint) = UtxoEntryFullAccessKey(Arc::new(key.to_vec())).extract_script_public_key_and_outpoint();
```
D-Stacks, line 180: Using an extract_outpoint function can be cheaper, no need to deserialize the script_public_key.

Can just take self.0[(self.0.len() - TRANSACTION_OUTPOINT_KEY_SIZE)..]

indexes/utxoindex/src/core/api/mod.rs
```diff
@@ -47,4 +54,4 @@ pub trait UtxoIndexApi: Send + Sync {
 // 2) there is no need for an inner Arc since we hold an Arc on the Option,
 // but alas, we need Sized for the option, hence it is in a Box.
 
-pub type DynUtxoIndexApi = Arc<Option<Box<RwLock<dyn UtxoIndexApi>>>>;
+pub type DynUtxoIndexApi = Option<Box<Arc<RwLock<dyn UtxoIndexApi>>>>;
```
D-Stacks, line 57: I think the box is redundant in this case, since the arc has sized. I am also unsure if cloning a box of an arc increases the arc reference count.

indexes/utxoindex/src/index.rs
```diff
@@ -29,8 +29,12 @@ pub struct UtxoIndex {
 
 impl UtxoIndex {
     /// Creates a new [`UtxoIndex`] within a [`RwLock`]
-    pub fn new(consensus: DynConsensus, db: Arc<DB>) -> RwLock<Self> {
-        RwLock::new(Self { consensus, store: Store::new(db) })
+    pub fn new(consensus: DynConsensus, db: Arc<DB>) -> UtxoIndexResult<Arc<RwLock<Self>>> {
+        let mut utxoindex = Self { consensus, store: Store::new(db) };
+        if !utxoindex.is_synced()? {
```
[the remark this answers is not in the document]
D-Stacks: Note: this is also one of the reasons why my native utxoindex test will fail with this pr, since the utxoindex is created with a consensus that does not call init() before it is passed.

indexes/utxoindex/src/stores/indexed_utxos.rs
D-Stacks: Also, the key bytes you retrieve from the seek iterator consists of transaction outpoint bytes (as it slices out the length of the bucket, i.e. the script_public_key bytes), as such the resulting key can simply be converted to an outpoint directly, there is no need for the UtxoEntryFullAccessKey here.

indexes/utxoindex/src/index.rs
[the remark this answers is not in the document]
someone235: This was fixed by processing genesis just when constructing consensus

indexes/utxoindex/src/stores/indexed_utxos.rs
```diff
@@ -24,6 +27,7 @@ impl From<&ScriptPublicKey> for ScriptPublicKeyBucket {
     fn from(script_public_key: &ScriptPublicKey) -> Self {
         let mut bytes: Vec<u8> = Vec::with_capacity(VERSION_TYPE_SIZE + script_public_key.script().len());
         bytes.extend_from_slice(&script_public_key.version().to_le_bytes());
+        bytes.extend_from_slice(&(script_public_key.script().len() as u64).to_le_bytes()); // TODO: Consider using a smaller integer
```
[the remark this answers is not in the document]
someone235: It's good practice to encode the size of a variable length slice, to guarantee any key is unique

michaelsutton (approved):

PR #131, opened 2023-02-21
Fix json_test sync issues

someone235:


michaelsutton:
Joining with `core.join()` might cause a race condition here because of how the `EventProcessor` manages the shutdown signal. It listens to the signal with a `tokio::select` so the shutdown can proceed events already in the channel and break the loop before processing them. In this case the test will fail due to utxo state mismatch between virtual and the index. In other consensus processors we passed the shutdown signal as a special enum variant through the same channel, hence ensuring processing consistency. 

I suggest we wait until @tiram88 completes the notification system PR which will effect the structure of `EventProcessor` anyhow, and revisit this issue once that is completed.

michaelsutton (approved):

PR #136, opened 2023-03-04
Generic notification system

tiram88:
Refactor the notification system:

- Move the notification library to a dedicated `kaspa-notify` crate
- Make the `Notifier` generic over 2 types implementing respectively the `Notification` and `Connection` traits, extending its usage beyond the RPC sub-system where it initially was placed.
- Add single and compounded `Subscription`s,
- Add scopes, adapted to each event type, and use those as subscription commands
- Build a notification backbone having its root in Consensus and made of a DAG of async services each providing a `Notifier` with following structure and roles:
    - `consensus-notify`: handle consensus notifications, relaying to `kaspa-rpc-core` and `kaspa-index-processor` (if activated)
    - `kaspa-index-processor`: optionally process consensus notifications related to the `utxoindex` and relays a converted version of them to `kaspa-rpc-core`
    - `kaspa-rpc-core`: handle the rpc notification tree
    - `kaspa-grpc-server`: handle the server-side gRPC notifications
    - `kaspa-grpc-client`: handle the client-side gRPC notifications

- **Rename 2 event types and matching notifications**:
     - `VirtualSelectedParentChainChanged` to `VirtualChainChanged`
     - `VirtualSelectedParentBlueScoreChanged` to `SinkBlueScoreChanged`

- Provide extensive unit tests of the notification system

The `event-processor` crate is replaced by (refactored into) `kaspa-index-processor`, featuring the processing of utxo-related notifications into the `utxoindex` and ready to host a future `txindex`.


consensus/src/pipeline/virtual_processor/processor.rs
```diff
@@ -407,21 +418,23 @@ impl VirtualStateProcessor {
...
-                            accepted_tx_ids: Arc::new(new_virtual_state.accepted_tx_ids),
-                        })))
-                        .expect("expected send");
-                };
+                // Emit notifications
+                let accumulated_diff = Arc::new(accumulated_diff);
+                let virtual_parents = Arc::new(new_virtual_state.parents);
+                let _ = self
+                    .notification_root()
+                    .notify(Notification::UtxosChanged(UtxosChangedNotification::new(accumulated_diff, virtual_parents)));
```
michaelsutton, line 426: Since `notify` calls do not require the cloned Arc, can we just directly access the `notification_root` field?

kaspad/src/main.rs
```diff
@@ -151,31 +152,34 @@ pub fn main() {
...
 
-    let utxoindex: DynUtxoIndexApi = if args.utxoindex {
+    let notify_service = Arc::new(NotifyService::new(consensus.notification_root(), notification_recv));
+
+    let index_service: Option<Arc<IndexService>> = if args.utxoindex {
         // Use only a single thread for none-consensus databases
         let utxoindex_db = database::prelude::open_db(utxoindex_db_dir, true, 1);
-        Some(UtxoIndex::new(consensus.clone(), utxoindex_db).unwrap())
+        let utxoindex: DynUtxoIndexApi = Some(UtxoIndex::new(consensus.clone(), utxoindex_db).unwrap());
+        Some(Arc::new(IndexService::new(notify_service.clone(), utxoindex)))
```
michaelsutton, line 168: I would pass the inner notifier (`Arc<ConsensusNotifier>`) to `IndexService` and not the `NotifyService` itself

kaspad/src/main.rs
```diff
@@ -151,31 +152,34 @@ pub fn main() {
...
     } else {
         None
     };
 
-    let event_processor = Arc::new(EventProcessor::new(utxoindex.clone(), consensus_recv, event_processor_send));
-
-    let rpc_core_server = Arc::new(RpcCoreServer::new(consensus.clone(), utxoindex, event_processor_recv));
+    let rpc_core_server = Arc::new(RpcCoreServer::new(consensus.clone(), notify_service.clone(), index_service.clone()));
```
michaelsutton, line 173: Same here. Pass the inner notifiers and `DynUtxoindex` if needed, and not the services themselves

consensus/notify/src/notification.rs
```diff
@@ -0,0 +1,173 @@
...
+        removed_chain_block_hashes: Arc<Vec<Hash>>,
+        accepted_transaction_ids: Arc<Vec<TransactionId>>,
+    ) -> Self {
+        Self { added_chain_block_hashes, removed_chain_block_hashes, accepted_transaction_ids }
+    }
+}
+
+#[derive(Debug, Clone, Default)]
+pub struct FinalityConflictNotification {
+    pub violating_block_hash: Arc<Hash>,
```
michaelsutton, line 114: `Hash` is `Copy` and should never be arced

consensus/notify/src/notification.rs
```diff
@@ -0,0 +1,173 @@
...
+    pub added_chain_block_hashes: Arc<Vec<Hash>>,
+    pub removed_chain_block_hashes: Arc<Vec<Hash>>,
+    pub accepted_transaction_ids: Arc<Vec<TransactionId>>,
+}
+
+impl VirtualChainChangedNotification {
+    pub fn new(
+        added_chain_block_hashes: Arc<Vec<Hash>>,
+        removed_chain_block_hashes: Arc<Vec<Hash>>,
+        accepted_transaction_ids: Arc<Vec<TransactionId>>,
```
michaelsutton, line 106: Technically this should be an array corresponding to `added_chain_block_hashes` with an `accepted_ids` vector per added chain block. Or better of, define an `AcceptanceData` struct with a field for `accepted_ids` and use it. However note that this data should be sent only if there is a listener for it (as it is heavy).

In golang we have this data pulled from consensus only if there's a listener. 

It seems like here we can check the subscription, but I emphasize that we should have a way to check the subscription and only then read this data from stores

consensus/notify/src/notification.rs
```diff
@@ -0,0 +1,173 @@
...
+    }
+
+    fn event_type(&self) -> EventType {
+        self.into()
+    }
+}
+
+#[derive(Debug, Clone)]
+pub struct BlockAddedNotification {
+    pub block: Arc<Block>,
```
michaelsutton, line 86: No need to Arc a block as it is a wrapper over header and txs Arcs

consensus/notify/src/notification.rs
```diff
@@ -0,0 +1,173 @@
...
+            true => Some(self.clone()),
+            false => None,
+        }
+    }
+
+    fn apply_virtual_chain_changed_subscription(&self, subscription: &VirtualChainChangedSubscription) -> Option<Self> {
+        match subscription.active() {
+            true => {
+                if let Notification::VirtualChainChanged(ref payload) = self {
+                    if !subscription.include_accepted_transaction_ids() && !payload.accepted_transaction_ids.is_empty() {
```
michaelsutton, line 59: Maybe explain in a comment that if no subscription + `payload.accepted_ids` is not empty, then we must re-create the object and drop the ids, otherwise we can clone it as is

consensus/src/consensus/mod.rs
```diff
@@ -318,6 +319,8 @@ impl Consensus {
         let (virtual_sender, virtual_receiver): (CrossbeamSender<BlockProcessingMessage>, CrossbeamReceiver<BlockProcessingMessage>) =
             unbounded_crossbeam();
 
+        let notification_root = Arc::new(ConsensusNotificationRoot::new(notification_sender));
```
michaelsutton, line 322: Maybe pass the root to consensus and not the sender? this way you can avoid the `notification_root(&self)` getter. I think it will play out better when we have a staging consensus and we need to switch the active consensus to it

michaelsutton: You can keep `TestConsensus` as is, so you will not need to change all tests

consensus/notify/src/notification.rs
tiram88: Partially done
`AcceptanceData` is moved to `consensus-core` and used in the notification

michaelsutton (approved):

PR #137, opened 2023-03-05
Implement connection manager and addresses flows

someone235:


components/addressmanager/src/lib.rs
```diff
@@ -48,19 +36,19 @@ impl AddressManager {
         }
 
         // We mark `connection_failed_count` as 0 only after first success
-        self.not_banned_address_store.set(address, ConnectionFailureCount(1));
+        self.not_banned_address_store.set(address, 1);
```
michaelsutton, line 39: I would switch `not_banned` to something like -> `nonbanned`/`allowed`/`legitimate ` in all struct/mod/var names

components/addressmanager/src/lib.rs
```diff
@@ -120,76 +107,89 @@ mod not_banned_address_store_with_cache {
...
-            let count = min(count, addresses.len());
-            let mut weights =
-                addresses.iter().map(|(_, count)| 64f64.powf((MAX_CONNECTION_FAILED_COUNT + 1 - count.0) as f64)).collect_vec();
+            let mut weights = addresses
+                .iter()
+                .map(|(_, entry)| 64f64.powf((MAX_CONNECTION_FAILED_COUNT + 1 - entry.connection_failed_count) as f64))
+                .collect_vec();
 
-            (0..count)
+            (0..addresses.len())
```
michaelsutton, line 180: If we go through all addresses (and eventually select all), it means we are only randomly sorting with priority, I would expect much more efficient methods. This imp makes sense to me only with `count` << `addresses.len()`

components/addressmanager/src/lib.rs
```diff
@@ -2,94 +2,82 @@ mod stores;
...
     }
 
     pub fn get_all_addresses(&self) -> impl Iterator<Item = NetAddress> + '_ {
-        self.not_banned_address_store.get_all_addresses()
+        self.address_store.get_all_addresses()
     }
 
-    pub fn get_random_addresses(&self, count: usize, exceptions: HashSet<NetAddress>) -> Vec<NetAddress> {
-        self.not_banned_address_store.get_random_addresses(count, exceptions)
+    pub fn get_random_addresses(&self, exceptions: HashSet<NetAddress>) -> impl Iterator<Item = NetAddress> {
```
michaelsutton, line 67: perhaps call it `iterate_prioritized_random_addresses` (prioritized is optional)

components/addressmanager/src/lib.rs
```diff
@@ -120,81 +107,117 @@ mod not_banned_address_store_with_cache {
...
-            (0..count)
-                .map(|_| {
-                    let dist = WeightedIndex::new(&weights).unwrap();
-                    let i = dist.sample(&mut rand::thread_rng());
-                    let addr = addresses[i];
-                    weights[i] = 0f64;
-                    *addr.0
-                })
-                .collect_vec()
+        pub fn get_randomized_addresses(&self, exceptions: HashSet<NetAddress>) -> impl Iterator<Item = NetAddress> {
```
michaelsutton, line 172: `unzip` can be used to save two vec collect ops as follows:

```
        pub fn get_randomized_addresses(&self, exceptions: HashSet<NetAddress>) -> impl Iterator<Item = NetAddress> {
            let exceptions: HashSet<AddressKey> = exceptions.into_iter().map(|addr| addr.into()).collect();
            let (weights, addresses) = self
                .addresses
                .iter()
                .filter(|(addr_key, _)| !exceptions.contains(addr_key))
                .map(|(_, e)| (64f64.powf((MAX_CONNECTION_FAILED_COUNT + 1 - e.connection_failed_count) as f64), e.address))
                .unzip();

            RandomWeightedIterator::new(weights, addresses)
        }
  ```

File: components/connectionmanager/src/lib.rs [hunk 5]
```diff
@@ -0,0 +1,179 @@
...
+            if is_connected && !request.is_permanent {
+                continue;
+            }
+
+            if !is_connected && request.next_attempt <= SystemTime::now() {
+                debug!("Connecting to a connection request to {}", address);
+                if self.p2p_adaptor.connect_peer(address.to_string()).await.is_none() {
+                    debug!("Failed connecting to a connection request to {}", address);
+                    if request.is_permanent {
+                        const MAX_RETRY_DURATION: Duration = Duration::new(600, 0);
```
michaelsutton, line 119: `Duration::from_secs` is more concise than `new` (here and in line below)

components/connectionmanager/src/lib.rs
```diff
@@ -0,0 +1,179 @@
...
+                        let retry_duration = max(Duration::new(30u64 * 2u64.pow(request.attempts), 0), MAX_RETRY_DURATION);
```
michaelsutton, line 120: `max` -> `min` (never exceed `MAX_RETRY_DURATION`)

components/connectionmanager/src/lib.rs
```diff
@@ -0,0 +1,179 @@
...
+                        debug!("Will retry to connect to {} in {}", address, DurationString::from(retry_duration));
+                        new_requests.insert(
+                            address,
+                            ConnectionRequest {
+                                next_attempt: request.next_attempt + retry_duration,
```
michaelsutton, line 125: makes more sense to recall `SystemTime::now()` and not use `request.next_attempt` as a start point. Also don't cache the now above to a variable bcs we have awaits in between

components/connectionmanager/src/lib.rs
```diff
@@ -0,0 +1,179 @@
...
+            let address = *address;
+            let request = request.clone();
+            let is_connected = peer_by_address.contains_key(&address);
+            if is_connected && !request.is_permanent {
+                continue;
+            }
+
+            if !is_connected && request.next_attempt <= SystemTime::now() {
+                debug!("Connecting to a connection request to {}", address);
+                if self.p2p_adaptor.connect_peer(address.to_string()).await.is_none() {
```
michaelsutton, line 116: Is this `connect_peer` (with the default 16 retries) deliberate? I suggest making the default w/o retries, and rename `connect_peer_with_retry_params` -> `connect_peer_with_retries` and fix current test locations that call `connect_peer` to use the latter

protocol/p2p/src/core/router.rs
```diff
@@ -231,3 +232,13 @@ impl Router {
         true
     }
 }
+
+fn msg_type_string(msg: &KaspadMessage) -> String {
```
michaelsutton, line 236: I prefer avoiding the String heap alloc. Maybe return `Option<KaspadMessagePayloadType>` and rely on a debug print above?

protocol/flows/src/service.rs
```diff
@@ -39,16 +60,18 @@ impl AsyncService for P2pService {
...
             if let Some(peer_address) = self.connect.clone() {
-                trace!("P2P, p2p::main - starting peer:{peer_address}");
-                let _peer_id = p2p_adaptor.connect_peer_with_retry_params(peer_address, 1, Duration::from_secs(1)).await;
+                connection_manager.add_connection_request(peer_address.to_socket_addrs().unwrap().next().unwrap(), true).await;
             }
 
             // Keep the P2P server running until a service shutdown signal is received
             shutdown_signal.await;
             p2p_adaptor.terminate_all_peers().await;
+            connection_manager.stop().await;
```
michaelsutton, line 74: remove the commented lines below

michaelsutton (approved):

PR #139, opened 2023-03-11
Add validate_pruning_point_proof

someone235:


consensus/src/consensus/mod.rs
```diff
@@ -662,8 +663,8 @@ impl ConsensusApi for Consensus {
         self.coinbase_manager.modify_coinbase_payload(payload, miner_data)
     }
 
-    fn validate_pruning_proof(self: Arc<Self>, _proof: &PruningPointProof) -> Result<(), PruningImportError> {
-        unimplemented!()
+    fn validate_pruning_proof(self: Arc<Self>, proof: &PruningPointProof) -> Result<(), PruningImportError> {
```
michaelsutton, line 666: change the interface and avoid the `Arc<Self>` and use `&self` (now all consensus api methods are like this)

protocol/flows/src/v5/ibd/flow.rs
```diff
@@ -109,8 +109,7 @@ impl IbdFlow {
         let proof: PruningPointProof = msg.try_into()?;
         debug!("received proof with overall {} headers", proof.iter().map(|l| l.len()).sum::<usize>());
 
-        // TODO: call validate_pruning_proof when implemented
-        // consensus.clone().validate_pruning_proof(&proof);
+        consensus.clone().validate_pruning_proof(&proof)?;
```
michaelsutton, line 112: once you change the api to `&self` you can avoid the clone here

consensus/src/model/stores/reachability.rs
```diff
@@ -58,19 +59,27 @@ pub struct DbReachabilityStore {
     db: Arc<DB>,
     access: CachedDbAccess<Hash, Arc<ReachabilityData>, BlockHasher>,
     reindex_root: CachedDbItem<Hash>,
+    level: Option<BlockLevel>,
 }
 
 impl DbReachabilityStore {
-    pub fn new(db: Arc<DB>, cache_size: u64) -> Self {
+    pub fn new(db: Arc<DB>, level: Option<BlockLevel>, cache_size: u64) -> Self {
```
michaelsutton, line 66: The `BlockLevel` arg might be a bit misleading (giving the impression that generally reachability is stored per level). Since this is only a specific use case for the pruning proof, perhaps make the arg `prefix_ending: &[u8]` and parse the block level by the caller. Also maybe avoid changing the `new` function and add another ctor function `with_prefix_ending` (which new will call with an empty slice).

michaelsutton (approved):

Issue #110, opened 2023-02-01
Transaction propagation flow


someone235:

someone235:
blocked by #111

PR #153, opened 2023-03-31
Transaction Relay Flow

tiram88:
Addresses [issue #110 ](https://github.com/kaspanet/rusty-kaspa/issues/110)


protocol/flows/src/v5/txrelay/flow.rs
```diff
@@ -0,0 +1,278 @@
...
+        self.ctx.mining_manager().get_transaction(transaction_id, true, true).is_some()
+    }
+
+    /// Returns the next Transaction or TransactionNotFound message in msg_route,
+    /// returning only one of the message types at a time.
+    ///
+    /// Populates the invs_route queue with any inv messages that meanwhile arrive.
+    async fn read_response(&mut self) -> Result<Response, ProtocolError> {
+        loop {
+            tokio::select! {
```
michaelsutton, line 156: There is no need to use select here, you only need to await on `self.msg_route.recv()`. The whole point of using two incoming routes (msg and invs) unlike go is that you don't need to do this. The main loop dequeues invs, and msg is used only after sending tx requests. This means that unlike block invs, you don't need the `TwoWayIncomingRoute` object here. The difference is that block invs requests can come indirectly from orphan roots, while txs don't have this mechanism

protocol/flows/src/v5/txrelay/flow.rs
```diff
@@ -0,0 +1,278 @@
...
+        Some(self.router.clone())
+    }
+
+    async fn start(&mut self) -> Result<(), ProtocolError> {
+        self.start_impl().await
+    }
+}
+
+impl RequestTransactionsFlow {
+    #[allow(dead_code)]
```
michaelsutton, line 253: delete

protocol/flows/src/flowcontext/transactions.rs
```diff
@@ -0,0 +1,76 @@
...
+    pb::{kaspad_message::Payload, InvTransactionsMessage, KaspadMessage},
+    Hub,
+};
+use std::time::{Duration, Instant};
+
+const REBROADCAST_INTERVAL: Duration = Duration::from_secs(30);
+const BROADCAST_INTERVAL: Duration = Duration::from_millis(500);
+const MAX_INV_PER_TX_INV_MSG: usize = 131_072;
+
+#[allow(dead_code)]
```
michaelsutton, line 17: delete

protocol/flows/src/v5/txrelay/flow.rs
```diff
@@ -0,0 +1,278 @@
...
+
+    async fn start_impl(&mut self) -> Result<(), ProtocolError> {
+        // trace!("Starting relay transactions flow with {}", self.router.identity());
+        loop {
+            // Loop over incoming block inv messages
+            let inv = self.invs_route.dequeue().await?;
+            // trace!("Receive an inv message from {} with {} transaction ids", self.router.identity(), inv.len());
+
+            // Transaction relay is disabled if the node is out of sync and thus not mining
+            if self.ctx.is_ibd_running() {
```
michaelsutton, line 103: Note that in go we do not test "IBD is running" but rather `flow.IsNearlySynced()`. See it blockrelay/flow.rs the lines 
```
                let sink_timestamp = consensus.get_sink_timestamp();
                if sink_timestamp.is_none_or(|t| unix_now() > t + self.ctx.config.expected_daa_window_duration_in_milliseconds()) {
                    continue;
                }
```
These lines are equal to `!flow.IsNearlySynced()` in go. You can consider making this logic a shared function somewhere

mining/src/mempool/validate_and_insert_transaction.rs
```diff
@@ -14,21 +14,23 @@ use kaspa_consensus_core::{
...
+use super::tx::{Orphan, Priority};
+
 impl<T: ConsensusMiningContext + ?Sized> Mempool<T> {
     pub(crate) fn validate_and_insert_transaction(
         &mut self,
         transaction: Transaction,
-        is_high_priority: bool,
-        allow_orphan: bool,
+        priority: Priority,
+        orphan: Orphan,
```
michaelsutton, line 24: Maybe struct name should be `Orphan` -> `Orphaning` ? not sure but not really important

protocol/flows/src/flow_context.rs
```diff
@@ -106,6 +154,80 @@ impl FlowContext {
...
+        let blocks = self.unorphan_blocks(hash).await;
+        // Use a ProcessQueue so we get rid of duplicates
+        let mut transactions_to_broadcast = ProcessQueue::new();
+        for block in once(block).chain(blocks.into_iter()) {
+            transactions_to_broadcast
+                .enqueue_chunk(self.mining_manager().handle_new_block_transactions(&block.transactions)?.iter().map(|x| x.id()));
+        }
+
+        // Don't relay transactions when in IBD
+        if self.is_ibd_running() {
```
michaelsutton, line 173: I guess it should be `is_nearly_synced` here as well

protocol/flows/src/v5/txrelay/flow.rs
tiram88: The reason why I kept the `TwoWayIncomingRoute` design is because processing the responses to a `RequestTransactions` with full capacity (thus an `InvTransaction` holding `MAX_INV_PER_TX_INV_MSG` ids) might take a significant time. Since the `IncomingRoute` provided by `Router::subscribe` is bounded to 128, the `invs_route` queue might fill up. So using the select prevents this risk.

That being said, this risk is probably low now by with network transaction rate increasing, it might materialize in the future.

WDYT?

michaelsutton: The two-way mechanism simply moves the size issue to the unlimited deque, so we can simply use a larger channel if needed.

michaelsutton: Perhaps add a method `Router::subscribe_with_capacity` and use a larger value for invs

protocol/flows/src/v5/txrelay/flow.rs
```diff
@@ -0,0 +1,278 @@
...
+                        Err(_) => Err(ProtocolError::Timeout(DEFAULT_TIMEOUT)),
+                    }
+                },
+            };
+        }
+    }
+
+    async fn receive_transactions(&mut self, requests: Vec<RequestScope<TransactionId>>) -> Result<(), ProtocolError> {
+        // trace!("Receive {} transaction ids from {}", requests.len(), self.router.identity());
+        for requested_id in requests.iter().map(|x| x.req.to_owned()) {
```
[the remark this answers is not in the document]
tiram88: yes, probably better

mining/src/mempool/validate_and_insert_transaction.rs
tiram88: let's keep `Orphan` if you don't mind

protocol/flows/src/flow_context.rs
tiram88: In original go codebase, [`is_ibd_running()` is used](https://github.com/tiram88/kaspad/blob/bd1420220a1c9f7ab253b2b120240351e9440146/app/protocol/flowcontext/blocks.go#L73).

Perhaps, yes, we might be more permissive and use `is_nearly_synced()`.

michaelsutton: so we changed this in the main tx relay flow, but missed this one

michaelsutton (approved):

ap501boii (commented):
Payment assistance

PR #157, opened 2023-04-09
Query DNS seeder when don't have enough addresses

someone235:


components/connectionmanager/src/lib.rs
```diff
@@ -203,4 +213,33 @@ impl ConnectionManager {
...
         join_all(futures).await;
     }
+
+    fn dns_seed(self: &Arc<Self>, mut min_addresses_to_fetch: usize) {
+        let mut dns_seeders = self.dns_seeders.to_vec();
+        dns_seeders.shuffle(&mut thread_rng());
+
+        for seeder in dns_seeders {
+            info!("Querying DNS seeder {}", seeder);
+            let addrs = match format!("{}:53", seeder).to_socket_addrs() {
```
elichai, line 223: FYI `(seeder, 53).to_socket_addrs()` should also work, not sure if `53` is the right port to put here BTW

components/connectionmanager/src/lib.rs
```diff
@@ -203,4 +213,33 @@ impl ConnectionManager {
         }
         join_all(futures).await;
     }
+
+    fn dns_seed(self: &Arc<Self>, mut min_addresses_to_fetch: usize) {
+        let mut dns_seeders = self.dns_seeders.to_vec();
+        dns_seeders.shuffle(&mut thread_rng());
```
elichai, line 219: FYI, You can use https://docs.rs/rand/latest/rand/seq/trait.SliceRandom.html#tymethod.choose_multiple to do this without copying the seeders to a Vec.

components/connectionmanager/src/lib.rs
```diff
@@ -203,4 +213,32 @@ impl ConnectionManager {
         }
         join_all(futures).await;
     }
+
+    fn dns_seed(self: &Arc<Self>, mut min_addresses_to_fetch: usize) {
+        let shuffled_dns_seeders = self.dns_seeders.choose_multiple(&mut thread_rng(), self.dns_seeders.len());
+        for seeder in shuffled_dns_seeders.cloned() {
```
elichai, line 219: I think you can use pattern matching i.e.: `for &seeder in shuffled_dns_seeders` without the need for `cloned()`? (this is the same thing, just more obvious that there aren't any clones here really)

elichai (dismissed):
LGTM

elichai (approved):

PR #159, opened 2023-04-10
WASM JS unit tests by CI

KaffinPX:
WIP


KaffinPX:
#154 should be merged first for passing tests

michaelsutton:
Seems extremally outdated, can we close this?

aspect:
I stalled on giving feedback on this as this doesn't really fit the approach we are using in WASM design, which is to use Rust as control mechanism for all functionality, including JavaScript unit tests. 

My policy is to avoid using JavaScript as much as possible, and if not, bias toward JS being injected/created via Rust. This also means that if push comes to shove and we absolutely need NodeJs for tests, we would start a NodeJs instance from within Rust unit tests.

So for now I think this should be closed and I will review and see what needs to be added for this in the coming weeks.

michaelsutton:
Closed as per discussion with @aspect

PR #161, opened 2023-04-11
RPC API

tiram88:
Implements the most important `RpcApi` functions, covering at least what is used by the wallet or miner (wip)


consensus/core/src/config/mod.rs
```diff
@@ -26,11 +26,19 @@ pub struct Config {
     // TODO:
     // is_archival: bool,
     // enable_sanity_check_pruning_utxoset: bool,
+    //
+    /// Enable the UTXO index
+    pub utxoindex: bool,
```
someone235, line 31: This is consensus config. It's a little bit weird to put it here. Maybe add a more general config struct that wraps it?

consensus/core/src/config/params.rs
```diff
@@ -273,3 +300,13 @@ pub const DEVNET_PARAMS: Params = Params {
     max_block_level: 250,
     pruning_proof_m: 1000,
 };
+
+#[cfg(test)]
+mod tests {
+    use crate::config::params::{DIFFICULTY_MAX, DIFFICULTY_MAX_AS_F64};
+
+    #[test]
+    fn test_difficulty_max_consts() {
```
someone235, line 309: Can you also test that `DIFFICULTY_MAX` is `Uint256::from_u64(1).wrapping_shl(255) - 1.into()`?

consensus/src/consensus/mod.rs
```diff
@@ -171,17 +173,16 @@ pub struct Consensus {
     pub counters: Arc<ProcessingCounters>,
 
     // Config
-    config: Config,
+    config: Arc<Config>,
```
someone235, line 176: Why you made it an Arc?

consensus/src/consensus/mod.rs
```diff
@@ -679,6 +681,11 @@ impl ConsensusApi for Consensus {
         }
     }
 
+    fn get_sink(&self) -> Option<Hash> {
+        // TODO: unwrap on virtual state read when staging consensus is implemented
```
someone235, line 685: I think we can just unwrap now. Right @michaelsutton ?

consensus/src/consensus/mod.rs
```diff
@@ -679,6 +681,11 @@ impl ConsensusApi for Consensus {
...
+        self.virtual_processor.virtual_stores.read().state.get().unwrap_option().map(|state| state.ghostdag_data.selected_parent)
+    }
+
     fn get_sink_timestamp(&self) -> Option<u64> {
         // TODO: unwrap on virtual state read when staging consensus is implemented
```
someone235, line 690: I think we can just unwrap now. Right @michaelsutton ?

consensus/core/src/block.rs
```diff
@@ -78,3 +78,67 @@ impl BlockTemplate {
         Self { block, miner_data, coinbase_has_red_reward, selected_parent_timestamp }
     }
 }
+
+/// A structure containing various information about a specific block
+#[derive(Debug, Clone)]
+pub struct BlockInfo {
```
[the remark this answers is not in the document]
someone235: If we decide to keep it, maybe wrap everything in an Option and drop the `exists` field?

someone235 (changes requested):

consensus/core/src/block.rs
```diff
@@ -78,3 +78,67 @@ impl BlockTemplate {
...
+            blue_score: Default::default(),
+            blue_work: Default::default(),
+            selected_parent: Default::default(),
+            mergeset_blues: Default::default(),
+            mergeset_reds: Default::default(),
+        }
+    }
+}
+
+pub struct BlockRelations {
```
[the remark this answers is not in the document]
tiram88: ok, I remove it and the matching `get_block_relations`

consensus/core/src/config/mod.rs
tiram88: yes, let's create a dedicated internal crate for that

consensus/src/consensus/mod.rs
tiram88: Because this struct is going to grow a lot and is shared by many sub-systems

consensus/src/consensus/mod.rs
michaelsutton: yes we can unwrap

consensus/src/consensus/mod.rs
michaelsutton: yup

consensus/core/src/block.rs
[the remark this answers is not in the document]
michaelsutton: I think I agree with @someone235, a `get_block_status` call + adding a direct `get_ghostdag_data` should due

consensus/core/src/config/mod.rs
michaelsutton: perhaps keep as is now and add a TODO

consensus/core/src/block.rs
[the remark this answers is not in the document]
michaelsutton: use `ExternalGhostdagData` as the return type for `get_ghostdag_data` (there are conversions from `consensus::GhostdagData` so simply use `.into()`)

consensus/src/consensus/mod.rs
```diff
@@ -842,10 +870,73 @@ impl ConsensusApi for Consensus {
         })
     }
 
+    fn get_block_even_if_header_only(&self, hash: Hash) -> ConsensusResult<Block> {
+        let Some(status) = self.statuses_store.read().get(hash).unwrap_option().filter(|&status| !status.has_block_header()) else {
```
michaelsutton, line 874: pretty sure you want `.filter(|&status| status.has_block_header())` without `!` negation

consensus/src/consensus/mod.rs
```diff
@@ -842,10 +870,73 @@ impl ConsensusApi for Consensus {
...
+            ghostdag_data.blue_score,
+            ghostdag_data.blue_work,
+            ghostdag_data.selected_parent,
+            (*ghostdag_data.mergeset_blues).clone(),
+            (*ghostdag_data.mergeset_reds).clone(),
+        ))
+    }
+
+    fn get_block_children(&self, hash: Hash) -> Option<Arc<Vec<Hash>>> {
+        self.relations_stores.read()[0].get_children(hash).unwrap_option()
```
michaelsutton, line 909: it's more strait forward to use `self.relations_service` (it wraps level 0 already)

consensus/src/consensus/mod.rs
```diff
@@ -842,10 +870,73 @@ impl ConsensusApi for Consensus {
...
+    }
+
+    fn get_block_parents(&self, hash: Hash) -> Option<Arc<Vec<Hash>>> {
+        self.relations_stores.read()[0].get_parents(hash).unwrap_option()
```
michaelsutton, line 913: same. use `self.relations_service`

consensus/src/consensus/mod.rs
```diff
@@ -842,10 +870,73 @@ impl ConsensusApi for Consensus {
...
+        Ok(self.acceptance_data_store.get(hash).unwrap())
+    }
+
+    fn get_blocks_acceptance_data(&self, hashes: &[Hash]) -> ConsensusResult<Vec<Arc<AcceptanceData>>> {
+        hashes
+            .iter()
+            .copied()
+            .map(|hash| {
+                self.validate_block_exists(hash)?;
+                Ok(self.acceptance_data_store.get(hash).unwrap())
```
michaelsutton, line 931: the logic here relies on the fact that `hashes` are chain blocks obtained by a call to `get_virtual_chain` (otherwise not all blocks have acceptance data). Given this fact I think we can skip the `validate_block_exists` part and add a doc over `get_blocks_acceptance_data` def in `ConsensusApi` explaining this assumption. Same goes for `get_block_acceptance_data`. 
Alternatively, do not unwrap on `self.acceptance_data_store.get(hash)` and propagate that error instead.

protocol/flows/src/flow_context.rs
```diff
@@ -128,6 +137,11 @@ impl FlowContext {
         self.is_ibd_running.load(Ordering::SeqCst)
     }
 
+    pub async fn is_nearly_synced(&self) -> bool {
+        // TODO: make sure relying on is_ibd_running() is safe in all cases
+        !self.is_ibd_running() || self.consensus().session().await.is_nearly_synced()
```
michaelsutton, line 142: I would remove `!self.is_ibd_running() ||`. Also I think callers should obtain the session themselves and make a direct call (in one of current usages a session is obtained anyway). So please remove it altogether

rpc/core/src/error.rs
```diff
@@ -30,6 +33,24 @@ pub enum RpcError {
...
+    RejectedTransaction(RpcTransactionId, String),
+
+    #[error("Block {0} is invalid. No verbose data can be built.")]
+    InvalidBlock(RpcHash),
+
+    #[error("If includeTransactions is set, then includeBlockVerboseData must be set as well.")]
+    InvalidGetBlocksRequest,
+
+    #[error("No sink exists.")]
+    NoSink,
```
michaelsutton, line 49: Remove when removing Option from `get_sink`

rpc/service/src/converter/consensus.rs
```diff
@@ -0,0 +1,205 @@
...
+            address.map(|address| RpcTransactionOutputVerboseData { script_public_key_type, script_public_key_address: address });
+        RpcTransactionOutput { value: output.value, script_public_key: output.script_public_key.clone(), verbose_data }
+    }
+
+    pub fn get_virtual_chain_accepted_transaction_ids(
+        &self,
+        consensus: &dyn ConsensusApi,
+        chain_path: &ChainPath,
+    ) -> RpcResult<Vec<RpcAcceptedTransactionIds>> {
+        // TODO: handle calls to consensus by chunks if still relevant
```
michaelsutton, line 163: not relevant bcs `acceptance_data_store` is lockless

consensus/core/src/config/mod.rs
tiram88: good for me

consensus/src/consensus/mod.rs
tiram88: pretty sure I agree :)
nice catch!

PR #166, opened 2023-04-14
Consensus reset handling 

michaelsutton:


protocol/flows/src/flow_context.rs
```diff
@@ -54,12 +54,17 @@ pub struct FlowContext {
     shared_block_requests: Arc<Mutex<HashSet<Hash>>>,
     transactions_spread: Arc<AsyncRwLock<TransactionsSpread>>,
```
tiram88, line 55: `orphans_pool`, `transactions_spread` seem to no longer need to be wrapped into an `Arc`:

tiram88 (changes requested):

tiram88 (approved):

PR #167, opened 2023-04-14
Disallow reorgs to the anticone of the pruning point

someone235:


michaelsutton:
Since this wasn't the cause of the bug we had (but rather #170), I prefer doing this with less pressure when implementing the virtual finality check. This issue should be very rare given that we process block bodies in GD order.

Issue #122, opened 2023-02-15
MUST DO BEFORE BETA: Address left comments in txscript PR


someone235:
Address all unresolved comments on #118 and cover txscript with many tests: Each opcode has to be tested

michaelsutton:
Is this covered by now @someone235 @tmrlvi @elichai ?

freshair18:
Is this covered now?

PR #173, opened 2023-04-20
Added check we're outside conditions in the end of script

tmrlvi:
Fixed one of the todos mentions in https://github.com/kaspanet/rusty-kaspa/issues/122

### DIFF
--- a/crypto/txscript/src/lib.rs
+++ b/crypto/txscript/src/lib.rs
@@ -231,8 +231,11 @@ impl<'a, T: VerifiableTransaction> TxScriptEngine<'a, T> {
             Ok(())
         });
 
-        // Moving between scripts
-        // TODO: Check that we are not in if when moving between scripts
+        // Moving between scripts - we can't be inside an if
+        if !self.cond_stack.is_empty() {
+            return Err(TxScriptError::ErrUnbalancedConditional);
+        }
+
         // Alt stack doesn't persist
         self.astack.clear();
         self.num_ops = 0; // number of ops is per script.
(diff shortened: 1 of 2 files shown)


crypto/txscript/src/lib.rs
```diff
@@ -559,6 +542,62 @@ mod tests {
...
+            ScriptTestCase {
+                script: b"\x00\x63", // opcodes::codes::OpFalse{data: ""}, opcodes::codes::OpIf{data: ""}
+                expected_result: Err(TxScriptError::ErrUnbalancedConditional),
+            },
+            ScriptTestCase {
+                script: b"\x51\x63\x51\x68", // opcodes::codes::OpTrue{data: ""}, opcodes::codes::OpIf{data: ""}
+                expected_result: Ok(()),
+            },
+            ScriptTestCase {
+                script: b"\x00\x63\x51\x68", // opcodes::codes::OpTrue{data: ""}, opcodes::codes::OpIf{data: ""}
```
someone235, line 593: Shouldn't it be `OpFalse OpIf OpTrue OpEndIf`?
And btw I think it's much more readable to write in the above way

(Some other comments seems wrong too, please check them)

someone235 (changes requested):

crypto/txscript/src/lib.rs
tmrlvi: Fixed

tmrlvi:
(Rebase from master - to be up to date)

someone235 (approved):

PR #178, opened 2023-04-28
CLI Peers Management

tiram88:
Enhance `kaspad` CLI with following repeatable arguments:
- `--connect=` and `--addpeer=` for peers management.
- `--uacomment=` for user-agent comments describing the launched node (see [BIP-014](https://en.bitcoin.it/wiki/BIP_0014) for more information)

### DIFF
--- a/kaspad/src/args.rs
+++ b/kaspad/src/args.rs
@@ -1,7 +1,9 @@
+use clap::ArgAction;
 #[allow(unused)]
 use clap::{arg, command, Arg, Command};
 use kaspa_consensus::config::Config;
 use kaspa_core::kaspad_env::version;
+use kaspa_utils::networking::ContextualNetAddress;
 
 pub struct Defaults {
     pub appdir: &'static str,
@@ -41,15 +43,16 @@ impl Default for Defaults {
 pub struct Args {
     // NOTE: it is best if property names match config file fields
     pub appdir: Option<String>,
-    pub rpclisten: Option<String>,
-    pub rpclisten_borsh: Option<String>,
-    pub rpclisten_json: Option<String>,
+    pub rpclisten: Option<ContextualNetAddress>,
+    pub rpclisten_borsh: Option<ContextualNetAddress>,
+    pub rpclisten_json: Option<ContextualNetAddress>,
     pub unsafe_rpc: bool,
     pub wrpc_verbose: bool,
     pub log_level: String,
     pub async_threads: usize,
-    pub connect: Option<String>,
-    pub listen: Option<String>,
+    pub connect_peers: Vec<ContextualNetAddress>,
+    pub add_peers: Vec<ContextualNetAddress>,
+    pub listen: Option<ContextualNetAddress>,
     pub utxoindex: bool,
     pub reset_db: bool,
     pub outbound_target: usize,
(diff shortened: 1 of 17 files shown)


kaspad/src/args.rs
```diff
@@ -111,23 +117,37 @@ pub fn cli(defaults: &Defaults) -> Command {
...
         )
         .arg(arg!(--unsaferpc "Enable RPC commands which affect the state of the node"))
         .arg(
-            Arg::new("connect")
+            Arg::new("connect-peers")
                 .long("connect")
                 .value_name("IP[:PORT]")
-                .num_args(0..=1)
+                .action(ArgAction::Append)
+                .num_args(0..=10) // TODO: define an upper bound
```
michaelsutton, line 129: Are you sure `num_args` is relevant when using `ArgAction::Append` ? we should manually test it.

kaspad/src/main.rs
```diff
@@ -188,7 +203,10 @@ pub fn main() {
     // DB used for addresses store and for multi-consensus management
     let meta_db = kaspa_database::prelude::open_db(meta_db_dir, true, 1);
 
-    let grpc_server_addr = args.rpclisten.unwrap_or_else(|| format!("0.0.0.0:{}", config.default_rpc_port())).parse().unwrap();
+    let connect_peers = args.connect_peers.iter().map(|x| x.normalize(config.default_p2p_port())).collect();
+    let add_peers = args.add_peers.iter().map(|x| x.normalize(config.default_p2p_port())).collect();
+    let listen = args.listen.unwrap_or_default().normalize(config.default_rpc_port());
```
michaelsutton, line 208: `default_rpc_port` -> `default_p2p_port`. Also probably better to rename `listen` -> `p2p_server_addr`

protocol/flows/src/service.rs
```diff
@@ -71,8 +75,11 @@ impl AsyncService for P2pService {
 
         // Launch the service and wait for a shutdown signal
         Box::pin(async move {
-            if let Some(peer_address) = self.connect.clone() {
-                connection_manager.add_connection_request(peer_address.to_socket_addrs().unwrap().next().unwrap(), true).await;
+            for peer_address in self.connect_peers.iter().cloned() {
```
michaelsutton, line 78: I think in go-kaspad having `connect` means exclusiveness. Meaning the node does not attempt any other outbound connections (can be achieved by setting `outbound_target=0` if `!self.connect_peers.is_empty()`. I'm not sure about inbound connections

kaspad/src/args.rs
michaelsutton: In general I'm not sure I understand the role of `.num_args(0..=1)` in all other flags

tiram88: I agree on both remarks.

I removed all `.num_args(iter)` and tested manually. The system exhibits the desired behavior as expected.

kaspad/src/main.rs
tiram88: Nice catch!

protocol/flows/src/service.rs
tiram88: Right, this exclusiveness is achieved in go-kaspad by setting outbound target to 0 and disabling DNS seeding. I added that.

I could find no impact of the presence of `connect` in go-kaspad on inbound connections.

I also discovered and fixed another little discrepancy: added peers have permanent connections.

michaelsutton (approved):

PR #181, opened 2023-05-05
Logger with parallel output to file with rotation & compression

tiram88:


supertypo:
Will it be possible to disable file logging completely (ie. only output to console)?

tiram88:
@supertypo, the new arg arg `--nologfiles` allows logging to console only

michaelsutton (approved):

PR #184, opened 2023-05-11
rename max diff to max pow

D-Stacks:
`max diff` is technically incorrect, and a `max diff` value, if defined, would be in the region of 2^223, not 2^255.

the value used here as max diff, is correctly labeled as max pow in go-kaspad. This PR rectifies this.   

I will also leave a quick reference to btc documentation regarding this: 

https://en.bitcoin.it/wiki/Difficulty#What_is_the_maximum_difficulty.3F

(only we are one power of two less, since our max PoW is one power of two less then btc).


D-Stacks:
closing as i am unsure about my assertions.

PR #187, opened 2023-05-23
add "Allow submit block when not synced" as "enable_unsynced_mining" command

D-Stacks:
no in-code edits required to activate devnet after this.


KaffinPX:
Isnt it a bit long, what about allow-desync-mining

D-Stacks:
> Isnt it a bit long, what about allow-desync-mining

I agree.

PR #190, opened 2023-06-01
Handle recovery of pruning workflows on process crashes

michaelsutton:
Adds support for batching and recovery for pruning operations in order to ensure coherent state also on system crashes. 
Namely:

1. Performs pruning point utxo-set update in DB batches which include the current position
2. Uses the above position to recover the process on system restart
3. Saves the last data pruned point in the DB
4. Uses this point to trigger data prune on system restart if a former operation was uncompleted
5. This point also marks the beginning of full history for archival nodes

### DIFF
--- a/consensus/src/pipeline/virtual_processor/processor.rs
+++ b/consensus/src/pipeline/virtual_processor/processor.rs
@@ -19,15 +19,15 @@ use crate::{
             daa::DbDaaStore,
             ghostdag::{DbGhostdagStore, GhostdagData, GhostdagStoreReader},
             headers::{DbHeadersStore, HeaderStoreReader},
-            past_pruning_points::{DbPastPruningPointsStore, PastPruningPointsStore},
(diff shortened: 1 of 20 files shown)


consensus/src/pipeline/virtual_processor/processor.rs
```diff
@@ -769,13 +776,22 @@ impl VirtualStateProcessor {
             ));
         }
 
+        {
+            // Set the pruning point utxoset position to the new point we just verified
```
someone235, line 780: Shouldn't we set it after we write the UTXO set?

someone235 (changes requested):

consensus/src/pipeline/virtual_processor/processor.rs
michaelsutton: exactly why the function name is misleading (see TODO). At this point the pruning point utxo-set is already written, the import is *from it* to virtual utxo-set

someone235 (approved):

PR #191, opened 2023-06-02
complete get_sync_info call

D-Stacks:


consensus/src/consensus/mod.rs
```diff
@@ -395,9 +395,17 @@ impl ConsensusApi for Consensus {
...
+            // we use the history root in archival cases.
+            return self.pruning_point_store.read().history_root().unwrap();
+        }
+        self.pruning_point_store.read().pruning_point().unwrap()
+    }
+
     fn get_sync_info(&self) -> SyncInfo {
-        // TODO: actually get those numbers
-        SyncInfo::default()
+        let count = self.get_virtual_daa_score() - self.get_header(self.get_source()).unwrap().daa_score;
```
michaelsutton, line 407: Perhaps rename to `estimated_count` and add a comment explaining that this might be off by the number of non-DAA blocks since `source` + it does not count all the headers comprising the pruning proof (roughly 14k on current mainnet)

D-Stacks: done, I also renamed the whole thing to block count, which 1) is also used in rpc core. 2) seems much more appropriate (as it doesn't actually give much sync info).

michaelsutton (approved):

Issue #132, opened 2023-02-21
Make sure event processor shutdown signal is processed in order


michaelsutton:
See https://github.com/kaspanet/rusty-kaspa/pull/131
And https://github.com/kaspanet/rusty-kaspa/pull/131#issuecomment-1438336145

PR #197, opened 2023-06-09
Fix sync issues. 

D-Stacks:
This Pr fixes sync issues between core, consensus, the utxoindex and within the notifier. 

1) Notifiers stop in order of hierarchy i.e. stop collectors => stop broadcasters => ~~downgrade listener's senders~~
2) Notifier's collectors / broadcasters stop via closing their receivers (not pure select! primitives), and wait for the que to finish. 
3) Consensus stop waits for all processors to exit 
4) AsyncRuntime blocks on calling stop, stops in order of binding. 
5) Pass core `running` atomic to index processor
6) indexer processor only exits with an empty que, if core is closed, and closes own receiver. 

This pr does not aim to fix stopping sync issues in the RPC realm, as select! exits are still present here, but besides for the use-case of signaling a shutdown back to the client, it may not really be necessary. 

Closes #131 #132


michaelsutton:
Handled by #202

PR #198, opened 2023-06-09
pruning notifications (+ tests)

D-Stacks:
This is required for the txindex, and possibly future external services, in order to sync with the "on-the-fly" pruning mechanism.

currently this is a draft proposal because:

1.  need define good batch sizes for pruned data (or we just leave it at the current "immediate" notify mechanic).
2.  no working pruning tests are currently available to test the implementation.


D-Stacks:
closing this one, outdated.

PR #200, opened 2023-06-10
Network Id

tiram88:
Adds a `NetworkId` type wrapping a `NetworkType` and an optional suffix, displaying a network name expressed as `kaspa-<NetworkType>[-<suffix>]`.

The suffix is normally used in testnet (currently running testnet 10, with network name `kaspa-testnet-10`).

This PR solves a bug in the gRPC implementation of API method `GetBlockDagInfo` reporting a network name as a string which could not be parsed as a network type when trying to convert back into the RPC core corresponding structure.


consensus/core/src/networktype.rs
```diff
@@ -104,3 +97,137 @@ impl Display for NetworkType {
...
+    #[error(transparent)]
+    InvalidNetworkType(#[from] NetworkTypeError),
+
+    #[error("Invalid network suffix: {0}. Only 32 bits unsigned integer (u32) are supported.")]
+    InvalidSuffix(String),
+}
+
+#[derive(Clone, Copy, Debug, Serialize, Deserialize, BorshSerialize, BorshDeserialize, BorshSchema, PartialEq, Eq)]
+#[serde(rename_all = "lowercase")]
+pub struct NetworkInstance {
```
michaelsutton, line 115: I think `NetworkId` might be a better name

consensus/core/src/networktype.rs
```diff
@@ -104,3 +97,124 @@ impl Display for NetworkType {
...
+impl From<NetworkId> for Prefix {
+    fn from(net: NetworkId) -> Self {
+        (*net).into()
+    }
+}
+
+impl FromStr for NetworkId {
+    type Err = NetworkIdError;
+    fn from_str(network_name: &str) -> Result<Self, Self::Err> {
+        let mut parts = network_name.split('-');
```
michaelsutton, line 157: maybe add `split('-').fuse()` to make it explicit that you might call the iterator multiple times also after returning `None`

consensus/core/src/networktype.rs
```diff
@@ -104,3 +97,124 @@ impl Display for NetworkType {
...
+        let prefix = parts.next().unwrap_or_default();
+        if prefix != "kaspa" {
+            return Err(NetworkIdError::InvalidPrefix(prefix.to_string()));
+        }
+        let network_type = NetworkType::from_str(parts.next().unwrap_or_default())?;
+        let suffix = parts.next().map(|x| u32::from_str(x).map_err(|_| NetworkIdError::InvalidSuffix(x.to_string()))).transpose()?;
+        Ok(Self::new(network_type, suffix))
```
michaelsutton, line 164: at this point, if `parts.next()` is still `Some` you should also return an error

consensus/core/src/networktype.rs
```diff
@@ -104,3 +97,124 @@ impl Display for NetworkType {
...
+
+#[derive(Clone, Copy, Debug, Serialize, Deserialize, BorshSerialize, BorshDeserialize, BorshSchema, PartialEq, Eq)]
+#[serde(rename_all = "lowercase")]
+pub struct NetworkId {
+    pub network_type: NetworkType,
+    pub suffix: Option<u32>,
+}
+
+impl NetworkId {
+    pub fn new(network_type: NetworkType, suffix: Option<u32>) -> Self {
```
michaelsutton, line 121: I'd call it `with_prefix` and make `new(network_type: NetworkType)` which defaults to no prefix. Also make both of them const and use them when initiating MAINNET/DEVNET params etc

consensus/core/src/networktype.rs
michaelsutton: (`fuse` is  a no op for already fused iterators)

consensus/core/src/networktype.rs
```diff
@@ -104,3 +97,124 @@ impl Display for NetworkType {
...
+        }
+        let network_type = NetworkType::from_str(parts.next().unwrap_or_default())?;
+        let suffix = parts.next().map(|x| u32::from_str(x).map_err(|_| NetworkIdError::InvalidSuffix(x.to_string()))).transpose()?;
+        Ok(Self::new(network_type, suffix))
+    }
+}
+
+impl Display for NetworkId {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        f.write_str(&self.name())
```
michaelsutton, line 170: Ideally you implement the name logic here (w/o String alloc), and in `name()` you just call `to_string`

consensus/core/src/networktype.rs
```diff
@@ -104,3 +97,124 @@ impl Display for NetworkType {
...
+    fn test_network_id_parse() {
+        struct Test {
+            name: &'static str,
+            expr: &'static str,
+            expected: Result<NetworkId, NetworkIdError>,
+        }
+
+        let tests = vec![
+            Test { name: "Valid mainnet", expr: "kaspa-mainnet", expected: Ok(NetworkId::new(NetworkType::Mainnet, None)) },
+            Test { name: "Valid testnet", expr: "kaspa-testnet-88", expected: Ok(NetworkId::new(NetworkType::Testnet, Some(88))) },
```
michaelsutton, line 200: after fixing the above, add a test with `kaspa-testnet-88-foo`

michaelsutton (changes requested):

consensus/core/src/networktype.rs
tiram88: Indeed! Nice catch :)

consensus/core/src/networktype.rs
tiram88: You mean `with_suffix`, right?

michaelsutton: yes sure

michaelsutton (approved):

PR #201, opened 2023-06-10
Proper gRPC Server shutdown

tiram88:


rpc/grpc/server/src/connection.rs
```diff
@@ -57,46 +541,66 @@ impl Connection for GrpcConnection {
...
-    pub fn register(&mut self, address: SocketAddr, sender: GrpcSender) -> ListenerId {
-        let connection = GrpcConnection::new(address, sender);
-        let id = self.notifier.clone().register_new_listener(connection.clone());
-        trace!("registering a new gRPC connection from: {address} with listener id {id}");
+    pub fn register(&self, connection: GrpcConnection) {
+        debug!("gRPC: Register a new connection from {connection}");
+        let prev_connection = self.connections.lock().insert(connection.identity(), connection);
+        if let Some(prev) = prev_connection {
+            if !prev.is_closed() {
+                prev.close();
```
michaelsutton, line 581: why not just call `close()` w/o checking, since duplicate calls are handled within

rpc/grpc/server/src/connection.rs
```diff
@@ -57,46 +541,66 @@ impl Connection for GrpcConnection {
...
-        self.connections.insert(address, connection);
-        id
+    pub fn is_full(&self) -> bool {
+        self.connections.lock().len() >= self.max_connections
+    }
+
+    pub fn unregister(&self, net_address: SocketAddr) {
+        if let Some(connection) = self.connections.lock().remove(&net_address) {
+            debug!("gRPC: Unregister the gRPC connection from {connection}");
+            if !connection.is_closed() {
```
michaelsutton, line 593: same

rpc/grpc/core/src/ext/kaspad.rs
```diff
@@ -11,6 +11,14 @@ impl KaspadRequest {
     pub fn from_notification_type(scope: &Scope, command: Command) -> Self {
         KaspadRequest { id: 0, payload: Some(kaspad_request::Payload::from_notification_type(scope, command)) }
     }
+
+    pub fn is_subscription(&self) -> bool {
+        if let Some(ref payload) = self.payload {
```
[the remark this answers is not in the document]
tiram88: Very cool!

You stand really always at the forefront of novelty with Rust.

rpc/grpc/server/src/connection.rs
tiram88: According to the implemented logic, `close()` is signaling shutdown, triggering the exit of the connection receive loop which itself calls `unregister()`. At this point, the connection is considered closed (`is_close()` is returning true). So calling close again would yield an extra log "gRPC: Connection close was called more than once, client-id ..." while this code path is the normal case.

michaelsutton: I'm not sure the log line should be there if it makes you avoid the path :)

michaelsutton (approved):

PR #202, opened 2023-06-14
Graceful shutdown sequence

michaelsutton:
Implements a controlled shutdown sequence of the notification system, starting from the notification root (consensus) and up through the various index and rpc services. 

Fixes P2P cyclic reference clean up, 

Inspired and applies lessons learned from #197 by @D-Stacks. Adapts Connection Manager shutdown from @someone235's daemon branch.    

Closes #131 #132.


tiram88 (approved):
Approved based on discussing with @michaelsutton.

I might come back later, after merge, if something crops up during a more formal review.

PR #209, opened 2023-06-22
Add P2P reject message support 

michaelsutton:
Sends an explicit reject message on P2P protocol errors for notifying the peer about the disconnect reason and for easier tracing of logical bugs causing protocol errors.

### DIFF
--- a/protocol/p2p/src/core/router.rs
+++ b/protocol/p2p/src/core/router.rs
@@ -1,8 +1,9 @@
 use crate::core::hub::HubEvent;
-use crate::pb::KaspadMessage;
-use crate::Peer;
+use crate::pb::RejectMessage;
+use crate::pb::{kaspad_message::Payload as KaspadMessagePayload, KaspadMessage};
 use crate::{common::ProtocolError, KaspadMessagePayloadType};
-use kaspa_core::{debug, error, info, trace};
+use crate::{make_message, Peer};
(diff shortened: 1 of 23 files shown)


utils/src/any.rs
```diff
@@ -0,0 +1,30 @@
+/// Provides a best-effort short type name
+pub fn type_name_short<T: ?Sized>() -> &'static str {
+    let s = std::any::type_name::<T>();
+    const SEP: &str = "::";
+    if s.find('<').is_some() {
+        // T is generic so would require more sophisticated processing
+        return s;
+    }
+    match s.rfind(SEP) {
```
tiram88, line 9: Perhaps remove the "Flow" suffix too, when present?

michaelsutton: This is a generic util method, so if done it should be on the `flow_trait` level. And perhaps we can convert the `CamelCaseText` to human readable "Camal case text". Let's leave it for another time.

protocol/p2p/src/core/router.rs
```diff
@@ -344,3 +367,26 @@ impl Router {
         true
     }
 }
+
+fn match_for_io_error(err_status: &tonic::Status) -> Option<&std::io::Error> {
```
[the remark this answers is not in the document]
michaelsutton: I agree on the conclusion :)

tiram88 (approved):

PR #217, opened 2023-07-03
Fixes 2 RPC bugs

tiram88:
- Fixes wRPC server direct subscriptions to notifications from the hosting node
- Fixes the gRPC client so that it handles the golang version of the network type in the response to `GetCurrentNetwork`

### DIFF
--- a/rpc/grpc/core/src/convert/message.rs
+++ b/rpc/grpc/core/src/convert/message.rs
@@ -474,7 +474,10 @@ try_from!(&protowire::NotifyNewBlockTemplateResponseMessage, RpcResult<kaspa_rpc
 
 try_from!(&protowire::GetCurrentNetworkRequestMessage, kaspa_rpc_core::GetCurrentNetworkRequest);
 try_from!(item: &protowire::GetCurrentNetworkResponseMessage, RpcResult<kaspa_rpc_core::GetCurrentNetworkResponse>, {
-    Self { network: RpcNetworkType::from_str(&item.current_network)? }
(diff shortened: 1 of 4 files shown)


rpc/grpc/core/src/convert/message.rs
```diff
@@ -474,7 +474,10 @@ try_from!(&protowire::NotifyNewBlockTemplateResponseMessage, RpcResult<kaspa_rpc
 
 try_from!(&protowire::GetCurrentNetworkRequestMessage, kaspa_rpc_core::GetCurrentNetworkRequest);
 try_from!(item: &protowire::GetCurrentNetworkResponseMessage, RpcResult<kaspa_rpc_core::GetCurrentNetworkResponse>, {
-    Self { network: RpcNetworkType::from_str(&item.current_network)? }
+    // Note that current_network is first converted to lowercase because the golang implementation
+    // returns a "human readable" version with a capital first letter while the rusty version
+    // is fully lowercase.
+    Self { network: RpcNetworkType::from_str(&item.current_network.to_lowercase())? }
```
michaelsutton, line 480: Maybe we should call `to_lowercase` within `from_str` to cover more future cases?

michaelsutton: if you move it to `from_str` perhaps just do
```
...
match network_type {
      "mainnet" | "Mainnet" => Ok(NetworkType::Mainnet),
...
```  
to save the String alloc

michaelsutton (approved):

PR #220, opened 2023-07-09
Weight ip prefix

D-Stacks:
This pr weights the prefix bytes in the address manager for the `iterate_prioritized_random_addresses` method. which should more evenly spread-out the network topology as well as increasing Sybil resistance. 

The improved selection will adjust weights by dividing the weights by the size of the prefix bucket the ip belongs to. this ensures other weighting metrics are still respected, but keep a priority of selecting evenly over all prefix bytes.

### DIFF
--- a/components/addressmanager/src/lib.rs
+++ b/components/addressmanager/src/lib.rs
@@ -204,7 +204,7 @@ mod address_store_with_cache {
 
             // Divide weights by size of bucket of the prefix bytes, to normalize distribution over global network.
             for (i, address) in filtered_addresses.iter().enumerate() {
-                weights[i] = weights[i] / (*prefix_counter.get_mut(&address.prefix_bytes()).unwrap() as f64);
+                *weights.get_mut(i).unwrap() /= *prefix_counter.get_mut(&address.prefix_bytes()).unwrap() as f64;
             }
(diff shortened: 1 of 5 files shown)


components/addressmanager/src/lib.rs
```diff
@@ -190,14 +190,24 @@ mod address_store_with_cache {
...
+                    let count = prefix_counter.entry(e.address.clone().prefix_bytes()).or_insert(0);
+                    *count += 1;
+                    (64f64.powf((MAX_CONNECTION_FAILED_COUNT + 1 - e.connection_failed_count) as f64), e.address)
+                })
                 .unzip();
 
-            RandomWeightedIterator::new(weights, addresses)
+            // Divide weights by size of bucket of the prefix bytes, to normalize distribution over global network.
+            for (i, address) in filtered_addresses.iter().enumerate() {
+                *weights.get_mut(i).unwrap() /= *prefix_counter.get_mut(&address.prefix_bytes()).unwrap() as f64;
```
michaelsutton, line 207: `prefix_counter.get_mut` can be a non-mut get: `prefix_counter.get`

michaelsutton (approved):

PR #222, opened 2023-07-10
Add base implementation of local address management

coderofstuff:
- If we can add local addresses, this assumes the first one is always the best
- If `--listen` is passed, it is added as the local address, if it's routable
- Adds the `externalip` flag
  - If an external IP is passed, it is assumed to be the best one to use
  - The value is added as a local address, if it's routable
- Passes the "best local IP address" to the version message
- If neither `--listen` nor `--externalip` is passed (most common default), then all interfaces are checked for local addresses that are routable and adds them to the list

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -146,6 +146,7 @@ wasm-bindgen-futures = "0.4.33"
 js-sys = "0.3.56"
 getrandom = { version = "0.2.8", features = ["js"] }
 uuid = { version = "1.2.2", features = ["v4", "fast-rng", "serde"] }
+getaddrs = "0.1.0"
 
 # bip32 dependencies
 rand_core = { version = "0.6", features = ["std"] }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -146,7 +146,6 @@ wasm-bindgen-futures = "0.4.33"
 js-sys = "0.3.56"
 getrandom = { version = "0.2.8", features = ["js"] }
 uuid = { version = "1.2.2", features = ["v4", "fast-rng", "serde"] }
-getaddrs = "0.1.0"
 
 # bip32 dependencies
 rand_core = { version = "0.6", features = ["std"] }
--- a/components/addressmanager/src/lib.rs
+++ b/components/addressmanager/src/lib.rs
@@ -54,7 +54,7 @@ impl AddressManager {
             None => {
                 // If listen_address === 0.0.0.0, bind all interfaces
                 // else, bind whatever was passed as listen address (if routable)
-                let listen_address = self.config.listen.normalize(self.config.default_p2p_port());
+                let listen_address = self.config.p2p_listen_address.normalize(self.config.default_p2p_port());
 
                 if listen_address.ip.is_unspecified() {
                     let network_interfaces = list_afinet_netifas();
--- a/components/addressmanager/src/lib.rs
+++ b/components/addressmanager/src/lib.rs
@@ -4,13 +4,13 @@ extern crate self as address_manager;
 
 use std::{collections::HashSet, sync::Arc};
 
+use getaddrs::{if_flags, InterfaceAddrs};
 use itertools::Itertools;
-use kaspa_core::{debug, time::unix_now, info};
 use kaspa_consensus_core::config::Config;
+use kaspa_core::{debug, info, time::unix_now};
 use kaspa_database::prelude::{StoreResultExtensions, DB};
 use kaspa_utils::networking::IpAddress;
 use parking_lot::Mutex;
-use getaddrs::{InterfaceAddrs, if_flags};
 
 use stores::banned_address_store::{BannedAddressesStore, BannedAddressesStoreReader, ConnectionBanTimestamp, DbBannedAddressesStore};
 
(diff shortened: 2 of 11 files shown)


Cargo.toml
```diff
@@ -146,6 +146,7 @@ wasm-bindgen-futures = "0.4.33"
 js-sys = "0.3.56"
 getrandom = { version = "0.2.8", features = ["js"] }
 uuid = { version = "1.2.2", features = ["v4", "fast-rng", "serde"] }
+getaddrs = "0.1.0"
```
elichai, line 149: `getaddrs` has a single release 6 years ago and their github link points no where

coderofstuff: @elichai I've removed `getaddrs` in the latest commit and replaced it with `local_ip_address` which is maintained.

Crate: https://docs.rs/local-ip-address/latest/local_ip_address/index.html
GitHub: https://github.com/EstebanBorai/local-ip-address

In addition, since the only user of this package would be kaspa-addressmanager, I've only added the crate in this package as opposed to adding it to the workspace.

components/addressmanager/src/lib.rs
```diff
@@ -20,14 +22,78 @@ const MAX_CONNECTION_FAILED_COUNT: u64 = 3;
...
+                            // TODO: Add Check IPv4 or IPv6 match from Go code
+                            if curr_ip.is_publicly_routable() {
+                                info!("Publicly routable local address {} added to store", curr_ip);
+                                self.local_net_addresses.push(NetAddress { ip: curr_ip, port: self.config.default_p2p_port() });
+                            } else {
+                                info!("Non-publicly routable interface address {} not added to store", curr_ip);
+                            }
+                        }
+                    } else {
+                        println!("Error getting network interfaces: {:?}", network_interfaces);
```
michaelsutton, line 76: `println` -> `error` (import from `kaspa_core`)

components/addressmanager/src/lib.rs
```diff
@@ -20,14 +22,78 @@ const MAX_CONNECTION_FAILED_COUNT: u64 = 3;
...
+                    }
+                } else if listen_address.ip.is_publicly_routable() {
+                    info!("Publicly routable local address {} added to store", listen_address.ip);
```
michaelsutton, line 79:
```suggestion
                    info!("Publicly routable P2P listen address {} added to store", listen_address);
```

components/addressmanager/src/lib.rs
```diff
@@ -20,14 +22,78 @@ const MAX_CONNECTION_FAILED_COUNT: u64 = 3;
...
     address_store: address_store_with_cache::Store,
+    config: Arc<Config>,
+    local_net_addresses: Vec<NetAddress>,
 }
 
 impl AddressManager {
-    pub fn new(db: Arc<DB>) -> Arc<Mutex<Self>> {
-        Arc::new(Mutex::new(Self {
+    pub fn new(config: Arc<Config>, db: Arc<DB>) -> Arc<Mutex<Self>> {
+        // TODO: do the things that local_address_manager would do here
```
michaelsutton, line 31: TODO seems obsolete

components/addressmanager/src/lib.rs
```diff
@@ -20,14 +22,78 @@ const MAX_CONNECTION_FAILED_COUNT: u64 = 3;
...
+                // An external IP was passed, we will try to bind that if it's valid
+                if local_net_address.is_publicly_routable() {
+                    info!("External ip {} added to store", local_net_address);
+                    self.local_net_addresses.push(NetAddress { ip: local_net_address, port: self.config.default_p2p_port() });
+                } else {
+                    info!("Non-publicly routable external ip {} not added to store", local_net_address);
+                }
+            }
+            None => {
+                // If localNetAddress === 0.0.0.0, bind all interfaces
```
michaelsutton, line 56: -> If listen_address === 0.0.0.0

components/addressmanager/src/lib.rs
```diff
@@ -20,14 +22,78 @@ const MAX_CONNECTION_FAILED_COUNT: u64 = 3;
...
+                }
+            }
+        }
+    }
+
+    pub fn best_local_address(&mut self) -> Option<NetAddress> {
+        if self.local_net_addresses.is_empty() {
+            None
+        } else {
+            // TODO: Add logic for finding the best.
```
michaelsutton, line 92: -> TODO: Add logic for finding the best as a function of a peer remote address.

components/addressmanager/src/lib.rs
```diff
@@ -20,14 +22,78 @@ const MAX_CONNECTION_FAILED_COUNT: u64 = 3;
...
+            // for now, putting in stub that allows it to work if externalip
```
michaelsutton, line 93: -> for now, returning the first one

consensus/core/src/config/mod.rs
```diff
@@ -46,6 +48,11 @@ pub struct Config {
     pub enable_mainnet_mining: bool,
 
     pub user_agent_comments: Vec<String>,
+
+    // If undefined, sets it to 0.0.0.0
+    pub listen: ContextualNetAddress,
```
michaelsutton, line 53: I think `listen` is out of context here. Should be `p2p_listen_address` maybe?

protocol/flows/src/flow_context.rs
```diff
@@ -507,8 +509,16 @@ impl ConnectionInitializer for FlowContext {
...
-        if router.is_outbound() {
-            self.address_manager.lock().add_address(router.net_address().into());
+        if router.is_outbound() || peer_version.address.is_some() {
+            let mut address_manager = self.address_manager.lock();
+
+            if router.is_outbound() {
+                address_manager.add_address(router.net_address().into());
+            }
+
+            if peer_version.address.is_some() {
```
michaelsutton, line 519: can use `if let Some(..) = peer_version.address { .. }` here

utils/src/networking.rs
```diff
@@ -333,4 +371,123 @@ mod tests {
         let addr_v6 = NetAddress::from_str("[2a01:4f8:191:1143::2]:5678");
         assert!(addr_v6.is_ok());
     }
+
+    #[test]
+    fn test_is_publicly_routable() {
+        // RFC 2544 tests
+        assert_eq!(false, IpAddress::from_str("198.18.0.0").unwrap().is_publicly_routable());
```
michaelsutton, line 378: clippy warns that assert should be used instead of assert_eq with a literal 

for all `assert_eq!(false, ..` -> `assert!(!..`
for all `assert_eq!(true, ..` -> `assert!(..`

components/addressmanager/src/lib.rs
coderofstuff: I changed the message to `Publicly routable P2P listen address ` but I left the variable as `listen_address.ip` to be consistent with the rest of the messages along these lines of "routable" related logs. All of the other ones log the IP address.

michaelsutton (approved):

michaelsutton:
We should consider using `IpAddr::is_global` within `is_publicly_routable`

PR #226, opened 2023-07-18
Enhance Mempool performance

tiram88:
This PR addresses the issue of the mempool being a bottleneck in the first testnet-11 experiment @ 10 BPS.

The previous monolithical design did lock both the `Mempool` and to some extent the virtual processor during every call to its public functions, through the `Manager` that owns it, leading to some long delays. **The new general design is to split the functions into simpler atomic steps logically protected by a lock on the mempool when the latter is implied**. This approach has following consequences:

- The `Manager` owning the `Mempool` instance no longer is a simple pass-through locking and exposing mempool functions publicly. An important part of the mempool logic gets moved into the manager.
- Some verification steps must sometimes be duplicated, particularly checking for double spends, since executing a `Manager` function is no longer atomic.
- Some processes are reorganized and some behaviors do change, taking advantage of the potential this new design offers.

Another performance improvements are brought:

- Processing batches of transaction in a topological order, forming sub-batches by levels of chained dependency
- Parallel validation of sub-batches of transactions into the virtual processor, processing those in chunks formed up to a maximal mass so the virtual processor does lock for too long,
- A more efficient algorithm for making room in the mempool when it is full and some transaction must be added. This process also has a modified behavior: the transaction with the lowest fee rate is removed, but (this is new) only among transactions having no chained dependency and not being a parent of the transaction getting added.

### Validate and insert a single transaction

In order to bring a much finer lock granularity the function of validating and inserting a transaction into the mempool is split in 4 steps:

1. pre-validation (read lock on `Mempool`)
2. validation by the virtual processor (no lock on `Mempool`)
3. post-validation and insertion (write lock on `Mempool`)
4. validation and insertion of unorphaned transactions (see below)

### Validate and insert unorphaned transactions

Following any insert of a transaction into the mempool, some orphan transactions may be unorphaned. Here again we take advantage of the new design. On insertions, room is made on the fly if necessary and then we keep looping as long as new transactions are unorphaned (both new behaviors).

1. parallel validation of batches of transactions (no lock)
2. post-validation and insertion (write lock)
3. in case of new unorphaned transactions, loop to step 1 (new behavior)

### Validate and insert transactions in batch

This function is involved in the transaction relay flow, where a bunch of transactions gets broadcasted to peers. The batch is processed in a topological order by level of chained dependency (new behavior). For each level:

1. pre-validation (read lock)
2. parallel validation of the sub-batch of transactions by the virtual processor (no lock)
3. post-validation and insertion (write lock)
4. validation and insertion of unorphaned transactions

### Handle the transactions of a block newly added to the DAG

Here again, a split in 3 steps adds granularity in the locks.

1. handling of transactions (write lock),
2. validation and insertion of unorphaned transactions
3. expiring the low level transactions (write lock)

### Revalidate high priority transactions

This is probably the most important bottleneck alleviation, together with tx relay broadcasting. This process occurs only every 30 seconds and for a node getting a very high rate of local transactions, it was implying a _very_ computer intensive processing all under a single atomic lock of the mempool.

This gets redesigned as:

1. getting all the high priority transactions (read lock)
2. processing the batch in a topological order by level of chained dependency, instead of a "standard" Kahn's in-degree algorithm (new behavior)
3. populating all transactions with UTXO entries found in the mempool (read lock)
4. parallel validation of the sub-batch of transactions by the virtual processor (no lock)
5. updating the transactions in the mempool and removing the invalid ones, along with all their redeemers (write lock)

Since the process is now much more granular in how it locks both the mempool and the consensus, it may happen that some high priority transactions present at the beginning got removed (mined in a block, invalidated, etc.) during the (relatively long) process. The transaction id may or may not be returned as accepted depending on the sequence of events. But no impact is expected on the node. It will simply try to rebroadcast a transaction id that is no longer present and the peers requesting it will eventually get a `TransactionNotFound`, which may and does occur anyway for other reasons.

The behavior changes compared to the golang version in that on transaction validation error the error is simply logged and the execution continues whereas in golang the execution halts returning the error.

## Performance gains

Cautious note: an upcoming PR will add a benchmarking infrastructure making it possible to measure the gains in performance, hence revealing to which extent this new design actually alleviates the bottleneck issue.

Depending on the measures, some further adjustments might be needed.


protocol/flows/src/flow_context.rs
```diff
@@ -397,6 +397,22 @@ impl FlowContext {
...
+            // The TransactionSpread instance ensures at most one expire running at any
+            // given time.
+            let mining_manager = self.mining_manager().clone();
+            let consensus_clone = consensus.clone();
+            let context = self.clone();
+            tokio::spawn(async move {
+                tokio::spawn(async move {
+                    mining_manager.expire_low_priority_transactions(&consensus_clone).await;
+                });
+                context.expire_done().await;
```
michaelsutton, line 412: Due to the double `tokio::spawn` this might run before `expire_low_priority_transactions` executes. Is that the intended behavior?

mining/src/monitor.rs
```diff
@@ -0,0 +1,94 @@
...
+
+            // Subtract the snapshots
+            let delta = &snapshot - &last_snapshot;
+            let now = Instant::now();
+            let elapsed = (now - last_log_time).as_secs_f64();
+
+            info!("Processed {} unique transactions in the last {:.2}s ({:.2} avg txs/s, in: {} via RPC, {} via P2P, out: {} via accepted blocks, {:.2}% e-tps)",
+                delta.tx_accepted_counts,
+                elapsed,
+                delta.tx_accepted_counts as f64 / elapsed,
```
michaelsutton, line 56: it feels like the log line has some repeating redundant info and is a little too long. I think `Processed {} unique transactions in the last {:.2}s` can be removed, wdyt? `avg txs/s` already contains this info, and the consensus monitor already prints raw tx counts 

I suggest the line looks like: `[Unique tx stats] {:.2} avg txs/s, in: {} via RPC, {} via P2P, out: {} via accepted blocks, {:.2}% e-tps`

mining/src/monitor.rs
```diff
@@ -0,0 +1,94 @@
...
+                delta.high_priority_tx_counts,
+                delta.low_priority_tx_counts,
+                delta.block_tx_counts,
```
michaelsutton, line 59: I see in `handle_new_block_txs` that `block_tx_counts` counts all txs, but I think you should print here only the unique/accepted count, bcs `in:` counts unique ones, so `out:` should reflect the same semantics

mining/src/lib.rs
```diff
@@ -1,10 +1,94 @@
...
 #[cfg(test)]
 pub mod testutils;
+
+#[derive(Default)]
+pub struct MiningCounters {
+    pub high_priority_tx_counts: AtomicU64,
+    pub low_priority_tx_counts: AtomicU64,
+    pub block_tx_counts: AtomicU64,
+    pub tx_accepted_counts: AtomicU64,
+    pub input_counts: AtomicU64,
```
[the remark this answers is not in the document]
tiram88: They are in the next commit. We should discuss what to keep/reformulate, I fully agree.

mining/src/monitor.rs
tiram88: Nice :)

But we still need the elapsed time somewhere. Consensus and Mining monitors are not synced. The log lines are switching order sometimes, their respective schedule may drift differently as the node runs for a long time and when grepping the log for this line, we loose the elapsed info altogether.

protocol/flows/src/flowcontext/transactions.rs
```diff
@@ -10,34 +10,62 @@ use kaspa_p2p_lib::{
...
-        let now = Instant::now();
-        if now - self.last_rebroadcast_time < REBROADCAST_INTERVAL {
+    /// Returns true if the time has come for running the task cleaning mempool transactions
+    /// and if so, mark the task as running.
+    pub fn should_run_cleaning_task(&mut self) -> bool {
+        if self.cleaning_task_running || Instant::now() < self.last_cleaning_time + CLEANING_TASK_INTERVAL {
             return false;
         }
-        self.last_rebroadcast_time = now;
+        // Keep the launching times aligned to exact intervals
```
michaelsutton, line 45: why do we need this alignment to exact intervals? this results in the weird while loop below for summing up time

tiram88: This is just me being a bit maniac with time precision and finding no better way to do calculations with `Instant`.

We can of course get back to the usual strategy of storing last time = now.

michaelsutton: I found a way to keep the alignment with a formula

michaelsutton (approved):

PR #229, opened 2023-07-20
Support a persistent channel connection in the notification system

tiram88:
This small PR covers a special use case required by the wRPC client where the listeners to the wRPC client do manage the life-cycle of their notification channel so need that the notification system do not close said channel.

### DIFF
--- a/notify/src/connection.rs
+++ b/notify/src/connection.rs
@@ -18,19 +18,31 @@ pub trait Connection: Clone + Debug + Send + Sync + 'static {
 }
 
 #[derive(Clone, Debug)]
+pub enum ChannelType {
+    Closable,
+    Persistent,
+}
+
+#[derive(Clone, Debug)]
 pub struct ChannelConnection<N>
 where
     N: Notification,
 {
     sender: Sender<N>,
+    channel_type: ChannelType,
 }
 
 impl<N> ChannelConnection<N>
 where
     N: Notification,
 {
-    pub fn new(sender: Sender<N>) -> Self {
-        Self { sender }
+    pub fn new(sender: Sender<N>, channel_type: ChannelType) -> Self {
+        Self { sender, channel_type }
+    }
+
+    /// Close the connection, ignoring the channel type
+    pub fn force_close(&self) -> bool {
+        self.sender.close()
     }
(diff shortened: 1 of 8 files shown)


notify/src/connection.rs
```diff
@@ -17,20 +17,32 @@ pub trait Connection: Clone + Debug + Send + Sync + 'static {
...
+    channel_type: ChannelType,
 }
 
 impl<N> ChannelConnection<N>
 where
     N: Notification,
 {
-    pub fn new(sender: Sender<N>) -> Self {
-        Self { sender }
+    pub fn new(sender: Sender<N>, channel_type: ChannelType) -> Self {
```
michaelsutton, line 39: since this seems like a temporary patch, I'd prefer that the usual `new` remains like before (defaulting to `closable`) and that we add a special `new_persistent` constructor

aspect: This is exactly what we want to avoid.  This behavior has to be explicit as this hides the problem under the rug.  Any developer supplying a channel should expect the channel to remain alive for the lifetime the developer expects it to have.  Handing a channel to an external API and then having it closed is quite unexpected (at least in my experience).  If an API wants to control the lifetime of the channel, I would expect it to provide one. If I am providing one, I expect to be in control.

notify/src/connection.rs
```diff
@@ -17,20 +17,32 @@ pub trait Connection: Clone + Debug + Send + Sync + 'static {
...
+        Self { sender, channel_type }
+    }
+
+    /// Close the connection, ignoring the channel type
+    pub fn force_close(&self) -> bool {
```
biryukovmaxim, line 44: Is this method should be public?
If it uses generic with PhantomData instead of Enum it will be possible to implement this method only for Closable ChannelConnection

notify/src/connection.rs
michaelsutton: ok

notify/src/connection.rs
michaelsutton: I think `force_close` will actually be used in the `Persistent` case

michaelsutton (approved):

PR #232, opened 2023-07-25
ScriptPubKey human-readable ser/deser (for json)

aspect:
This PR implements ScriptPubKey serialization/deserialization for JSON.

### DIFF
--- a/consensus/core/src/tx.rs
+++ b/consensus/core/src/tx.rs
@@ -43,6 +43,7 @@ pub struct ScriptPublicKey {
 
 #[derive(Default, Debug, PartialEq, Eq, Serialize, Deserialize, Clone, Hash)]
 #[serde(rename_all = "camelCase")]
+#[serde(rename = "ScriptPublicKey")]
 struct ScriptPublicKeyInternal<'a> {
     version: ScriptPublicKeyVersion,
     script: &'a [u8],
@@ -57,7 +58,7 @@ impl Serialize for ScriptPublicKey {
             let mut hex = [0u8; SCRIPT_VECTOR_SIZE * 2 + 4];
             faster_hex::hex_encode(&self.version.to_be_bytes(), &mut hex).expect("Unable to serialize ScriptPubKey version");
             faster_hex::hex_encode(&self.script, &mut hex[4..]).expect("Unable to serialize ScriptPubKey script");
-            serializer.serialize_str(str::from_utf8(&hex).expect("hex is must be a valid UTF-8"))
+            serializer.serialize_str(unsafe { str::from_utf8_unchecked(&hex) })
         } else {
             ScriptPublicKeyInternal { version: self.version, script: &self.script }.serialize(serializer)
         }
@@ -70,8 +71,8 @@ impl<'de: 'a, 'a> Deserialize<'de> for ScriptPublicKey {
         D: Deserializer<'de>,
     {
         if deserializer.is_human_readable() {
-            let s = <std::string::String as Deserialize>::deserialize(deserializer)?;
-            FromStr::from_str(&s).map_err(serde::de::Error::custom)
+            let s = <&str as Deserialize>::deserialize(deserializer)?;
+            FromStr::from_str(s).map_err(serde::de::Error::custom)
         } else {
             ScriptPublicKeyInternal::deserialize(deserializer)
                 .map(|ScriptPublicKeyInternal { script, version }| Self { version, script: SmallVec::from_slice(script) })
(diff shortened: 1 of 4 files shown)


consensus/core/src/tx.rs
```diff
@@ -30,27 +35,57 @@ pub use smallvec::smallvec as scriptvec;
...
+impl Serialize for ScriptPublicKey {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        if serializer.is_human_readable() {
+            let mut hex = [0u8; SCRIPT_VECTOR_SIZE * 2 + 4];
+            faster_hex::hex_encode(&self.version.to_be_bytes(), &mut hex).expect("Unable to serialize ScriptPubKey version");
+            faster_hex::hex_encode(&self.script, &mut hex[4..]).expect("Unable to serialize ScriptPubKey script");
+            serializer.serialize_str(str::from_utf8(&hex).expect("hex is must be a valid UTF-8"))
```
biryukovmaxim, line 60: It's safe to use unsafe from_utf8_unchecked, because hex encoding will never produce non-utf8. Btw have you tried serialize_bytes, it should produce the same outcome?

biryukovmaxim, line 60: `.expect` -> `.map_err(serde::de::Error::custom)`

consensus/core/src/tx.rs
```diff
@@ -30,27 +35,57 @@ pub use smallvec::smallvec as scriptvec;
...
+        } else {
+            ScriptPublicKeyInternal { version: self.version, script: &self.script }.serialize(serializer)
```
biryukovmaxim, line 62:
```
let mut state = serializer.serialize_struct("ScriptPublicKey", 2)?;
            state.serialize_field("version", &self.version)?;
            state.serialize_field("script", self.script.as_ref())?;
            state.end()
 ```
 it can be done without `ScriptPublicKeyInternal`, but I dont think it's a big deal

File: consensus/core/src/tx.rs [hunk 3]
```diff
@@ -30,27 +35,57 @@ pub use smallvec::smallvec as scriptvec;
...
     where
         D: Deserializer<'de>,
     {
-        #[derive(Default, Debug, PartialEq, Eq, Serialize, Deserialize, Clone, Hash)]
-        #[serde(rename_all = "camelCase")]
-        struct ScriptPublicKeyInternal<'a> {
-            version: ScriptPublicKeyVersion,
-            script: &'a [u8],
+        if deserializer.is_human_readable() {
+            let s = <std::string::String as Deserialize>::deserialize(deserializer)?;
```
biryukovmaxim, line 73:
```suggestion
            let s = <&str as Deserialize>::deserialize(deserializer)?;
```
It needs to be tested, I think it can prevent String allocation, but I'm not sure

consensus/core/src/tx.rs
```diff
@@ -30,27 +35,57 @@ pub use smallvec::smallvec as scriptvec;
...
-#[serde(rename_all = "camelCase")]
+#[derive(Default, Debug, PartialEq, Eq, Clone, Hash)]
 pub struct ScriptPublicKey {
     version: ScriptPublicKeyVersion,
     script: ScriptVec, // Kept private to preserve read-only semantics
 }
 
+#[derive(Default, Debug, PartialEq, Eq, Serialize, Deserialize, Clone, Hash)]
+#[serde(rename_all = "camelCase")]
+struct ScriptPublicKeyInternal<'a> {
```
biryukovmaxim, line 46:
```suggestion
#[derive(Default, Debug, PartialEq, Eq, Serialize, Deserialize, Clone, Hash)]
#[serde(rename_all = "camelCase")]
#[serde(rename = "ScriptPublicKey")]
struct ScriptPublicKeyInternal<'a> {
```

Could you add this renaming? It was my fault. I think without renaming it xml and other structure-name sensitive formats can work incorrectly

consensus/core/src/tx.rs
aspect: replaced

aspect: done

consensus/core/src/tx.rs
aspect: let's keep it as-is

consensus/core/src/tx.rs
aspect: fixed

michaelsutton (approved):

PR #234, opened 2023-07-28
Rocksdb config

biryukovmaxim:
implement rocksdb connect builder that  allows to set different settings

### DIFF
--- a/database/src/db/conn_builder.rs
+++ b/database/src/db/conn_builder.rs
@@ -1,0 +1,168 @@
+use crate::db::DB;
+use rlimit::Resource;
+use std::cmp::min;
+use std::path::PathBuf;
+use std::sync::Arc;
+
+#[derive(Debug, Copy, Clone)]
+pub struct Unspecified;
+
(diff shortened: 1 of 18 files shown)


database/src/db/conn_builder.rs
aspect, line 1: Perhaps consider destructuring `self` to reduce the footprint? I.e. replacing:
```rust
pub fn with_parallelism(self, parallelism: impl Into<usize>) -> ConnBuilder<Path, STATS_ENABLED, StatsPeriod> {
        ConnBuilder {
            db_path: self.db_path,
            create_if_missing: self.create_if_missing,
            parallelism: parallelism.into(),
            files_limit: self.files_limit,
            mem_budget: self.mem_budget,
            stats_period: self.stats_period,
        }
    }
```
with:
```rust
pub fn with_parallelism(self, parallelism: impl Into<usize>) -> ConnBuilder<Path, STATS_ENABLED, StatsPeriod> {
        ConnBuilder {
            parallelism: parallelism.into(),
            .. self
        }

```

PR #240, opened 2023-08-13
Restrict subscription to blanket `UtxosChange` notification

tiram88:
The subscription to blanket `UtxosChanged` notifications is restricted to unsafe mode only since the notifications yielded are highly resource intensive.

Please note that unsubscribing to blanket `UtxosChanged` is always allowed and cancels the whole subscription no matter if blanket or targeting specified addresses.

This PR documents the [subscriptions processing in the notification system](https://github.com/kaspanet/rusty-kaspa/blob/master/notify/src/subscription/processing.md).

Submitting a transaction allowing orphans is blocked in safe mode.


rpc/service/src/service.rs
```diff
@@ -634,6 +634,18 @@ impl RpcApi for RpcCoreService {
...
+                //
+                // Please note that unsubscribing to blanket UtxosChanged is always allowed and cancels
+                // the whole subscription no matter if blanket or targeting specified addresses.
+
+                warn!("RPC subscription to blanket UtxosChanged called while node in safe RPC mode -- ignoring.");
+                return Err(RpcError::UnavailableInSafeMode);
+            }
+        }
         self.notifier.clone().start_notify(id, scope).await?;
         Ok(())
```
biryukovmaxim, line 650: with `match` it would be less nested

```suggestion
        match scope {
            Scope::UtxosChanged(ref utxos_changed_scope) if !self.config.unsafe_rpc && utxos_changed_scope.addresses.is_empty() => {
                // The subscription to blanket UtxosChanged notifications is restricted to unsafe mode only
                // since the notifications yielded are highly resource intensive.
                //
                // Please note that unsubscribing to blanket UtxosChanged is always allowed and cancels
                // the whole subscription no matter if blanket or targeting specified addresses.

                warn!("RPC subscription to blanket UtxosChanged called while node in safe RPC mode -- ignoring.");
                Err(RpcError::UnavailableInSafeMode)
            }
            _ => self.notifier.clone().start_notify(id, scope).await,
        }
```

rpc/service/src/service.rs
```diff
@@ -634,8 +634,22 @@ impl RpcApi for RpcCoreService {
 
     /// Start sending notifications of some type to a listener.
     async fn start_notify(&self, id: ListenerId, scope: Scope) -> RpcResult<()> {
-        self.notifier.clone().start_notify(id, scope).await?;
-        Ok(())
+        match scope {
```
biryukovmaxim, line 637: with `matches` macro
```
if matches!(scope, Scope::UtxosChanged(ref utxos_changed_scope) if !self.config.unsafe_rpc && utxos_changed_scope.addresses.is_empty()) {...
```

tiram88: Nah, good enough as is.

Though it's always fun, let's stop for now exploring the 50 nuances of Rust syntax :)

michaelsutton:
What do you think about blocking tx submission with `allow_orphan=true` on safe rpc mode? the reasoning would be that this is a sensitive operation which fills the orphan tx pool

tiram88:
Looks legit, let's do that

michaelsutton (approved):

PR #241, opened 2023-08-16
Implement standalone daemon for integration tests

someone235:


testing/integration/src/integration_tests.rs
```diff
@@ -1703,3 +1709,77 @@ async fn staging_consensus_test() {
...
+        drop(socket1);
+        drop(socket2);
+        drop(socket3);
+        drop(socket4);
+
+        args.rpclisten = Some(format!("0.0.0.0:{rpc_port}").try_into().unwrap());
+        args.listen = Some(format!("0.0.0.0:{p2p_port}").try_into().unwrap());
+        args.rpclisten_json = Some(format!("0.0.0.0:{rpc_json_port}").parse().unwrap());
+        args.rpclisten_borsh = Some(format!("0.0.0.0:{rpc_borsh_port}").parse().unwrap());
+        args.appdir = Some(tempdir().unwrap().path().to_str().unwrap().to_owned());
```
michaelsutton, line 1764: I think we need to store `tempdir` within `DaemonWithRpc ` so that it is dropped only at the end

kaspad/src/daemon.rs
```diff
@@ -0,0 +1,357 @@
...
+        );
+
+        // Make sure config and args form a valid set of properties
+        if let Err(err) = validate_config_and_args(&config, args) {
+            println!("{}", err);
+            exit(1);
+        }
+
+        // TODO: Refactor all this quick-and-dirty code
+        let app_dir = args
```
michaelsutton, line 110: There's an identical code block below at `create_core_with_runtime`. Depending on the intended usage we can either store `app_dir` within `Runtime` and reuse it below, or we can extract the logic to a function and call from both locations (in case we want `app_dir` to be different between core instances). I would also remove the `// TODO: Refactor all this quick-and-dirty code`

kaspad/src/args.rs
```diff
@@ -132,17 +147,18 @@ pub fn cli(defaults: &Defaults) -> Command {
                 .value_name("IP[:PORT]")
                 .num_args(0..=1)
                 .require_equals(true)
-                .default_missing_value("default")
+                .default_missing_value("default") // TODO: Find a way to use defaults.rpclisten_borsh
```
michaelsutton, line 150: this TODO is not clear

kaspad/src/args.rs
```diff
@@ -132,17 +147,18 @@ pub fn cli(defaults: &Defaults) -> Command {
...
+
         )
         .arg(
             Arg::new("rpclisten-json")
                 .long("rpclisten-json")
                 .value_name("IP[:PORT]")
                 .num_args(0..=1)
                 .require_equals(true)
-                .default_missing_value("default")
+                .default_missing_value("default") // TODO: Find a way to use defaults.rpclisten_json
```
michaelsutton, line 161: same

testing/integration/src/integration_tests.rs
michaelsutton: from the tempdir crate docs:


[Resource Leaking](https://docs.rs/tempfile/latest/tempfile/struct.TempDir.html#resource-leaking)
Various platform-specific conditions may cause TempDir to fail to delete the underlying directory. It’s important to ensure that handles (like [File](http://doc.rust-lang.org/std/fs/struct.File.html) and [ReadDir](http://doc.rust-lang.org/std/fs/struct.ReadDir.html)) to files inside the directory are dropped before the TempDir goes out of scope. The TempDir destructor will silently ignore any errors in deleting the directory; to instead handle errors call [TempDir::close()](https://docs.rs/tempfile/latest/tempfile/struct.TempDir.html#method.close).

kaspad/src/daemon.rs
```diff
@@ -0,0 +1,357 @@
...
+pub struct Runtime {
+    log_dir: Option<String>,
+}
+
+impl Runtime {
+    pub fn from_args(args: &Args) -> Self {
+        // Configure the panic behavior
+        kaspa_core::panic::configure_panic();
+
+        let network = args.network();
```
michaelsutton, line 94: Seems like lines 94-107 can be removed from here. They are repeated at `create_core_with_runtime`

michaelsutton: * network should be an argument to Runtime, and then you can use `network.name()` instead of `config.network_name()`

michaelsutton (approved):

PR #243, opened 2023-08-24
Add pruning point list validations

someone235:


protocol/flows/src/v5/ibd/flow.rs
```diff
@@ -107,6 +107,16 @@ impl IbdFlow {
                 .await?;
             }
             IbdType::DownloadHeadersProof => {
+                if unix_now() - session.async_creation_timestamp().await > self.ctx.config.finality_duration() {
```
michaelsutton, line 110: 1. Better to do `unix_now() > self.ctx.config.finality_duration() + session.async_creation_timestamp().await` in order to avoid even a theoretical underflow  
2. Same goes for line 113
3. I think this check should be part of `determine_ibd_type` (inserted at line 168; returning `IbdType::None` if needed)
4. Missing a comment explaining this condition. "We reject the headers proof if the node has a relatively up-to-date finality point and current consensus has matured for long enough (and not recently synced). This is mostly a spam-protector since subsequent checks identify these violations as well"

protocol/flows/src/v5/ibd/flow.rs
```diff
@@ -107,6 +107,16 @@ impl IbdFlow {
...
+                    let f = session.async_finality_point().await;
```
michaelsutton, line 111: `f, f_ts` -> `fp, fp_ts`

consensus/src/processes/pruning.rs
```diff
@@ -169,6 +180,86 @@ impl<S: GhostdagStoreReader, T: ReachabilityStoreReader, U: HeaderStoreReader, V
...
+        // we expect to see it later on the list.
+        // The first stage is important because the most recent pruning point is pointing to a few
+        // pruning points before, so the first few pruning points on the list won't be pointed by
+        // any other pruning point in the list, so we are compelled to check if it's refereced by
+        // the selected chain.
+        let mut expected_pps_queue = VecDeque::new();
+        let hst = self.header_selected_tip_store.read().get().unwrap().hash;
+        for current in self.reachability_service.backward_chain_iterator(hst, pruning_info.pruning_point, false) {
+            let current_header = self.headers_store.get_header(current).unwrap();
+            if expected_pps_queue.is_empty() || *expected_pps_queue.back().unwrap() != current_header.pruning_point {
```
michaelsutton, line 219: `if expected_pps_queue.back().is_none_or(|&&h| h != current_header.pruning_point)` is a bit more readable

consensus/src/consensus/mod.rs
```diff
@@ -508,6 +514,19 @@ impl ConsensusApi for Consensus {
         self.virtual_processor.import_pruning_point_utxo_set(new_pruning_point, imported_utxo_multiset)
     }
 
+    fn validate_pruning_points(&self) -> ConsensusResult<()> {
+        let pp_info = self.pruning_point_store.read().get().unwrap();
+        if !self.services.pruning_point_manager.is_valid_pruning_point(pp_info.pruning_point) {
```
michaelsutton, line 519: I think we should read `hst` here and pass it to both `is_valid_pruning_point` and `are_pruning_points_in_valid_chain`.
Because (i) they both use it, (ii) the second relies on the chain check performed in the first

protocol/flows/src/v5/ibd/flow.rs
```diff
@@ -146,14 +146,30 @@ impl IbdFlow {
                 return Ok(IbdType::Sync(highest_known_syncer_chain_hash));
             }
 
-            // TODO: in this case we know a syncer chain block, but it violates our current finality. In some cases
-            // this info should possibly be used to reject the IBD despite having more blue work etc.
+            // If the pruning point is not in the chain of `highest_known_syncer_chain_hash`, it
+            // means it's in its antichain (because if `highest_known_syncer_chain_hash` was in
+            // the pruning point's past the pruning point itself would be
+            // `highest_known_syncer_chain_hash`). So it means there's a finality conflict.
+            // TODO: Find a better way to handle finality conflicts.
```
michaelsutton, line 153: 1. `TODO: Find a better way to handle finality conflicts.` appears 3 times in this file now. I think once in the correct location is enough 
2. while we are at it, there are 2 more simple `TODO` checks in this file which we should cover:
  `TODO: make sure that the proof pruning point is not genesis`
  `TODO: verify the proof pruning point is different than current consensus pruning point`

consensus/src/pipeline/virtual_processor/processor.rs
```diff
@@ -932,6 +933,32 @@ impl VirtualStateProcessor {
 
         Ok(())
     }
+
+    pub fn are_pruning_points_violating_finality(&self, pp_list: PruningPointsList) -> bool {
+        // Ideally we would want to check if the last known pruning point has the finality point
```
michaelsutton, line 938: Update comment following our offline conversation

michaelsutton (approved):

PR #271, opened 2023-08-28
Allow running devnet with fake UTXO set

someone235:


consensus/src/consensus/utxo_set_override.rs
```diff
@@ -0,0 +1,34 @@
...
+    use kaspa_consensus_core::{
+        api::ConsensusApi, config::Config, header::Header, muhash::MuHashExtensions, utxo::utxo_collection::UtxoCollection,
+    };
+    use kaspa_hashes::Hash;
+    use kaspa_muhash::MuHash;
+
+    use crate::consensus::Consensus;
+
+    pub fn set_genesis_utxo_commitment_from_config(config: &mut Config) {
+        let mut genesis_ms = MuHash::new();
```
michaelsutton, line 15: I prefer `genesis_multiset`. `ms` is automatically interoperated as milliseconds

consensus/src/consensus/utxo_set_override.rs
```diff
@@ -0,0 +1,34 @@
...
+        }
+
+        config.params.genesis.utxo_commitment = genesis_ms.finalize();
+        let genesis_header: Header = (&config.params.genesis).into();
+        config.params.genesis.hash = genesis_header.hash;
+    }
+
+    pub fn set_initial_utxo_set(initial_utxo_set: &UtxoCollection, consensus: Arc<Consensus>, genesis_hash: Hash) {
+        let utxo_slice = &initial_utxo_set.iter().map(|(op, entry)| (*op, entry.clone())).collect_vec()[..];
+        let mut ms = MuHash::new();
```
michaelsutton, line 27: `ms` -> `genesis_multiset`

consensus/src/consensus/factory.rs
```diff
@@ -192,6 +195,15 @@ impl ConsensusFactory for Factory {
...
             }
         };
+
+        #[cfg(feature = "developer-mode")]
+        if is_new_consensus {
+            set_genesis_utxo_commitment_from_config(&mut config);
+        }
+
+        #[cfg(feature = "developer-mode")]
+        let genesis_hash = config.params.genesis.hash;
```
michaelsutton, line 205: why is this variable needed? seems like it can be inlined below

consensus/src/consensus/factory.rs
```diff
@@ -192,6 +195,15 @@ impl ConsensusFactory for Factory {
                 entry
             }
         };
+
+        #[cfg(feature = "developer-mode")]
+        if is_new_consensus {
+            set_genesis_utxo_commitment_from_config(&mut config);
```
michaelsutton, line 201: I think you need to call this also if `is_new_consensus == false`. Otherwise genesis hash will be wrong if the node restarted

michaelsutton (approved):

IgorKhomenko:
Is there any guide re how to use this feature? 

I tried build the node

```
cargo build --release --features devnet-prealloc --workspace
```

then run it

```
cargo run --release --bin kaspad -- --devnet --prealloc-utxos 10
```

but it fails with an error:

```
error: unexpected argument '--prealloc-utxos' found

```

PR #277, opened 2023-09-08
Add warning for negative IO delta values

biryukovmaxim:
Added a warning message in case the new IO read or write bytes values are less than the previous values. This was done using the unwrap_or_else() function to prevent the program from potentially crashing when the subtraction operation results in a negative value.


coderofstuff (approved):
I no longer observe the issue I found

michaelsutton (approved):

PR #278, opened 2023-09-11
Implement Message Signing

coderofstuff:
Implements the message signing functionality as specified by https://github.com/kaspanet/kips/pull/8. In summary, an arbitrary message can be signed to prove access to some given address. Supports only Schnorr addresses.

Follow-up work:
- `signMessage` and `verifyMessage` to wallet
- expose them to WASM
- create sample JS implementations


wallet/core/src/message.rs
```diff
@@ -0,0 +1,265 @@
...
+    Ok(sig.to_vec())
+}
+
+pub fn verify_message(msg: PersonalMessage, signature: Vec<u8>, pubkey: XOnlyPublicKey) -> Result<bool, Error> {
+    let hash = calc_personal_message_hash(msg);
+    let msg = secp256k1::Message::from_slice(hash.as_bytes().as_slice())?;
+    let sig = secp256k1::schnorr::Signature::from_slice(signature.as_slice())?;
+    let sig_res = sig.verify(&msg, &pubkey);
+
+    match sig_res {
```
aspect, line 40: idiomatic: `sig_res.map(|_|true)`

wallet/core/src/message.rs
```diff
@@ -0,0 +1,265 @@
...
+        let pubkey = XOnlyPublicKey::from_slice(&[
+            0xF9, 0x30, 0x8A, 0x01, 0x92, 0x58, 0xC3, 0x10, 0x49, 0x34, 0x4F, 0x85, 0xF8, 0x9D, 0x52, 0x29, 0xB5, 0x31, 0xC8, 0x45,
+            0x83, 0x6F, 0x99, 0xB0, 0x86, 0x01, 0xF1, 0x13, 0xBC, 0xE0, 0x36, 0xF9,
+        ])
+        .unwrap();
+        let sig_result = sign_message(pm.clone(), &privkey);
+
+        assert!(sig_result.is_ok());
+
+        if let Ok(sig_result) = sig_result {
```
aspect, line 72: opinions might differ, but although I myself do `if let Ok()`, pre-checking `assert!()` and then doing `if let Ok()` is a bit redundant. Following are few suggestions on alternate handling:

since `assert!()` triggers a `panic!()` (which is what `expect()` does as well, may I suggest something like:
```
let sig = sig_result.expect("signature result error");
verify_ message(pm,sig_result,pubkey).expect("signature verification error");
```
this can be further collapsed to:
```
verify_message(
    pm.clone(), 
    sign_message(pm, &privkey).expect("sign_message failure"), 
    pubkey)
.expect("verify_message failure");
```
this is more rusty.

I think `expect()` would be better here as it signals what is actually wrong.

wallet/core/src/message.rs
```diff
@@ -0,0 +1,265 @@
...
+
+    let msg = secp256k1::Message::from_slice(hash.as_bytes().as_slice())?;
+    let schnorr_key = secp256k1::KeyPair::from_seckey_slice(secp256k1::SECP256K1, privkey)?;
+    let curve = secp256k1::Secp256k1::new();
+    let sig: [u8; 64] = *curve.sign_schnorr_with_aux_rand(&msg, &schnorr_key, aux_rand).as_ref();
+
+    Ok(sig.to_vec())
+}
+
+pub fn verify_message(msg: PersonalMessage, signature: Vec<u8>, pubkey: XOnlyPublicKey) -> Result<bool, Error> {
```
aspect, line 34: You should pass a reference for pubkey `&XOnlyPublicKey` here - because to use this function, you have to clone the pubkey, just for it to be used later as a reference in `sig.verify()`...  (You don't feel it in the unit tests because the pubkey is local).

aspect (commented):
Looks very good.  The "pubkey as reference" comment is the only important thing really.

wallet/core/src/message.rs
biryukovmaxim: Whst does it mean if it returns Ok(false)?

coderofstuff: @biryukovmaxim it can't return `Ok(false)`. The underlying schnorr verify will return an `Err` if the signature is anything other than a legit one.

I think the correct return signature of `verify_message` should be changed to just bool, since the only error that can be thrown in the underlying command happens when the signature is not legit

coderofstuff: I'll just go with that

aspect (approved):
looks good

PR #279, opened 2023-09-12
Adding multisig capabilities to wallet

biryukovmaxim:
add functionality of creating, importing, exporting multisig wallet


aspect (approved):
As I have been following this development, everything looks good.  Merging so that we can continue the integration of this codebase against other ongoing work in the wallet.

PR #280, opened 2023-09-12
Add cli command for message sign/verify

coderofstuff:
New CLI command: `message`

Subcommands:
- sign <kaspa_address> - Sign a message with the private key that matches the given address. Prompts for message.
- verify <kaspa_address> <signature> - Verify the signature against the message and kaspa_address. Prompts for message.

### DIFF
--- a/cli/src/modules/message.rs
+++ b/cli/src/modules/message.rs
@@ -31,22 +31,22 @@ impl Message {
 
         match argv.get(0).unwrap().as_str() {
             "sign" => {
-                if argv.len() < 2 {
+                if argv.len() != 2 {
                     return self.display_help(ctx, argv).await;
                 }
 
-                let kaspa_address = argv.get(1).unwrap().as_str();
+                let kaspa_address = argv[1].as_str();
                 let asked_message = ctx.term().ask(false, "Message: ").await?;
                 let message = asked_message.as_str();
 
                 self.sign(ctx, kaspa_address, message).await?;
             }
             "verify" => {
-                if argv.len() < 3 {
+                if argv.len() != 3 {
                     return self.display_help(ctx, argv).await;
                 }
-                let kaspa_address = argv.get(1).unwrap().as_str();
-                let signature = argv.get(2).unwrap().as_str();
+                let kaspa_address = argv[1].as_str();
+                let signature = argv[2].as_str();
                 let asked_message = ctx.term().ask(false, "Message: ").await?;
                 let message = asked_message.as_str();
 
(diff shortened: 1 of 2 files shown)


cli/src/modules/message.rs
```diff
@@ -0,0 +1,157 @@
...
+            AccountKind::Bip32 => {
+                let (wallet_secret, payment_secret) = ctx.ask_wallet_secret(Some(&account)).await?;
+                let keydata = account.prv_key_data(wallet_secret).await?;
+                let mut keys = HashMap::new();
+                let account = account.clone().as_derivation_capable().expect("expecting derivation capable");
+
+                let (receive, change) = account.derivation().addresses_indexes(&[&kaspa_address])?;
+                let private_keys = account.create_private_keys(&keydata, &payment_secret, &receive, &change)?;
+                for (address, private_key) in private_keys {
+                    keys.insert(address.clone(), private_key.secret_bytes());
```
biryukovmaxim, line 142: Instead of insertion you can compare addresses and return it directly

coderofstuff: Fixed in [69020cb](https://github.com/kaspanet/rusty-kaspa/pull/280/commits/69020cbc0ef03a807a05554acb96ffca57ebb281)

cli/src/modules/message.rs
```diff
@@ -0,0 +1,158 @@
...
+            return self.display_help(ctx, argv).await;
+        }
+
+        match argv.get(0).unwrap().as_str() {
+            "sign" => {
+                if argv.len() < 2 {
+                    return self.display_help(ctx, argv).await;
+                }
+
+                let kaspa_address = argv.get(1).unwrap().as_str();
```
biryukovmaxim, line 38: If there's no argument, you can try to get current active account or ask to select it

cli/src/modules/message.rs
```diff
@@ -0,0 +1,158 @@
...
+        let pm = PersonalMessage(message);
+        let privkey = self.get_address_private_key(&ctx, kaspa_address).await?;
+
+        let sig_result = sign_message(&pm, &privkey);
+
+        match sig_result {
+            Ok(signature) => {
+                let sig_hex = faster_hex::hex_string(signature.as_slice());
+                tprintln!(ctx, "Signature: {}", sig_hex);
+            }
```
biryukovmaxim, line 94:
```
Ok(())
```
Then you will be able to remove `return` word in the second condition branch and 'Ok(())' below

aspect (approved):
Everything looks good.

PR #281, opened 2023-09-20
Implement Message Signing WASM

coderofstuff:
New JS functions:
- `signMessage(message, privkey)` - privkey can be a hex string or `PrivateKey` object
- `verifyMessage(message, signature, pubkey)` - pubkey can be a hex string or `PublicKey` object

### DIFF
--- a/wallet/core/src/wasm/message.rs
+++ b/wallet/core/src/wasm/message.rs
@@ -1,29 +1,46 @@
+use crate::imports::*;
 use crate::message::*;
 use kaspa_consensus_wasm::{PrivateKey, PublicKey};
-use wasm_bindgen::prelude::*;
 
-#[wasm_bindgen(js_name = signMessage)]
-pub fn js_sign_message(raw_msg: String, privkey: JsValue) -> String {
-    let mut privkey_bytes = [0u8; 32];
+/// Signs a message with the given private key
+/// @param {object} value - an object containing { message: String, privateKey: String|PrivateKey }
+/// @returns {String} the signature, in hex string format
+#[wasm_bindgen(js_name = signMessage, skip_jsdoc)]
+pub fn js_sign_message(value: JsValue) -> Result<String, Error> {
+    if let Some(object) = Object::try_from(&value) {
+        let private_key = object.get::<PrivateKey>("privateKey")?;
+        let raw_msg = object.get_string("message")?;
+        let mut privkey_bytes = [0u8; 32];
 
-    let private_key = PrivateKey::try_from(privkey).unwrap_throw();
-    privkey_bytes.copy_from_slice(&private_key.secret_bytes());
+        privkey_bytes.copy_from_slice(&private_key.secret_bytes());
 
-    let pm = PersonalMessage(&raw_msg);
+        let pm = PersonalMessage(&raw_msg);
 
-    let sig_result = sign_message(&pm, &privkey_bytes);
-    let sig_vec = sig_result.unwrap_throw();
+        let sig_result = sign_message(&pm, &privkey_bytes);
+        let sig_vec = sig_result.unwrap_throw();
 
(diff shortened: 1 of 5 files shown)


wallet/core/src/wasm/message.rs
```diff
@@ -0,0 +1,40 @@
...
+use secp256k1::XOnlyPublicKey;
+use std::str::FromStr;
+use wasm_bindgen::prelude::*;
+use workflow_wasm::prelude::ref_from_abi;
+
+#[wasm_bindgen(js_name = signMessage)]
+pub fn js_sign_message(raw_msg: String, privkey: JsValue) -> String {
+    let mut privkey_bytes = [0u8; 32];
+
+    if let Ok(privkey) = ref_from_abi!(PrivateKey, &privkey) {
```
aspect, line 12: Most primitives like Private and Public key have `TryFrom<JsValue>` to automate their conversion:
https://github.com/kaspanet/rusty-kaspa/blob/master/consensus/wasm/src/keypair.rs#L170
So intake can be reduced to `let privkey = PrivateKey::try_from(privkey)?;

coderofstuff: Simplified. I can't use `?` here because the return values aren't `Result` or `Option` so I use `unwrap_throw()` instead.

KaffinPX:
I think signMessage should be on private key class?

wallet/core/src/wasm/message.rs
aspect: The result of the function should be `Result<String>`, then you can use `?`.  This is the common practice (wasm_bindgen will resolve ordinary types vs Results and setup appropriate handling).

aspect:
@coderofstuff Before moving forward with merging this PR, I wanted to discuss/suggest perhaps converting these functions to receive not multiple arguments but a single `JsValue` object containing these arguments.  I have been converting all WASM functions (where applicable), removing arguments to have them receive a single `Object` argument.

For example, the following function receives arguments via an object: https://github.com/kaspanet/rusty-kaspa/blob/master/wallet/core/src/wasm/utxo/context.rs#L114-L128

This allows us to "future proof" these functions, allowing them to, in the future, if the need arises, to receive different types of arguments, without having to change function signatures.

For example, if in the future, we want to add other types of signatures (as discussed during our KIP-005 debate) the function can decide what type of signature is to be used by examining the supplied object data...

Arguments would need to be documented in the rustdoc section for each function.

--

One more note - if one wants to document types on JS side, instead of receiving a JsValue, the function can receive a custom data type which is declared as JS type on the Rust side (example: https://github.com/kaspanet/rusty-kaspa/blob/master/wallet/core/src/wasm/tx/generator/generator.rs#L22)

The only reason this is done is that all type documentation propagated to TypeScript, so when using such type in TypeScript/JavaScript, the user immediately gets full inline documentation of the data type in question.

This is not currently used in many places but the long-term goal is to convert the majority of API calls to use custom types in such manner. This last part is not really needed at this point, just letting you know that this can be done.

aspect:
> I think signMessage should be on private key class?
@KaffinPX if you see my recent feedback, I am trying to convert and make most API calls to accept a single Object argument that contains different properties as arguments. This allows us to make these functions accept additional arguments in the future without changing their signatures. 

There have been a number of side discussions to support other types of signatures in the future, such approach allows us to extend these fns in the future.

However, affixing these functions to ProvateKey plays against that (for example, in the future such function may accept a PrivateKey parameter or a wallet Account parameter).

So I prefer to keep these functions disassociated from any specific data types (classes).

wallet/core/src/wasm/message.rs
coderofstuff: Fixed

aspect (approved):
Looks great. Merging.

PR #282, opened 2023-09-20
Revert "Implement Message Signing"

someone235:
Reverts kaspanet/rusty-kaspa#278
We should wait until https://github.com/kaspanet/kips/pull/8 becomes active.


someone235:
KIP 5 was activated, so I'm closing this PR

PR #288, opened 2023-09-30
Add UPnP support and configurations to Kaspad

biryukovmaxim:
Introduced Universal Plug and Play (UPnP) configurations in the Kaspad system to automatically forward ports on a broadband router. The UPnP setting is optional and can be configured via the 'kaspad/src/args.rs' file. The 'upnp' field was added to 'Config' structs and default configuration was set as 'false'. A function was added to map ports if UPnP is enabled and the 'igd-next' library was incorporated to facilitate UPnP support. The changes improve Kaspad's peer discovery mechanism by allowing it to be accessed by more peers especially when behind a NAT firewall.


kaspad/src/args.rs
```diff
@@ -306,7 +311,7 @@ pub fn cli() -> Command {
             .require_equals(true)
             .value_parser(clap::value_parser!(u64))
             .help("Interval in seconds for performance metrics collection."),
-    );
+    ).arg(arg!(--upnp "Enable upnp"));
```
coderofstuff, line 314: Can we make upnp enabled by default?

Then flag here can be `--disable-upnp` which disables the implementation. Ideally, the base `kaspad --utxoindex` will do the upnp check.

coderofstuff (changes requested):

components/addressmanager/src/port_mapping_extender.rs
```diff
@@ -0,0 +1,82 @@
...
+            self.gateway
+                .add_port(
+                    igd_next::PortMappingProtocol::TCP,
+                    self.external_port,
+                    self.local_addr,
+                    self.deadline_sec as u32,
+                    "Kaspad-rusty",
+                )
+                .await
+                .unwrap(); // todo handle??
```
coderofstuff, line 48: Definitely needs to be handled. Currently can cause the node to panic.

Replication:
1. Start the node
2. Turn off your wifi

someone235 (changes requested):

someone235 (approved):

PR #289, opened 2023-10-05
Fix for mnemonic import in existing wallet

surinder83singh:


aspect (approved):
Thanks for putting this together.

I am approving this PR - it looks fine. 

We will have a separate discussion regarding `ktub` types of prefixes and if needed will address this topic in a separate PR.

PR #295, opened 2023-10-20
Optimize P2P block processing latency -- Broadcast blocks after Body Processor

coderofstuff:
Changes:
1. Refactor validate_and_insert_block to return two futures - Added in [b134d61](https://github.com/kaspanet/rusty-kaspa/pull/295/commits/b134d6128b0fe11803db77dc42218e3393804460)
- one future resolves when body processor is complete
- the other future resolves when virtual state processor is complete

  The second future is used by all original callers, making the current implementation behave the same as before.

2. Change BodyProcessorMessage to VirtualStateProcessorMessage when passing the block from body-processor to virtual-processor - Added in [af066c5](https://github.com/kaspanet/rusty-kaspa/pull/295/commits/af066c50ca4e65688e19cb7be9ec797a156e286e)
3. Add conditions/checks to see if we can broadcast
4. Actually broadcast the block after Body Processor

### DIFF
--- a/components/consensusmanager/src/session.rs
+++ b/components/consensusmanager/src/session.rs
@@ -142,11 +142,11 @@ impl ConsensusSessionOwned {
 }
 
 impl ConsensusSessionOwned {
-    pub fn validate_and_insert_block(&self, block: Block) -> BlockValidationFuture {
+    pub fn validate_and_insert_block(&self, block: Block) -> (BlockValidationFuture, BlockValidationFuture) {
         self.consensus.validate_and_insert_block(block)
     }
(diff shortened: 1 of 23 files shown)


michaelsutton (approved):

michaelsutton:
The main aim of this PR is to minimize peer to peer latency of block processing by sharing a block with peers ASAP. Tests and benchmarks have shown that this is especially important with higher BPS network configurations.

someone235 (changes requested):

someone235 (approved):

PR #296, opened 2023-10-20
gRPC server enhancements

tiram88:
Enhance the gRPC server with:

- Connections (clients) identified with Uuids, allowing multiple connections from the same origin.
- Handling of concurrent requests and subscriptions. However, in order to preserve the behavior existing in golang version, requests of a given connection and method are still processed in queue.
- Cleanly shuts down services from core and async runtime.
- Fix integration tests not shutting down the gRPC server correctly.
- Add an extra safeguard forcing the gRPC server shutdown on timeout.


rpc/grpc/server/src/connection.rs
```diff
@@ -160,354 +168,74 @@ impl Connection {
...
-        Ok(response)
+    async fn route_to_handler(&self, core_service: &DynRpcService, request: KaspadRequest) -> GrpcServerResult<()> {
+        // TODO: add appropriate error
+        if request.payload.is_none() {
+            debug!("GRPC, Route to handler got empty payload, client: {}", self);
+            return Err(GrpcServerError::InvalidRequestPayload);
+        }
+        let rpc_op = request.payload.as_ref().unwrap().into();
+        let sender = self.inner.routing_map.read().get(&rpc_op).cloned();
+        let sender = sender.unwrap_or_else(|| self.subscribe(core_service, rpc_op));
```
michaelsutton, line 224: There is a theoretic race-condition here if 2 threads try to add a new handler for the same op concurrently. `sender` will be `None` and they both will subscribe with the second one panicking. However this is not possible with current logic since a single task calls `route_to_handler` sequentially. Yet we should either document it or fix this by changing subscribe to use map `entry` api and adding the entry only if needed

tiram88: Let's opt for the entry API

rpc/grpc/server/src/connection.rs
```diff
@@ -160,354 +226,42 @@ impl Connection {
...
+    }
 
-        Ok(response)
+    /// Enqueues a response to be sent to the client
+    pub async fn enqueue(&self, response: KaspadResponse) -> GrpcServerResult<()> {
+        assert!(response.payload.is_some(), "Kaspad gRPC message should always have a value");
+        match self.inner.outgoing_route.try_send(response) {
+            Ok(_) => Ok(()),
+            Err(TrySendError::Closed(_)) => Err(GrpcServerError::ConnectionClosed),
+            Err(TrySendError::Full(_)) => Err(GrpcServerError::OutgoingRouteCapacityReached(self.to_string())),
```
michaelsutton, line 263: why not `await` if channel is full? (meaning use `.outgoing_route.send(response).await`)
this was the previous behavior as far as I can tell

tiram88: If undelivered messages keep accumulating up the channel capacity, there might be an issue with the connection. As discussed, better increase the capacity and disconnect the client when reaching full capacity.

michaelsutton (approved):

PR #300, opened 2023-10-24
Refactor 'addressmanager', reorder fetching sources for-external ip

biryukovmaxim:
reorder fetching sources for external ip according to: 
   external ip -> provided local address -> net interfaces -> upnp

The 'addressmanager' component of the codebase has been thoroughly refined for improved readability and maintainability. Firstly, the logic for assigning local network addresses has been streamlined and broken down into smaller methods for more clarity. Old methods 'add_routable_addresses_from_net_interfaces' and 'configured_address' have been replaced by 'routable_addresses_from_net_interfaces', 'local_addresses', and 'upnp'. This also led to the removal of redundant code.

### DIFF
--- a/components/addressmanager/src/lib.rs
+++ b/components/addressmanager/src/lib.rs
@@ -72,7 +72,7 @@ impl AddressManager {
         let extender = if self.local_net_addresses.is_empty() && !self.config.disable_upnp {
             let (net_address, ExtendHelper { gateway, local_addr, external_port }) = match self.upnp() {
                 Err(err) => {
-                    warn!("Error adding port mapping: {err}");
+                    warn!("[UPnP] Error adding port mapping: {err}");
                     return None;
                 }
                 Ok(None) => return None,
@@ -148,14 +148,14 @@ impl AddressManager {
     }
 
     fn upnp(&self) -> Result<Option<(NetAddress, ExtendHelper)>, UpnpError> {
-        info!("Attempting to register upnp... (to disable run the node with --disable-upnp)");
+        info!("[UPnP] Attempting to register upnp... (to disable run the node with --disable-upnp)");
         let gateway = igd::search_gateway(Default::default())?;
         let ip = IpAddress::new(gateway.get_external_ip()?);
         if !ip.is_publicly_routable() {
-            info!("Non-publicly routable external ip from gateway using upnp {} not added to store", ip);
+            info!("[UPnP] Non-publicly routable external ip from gateway using upnp {} not added to store", ip);
             return Ok(None);
         }
-        info!("Got external ip from gateway using upnp: {ip}");
+        info!("[UPnP] Got external ip from gateway using upnp: {ip}");
 
         let default_port = self.config.default_p2p_port();
 
@@ -174,7 +174,7 @@ impl AddressManager {
             UPNP_REGISTRATION_NAME,
         ) {
             Ok(_) => {
-                info!("Added port mapping to default external port: {ip}:{default_port}");
+                info!("[UPnP] Added port mapping to default external port: {ip}:{default_port}");
                 Ok(Some((NetAddress { ip, port: default_port }, ExtendHelper { gateway, local_addr, external_port: default_port })))
             }
             Err(AddPortError::PortInUse {}) => {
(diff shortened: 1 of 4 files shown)


components/addressmanager/src/lib.rs
```diff
@@ -55,124 +55,123 @@ impl AddressManager {
...
-                }
-            }
-            None => None,
+    fn routable_addresses_from_net_interfaces(&self) -> impl Iterator<Item = NetAddress> + '_ {
+        // check whatever was passed as listen address (if routable)
+        // otherwise(listen_address === 0.0.0.0) check all interfaces
+        let listen_address = self.config.p2p_listen_address.normalize(self.config.default_p2p_port());
+        if listen_address.ip.is_publicly_routable() {
+            info!("Publicly routable local address found: {}", listen_address.ip);
+            itertools::Either::Left(Some(listen_address).into_iter())
```
someone235, line 112: It's nicer to use `iter::once(listen_address)`

components/addressmanager/src/lib.rs
```diff
@@ -55,124 +55,123 @@ impl AddressManager {
...
+        } else if listen_address.ip.is_unspecified() {
+            let network_interfaces = list_afinet_netifas();
+            let Ok(network_interfaces) = network_interfaces else {
+                warn!("Error getting network interfaces: {:?}", network_interfaces);
+                return itertools::Either::Left(None.into_iter());
```
someone235, line 117: It's nicer to use `iter::empty()`

components/addressmanager/src/lib.rs
```diff
@@ -55,124 +55,123 @@ impl AddressManager {
...
-                    external_port,
-                    local_addr,
-                ))
-            } else {
-                None
-            }
+        self.local_net_addresses = self.local_addresses().collect();
+
+        let extender = if self.local_net_addresses.is_empty() && !self.config.disable_upnp {
+            let (net_address, ExtendHelper { gateway, local_addr, external_port }) = self.upnp()?;
```
someone235, line 61: log in case of upnp error

someone235 (changes requested):

components/addressmanager/src/lib.rs
michaelsutton: some of the exit paths within the `upnp` method already log an error, but not all

components/addressmanager/src/lib.rs
biryukovmaxim: unfortunately it will produce the error of incompatible types
`Option.into_iter()` gives me
 ```
struct Item<A> {
    opt: Option<A>,
}
```
for both None and Some. `iter::empty` and `iter::once` gives 2 different types.
to make it works I have to implement or find Either3 with predefined iterator implementation


and actually my first attempt was with `iter::` and then I changed it to Option

File: components/addressmanager/src/lib.rs (hunk 3 above)
Comment (biryukovmaxim) [reply components/addressmanager/src/lib.rs:61]: it is already logged here:
https://github.com/biryukovmaxim/rusty-kaspa/blob/a76413f5b1ce6509ea6f0d8c63009af08c748e0b/components/addressmanager/src/lib.rs#L136
and here: https://github.com/biryukovmaxim/rusty-kaspa/blob/a76413f5b1ce6509ea6f0d8c63009af08c748e0b/components/addressmanager/src/lib.rs#L169
should I change it from warning to error?

Comment (michaelsutton) [reply components/addressmanager/src/lib.rs:61]: lines 133 and 134 are not logged

Comment (michaelsutton) [reply components/addressmanager/src/lib.rs:61]: btw line 169 `warn!("error adding port: {err}");` should start with a capital `Error` (keep it warn though). Also I think perhaps all upnp log messages should be prefixed with `[UPnP] `

File: components/addressmanager/src/lib.rs (hunk 1 above)
Comment (biryukovmaxim) [reply components/addressmanager/src/lib.rs]: in the last commit I changed it to Left(Left(...) Left(Right(..._) and Right(...). not sure which variant is better/cleaner/nicer

File: components/addressmanager/src/lib.rs [hunk 4]
```diff
@@ -147,6 +172,34 @@ impl AddressManager {
             normalized_p2p_listen_address.into()
         };
 
+        let mut index = 0;
+        let already_in_use = loop {
```
someone235, line 176: Add a comment that explains why this code is needed in addition to handling the error in `AddPortError::PortInUse`

someone235 (changes requested):

someone235 (approved):

PR #301, opened 2023-10-24
Delete inactive entries after staging commit

someone235:


consensus/src/consensus/factory.rs
```diff
@@ -307,4 +287,34 @@ impl ConsensusFactory for Factory {
...
+                    }
+                } else {
+                    Some(entry)
+                }
+            })
+            .collect_vec();
+
+        for entry in entries_to_delete {
+            write_guard.delete_entry(entry).unwrap();
+        }
```
biryukovmaxim, line 318: you can avoid vector allocation by calling `for_each` directly on the iterator

michaelsutton (approved):

PR #302, opened 2023-10-28
Update WRS to 0.8.0 and async_channel to 2.0.0

aspect:
This PR updates `workflow-rs` to `0.8.0` to enable `accept()` functionality in wRPC needed for the file descriptor budget pool being implemented in https://github.com/kaspanet/rusty-kaspa/pull/284

As `workflow-rs` updates the `async_channel` dependency to `2.0.0`, it is also updated.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -222,8 +222,21 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "81953c529336010edd6d8e358f886d9581267795c61b19475b71314bffa46d35"
 dependencies = [
  "concurrent-queue",
- "event-listener",
+ "event-listener 2.5.3",
  "futures-core",
+]
+
+[[package]]
+name = "async-channel"
+version = "2.0.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "336d835910fab747186c56586562cb46f42809c2843ef3a84f47509009522838"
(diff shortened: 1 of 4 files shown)


biryukovmaxim (approved):

Cargo.lock
```diff
@@ -291,7 +304,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "62565bb4402e926b29953c785397c6dc0391b7b446e45008b0049eb43cec6f5d"
 dependencies = [
  "async-attributes",
- "async-channel",
+ "async-channel 1.9.0",
```
coderofstuff, line 307: Why do some of the references in this lock refer to 1.9.0 and why some 2.0.0?

aspect: async_std still uses an old version of async_channel and is thus a dependency of the project, however we make it a point that we use async_channel everywhere directly (I actually went over the code and corrected this where it was not so).

aspect: Just to add: we can’t use same crates with different versions in any code that builds to WASM as wasm-pack will error out with naming collisions. To make sure this doesn’t occur by an accident, we have /test script build WASM after running tests. (and CI itself builds WASM as a separate stage). So the noted dependency is benign.

aspect:
Closing as same changes are coming in with #303 and #305.

PR #310, opened 2023-11-02
Create RPC Integration Tests

coderofstuff:
# Introduction
Test currently covers all RPC calls, but those that are failing or those that are unimplemented are commented out

To execute this test, run:
```
cargo test --release --package kaspa-testing-integration --lib --features devnet-prealloc -- rpc_tests::base_test --exact --nocapture --ignored
```

# TODO
- [x] Add calling the above to a github action
- [x] Fix the commented out sections for failing tests

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -110,6 +110,9 @@ jobs:
         
       - name: Run cargo test regular features
         run: cargo test --workspace
+      
+      - name: Run RPC tests
+        run: cargo test --release --package kaspa-testing-integration --lib --features devnet-prealloc -- rpc_tests::base_test
 
       - name: Run cargo test on kaspa-hashes without asm
         run: cargo test -p kaspa-hashes --features=no-asm --benches
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -123,8 +123,8 @@ jobs:
       - name: Run cargo test on kaspa-hashes without asm
         run: cargo nextest run --release -p kaspa-hashes --features=no-asm --benches
 
-      - name: Run cargo doc tests with features=no-asm
-        run: cargo test --doc --release --features=no-asm
+      - name: Run cargo doc tests with features=no-asm on kaspa-hashes
+        run: cargo test --doc --release -p kaspa-hashes --features=no-asm
 
   # test-release:
   #   name: Test Suite Release
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -117,12 +117,11 @@ jobs:
       - name: Run cargo test regular features
         run: cargo nextest run --release --workspace
 
-      - name: Run cargo doc tests
-        run: cargo test --release --doc --workspace
-
       - name: Run cargo test on kaspa-hashes without asm
         run: cargo nextest run --release -p kaspa-hashes --features=no-asm --benches
 
+      - name: Run cargo doc tests
+        run: cargo test --release --doc --workspace --features=no-asm
 
   # test-release:
   #   name: Test Suite Release
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -113,9 +113,12 @@ jobs:
 
       - name: Run cargo build with devnet-prealloc feature
         run: cargo build --release --features devnet-prealloc --workspace --all --tests --benches
-        
+
       - name: Run cargo test regular features
         run: cargo nextest run --release --workspace
+
+      - name: Run cargo doc tests
+        run: cargo test --release --doc --workspace
 
       - name: Run cargo test on kaspa-hashes without asm
         run: cargo nextest run --release -p kaspa-hashes --features=no-asm --benches
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -112,7 +112,7 @@ jobs:
         run: cargo test --workspace
       
       - name: Run RPC tests
-        run: cargo test --release --package kaspa-testing-integration --lib --features devnet-prealloc -- rpc_tests::base_test
+        run: cargo test --package kaspa-testing-integration --lib --features devnet-prealloc -- rpc_tests::base_test
 
       - name: Run cargo test on kaspa-hashes without asm
         run: cargo test -p kaspa-hashes --features=no-asm --benches
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -116,9 +116,6 @@ jobs:
         
       - name: Run cargo test regular features
         run: cargo nextest run --release --workspace
-      
-      - name: Run RPC tests
-        run: cargo nextest run --release -p kaspa-testing-integration --lib --features devnet-prealloc -- rpc_tests::base_test
 
       - name: Run cargo test on kaspa-hashes without asm
         run: cargo nextest run --release -p kaspa-hashes --features=no-asm --benches
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -112,16 +112,16 @@ jobs:
           key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
 
       - name: Run cargo build with devnet-prealloc feature
-        run: cargo build --features devnet-prealloc --workspace --all --tests --benches
+        run: cargo build --release --features devnet-prealloc --workspace --all --tests --benches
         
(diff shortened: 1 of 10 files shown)


aspect (dismissed):
Looks very good for the initial pass.

testing/integration/src/rpc_tests.rs
```diff
@@ -0,0 +1,367 @@
...
+    let prealloc_address =
+        Address::new(NetworkType::Simnet.into(), kaspa_addresses::Version::PubKey, &prealloc_pk.x_only_public_key().0.serialize());
+    let schnorr_key = secp256k1::KeyPair::from_secret_key(secp256k1::SECP256K1, &prealloc_sk);
+    let spk = pay_to_address_script(&prealloc_address);
+
+    let args = Args {
+        simnet: true,
+        disable_upnp: true, // UPnP registration might take some time and is not needed for this test
+        enable_unsynced_mining: true,
+        num_prealloc_utxos: Some(TX_LEVEL_WIDTH as u64 * 1),
```
tiram88, line 51: the * 1 is not useful

testing/integration/src/rpc_tests.rs
```diff
@@ -0,0 +1,367 @@
...
+        prealloc_address: Some(prealloc_address.to_string()),
+        prealloc_amount: 500 * SOMPI_PER_KASPA,
+        block_template_cache_lifetime: Some(0),
+        utxoindex: true,
+        unsafe_rpc: true,
+        ..Default::default()
+    };
+
+    let utxoset = args.generate_prealloc_utxos(args.num_prealloc_utxos.unwrap());
+    let txs = common::utils::generate_tx_dag(utxoset.clone(), schnorr_key, spk, TX_COUNT / TX_LEVEL_WIDTH, TX_LEVEL_WIDTH);
```
tiram88, line 61: this doesn't seem to be used in the test right?

testing/integration/src/rpc_tests.rs
```diff
@@ -0,0 +1,367 @@
...
+        $tasks.push(task);
+    };
+}
+
+/// `cargo test --release --package kaspa-testing-integration --lib --features devnet-prealloc -- rpc_tests::base_test`
+#[tokio::test]
+#[cfg(feature = "devnet-prealloc")]
+async fn base_test() {
+    kaspa_core::panic::configure_panic();
+    kaspa_core::log::try_init_logger("info,kaspa_core::time=debug,kaspa_mining::monitor=debug");
```
tiram88, line 28: `info,kaspa_rpc_core=debug,kaspa_rpc_service=debug,kaspa_grpc_client=debug,kaspa_grpc_server=debug` might make more sense I guess

tiram88 (changes requested):

.github/workflows/ci.yaml
```diff
@@ -106,13 +112,13 @@ jobs:
           key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
 
       - name: Run cargo build with devnet-prealloc feature
-        run: cargo build --features devnet-prealloc --workspace --all --tests --benches
+        run: cargo build --release --features devnet-prealloc --workspace --all --tests --benches
         
       - name: Run cargo test regular features
-        run: cargo test --workspace
+        run: cargo nextest run --release --workspace
```
biryukovmaxim, line 118: Add doc test

.github/workflows/ci.yaml
```diff
@@ -106,13 +112,13 @@ jobs:
...
 
       - name: Run cargo test on kaspa-hashes without asm
-        run: cargo test -p kaspa-hashes --features=no-asm --benches
+        run: cargo nextest run --release -p kaspa-hashes --features=no-asm --benches
```
biryukovmaxim, line 121: Add doc test

test
```diff
@@ -6,7 +6,7 @@ if [ $? -ne 0 ]; then
 fi
 
 # integration tests
-cargo test --release --workspace -p kaspa-testing-integration --lib
+cargo nextest run --release --workspace -p kaspa-testing-integration --lib
```
biryukovmaxim, line 9: Add doctest

biryukovmaxim (changes requested):

test
aspect: We don't want doc tests here. Unless we add --full argument (or also --doc).

These tests are "pre-checks", waited on by developers before commits, so we want them to run as fast as possible and defer actual potential failures to CI. Since doc tests will be in CI, there is no reason to have them here unless manually triggered.

biryukovmaxim: It was enabled before by default. Is it time to disable it?

aspect: Oh, sorry about the confusion.

@coderofstuff cargo test runs doc tests by default, while nextest doesn't. This means that if we switch to nextest we should add cargo test --doc --workspace as a separate step.

.github/workflows/ci.yaml
coderofstuff: This should be covered by the `cargo test --release --doc --workspace --features=no-asm` I added

test
coderofstuff: This should be covered by the `cargo test --release --doc --workspace --features=no-asm` I added

.github/workflows/ci.yaml
```diff
@@ -106,14 +112,19 @@ jobs:
...
+
+      - name: Run cargo doc tests with features=no-asm
+        run: cargo test --doc --release --workspace
 
       - name: Run cargo test on kaspa-hashes without asm
-        run: cargo test -p kaspa-hashes --features=no-asm --benches
+        run: cargo nextest run --release -p kaspa-hashes --features=no-asm --benches
 
+      - name: Run cargo doc tests with features=no-asm
+        run: cargo test --doc --release --features=no-asm
```
tiram88, line 127: IMHO this one can be narrowed down to
`cargo test --doc --release -p kaspa-hashes --features=no-asm --benches`

.github/workflows/ci.yaml
```diff
@@ -106,14 +112,19 @@ jobs:
...
       - name: Run cargo build with devnet-prealloc feature
-        run: cargo build --features devnet-prealloc --workspace --all --tests --benches
-        
+        run: cargo build --release --features devnet-prealloc --workspace --all --tests --benches
+
       - name: Run cargo test regular features
-        run: cargo test --workspace
+        run: cargo nextest run --release --workspace
+
+      - name: Run cargo doc tests with features=no-asm
```
tiram88, line 120: remove "with features=no-asm"

.github/workflows/ci.yaml
```diff
@@ -106,14 +112,19 @@ jobs:
...
+        run: cargo test --doc --release --workspace
 
       - name: Run cargo test on kaspa-hashes without asm
-        run: cargo test -p kaspa-hashes --features=no-asm --benches
+        run: cargo nextest run --release -p kaspa-hashes --features=no-asm --benches
 
+      - name: Run cargo doc tests with features=no-asm
```
tiram88, line 126: add "on kaspa-hashes"

tiram88 (changes requested):

.github/workflows/ci.yaml
coderofstuff: It can't because `--benches` and `-p kaspa-hashes` is another target:

```
rusty-kaspa git:(rpc-integration-tests) ✗ cargo test --doc --release -p kaspa-hashes --features=no-asm --benches
error: Can't mix --doc with other target selecting options
```

.github/workflows/ci.yaml
coderofstuff: We can't add it since `-p kaspa-hashes` is not an option we can add when doing `--doc`

.github/workflows/ci.yaml
coderofstuff: Done

.github/workflows/ci.yaml
coderofstuff: Nvm, `cargo test --doc --release -p kaspa-hashes --features=no-asm` would actually work. but with this result:

```
test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s
```

Assuming that in the future this may have docs, I'll add this step in.

.github/workflows/ci.yaml
coderofstuff: Removing `--benches` and keeping `-p kaspa-hashes` works so doing that now.

tiram88 (approved):

PR #312, opened 2023-11-02
Legacy (PWA, KDX) address derivation support

surinder83singh:
This PR adds legacy address derivation support (custom derivation path `972`, used by the legacy JavaScript framework and KDX/PWA applications) within the Rusty Kaspa wallet framework infrastructure. 

Other changes include:
- Fix for account import issues requiring wallet reload after account import
- Change `/test` script to use nextest to prevent integration test failures on MacOS
- Add various unit tests for account derivation processing
- Fix a panic in `checked_sub()` when transactions are created within 180 seconds from the web application start (due to instant update that switched to using `performance.now()` that measures time as a duration from the application start)
- Change `sendTransaction()` to use `allowOrphans: false` (as `allowOrphans` is now shielded by `unsafe-rpc`)


wallet/core/src/runtime/account/kind.rs
```diff
@@ -21,6 +21,32 @@ u8_try_from! {
     }
 }
 
+impl AccountKind {
+    pub fn is_legacy(&self) -> bool {
```
biryukovmaxim, line 25: I think deriving  eq+partial eq is much easier to implement and use than these helper functions

wallet/core/src/runtime/wallet.rs
```diff
@@ -725,10 +755,17 @@ impl Wallet {
     }
 
     pub async fn find_accounts_by_name_or_id(&self, pat: &str) -> Result<Vec<Arc<dyn Account>>> {
-        let accounts = self.active_accounts().inner().values().cloned().collect::<Vec<_>>();
+        let mut accounts = self.active_accounts().inner().values().cloned().collect::<Vec<_>>();
+        let legacy_accounts = self.legacy_accounts().inner().values().cloned().collect::<Vec<_>>();
+        accounts.extend(legacy_accounts);
```
biryukovmaxim, line 760: It can be done without allocating accounts and legacy accounts vectors.
Accounts_iter.chain(legacy_iter).filter().collect()

wallet/core/src/runtime/account/kind.rs
aspect: changes: https://github.com/kaspanet/rusty-kaspa/pull/312/commits/c56940a79f0a2627b55a05e1076f9990900cc9f3

kept this one though due to it being a complex type: https://github.com/aspectron/rusty-kaspa/blob/legacy/wallet/core/src/storage/account.rs#L154

wallet/core/src/runtime/wallet.rs
aspect: Thanks for noting this. This was actually a bug (albeit benign). We recently changed the implementation pattern to have all legacy accounts be active and thus present in the `active_accounts` list. As such this iteration was not even necessary (old code that was benign in nature). Fixed in https://github.com/kaspanet/rusty-kaspa/pull/312/commits/abc0298866e878c5b303efdff4ab870303717e98

aspect (approved):

PR #320, opened 2023-11-12
TN11 -- various fixes concluded from ongoing internal experiment

michaelsutton:
Partial list:

1. Bump version to 0.13.0 which is higher in order than current kaspad-go version (important for miners/tools who rely on node version for selecting the GRPC version)
2. Make RPC submit block not wait for virtual processing and fix corresponding tests which assume it does
3. IBD: fix fetch relay block past to include blocks in future(sink)
4. IBD: update pending relay block to the latest block, this should shorten the usual IBD loop when syncing a node (loop meaning there is a long IBD followed by a shorter one and so on until it decays and the node finally syncs; with this change it should decay faster)
5. IBD: syncer: fill DAG gaps in trusted blocks set; syncee: verify all blocks w/o body are in pruning point past

### DIFF
--- a/consensus/src/processes/pruning_proof/mod.rs
+++ b/consensus/src/processes/pruning_proof/mod.rs
@@ -1,7 +1,7 @@
 use std::{
     cmp::{max, Reverse},
-    collections::hash_map::Entry::Vacant,
-    collections::BinaryHeap,
+    collections::{hash_map::Entry, BinaryHeap},
+    collections::{hash_map::Entry::Vacant, VecDeque},
     ops::{Deref, DerefMut},
     sync::Arc,
 };
@@ -12,7 +12,7 @@ use parking_lot::{Mutex, RwLock};
 use rocksdb::WriteBatch;
 
 use kaspa_consensus_core::{
-    blockhash::{BlockHashExtensions, BlockHashes, ORIGIN},
+    blockhash::{self, BlockHashExtensions, BlockHashes, ORIGIN},
     errors::{
         consensus::{ConsensusError, ConsensusResult},
         pruning::{PruningImportError, PruningImportResult},
@@ -179,7 +179,7 @@ impl PruningProofManager {
         drop(pruning_point_write);
     }
 
-    pub fn apply_proof(&self, mut proof: PruningPointProof, trusted_set: &[TrustedBlock]) {
+    pub fn apply_proof(&self, mut proof: PruningPointProof, trusted_set: &[TrustedBlock]) -> PruningImportResult<()> {
         let pruning_point_header = proof[0].last().unwrap().clone();
         let pruning_point = pruning_point_header.hash;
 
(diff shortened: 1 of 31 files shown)


consensus/src/processes/traversal_manager.rs
```diff
@@ -85,13 +105,13 @@ impl<T: GhostdagStoreReader, U: ReachabilityStoreReader, V: RelationsStoreReader
...
                     queue.len(),
-                    anticone.len(),
+                    output.len(),
                     visited.len()
                 );
             }
 
-            if !self.reachability_service.is_dag_ancestor_of(block, current) {
-                anticone.push(current);
+            if !return_anticone_only || !self.reachability_service.is_dag_ancestor_of(block, current) {
```
coderofstuff, line 113: For code clarity, can we do:
```
let is_block_anticone_of_current = !self.reachability_service.is_dag_ancestor_of(block, current);

if !return_anticone_only || is_block_anticone_of_current {
  output.push(current);
}
```

It'd make it easier for whoever is reading this to understand the intent with self-documenting code.

protocol/flows/src/v5/ibd/flow.rs
```diff
@@ -385,8 +385,9 @@ impl IbdFlow {
...
         }
 
-        // Send a special header request for the selected tip anticone. This is expected to
-        // be a small set, as it is bounded to the size of virtual's mergeset.
+        // Send a special header request for the sink antipast. This is expected to
+        // be a relatively small set since virtual and relay blocks should be close topologically.
+        // See server-side handling of `RequestAnticone` for further details.
         self.router
             .enqueue(make_message!(
                 Payload::RequestAnticone,
```
someone235, line 393: Maybe change to `RequestAntiPast` and add a comment in the proto file that old versions can return only anticone?

testing/integration/src/common/client_notify.rs
```diff
@@ -0,0 +1,15 @@
+use async_channel::Sender;
+use kaspa_notify::notifier::Notify;
+use kaspa_rpc_core::Notification;
+
+#[derive(Debug)]
+pub struct ChannelNotify {
+    pub sender: Sender<Notification>,
```
someone235, line 7: I would make it private (accessible only via `new`)

someone235 (changes requested):

consensus/src/processes/pruning_proof/mod.rs
```diff
@@ -670,6 +683,26 @@ impl PruningProofManager {
         }
     }
 
+    /// Returns the k + 1 chain blocks below this hash (inclusive). If data is missing
+    /// the search is halted and a partial chain is returned.
+    ///
+    /// The returned hashes are guaranteed to have GHOSTDAG data
+    pub(crate) fn get_ghostdag_chain_k_depth(&self, hash: Hash) -> Vec<Hash> {
```
coderofstuff, line 690: Do we need this specifically to be just about `get_ghostdag_chain_k_depth`? A generic `get_ghostdag_chain_with_depth(&self, hash: Hash, depth: u64)` may make it useful for other possible use-cases, like if we ever need to dig deeper.

protocol/flows/src/v6/request_pruning_point_and_anticone.rs
```diff
@@ -0,0 +1,100 @@
...
+                .await?;
+
+            for hashes in trusted_data.anticone.chunks(IBD_BATCH_SIZE) {
+                for &hash in hashes {
+                    let block = session.async_get_block(hash).await?;
+                    self.router
+                        .enqueue(make_response!(
+                            Payload::BlockWithTrustedDataV4,
+                            // No need to send window indices in v6
+                            BlockWithTrustedDataV4Message { block: Some((&block).into()), ..Default::default() },
```
someone235, line 79: Add TODO in the proto file to remove the indices fields once v5 is obsolete

someone235, line 79: Add TODO in the proto file to remove the indices fields once v5 is obsolete

someone235 (changes requested):

someone235 (approved):

PR #324, opened 2023-11-16
Build pruning proof -- Bug fix

michaelsutton:
The following line of logic in [kaspad-go](https://github.com/kaspanet/kaspad/blob/master/domain/consensus/processes/pruningproofmanager/pruningproofmanager.go#L199) was missing from rust impl. This means that we were searching for common ancestor when not needed hence loosing contiguous topology guarantees known from proof validation or from ongoing relation maintenance. This in turn lead to not including `block_at_depth_2m` in the proof in some cases, thus deleting it prematurely

### DIFF
--- a/consensus/src/processes/pruning_proof/mod.rs
+++ b/consensus/src/processes/pruning_proof/mod.rs
@@ -480,11 +480,13 @@ impl PruningProofManager {
             }
 
             if level < self.max_block_level {
-                let block_at_depth_m_at_next_level = self.block_at_depth(
-                    &*ghostdag_stores[level_idx + 1],
-                    selected_tip_by_level[level_idx + 1].unwrap(),
-                    self.pruning_proof_m,
-                );
+                let block_at_depth_m_at_next_level = self
+                    .block_at_depth(
+                        &*ghostdag_stores[level_idx + 1],
+                        selected_tip_by_level[level_idx + 1].unwrap(),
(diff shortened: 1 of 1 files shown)


consensus/src/processes/pruning_proof/mod.rs
```diff
@@ -656,32 +665,39 @@ impl PruningProofManager {
             .collect_vec()
     }
 
-    fn block_at_depth(&self, ghostdag_store: &impl GhostdagStoreReader, high: Hash, depth: u64) -> Hash {
-        let high_gd = ghostdag_store.get_compact_data(high).unwrap();
+    fn block_at_depth(&self, ghostdag_store: &impl GhostdagStoreReader, high: Hash, depth: u64) -> Result<Hash, String> {
```
someone235, line 668: It's best practice to return an error that implements the `Error` trait. You can also keep the previous function signature and use `unwrap_or_else` instead of `map_err` so you can format the panic message. (same comment for the other functions)

someone235 (changes requested):

consensus/src/processes/pruning_proof/mod.rs
michaelsutton: I want the panic at the caller for additional info. Added a small Error wrapper enum

someone235 (approved):

PR #327, opened 2023-11-20
Allow addition of path when specifying RPC url

starkbamse:
**Changes**

- Changed the struct `ParseHostOutput` to include path
- Modified the format of the returned rpc url after parsing to include path

**Why**

To ensure Kaspa adoption and increased ecosystem development it is important to allow the addition of path when specifying a RPC url. To understand this from another perspective, as a node provider that is running multiple blockchains from the same hostname, it is crucial for me to be able to redirect requests to different chains based on path. E.g. abc.com/kaspa, abc.com/eth.

- [ ] /check no errors.


aspect (approved):
Looks good.

PR #333, opened 2023-11-24
Initialize FD limits on MacOS and Linux platforms at startup

aspect:
MacOS allows applications to set FD limits. The same should work in Linux if you are running as a root.

This PR adds an initializer that passively attempts to configure the needed settings and warns the user if it cannot do so (the output needs to be `println!()` as the logging subsystem is not yet initialized.

### DIFF
--- a/kaspad/src/daemon.rs
+++ b/kaspad/src/daemon.rs
@@ -37,7 +37,7 @@ pub const DESIRED_DAEMON_SOFT_FD_LIMIT: u64 = 16 * 1024;
 /// process. (Rusty Kaspa will operate with the minimal
 /// acceptable limit of `4096`, but a setting below
 /// this value may impact the database performance).
-pub const MINIMUM_DAEMON_SOFT_FD_LIMIT: u64 = 4096;
+pub const MINIMUM_DAEMON_SOFT_FD_LIMIT: u64 = 4*1024;
 
 use crate::args::Args;
 
--- a/kaspad/src/daemon.rs
+++ b/kaspad/src/daemon.rs
@@ -35,7 +35,7 @@ use kaspa_wrpc_server::service::{Options as WrpcServerOptions, ServerCounters as
 pub const DESIRED_DAEMON_SOFT_FD_LIMIT: u64 = 16 * 1024;
 /// Minimum acceptable soft FD limit for the kaspad
 /// process. (Rusty Kaspa will operate with the minimal
-/// acceptable limit of `1024`, but a setting below
+/// acceptable limit of `4096`, but a setting below
 /// this value may impact the database performance).
 pub const MINIMUM_DAEMON_SOFT_FD_LIMIT: u64 = 4096;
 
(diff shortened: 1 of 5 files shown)


utils/src/fd_budget.rs
```diff
@@ -50,6 +50,17 @@ pub fn acquire_guard(value: i32) -> Result<FDGuard, Error> {
     }
 }
 
+#[cfg(not(any(target_arch = "wasm32", target_os = "windows")))]
+pub fn ensure_os_limits(user_soft_limit: u64, user_hard_limit: u64) -> std::io::Result<(u64, u64)> {
+    let (soft, hard) = rlimit::Resource::NOFILE.get()?;
+    if soft < user_soft_limit || hard < user_hard_limit {
+        rlimit::Resource::NOFILE.set(user_soft_limit, user_hard_limit)?;
```
biryukovmaxim, line 57: It should be max(user_hard/soft_limit, soft/hard) otherwise it can decrease hard limit even if it was okay.

kaspad/src/main.rs
```diff
@@ -17,6 +20,22 @@ pub fn main() {
...
 
     let args = parse_args();
+
+    match fd_budget::try_set_fd_limit(DESIRED_DAEMON_SOFT_FD_LIMIT) {
+        Ok(limit) => {
+            if limit < MINIMUM_DAEMON_SOFT_FD_LIMIT {
+                println!("Current OS file descriptor limit (soft FD limit) is set to {limit}");
+                println!("The kaspad node requires a setting of at least {DESIRED_DAEMON_SOFT_FD_LIMIT} to operate properly.");
+                println!("Please increase the limits using the following command:");
+                println!("ulimit -n {DESIRED_DAEMON_SOFT_FD_LIMIT}");
```
coderofstuff, line 30: Is it worth making a recommendation in README.md so that this can be done intentionally before the user uses kaspad?

kaspad/src/daemon.rs
```diff
@@ -30,6 +30,15 @@ use kaspa_perf_monitor::builder::Builder as PerfMonitorBuilder;
...
 use kaspa_wrpc_server::service::{Options as WrpcServerOptions, ServerCounters as WrpcServerCounters, WrpcEncoding, WrpcService};
 
+/// Desired soft FD limit that needs to be configured
+/// for the kaspad process.
+pub const DESIRED_DAEMON_SOFT_FD_LIMIT: u64 = 16 * 1024;
+/// Minimum acceptable soft FD limit for the kaspad
+/// process. (Rusty Kaspa will operate with the minimal
+/// acceptable limit of `1024`, but a setting below
+/// this value may impact the database performance).
+pub const MINIMUM_DAEMON_SOFT_FD_LIMIT: u64 = 4096;
```
coderofstuff, line 40: The comment says `1024` but this value is `4096`. Why the difference? And if it really is `4096` can we express it as `4 * 1024` to be consistent with the above?

aspect: I made the comment say 4096 and the declaration to be 4*1024.  I use full numbers below 10k at which point it becomes hard to read.

These values perhaps should be changed down the road.  I am defining 4096 from my local observations on T10 where a typical rusty kaspa consumption without limit currently ends up being slightly over 3k.  It is possible that it will be more on the current mainnet.

kaspad/src/main.rs
aspect: Perhaps.  I don't want to touch the readme right now because of another outstanding PR that needs a review.  Now that we have discovered that this works I am much less concerned about this as it appears that rusty will now self-configure in most cases and will only fail to do this in some unusual circumstances, which I am currently not sure what they are... I think the upcoming T11 should show if there are any issues with any types of exotic hardware/setups.

michaelsutton (approved):

PR #335, opened 2023-11-26
Add request/response size middlewares and dependencies

biryukovmaxim:
Introduced a new module `request_response_size_middlewares` to utils library to count the sizes of request and response bodies. Furthermore, added several dependencies to Cargo.toml and updated Cargo.lock necessary for new functionality. This includes libraries like 'hyper', 'tower', and 'pin-project-lite'. The new implementation will provide us with exact sizes of data transferred.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1715,6 +1715,12 @@ dependencies = [
 ]
 
 [[package]]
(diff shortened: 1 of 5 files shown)


aspect (approved):
Looks good.

PR #344, opened 2023-12-02
Remove File Descriptor init on Windows

aspect:
It appears that instead of gracefully handling errors, on Windows `fdlimit` is causing a crash resulting in the daemon halting with `buffer overflow` error without any kind of a panic, even setting `RUST_BACKTRACE=1` is not helping.  With the help of `@Callidon` on Discord #development we narrowed it down to the changes in #333.  In this PR I've isolated the code to run on MacOS and Linux only (as me and @biryukovmaxim have tested it on these platforms manually).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3081,6 +3081,7 @@ name = "kaspad"
 version = "0.13.0"
 dependencies = [
  "async-channel 2.0.0",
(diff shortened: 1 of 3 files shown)


aspect:
We discovered that windows max fd value is 8192. Prepping another PR to handle this.

PR #349, opened 2023-12-07
[WIP] Implementation of WASM Bindings for TypeScript

andreidev1:
At this moment most of the rust bindings that have been already implemented point to Nodejs. By default these types are declared as being `any`.
The purpose of this PR is to create rust bindings for TypeScript so other developers will not get confused anymore.

### DIFF
--- a/wallet/core/src/wasm/tx/generator/generator.rs
+++ b/wallet/core/src/wasm/tx/generator/generator.rs
@@ -7,6 +7,19 @@ use crate::wasm::tx::generator::*;
 use crate::wasm::wallet::Account;
 use crate::wasm::UtxoContext;
 
+#[wasm_bindgen(typescript_custom_section)]
+const IGeneratorSettingsObject: &'static str = r#"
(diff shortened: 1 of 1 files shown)


aspect:
This is the correct solution. There are two ways of doing this. One by binding rust generated types and another is by using this approach. Rust generated types are very inflexible (you can't have Address[] | string[] bounds to a single type). Due to that inflexibility we had to use JsValue with run-time analysis of arguments. JsValue comes across as 'any'. So we are gradually working to re-seat all WASM32 bindings across the entire project onto TS-declared type definitions.

The problem presented itself when developers started to integrate bindings into TypeScript projects. Our examples were built in JS assuming that TS will just work (which it does) but the lack of TS types is problematic, hence this refactoring is gradually taking place across multiple branches.

Also in case of RPC that uses serde-wasm-bindgen typing is not possible in any other way than this.

aspect:
This was folded into #349. Closing.

PR #351, opened 2023-12-11
Bounding node RAM -- implementing a cache policy which tracks dynamic data sizes

michaelsutton:
The current cache implementation bounds the number of cached elements by *unit* count. This works perfectly for constant-size elements, however when dealing with data with varying size (e.g., headers, ghostdag data, relations etc) it suffers from a size-estimation tradeoff -- on the one hand you want to estimate a unit size by its average case so that you allocate enough units for normal node operation, and on the other hand doing so might overfill the RAM when abnormal data arrives. 

This pull-request implements a cache policy which includes a `tracked` mode which tracks dynamic data sizes. This way cache allocations can be monitored accurately resulting in tight management of resources also under abnormal conditions. The cache can now define a hard-bound limit in bytes and inserted/removed elements are tracked in real-time by their actual runtime memory requirements.   

Remaining tasks: 
- Ideally add a cmd flag with an overall RAM bound which the user can provide (will be respected by consensus storage; will not guarantee other parts of the system)


coderofstuff:
Fixes https://github.com/kaspanet/rusty-kaspa/issues/352

consensus/src/model/stores/relations.rs
```diff
@@ -419,7 +419,7 @@ mod tests {
     #[test]
     fn test_db_relations_store() {
         let (lt, db) = create_temp_db!(kaspa_database::prelude::ConnBuilder::default().with_files_limit(10));
-        test_relations_store(DbRelationsStore::new(db, 0, 2));
+        test_relations_store(DbRelationsStore::new(db, 0, CachePolicy::Tracked(2)));
```
coderofstuff, line 422: What's the logic between choose where we use `Tracked` and where we'd use `Unit`?

utils/src/mem_size.rs
```diff
@@ -0,0 +1,66 @@
+use std::{collections::HashSet, sync::Arc};
+
+use parking_lot::RwLock;
+
+#[derive(Clone, Copy)]
+pub enum MemSize {
```
coderofstuff, line 6: Please add comments about intended use and where/when using this would be relevant.

consensus/src/model/stores/headers.rs
```diff
@@ -25,6 +27,14 @@ pub struct HeaderWithBlockLevel {
     pub block_level: BlockLevel,
 }
 
+impl MemSizeEstimator for HeaderWithBlockLevel {
+    fn estimate_mem_size(&self) -> MemSize {
```
coderofstuff, line 31: Please add comments why this is the correct estimation for the memsize of this object

consensus/src/model/stores/ghostdag.rs
```diff
@@ -35,6 +37,17 @@ pub struct CompactGhostdagData {
     pub selected_parent: Hash,
 }
 
+impl MemSizeEstimator for GhostdagData {
+    fn estimate_mem_size(&self) -> MemSize {
```
coderofstuff, line 41: Please add comments why this is the correct estimation for this object

consensus/src/consensus/storage.rs
```diff
@@ -74,91 +74,121 @@ impl ConsensusStorage {
...
+        ));
+        let depth_store = Arc::new(DbDepthStore::new(db.clone(), CachePolicy::Unit(noise(perf_params.header_data_cache_size))));
         let selected_chain_store =
-            Arc::new(RwLock::new(DbSelectedChainStore::new(db.clone(), noise(perf_params.header_data_cache_size))));
+            Arc::new(RwLock::new(DbSelectedChainStore::new(db.clone(), CachePolicy::Unit(noise(perf_params.header_data_cache_size)))));
 
         // Pruning
         let pruning_point_store = Arc::new(RwLock::new(DbPruningStore::new(db.clone())));
-        let past_pruning_points_store = Arc::new(DbPastPruningPointsStore::new(db.clone(), 4));
+        let past_pruning_points_store = Arc::new(DbPastPruningPointsStore::new(db.clone(), CachePolicy::Unit(1024)));
```
coderofstuff, line 166: Why did this change from 4 to 1024?

coderofstuff (changes requested):
These changes have been running on mainnet and tn11 stable for the last 24hrs

utils/src/mem_size.rs
```diff
@@ -0,0 +1,117 @@
...
+    Units,
+}
+
+/// The contract for estimating deep memory size owned by this object. Implementors
+/// are expected to support only a single function - bytes or units. Objects with pre-compliation
+/// known static size or which are containers of items with static size should implement the `_units`
+/// estimation and return the number of logical items (usually 1 or the number of items in
+/// the container). Objects with varying runtime sizes should implement the `_bytes` estimation.
+///
+/// By panicking on the remaining unimplemented function we ensure that tests will catch any inconsistency over the
```
coderofstuff, line 22: > By panicking on the remaining unimplemented function

The remaining function is not necessarily always unimplemented (such as in the implementations for `Arc<T>` and `RwLock<T>` below). Please add a comment about in what instances it would be acceptable that both are implemented.

coderofstuff (changes requested):
I have very minor comments here. Functionality-wise this has been running stable on TN11 for a few hours now and I'll keep it running.

components/addressmanager/src/lib.rs
```diff
@@ -359,7 +359,7 @@ mod address_store_with_cache {
 
     impl Store {
         fn new(db: Arc<DB>) -> Self {
-            let db_store = DbAddressesStore::new(db, 0);
+            let db_store = DbAddressesStore::new(db, CachePolicy::Empty);
```
coderofstuff, line 362: Not specifically related to this PR, but the fact that this store's cache is empty, that it's not a test AND it's very specifically a mod that says `address_store_with_cache` makes me think this use of cache here needs to be re-examined (even outside the context of this PR).

utils/src/mem_size.rs
michaelsutton: Arc and RwLock are simple delegations. The underlying object is expected to implement only a single method. I'll comment

components/addressmanager/src/lib.rs
michaelsutton: Explained in a comment

database/src/cache.rs
```diff
@@ -1,75 +1,214 @@
...
+where
+    TKey: Clone + std::hash::Hash + Eq + Send + Sync,
+    TData: Clone + Send + Sync + MemSizeEstimator,
+    S: BuildHasher + Default,
+{
+    /// Evicts items until meeting cache policy requirements
+    fn evict(&mut self, policy: &CachePolicyInner) {
+        // We allow passing tracked size limit as long as there are no more than min_units units
+        while self.tracked_size > policy.max_size && self.map.len() > policy.min_units {
+            if let Some((_, v)) = self.map.swap_remove_index(rand::thread_rng().gen_range(0..self.map.len())) {
```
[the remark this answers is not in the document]
michaelsutton: Our experience shows that randomness is good enough at not hitting recent elements (w.h.p. if `cache size >> hot cache area size`) while any other form of maintaining sorted timestamps etc becomes more expensive (much more expensive)

database/src/cache.rs
```diff
@@ -1,75 +1,214 @@
 use indexmap::IndexMap;
+use kaspa_utils::mem_size::{MemMode, MemSizeEstimator};
 use parking_lot::RwLock;
 use rand::Rng;
 use std::{collections::hash_map::RandomState, hash::BuildHasher, sync::Arc};
 
+#[derive(Debug, Clone, Copy)]
+pub enum CachePolicy {
+    /// An empty cache (avoids aqcuiring locks etc so considred perf-free)
```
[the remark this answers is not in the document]
michaelsutton: Fixed

database/src/cache.rs
```diff
@@ -1,75 +1,214 @@
...
+    fn insert(&mut self, policy: &CachePolicyInner, key: TKey, data: TData) {
+        if policy.tracked {
+            let new_data_size = data.estimate_size(policy.mem_mode);
+            self.tracked_size += new_data_size;
+            if let Some(removed) = self.map.insert(key, data) {
+                self.tracked_size -= removed.estimate_size(policy.mem_mode);
+            }
+            self.evict(policy);
+        } else {
+            if self.map.len() == policy.max_size {
```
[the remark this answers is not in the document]
michaelsutton: I added a `MemMode::Undefined` and renamed `CachePolicy::Units` to `CachePolicy::Count` to avoid confusion. Non tracking policy means by definition that we don't care about the memory mode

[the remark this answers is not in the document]
michaelsutton: Re `self.map.len() > policy.max_size` the current logic is tight in the sense that it never passes the limit (bcs we check and evict before insertion). If we would change to `>=` then we also need a `while` loop. To me it seems more safe to keep the logic tight since this code should rarely change. Wdyt @someone235 ?

consensus/src/consensus/storage.rs
michaelsutton: They are simple hashes and might be accessed by `estimate_timestamp_from_daa_score` rpc calls. So I don't see a reason not to cover `~1000` points

database/src/cache.rs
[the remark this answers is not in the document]
someone235: I agree that we should keep it as is to avoid the `while`

database/src/cache.rs
```diff
@@ -1,75 +1,216 @@
...
+
+impl<TKey, TData, S> Inner<TKey, TData, S>
+where
+    TKey: Clone + std::hash::Hash + Eq + Send + Sync,
+    TData: Clone + Send + Sync + MemSizeEstimator,
+    S: BuildHasher + Default,
+{
+    /// Evicts items until meeting cache policy requirements
+    fn evict(&mut self, policy: &CachePolicyInner) {
+        // We allow passing tracked size limit as long as there are no more than min_items items
```
someone235, line 67: nit: Add `assert(self.tracked)`

someone235 (changes requested):

database/src/cache.rs
```diff
@@ -1,75 +1,192 @@
...
+            }
+            CachePolicy::Tracked(max_size) => {
+                let new_data_size = data.estimate_mem_size().agnostic_size();
+                inner.tracked_size += new_data_size;
+                if let Some(removed) = inner.map.insert(key, data) {
+                    inner.tracked_size -= removed.estimate_mem_size().agnostic_size();
+                }
+
+                while inner.tracked_size > max_size {
+                    if let Some((_, v)) = inner.map.swap_remove_index(rand::thread_rng().gen_range(0..inner.map.len())) {
```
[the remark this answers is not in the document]
michaelsutton: Perhaps in some other chance

consensus/core/src/tx.rs
```diff
@@ -44,6 +45,12 @@ impl UtxoEntry {
     }
 }
 
+impl MemSizeEstimator for UtxoEntry {
+    fn estimate_mem_units(&self) -> usize {
```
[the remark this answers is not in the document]
michaelsutton: Added TODO

consensus/core/src/utxo/utxo_diff.rs
```diff
@@ -17,6 +18,12 @@ pub struct UtxoDiff {
     pub remove: UtxoCollection,
 }
 
+impl MemSizeEstimator for UtxoDiff {
+    fn estimate_mem_bytes(&self) -> usize {
+        size_of::<Self>() + (self.add.len() + self.remove.len()) * (size_of::<TransactionOutpoint>() + size_of::<UtxoEntry>())
```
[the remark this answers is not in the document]
michaelsutton: Added TODO

database/src/cache.rs
michaelsutton: Will panic on `MemMode::Undefined`

consensus/src/processes/pruning_proof/mod.rs
```diff
@@ -383,19 +382,25 @@ impl PruningProofManager {
         if proof.len() != self.max_block_level as usize + 1 {
             return Err(PruningImportError::ProofNotEnoughLevels(self.max_block_level as usize + 1));
         }
+        if proof[0].is_empty() {
```
coderofstuff, line 385: This change doesn't seem to be related to caches. Was it a bug that this wasn't here before, or did the cache changes introduce the need to add this check?

coderofstuff (approved):
I've had this running stable for 24hrs+ on my TN11 node.

consensus/src/consensus/storage.rs
```diff
@@ -74,91 +73,158 @@ impl ConsensusStorage {
...
-            200_000_000u64,
-            size_of::<GhostdagData>() + perf::approx_mergeset_size(params) * size_of::<Hash>() * 2,
-        );
+        // Lower and upper bounds
+        let pruning_depth = params.pruning_depth as usize;
+        let pruning_size_for_caches = (params.pruning_depth + params.finality_depth) as usize; // Upper bound for any block/header related data
+        let level_lower_bound = 2 * params.pruning_proof_m as usize; // Number of items lower bound for level-related caches
+
+        // Budgets in bytes. All byte budgets overall sum up to ~1GB of memory (which obviously takes more low level alloc space)
+        let daa_excluded_budget = 30_000_000;
```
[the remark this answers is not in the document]
michaelsutton: Depending on the importance of the store (frequency and where during processing it is accessed) and on the expected sizes. More or less. + Trial and error

consensus/src/processes/pruning_proof/mod.rs
michaelsutton: I'm pretty sure it is the first option, but even if it is checked before this call it's better to check it here as well

consensus/src/consensus/cache_policy_builder.rs
```diff
@@ -0,0 +1,144 @@
...
+impl Default for CachePolicyBuilder {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+impl CachePolicyBuilder {
+    pub fn new() -> Self {
+        Self {
+            bytes_budget: usize::MAX,
```
someone235, line 39: nit: It seems the `MAX` value is never passed to the resulted `CachePolicy`, so I would use `Option<usize>` instead

someone235 (approved):

Issue #354, opened 2023-12-13
Revisit definition of "Synced"


coderofstuff:
For now make shorter as function of bps. Consider conditioning config on testnet due to hash instability.

Another possibility: identify the case where headers selected tip is too far above sink (which happens in IBD headers phase) and consider it not synced.

coderofstuff:
Fixed in #356

PR #356, opened 2023-12-13
TN11 parameter fine-tuning

michaelsutton:


protocol/flows/src/v5/blockrelay/flow.rs
```diff
@@ -251,7 +251,10 @@ impl HandleRelayInvsFlow {
             .await?;
         let msg = dequeue_with_timeout!(self.msg_route, Payload::BlockLocator)?;
         let locator_hashes: Vec<Hash> = msg.try_into()?;
-        for h in locator_hashes {
+        // Locator hashes are sent from later to earlier, so it makes sense to query consensus in reverse. Technically
+        // with current syncer-side implementations (in both go-kaspa and this codebase) we could query only the last one,
+        // but we prefer not relying on such details for correctness
```
someone235, line 256: Should we add a TODO to send only one block in the next protocol version?

someone235 (changes requested):

protocol/flows/src/v5/blockrelay/flow.rs
michaelsutton: not sure you need a protocol version

consensus/core/src/blockstatus.rs
```diff
@@ -43,4 +43,8 @@ impl BlockStatus {
     pub fn is_valid(self) -> bool {
         self != BlockStatus::StatusInvalid
     }
+
+    pub fn is_invalid(self) -> bool {
+        self == BlockStatus::StatusInvalid
```
coderofstuff, line 48: Optional: seems strange that this isn't just `!is_valid()`

coderofstuff:
Related to #354

consensus/core/src/blockstatus.rs
michaelsutton: I find it more readable

someone235 (approved):

Issue #355, opened 2023-12-13
Optimize RPC Submit Block with more handlers


coderofstuff:
Have a small pool of handlers (~5) for this message type.

If handlers are not keeping up, reject block submissions (instead of awaiting the distribution channel which blocks other message types such as get block template hence getting the miner stuck on old template)

coderofstuff:
Fixed by  #357

PR #357, opened 2023-12-13
Process SubmitBlock requests in parallel

tiram88:
Following changes to the gRPC server:

1. When the node BPS is greater than 1 (ie. 10 BPS in testnet-11), processes `SubmitBlock` requests in parallel with as many workers as BPS.
2. Rejects incoming blocks when the `SubmitBlock` handler queue is full.


rpc/grpc/server/src/connection.rs
```diff
@@ -121,11 +146,19 @@ impl Router {
...
-            Ok(_) => Ok(()),
-            Err(_) => Err(GrpcServerError::ClosedHandler(rpc_op)),
+        match method.routing_policy() {
+            RoutingPolicy::Enqueue => match sender.send(request).await {
+                Ok(_) => Ok(()),
+                Err(_) => Err(GrpcServerError::ClosedHandler(rpc_op)),
+            },
+            RoutingPolicy::DropIfFull => match sender.try_send(request) {
+                Ok(_) => Ok(()),
+                Err(MpmcTrySendError::Full(_)) => Err(GrpcServerError::RouteIsFull(rpc_op)),
```
michaelsutton, line 159: I think this should not lead to a disconnection but rather returned to the client as RpcError

rpc/grpc/server/src/connection.rs
```diff
@@ -121,11 +146,19 @@ impl Router {
             return Err(GrpcServerError::InvalidRequestPayload);
         }
         let rpc_op = request.payload.as_ref().unwrap().into();
+        let method = self.interface.get_method(&rpc_op);
```
michaelsutton, line 149: getting the method per request seems expensive to me. Maybe keep the policy in the routing map or do a static mapping from rpc_op to policy

rpc/grpc/server/src/connection.rs
michaelsutton: actually as `SubmitBlockReport::Reject`  @tiram88

rpc/grpc/server/src/service.rs
```diff
@@ -13,6 +13,7 @@ const GRPC_SERVICE: &str = "grpc-service";
 
 pub struct GrpcService {
     net_address: NetAddress,
+    bps: u64,
```
coderofstuff, line 16: Optional: Is it possible that the underlying implementation will care about anything else in `config` other than bps? Maybe some other config later on that may be used to drive some other behavior. If so, it might be worth passing `config` here instead.

It's kinda strange to see `bps` being a specific parameter of a grpc service.

rpc/grpc/server/src/request_handler/factory.rs
```diff
@@ -122,6 +127,10 @@ impl Factory {
             });
         interface.replace_method(KaspadPayloadOps::NotifyFinalityConflict, method);
 
+        // Methods with special properties
+        let bps = bps as usize;
+        interface.replace_method_properties(KaspadPayloadOps::SubmitBlock, bps, 10.max(bps * 2), RoutingPolicy::DropIfFull);
```
coderofstuff, line 132: `tasks` is directly set using bps. Should there be some upper bound for `tasks` (consequently the handlers which cares about this number) or is it expected that it must grow linearly with bps?

coderofstuff:
Fixes #355

rpc/grpc/server/src/adaptor.rs
```diff
@@ -33,13 +33,14 @@ impl Adaptor {
 
     pub fn server(
         serve_address: NetAddress,
+        bps: u64,
```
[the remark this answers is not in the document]
tiram88: ok, makes sense

rpc/grpc/server/src/service.rs
tiram88: This was saving a direct crate dependency but I also like the config approach. Let's do that.

rpc/grpc/server/src/request_handler/factory.rs
tiram88: This will need further fine tuning when we explore higher BPS. For now, it's ok like that I guess.

rpc/grpc/server/src/connection.rs
tiram88: ok, let's store policies in the routing map

rpc/grpc/server/src/connection.rs
tiram88: done

michaelsutton (approved):

PR #361, opened 2023-12-19
Move to mimalloc

someone235:
Since rusty-kaspa uses a lot of allocations, the RSS remains high even after the memory is free. This should be mitigated by using mimalloc which utilizes the use of MADV_FREE and MADV_DONTNEED (and equivalent techniques on non unix-like systems) which tend to release memory earlier and help in keeping the RSS small.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1132,16 +1132,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "ctor"
-version = "0.2.6"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "30d2b3721e861707777e3195b0158f950ae6dc4a27e4d02ff9f67e3eb3de199e"
-dependencies = [
- "quote",
- "syn 2.0.41",
-]
-
-[[package]]
 name = "ctrlc"
 version = "3.4.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2870,7 +2860,6 @@ dependencies = [
  "bincode",
  "criterion",
  "crossbeam-channel",
- "ctor",
  "faster-hex 0.6.1",
  "flate2",
  "futures-util",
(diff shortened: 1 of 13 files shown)


elichai:
Why only unix?

michaelsutton:
1. we should also move the C API call to a more accessible module and should probably set it (the allocator and the `purge_decommits` setting) for tests as well
2. should we leave the version free `version = "*"` @elichai ?

elichai:
> 2. should we leave the version free `version = "*"` @elichai ?

Hmm no, in case some of the API(the C API or the allocator name) changes we don't want our code to just break

aspect:
We don't want any dependencies to be `"*"` as that exposes the project to supply-chain attacks whenever `Cargo.lock` is reset.

aspect:
If mimalloc is to be kept, can its initialization please be moved to `daemon.rs` or better, a dedicated `kaspad/src/memory.rs` module (which is then invoked from `main.rs` via a public fn). If this becomes a requirement for rusty kaspa, it should be loadable externally alongside the daemon itself (with pre-configured parameters within this module/loader fn, not in main).

elichai:
> If mimalloc is to be kept, can its initialization please be moved to `daemon.rs` or better, a dedicated `kaspad/src/memory.rs` module (which is then invoked from `main.rs` via a public fn). If this becomes a requirement for rusty kaspa, it should be loadable externally alongside the daemon itself (with pre-configured parameters within this module/loader fn, not in main).

that might not work well if you have C dependencies, if you want the `override` feature to work properly then mimalloc has to be passed before any C dependency, meaning it should be the first dependency of the top-most crate

Note that this is only for the C deps, for the Rust deps it shouldn't matter where in the dep graph are you setting the global allocator

elichai (dismissed):
LGTM

michaelsutton (approved):

PR #365, opened 2023-12-26
remove redundant files.

D-Stacks:
These seem redundant and can probaby be removed


coderofstuff (approved):
Yup, these files aren't in the `lib.rs`. There's also something else in `indexes/core` that look to be the updated and in-use versions of these.

michaelsutton (approved):

PR #366, opened 2023-12-27
cargo-machete: clean-up  and integration into the `check` workflow.

D-Stacks:
This PR:

1) Removes unused deps, as found by cargo-machete,
2) But Keeps false positives, which where manually tested to be required for compilation
3) Adds ignores for these false positives. 
4) Integrates cargo machete into the `checks` workflow in github.

Future use:

To ignore a false positive, which one thinks is errnously flagged by cargo machete, place the deps name under `[package.metadata.cargo-machete]` in the corrosponding `Cargo.toml` file.


D-Stacks:
probably out of date by now..

PR #369, opened 2023-12-28
fixed parse failure for subdomains on parse_host function

elertan:


aspect (approved):
Many thanks for addressing this promptly!  Much appreciated!

PR #373, opened 2024-01-03
Orphans improvements part 2

michaelsutton:


protocol/flows/src/flowcontext/orphans.rs
```diff
@@ -63,11 +78,45 @@ impl OrphanBlocksPool {
...
-            debug!("Orphan blocks pool size exceeded. Evicting a random orphan block.");
-            // Evict a random orphan in order to keep pool size under the limit
-            if let Some((evicted, _)) = self.orphans.swap_remove_index(rand::thread_rng().gen_range(0..self.max_orphans)) {
-                debug!("Evicted {} from the orphan blocks pool", evicted);
+            let mut eviction_succeeded = false;
+            debug!("Orphan blocks pool size exceeded. Trying to evict a random orphan block.");
+            // Retry up to a logarithmic number of times
+            for i in 0..self.max_orphans_log {
+                // Evict a random orphan in order to keep pool size under the limit
+                let rand_index = rand::thread_rng().gen_range(0..orphan_ancestors.len());
```
someone235: Switch to orphan pool size

protocol/flows/src/flowcontext/orphans.rs
```diff
@@ -90,19 +139,24 @@ impl OrphanBlocksPool {
...
+        } else {
+            OrphanOutput::Unknown
         }
-        self.get_orphan_roots(consensus, orphan).await
     }
 
-    /// Internal get roots method. Assumes 'orphan' is within the pool
-    async fn get_orphan_roots(&self, consensus: &ConsensusProxy, orphan: Hash) -> OrphanOutput {
+    /// Internal get roots method
+    async fn get_orphan_roots(&self, consensus: &ConsensusProxy, mut queue: VecDeque<Hash>) -> FindRootsOutput {
```
someone235, line 153: Some explanation about what is `queue` would be nice

components/consensusmanager/src/batch.rs
```diff
@@ -0,0 +1,30 @@
+use kaspa_consensus_core::{api::BlockValidationFuture, block::Block};
+use std::fmt::Debug;
+
+pub struct BlockProcessingBatch {
```
someone235, line 4: I think it's more readable if you wrap the whole thing with an `Option`, instead of internal fields?

someone235 (changes requested):

someone235 (approved):

components/consensusmanager/src/batch.rs
michaelsutton: Sometimes only one of the task vecs is taken and the others remain for future processing

PR #374, opened 2024-01-04
Add a RAM usage scale factor configurable from the command line

michaelsutton:
- Adds a RAM usage scale factor configurable from the command line. The scale factor is applied to memory allocation bounds. Nodes with limited RAM (~4-8GB) should set this to ~0.3-0.5 respectively. Nodes with a large RAM (~64GB) can set this value to ~3.0-4.0 and gain superior performance especially for syncing peers faster on high-BPS networks
- Changes block txs and acceptance data stores to use tracked bytes mode
- Reduces orphan block log flood (appears in some cases on start-up)
- Benchmarks using simpa show that normal operation remains performant also if setting `--ram-scale=0.3` (but below that perf starts degrading)

### DIFF
--- a/components/addressmanager/src/stores/address_store.rs
+++ b/components/addressmanager/src/stores/address_store.rs
@@ -18,11 +18,7 @@ pub struct Entry {
     pub address: NetAddress,
 }
 
-impl MemSizeEstimator for Entry {
-    fn estimate_mem_units(&self) -> usize {
-        1
-    }
-}
+impl MemSizeEstimator for Entry {}
 
 pub trait AddressesStoreReader {
     fn get(&self, key: AddressKey) -> Result<Entry, StoreError>;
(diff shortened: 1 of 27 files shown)


biryukovmaxim (approved):

someone235 (approved):

humoud19801:
In summary, this pull request provides users with the ability to configure a scale factor for RAM usage, aims to improve the efficiency of block transactions and acceptance data stores, and addresses log flooding issues related to orphan blocks. The provided benchmarks suggest that the system remains performant under normal operation, but caution is advised when setting the RAM scale factor below a certain threshold.

jorchjorch:
Does deserve to use this parameter in TN11 tests?

Maybe only should run hardware with enough memory to run testnet?

PR #376, opened 2024-01-05
Add timestamps to IBD progress

someone235:


protocol/flows/src/v5/ibd/flow.rs
```diff
@@ -538,9 +546,10 @@ staging selected tip ({}) is too small or negative. Aborting IBD...",
         &mut self,
         consensus: &ConsensusProxy,
         chunk: &[Hash],
-    ) -> Result<(Vec<BlockValidationFuture>, u64), ProtocolError> {
+    ) -> Result<(Vec<BlockValidationFuture>, u64, u64), ProtocolError> {
```
biryukovmaxim, line 549: what about using type alias or wrapping to struct with fields to make function signature clear?
what is u64 and u64? that's not clear without reading function body

michaelsutton (dismissed):

michaelsutton (approved):

PR #378, opened 2024-01-07
Add a pruning exit point on advance_pruning_utxoset

someone235:


consensus/src/pipeline/pruning_processor/processor.rs
```diff
@@ -211,8 +221,10 @@ impl PruningProcessor {
         drop(pruning_utxoset_write);
 
         if self.config.enable_sanity_checks {
+            info!("Doing a sanity check that the new UTXO set has the expected UTXO commitment");
```
michaelsutton, line 224: maybe Doing -> Performing

michaelsutton (dismissed):

michaelsutton (approved):

PR #384, opened 2024-01-07
Feature gate MiMalloc use in MacOS

coderofstuff:
MiMalloc will still be the default to use for `linux` and `windows` regardless of whether the `mimalloc` feature is set or not.

However, for `macos`, MiMalloc will only be used if the feature `mimalloc` is specified. This is to alleviate the recent segfaulting issue found in mac related to mimalloc.


coderofstuff:
Superseded by #388

PR #387, opened 2024-01-11
Implement "nogrpc" flag and change default rpc listen ip

coderofstuff:
- Add nogrpc flag: When set, the gRPC service is not initialized
- Set default grpc listen to 127.0.0.1

### DIFF
--- a/kaspad/src/daemon.rs
+++ b/kaspad/src/daemon.rs
@@ -469,12 +469,12 @@ do you confirm? (answer y/n or pass --yes to the Kaspad command line to confirm 
     if let Some(port_mapping_extender_svc) = port_mapping_extender_svc {
         async_runtime.register(Arc::new(port_mapping_extender_svc))
     };
+    async_runtime.register(rpc_core_service.clone());
     if !args.disable_grpc {
         let grpc_service =
             Arc::new(GrpcService::new(grpc_server_addr, config, rpc_core_service.clone(), args.rpc_max_clients, grpc_tower_counters));
         async_runtime.register(grpc_service);
     }
-    async_runtime.register(rpc_core_service.clone());
     async_runtime.register(p2p_service);
     async_runtime.register(consensus_monitor);
     async_runtime.register(mining_monitor);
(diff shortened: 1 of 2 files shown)


kaspad/src/daemon.rs
```diff
@@ -471,8 +469,12 @@ do you confirm? (answer y/n or pass --yes to the Kaspad command line to confirm
     if let Some(port_mapping_extender_svc) = port_mapping_extender_svc {
         async_runtime.register(Arc::new(port_mapping_extender_svc))
     };
+    if !args.disable_grpc {
+        let grpc_service =
+            Arc::new(GrpcService::new(grpc_server_addr, config, rpc_core_service.clone(), args.rpc_max_clients, grpc_tower_counters));
+        async_runtime.register(grpc_service);
+    }
     async_runtime.register(rpc_core_service.clone());
```
michaelsutton, line 477: Probably makes no difference but perhaps keep  `rpc_core_service` registration before grpc since that's how it is now

kaspad/src/daemon.rs
```diff
@@ -348,7 +348,7 @@ do you confirm? (answer y/n or pass --yes to the Kaspad command line to confirm
     let outbound_target = if connect_peers.is_empty() { args.outbound_target } else { 0 };
     let dns_seeders = if connect_peers.is_empty() && !args.disable_dns_seeding { config.dns_seeders } else { &[] };
 
-    let grpc_server_addr = args.rpclisten.unwrap_or(ContextualNetAddress::unspecified()).normalize(config.default_rpc_port());
+    let grpc_server_addr = args.rpclisten.unwrap_or(ContextualNetAddress::loopback()).normalize(config.default_rpc_port());
```
michaelsutton, line 351: Did you sanity test local access works fine?

coderofstuff: Yes:

```
kaspa-utils ./kaspactl -s=127.0.0.1 -a GetInfo     
{
    "getInfoResponse":  {
        "p2pId":  "0a1d9764-baff-4267-b381-ca241d128963",
        "mempoolSize":  "0",
        "serverVersion":  "0.13.3",
        "isUtxoIndexed":  true,
        "isSynced":  true,
        "error":  null
    }
}
```

kaspad/src/daemon.rs
coderofstuff: If it makes no difference, I prefer to keep it this way so that the things that have conditionals are all at the top followed by a series of one-liners.

michaelsutton (approved):

PR #388, opened 2024-01-12
Remove mimalloc override feature for macos

coderofstuff:
The `override` mimalloc feature causes instability in macos builds


michaelsutton:
Is it possible to remove `override` only for macos? Since we kind of reached stability on Linux with override
cc: @elichai

michaelsutton (approved):

PR #392, opened 2024-01-14
Feat/indexed db transaction store

elertan:
The indexed db transaction store implementation except for the `load_range` function.

Still needs some good testing, mostly on the deserialisation process to ensure this functions as expected (so `load_single`. `load_multiple`).
Same goes for `remove`, but this is more likely to work properly.
`store_transaction_note` and `store_transaction_metadata` will most likely work properly when `load_single`/`load_multiple` work as well, since it relies on similar logic.


elertan:
Fixed clippy warnings

aspect (approved):
Many thanks for implementing this. This is a great help.

PR #412, opened 2024-01-26
cache utxoset size, and add to logs. 

D-Stacks:
1) adds utxoset size to the db
2) propagates via p2p
3) expand logs pertaining to utxo syncing
4) adds  corresponding regex capture to @aspect 's wallet.

### DIFF
--- a/cli/src/cli.rs
+++ b/cli/src/cli.rs
@@ -717,10 +717,36 @@ impl KaspaCli {
                         .join(" "),
                     )
                 }
-                SyncState::UtxoSync { total, .. } => {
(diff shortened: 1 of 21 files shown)


D-Stacks:
As I noticed with sanity checks, utxo diffs may hold the same outpoint in added and removed at different daa scores, as such this either 1) requires extra processing / looping, which is probably not worth the strain, or 2) expansion of the utxo algebra / diff to extract unique utxo growth in-place, which is a deeper incursion on sensitive consensus processing and requires more discussion. As such putting this on ice and closing it... potentially will re-open it with a PoC for 2) that can be discussed.

PR #419, opened 2024-01-31
Add optional Multiprogressbars / counters and spinners, for an alternate logging experience with the  `--progressions` flag.

D-Stacks:
Draft for now, waiting for #401 and #412, in order to add more respective sync bars and perhaps work on the formatting a bit.. 

Adds modern [indicatif](https://crates.io/crates/indicatif) counters / spinners and progress bars for an alternative logging experience with a new `--progressions` flag:

example: 

Screenshot from 2024-01-31 23-57-53

will display bars for ibd, connection searches, pruning level validations, pruned blocks (although didn't have a chance to test the last one), and counters form the consensus monitor, as well as utxo resyncs. It will also report on the last info / warn and error messages. trace and debug are also easily possible, but simply commented out in-code as they may be too fast moving for some...  

It also plays nicely with the current logging system.  Only better space-aware formatting, and expanding into other components may require some attention, but it implements a nice starting point to expand on.


D-Stacks:
i probably won't work on this for a while.

PR #420, opened 2024-02-06
Confindex

D-Stacks:
Adds a `confindex` index which stores `accepting blue score => chain_block` which can be used to query acceptance data within a confirmation window, or by accepting blue score (i.e. imitates conventional block "height" scanning, as found in regular block chains).  Such rpc calls are also fully implemented. 

borrows some consensus api as found #401 and will be kept as draft until #401 is merged.


D-Stacks:
will revisit after txindex. closing for now, as i won't maintain until then.

PR #426, opened 2024-02-26
rocksdb options tuning

biryukovmaxim:
enable stats by default.
store logs into the single file


michaelsutton:
> enable stats by default

why do we need to enable stats by default?

biryukovmaxim:
> > enable stats by default
> 
> why do we need to enable stats by default?

Why not? In case of any problem related to db we will be able to analyze logs

PR #427, opened 2024-02-28
Address Tracker for subscriptions to changed UTXOs notifications

tiram88:
Subscriptions to `UtxosChanged` notifications is getting a deep revamp optimizing memory consumption (highly) and execution speed (moderately).

## Address Tracker

The subscription sub-system uses a new ubiquitous `SubscriptionContext`, containing essentially an address `Tracker` whose role is to track addresses involved in `UtxosChangedSubscriptions`, indexing them and counting references. All subscriptions use now address indexes instead of `ScriptPublicKeys`, reducing memory consumption significantly.

The `Tracker` features an optional maximum address count. When provided, it allows to

- pre-allocate memory, effectively reducing memory fragmentation
- limit and protect the node ressources allocated to this feature

### Kaspad Argument

The maximum tracked addresses can be defined in the kaspad command with argument `--max-tracked-addresses` with default value set to 1,835,007.

The component used internally (an `IndexMap`) allocates maps with a capacity computed with following formula:

```rust
((max_addresses + 1) * 8 / 7).next_power_of_two() * 7 / 8
``` 

The value passed by argument always gets expanded to the actual usable allocated capacity. For a target of 1M addresses the expanded capacity is 1,835,008. The tracker reserves the last entry for address recycling so the actual maximum is 1,835,007.

## Subscription propagation and memory optimizations

Every `Notifier` is provided a new `UtxosChangedMutationPolicy` that defines how an incoming `UtxosChangedScope` mutations will be propagated to its parents. The policy supports two modes: fine-grained `AddressSet` or all or nothing `Wildcard`.

The notification system is configured to propagate address sets upstream from all RPC Servers up to the Index Processor. This appears to be the optimal balance in memory consumption + processing effort between upstream subscriptions and downstream notifications.

The processing of an incoming `UtxosChangedScope` mutation with its embedded address vector is refactored in a way that retains the same vector all the way up from RPC server to the index processor, helping reduce memory fragmentation.

A `Notifier` and its `Broadcaster`s now share the same subscriptions instances, significantly reducing the allocated memory for large address sets.

### UtxosChangedSubscription

The struct is refactored so that it gets inner mutability of its data field (containing state & address index set) and an immutable listener ID (used to identify the instance, notably for equality).

A subscription has 3 states: `None`, `Selected` and `All`. The last 2 are indicative of an active subscription. `Selected` indicates an address set.

In a `Broadcaster`, the subscriptions with `Selected` mode are used as such in the broadcasting `Plan`, meaning that 2 listeners of this state are always considered distinct. The rationale here for not trying to compare their respective address set for possible equality is that the this outcome has a probability close to null while the computing cost of the comparison is high.

In a `Broadcaster`, all subscriptions with `All` state are grouped under a unique ubiquitous instance provided by the `SubscriptionContext`, allowing to mutate the state of a subscription without affecting the broadcasting `Plan`.

### New framework

A new test framework allows the quick and idiomatic setup of an integration test featuring ie. a node, a miner and any additional task the test needs, like submitting large sets of transactions (mempool benchmarking) or cycles of subscriptions/unsubscriptions for many clients at once.

As an example, the mempool benchmark setup gets refactored as:

```rust
let client_manager = Arc::new(ClientManager::new(args));
let mut tasks = TasksRunner::new(Some(DaemonTask::build(client_manager.clone())))
    .launch().await
    .task(
        MinerGroupTask::build(
            network,
            client_manager.clone(),
            SUBMIT_BLOCK_CLIENTS,
            params.bps(),
            BLOCK_COUNT,
            Stopper::Signal
        ).await,
    )
    .task(
        TxSenderGroupTask::build(
            client_manager.clone(),
            SUBMIT_TX_CLIENTS,
            false,
            txs,
            TPS_PRESSURE,
            MEMPOOL_TARGET,
            Stopper::Signal,
        ).await,
    );
tasks.run().await;
tasks.join().await;
```

### Memory benchmarking of UTXOs changed subscriptions

A set of benchmarks is added, dedicated to measuring the memory consumption in various UTXOs changed subscriptions setups. In all cases, 500 clients connect to a mining node, processing ~100 TPS. 495 client are considered "wallets" and each do subscribe to 800 unique addresses. The 5 clients left are "monitoring services" that subscribe to addresses sets of length 200K to 1M by 200K increments, sets mainly overlapping with the wallets addresses.

Clients run subscription cycles of predefined duration. During 2/3 of the cycle, the client is subscribed to its addresses and the remaining 1/3 of the time, the client is fully unsubscribed.

The benchmark runs until 1.5M transactions are mined. The cycle duration appears to have some moderate impact on the overall benchmark duration. The sending and processing of the 500 subscriptions all in a row slows down the node TPS a bit during ~30 seconds, probably because the network bandwidth and gRPC server are getting briefly saturated.

The benchmark is structured into a client process spawning a child server process. The parent process runs all client activities: mining, submitting transactions and the 500 gRPC clients subscribing to UtxosChanged notifications cyclically. The server runs a node. This architecture has the desired property of having the client and the server running each in isolation, in particular each managing its own memory.

Memory consumption of the server process is tracked every 5 seconds all along the benchmark run. An average value is then computed on the data excluding the first 60 minutes considered as warmup.

The various benchmark setups are:

| **#**     | **Subscriptions cycle**                              | **Average memory consumption in GB** | **Delta vs A (GB)** |
|-----------|------------------------------------------------------|--------------------------------------|---------------------|
| **A**     | No subscriptions at all (aka. control sample)        | 2.7                                  |                     |
| **B**     | Single cycle of infinite duration (no unsubscribing) | 3.1                                  | 0.5                 |
| **C**     | 2 hours cycles                                       | 3.8                                  | 1.1                 |
| **D**     | 30 minutes cycles                                    | 4.0                                  | 1.3                 |
| **E**     | 3 minutes cycles                                     | 5.4                                  | 2.8                 |


notify/src/address/tracker.rs
```diff
@@ -0,0 +1,649 @@
...
+#[derive(Debug)]
+struct Inner {
+    script_pub_keys: IndexMap<ScriptPublicKey, RefCount>,
+    max_addresses: Option<usize>,
+    empty_entries: HashSet<Index>,
+}
+
+impl Inner {
+    const fn expand_max_addresses(max_addresses: usize) -> usize {
+        ((max_addresses + 1) * 8 / 7).next_power_of_two() * 7 / 8 - 1
```
coderofstuff, line 209: When `--max-tracked-addresses=0`, you get this panic:

```
rusty-kaspa git:(pr/tiram88/427-1) cargo run --bin kaspad -- --utxoindex --disable-upnp --max-tracked-addresses=0                       
    Finished dev [unoptimized + debuginfo] target(s) in 0.18s
     Running `target/debug/kaspad --utxoindex --disable-upnp --max-tracked-addresses=0`
2024-03-11 22:34:47.007-06:00 [INFO ] kaspad v0.13.5
2024-03-11 22:34:47.007-06:00 [INFO ] Application directory: /Users/dev/.rusty-kaspa
2024-03-11 22:34:47.007-06:00 [INFO ] Data directory: /Users/dev/.rusty-kaspa/kaspa-mainnet/datadir
2024-03-11 22:34:47.007-06:00 [INFO ] Logs directory: /Users/dev/.rusty-kaspa/kaspa-mainnet/logs
2024-03-11 22:34:47.007-06:00 [INFO ] Utxoindex Data directory /Users/dev/.rusty-kaspa/kaspa-mainnet/datadir/utxoindex
thread 'main' panicked at notify/src/address/tracker.rs:209:9:
attempt to subtract with overflow
Exiting...
```

Please either make the param disallow 0, or do some handling when 0 is passed.

coderofstuff (changes requested):
Still reviewing but I thought I should get some of these out.

consensus/src/processes/transaction_validator/transaction_validator_populated.rs
```diff
@@ -35,7 +35,7 @@ impl TransactionValidator {
             // Storage mass hardfork was activated
             self.check_mass_commitment(tx)?;
 
-            if pov_daa_score < self.storage_mass_activation_daa_score + 10 {
+            if pov_daa_score < self.storage_mass_activation_daa_score + 10 && self.storage_mass_activation_daa_score > 0 {
```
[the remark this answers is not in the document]
michaelsutton: I think he wanted new nets such as dev net which activate storage mass from the start to avoid printing hf messages

notify/src/address/tracker.rs
```diff
@@ -0,0 +1,649 @@
...
+    }
+
+    fn new(max_addresses: Option<usize>) -> Self {
```
michaelsutton, line 212: `max_addresses` must have a strict upper-bound, not an optional one. Note that even `Index::MAX * size_of<SPK>()` is too large (~250GB). Of course, when using this default bound, the map should not be pre-allocated. 
Overall it's a chance to make the field `self.max_addresses` not optional which will simplify logic in a few places.

notify/src/address/tracker.rs
```diff
@@ -0,0 +1,649 @@
...
+    }
+
+    fn is_empty(&self) -> bool {
+        self.is_empty()
+    }
+}
+
+#[derive(Debug)]
+struct Inner {
+    script_pub_keys: IndexMap<ScriptPublicKey, RefCount>,
```
michaelsutton, line 202: I would add a noticeable comment/warning here emphasizing that every mutable access to `script_pub_keys` must be careful to not use `IndexMap` APIs which alter the index.

notify/src/address/tracker.rs
```diff
@@ -0,0 +1,649 @@
...
+                    let _ = *entry.insert(0);
+
+                    // Recycle empty entries
+                    let mut recycled = false;
+                    if (index + 1) as usize == self.script_pub_keys.len() && !self.empty_entries.is_empty() {
+                        let empty_index = self.empty_entries.iter().cloned().next();
+                        if let Some(empty_index) = empty_index {
+                            self.script_pub_keys.swap_remove_index(empty_index as usize);
+                            index = empty_index;
+                            recycled = true;
```
michaelsutton, line 268: It took me a while to figure out that `empty_index` is removed from `self.empty_entries` only when the count is increased via `inc_count`. Perhaps a comment here would help

michaelsutton (commented):
Just a few initial comments on inner tracker logic. Still reviewing otherwise.

consensus/src/processes/transaction_validator/transaction_validator_populated.rs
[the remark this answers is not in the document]
tiram88: Right, that's exactly the reason. Getting rid of those annoying logs in the memory benchmark.

notify/src/address/tracker.rs
tiram88: We might take advantage of `--max-tracked-addresses=0` to disable the max altogether, wdyt?

coderofstuff: Sounds good to me. Much better than a panic. Please make sure that's documented somewhere.

coderofstuff (commented):
Lol, I had comments here but it looks like I never submitted them.

notify/src/subscription/compounded.rs
```diff
@@ -144,79 +144,81 @@ impl Subscription for VirtualChainChangedSubscription {
...
                         // Remove All
                         assert!(self.all > 0);
                         self.all -= 1;
                         if self.all == 0 {
-                            if !self.addresses.is_empty() {
-                                return Some(Mutation::new(
-                                    Command::Start,
-                                    Scope::UtxosChanged(UtxosChangedScope::new(self.addresses.keys().cloned().collect())),
-                                ));
+                            let addresses = self.to_addresses(Prefix::Mainnet, context);
```
michaelsutton, line 217: `Prefix::Mainnet`: the prefix should be dynamic and based on actual network (this should be fixed in multiple places)

notify/src/subscription/compounded.rs
```diff
@@ -234,20 +236,25 @@ impl Subscription for UtxosChangedSubscription {
...
 
     fn active(&self) -> bool {
-        self.all > 0 || !self.addresses.is_empty()
+        self.all > 0 || !self.indexes.is_empty()
     }
 
-    fn scope(&self) -> Scope {
-        let addresses = if self.all > 0 { vec![] } else { self.addresses.keys().cloned().collect() };
+    fn scope(&self, context: &SubscriptionContext) -> Scope {
+        let addresses = if self.all > 0 { vec![] } else { self.to_addresses(Prefix::Mainnet, context) };
```
michaelsutton, line 243: `Prefix::Mainnet`: the prefix should be dynamic and based on actual network

notify/src/subscription/single.rs
```diff
@@ -143,162 +167,386 @@ impl Subscription for VirtualChainChangedSubscription {
...
+        Ok(context.address_tracker.register(&mut self.indexes, addresses)?)
+    }
+
+    pub fn unregister(&mut self, addresses: Vec<Address>, context: &SubscriptionContext) -> Vec<Address> {
+        context.address_tracker.unregister(&mut self.indexes, addresses)
+    }
+
+    pub fn unregister_indexes(&mut self, context: &SubscriptionContext) -> Vec<Address> {
+        // TODO: consider using a provided prefix
+        let removed = self.to_addresses(Prefix::Mainnet, context);
```
michaelsutton, line 290: `Prefix::Mainnet`: the prefix should be dynamic and based on actual network. After fixing, the TODO above should be removed

notify/src/subscription/single.rs
```diff
@@ -308,11 +556,27 @@ impl Subscription for UtxosChangedSubscription {
     }
 
     fn active(&self) -> bool {
-        self.active
+        self.state().active()
+    }
+
+    fn scope(&self, context: &SubscriptionContext) -> Scope {
+        // TODO: consider using a provided prefix
+        UtxosChangedScope::new(self.data().to_addresses(Prefix::Mainnet, context)).into()
```
michaelsutton, line 564: `Prefix::Mainnet`: the prefix should be dynamic and based on actual network. Remove TODO after fixing

rpc/grpc/client/src/lib.rs
```diff
@@ -81,6 +87,7 @@ impl GrpcClient {
     pub async fn connect(
```
michaelsutton, line 87: Missing some documentation of the arguments for this method. Most devs building this object will be unfamiliar with internals. Perhaps also a simplified `connect_with_defaults` where we select the most basic config would be in place (for instance the subscription context should be `None` )

rpc/grpc/client/src/lib.rs
```diff
@@ -99,28 +106,39 @@ impl GrpcClient {
             counters,
         )
         .await?;
+        let client_id = u64::from_le_bytes(rand::random::<[u8; 8]>());
```
michaelsutton, line 109: what's the worst-case scenario if this id gets duplicated?

tiram88: No impact at all.

This id is only used in direct mode for building `subscriptions` and is actually ignored by `start_notify` and `stop_notify`. It can be removed and replaced by a default 0 at all occurrences.

notify/src/address/tracker.rs
tiram88: Done impl and doc, together with a revised low default (forcing to explicitly ask for a large count) and an upper bound

notify/src/address/tracker.rs
tiram88: `max_addresses` remains optional, leading to no preallocation on `None`.

A new `Tracker::MAX_ADDRESS_UPPER_BOUND` is declared and enforced both in preallocated and unallocated scenarios.

Currently the upper bound is 14,680,063 (expansion of 10M). We might want to update that in the future.

testing/integration/src/subscribe_benchmarks.rs
```diff
@@ -0,0 +1,305 @@
...
+    client.shutdown().await.unwrap();
+    kaspa_core::warn!("Disconnect the main client");
+    client.disconnect().await.unwrap();
+    drop(client);
+
+    kaspa_core::warn!("Waiting for the daemon to exit...");
+    daemon_process.wait().await.expect("failed to wait for the daemon process");
+}
+
+/// `cargo test --package kaspa-testing-integration --lib --features devnet-prealloc -- subscribe_benchmarks::bench_utxos_changed_subscriptions_daemon --exact --nocapture --ignored -- --rpc=16610 --p2p=16611`
```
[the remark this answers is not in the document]
tiram88: Right, but the test is not meant to be run on its own.

Fixed and documented.

testing/integration/src/subscribe_benchmarks.rs
```diff
@@ -0,0 +1,305 @@
...
+        Some("ucs-server".to_owned()),
+    );
+    let server_start_time = std::time::Instant::now();
+    let mut daemon_process = tokio::process::Command::new("cargo")
+        .args(daemon_args.to_command_args("subscribe_benchmarks::bench_utxos_changed_subscriptions_daemon"))
+        .spawn()
+        .expect("failed to start daemon process");
+
+    // Make sure that the server was given enough time to start
+    let client_start_time = server_start_time + Duration::from_secs(DAEMON_LAUNCH_SECS_DELAY);
```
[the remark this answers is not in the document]
tiram88: It's a child process, so we have no access to its internal memory from the parent process.

There might be some fancy interprocess data sharing tricks that we might implement but it seems overkill here.

Just relying on RPC giving some time so the server responds seems ok, especially since the test is ignored for belonging to a benchmark, thus not executed either locally with nextest or in CI.

kaspad/src/daemon.rs
```diff
@@ -458,6 +462,19 @@ do you confirm? (answer y/n or pass --yes to the Kaspad command line to confirm
         p2p_tower_counters.clone(),
         grpc_tower_counters.clone(),
     ));
+    let grpc_service_broadcasters: usize = 3; // TODO: add a command line argument or derive from other arg/config/host-related fields
```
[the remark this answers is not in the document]
tiram88: Just a rule of thumb.

This is highly correlated to the expected amount of notification traffic, hence the TODO, allowing the user to decide as function of its use case.

notify/src/subscription/compounded.rs
tiram88: Fully aware of this issue.

This has no impact on the expected correct behavior since the tracker stores `ScriptPublicKey` instead of `Address` internally, stripping away the `Prefix` of the addresses.

Anyway I agree that is looks weird and even buggy (though functionally correct) and it should eventually be fixed. This though requires changes in a lot of places in the code since we do not have the current network and its matching address prefix handy all over the place and there is some internal challenge in gRPC & wRPC to rpc-core conversions.

We might postpone that for a following PR.

rpc/grpc/client/src/lib.rs
tiram88: done

notify/src/address/tracker.rs
biryukovmaxim: Use `Option<NonZeroUsize>` it has isize layout and reflect purpose of its use better

kaspad/src/daemon.rs
```diff
@@ -357,7 +357,7 @@ do you confirm? (answer y/n or pass --yes to the Kaspad command line to confirm
 
     let tick_service = Arc::new(TickService::new());
     let (notification_send, notification_recv) = unbounded();
-    let max_tracked_addresses = if args.utxoindex { args.max_tracked_addresses } else { None };
+    let max_tracked_addresses = if args.utxoindex && args.max_tracked_addresses > 0 { Some(args.max_tracked_addresses) } else { None };
```
biryukovmaxim, line 360: Make it Option<NonZeroUsize>. It has the same layout as usize, and better reflects purpose of its use

michaelsutton: It's a cool structure to know about, but I'm not sure it's needed here (should definitely be used when a vector of such values is created to reduce mem, but here it's over-fancy imho).

michaelsutton (approved):
Approving this, but there are a few technical debts to address in following PRs:

1. Change `UtxosChangedScope` to an enum and allow a variant with a vector of indexes. This way internal mutations can avoid the translation back and forth to an address. This is both an optimization and a way to avoid the hacked usage of the mainnet address prefix in the translation 
2. In `GrpcClient`, obtain the network from the server on start-up and use it to maintain the correct address prefix (important for re-registration on client re-connection).  This too will avoid the usage of mainnet address prefix during re-connection
3. In `AddressTracker`, use a Guard-style mechanism to manage registrations. The guard will allow modifying the set of registered indexes, but most importantly, will unregister on `Drop`, which makes it safer and will avoid future errors

PR #433, opened 2024-03-13
Fix ContextualNetAddress parsing

orgr:


orgr:
add-peers expects an array so it would be
```
add-peers = ["123.4.5.6", "192.168.1.1"]
```
or single value `["123.4.5.6"]`

Maybe ini format is closer to what you're expecting, I'll check the library aspect mentioned

orgr:
the `perf-meters` issue is because of an issue with the log-level,
@callid0n should be working now, any more sneaky options I might have missed?

TODO
[ ] - align `Args` struct names with CLI arg names

michaelsutton:
Perhaps serde rename should be applied to:
`inbound_limit` -> `maxinpeers`
`outbound_target` -> `outpeers`

And perhaps others I'm missing. Ideally we find a way to avoid maintaining these multiple renames (both on args definition and serde attributes), but that can be a future fix

coderofstuff (dismissed):
Looks good to me

README.md
```diff
@@ -259,7 +259,7 @@ cargo run --release --bin kaspad -- --configfile /path/to/configfile.toml
 # or
 cargo run --release --bin kaspad -- -C /path/to/configfile.toml
   ```
-  The config file should be a list of \<CLI argument\> = \<value\> separated by newlines.
+  The config file should be a list of \<CLI argument\> = \<value\> separated by newlines, values with special characters like '.' or '=' will require quoting the value i.e \<CLI argument\> = "\<value\>".
```
michaelsutton, line 262: The readme should reflect the latest changes. Should add a list example + fix the arg names (like testnet-suffix -> netsuffix). Title of this section should be "Using a configuration file"

coderofstuff (approved):

PR #436, opened 2024-03-21
Implement UTXO Return Address RPC command

coderofstuff:
`GetUtxoReturnAddress`
Input: `txid` and `accepting_block_daa_score`
Output: `return_address` -> `Option<Address>`

Notes:
- When this call finds an address, it is guaranteed that the address is from the first input of the transaction that produced your UTXO. If the algorithm cannot find a return address (there are many reasons why it might not be able to), it will return `None`

Proto
```
GetUtxoReturnAddressRequestMessage getUtxoReturnAddressRequest = 1112;
GetUtxoReturnAddressResponseMessage getUtxoReturnAddressResponse = 1113;
```

### DIFF
--- a/consensus/src/consensus/mod.rs
+++ b/consensus/src/consensus/mod.rs
@@ -66,7 +66,7 @@ use kaspa_consensus_core::{
     network::NetworkType,
     pruning::{PruningPointProof, PruningPointTrustedData, PruningPointsList},
     trusted::{ExternalGhostdagData, TrustedBlock},
-    tx::{MutableTransaction, Transaction, TransactionOutpoint, UtxoEntry},
+    tx::{MutableTransaction, ScriptPublicKey, Transaction, TransactionOutpoint, UtxoEntry},
     BlockHashSet, BlueWorkType, ChainPath, HashMapCustomHasher,
 };
 use kaspa_consensus_notify::root::ConsensusNotificationRoot;
@@ -682,6 +682,69 @@ impl ConsensusApi for Consensus {
         }
 
         sample_headers
+    }
+
+    fn get_utxo_return_address(&self, txid: Hash, daa_score: u64) -> Option<ScriptPublicKey> {
+        // We need consistency between the past pruning points, selected chain and header store reads
+        let _guard = self.pruning_lock.blocking_read();
+
+        let sc_read = self.selected_chain_store.read();
+
+        let low = self.pruning_point_store.read().get().unwrap().pruning_point;
+        let high = sc_read.get_tip().unwrap().1;
+
+        let mut low_index = sc_read.get_by_hash(low).unwrap();
+        let mut high_index = sc_read.get_by_hash(high).unwrap();
+
+        // let mut locator = Vec::with_capacity((high_index - low_index).ceiling_log_base_2() as usize);
+        // let mut current_index = high_index;
+        let mut matching_chain_block_hash: Option<Hash> = None;
+        while low_index < high_index {
+            let mid = low_index + (high_index - low_index) / 2;
+
+            if let Ok(hash) = sc_read.get_by_index(mid) {
+                if let Ok(compact_header) = self.headers_store.get_compact_header_data(hash) {
+                    if compact_header.daa_score == daa_score {
+                        // We found the chain block we need
+                        matching_chain_block_hash = Some(hash);
+                        break;
+                    } else if compact_header.daa_score < daa_score {
+                        low_index = mid + 1;
+                    } else {
+                        high_index = mid - 1;
+                    }
+                } else {
+                    println!("Did not find a compact header with hash {}", hash);
+                    break;
+                }
+            } else {
+                println!("Did not find a hash at index {}", mid);
+                break;
+            }
+        }
+
+        if matching_chain_block_hash.is_none() {
+            println!("No matching chain block hash found");
+            return None;
+        }
+
+        if let Ok(acceptance_data) = self.acceptance_data_store.get(matching_chain_block_hash.unwrap()) {
+            let containing_acceptance = acceptance_data
+                .iter()
+                .find(|&mbad| mbad.accepted_transactions.iter().find(|&tx| tx.transaction_id == txid).is_some())
+                .cloned();
+
+            if let Some(containing_acceptance) = containing_acceptance {
+                // I found the merged block containing the TXID
+                // let txs = self.block_transactions_store.get(containing_acceptance.block_hash).unwrap().iter().find(|tx| => )
+            } else {
+                return None;
+            }
+
+            return None;
+        } else {
+            return None;
+        };
     }
 
     fn get_virtual_parents(&self) -> BlockHashSet {
--- a/consensus/src/consensus/mod.rs
+++ b/consensus/src/consensus/mod.rs
@@ -77,6 +77,7 @@ use crossbeam_channel::{
 use itertools::Itertools;
 use kaspa_consensusmanager::{SessionLock, SessionReadGuard};
 
+use kaspa_core::{trace, warn};
 use kaspa_database::prelude::StoreResultExtensions;
 use kaspa_hashes::Hash;
 use kaspa_muhash::MuHash;
@@ -101,6 +102,9 @@ use self::{services::ConsensusServices, storage::ConsensusStorage};
 use crate::model::stores::selected_chain::SelectedChainStoreReader;
 
 use std::cmp;
+
+use crate::model::stores::utxo_diffs::UtxoDiffsStoreReader;
+use kaspa_consensus_core::utxo::utxo_diff::ImmutableUtxoDiff;
 
 pub struct Consensus {
     // DB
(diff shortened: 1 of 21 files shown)


consensus/src/consensus/mod.rs
```diff
@@ -610,6 +615,97 @@ impl ConsensusApi for Consensus {
...
+                let first_input_prev_outpoint = &tx.inputs[0].previous_outpoint;
+                // Expected to never fail, since
+                let utxo_diff = self.utxo_diffs_store.get(matching_chain_block_hash).unwrap();
+                let removed_diffs = utxo_diff.removed();
+
+                return Some(removed_diffs.get(first_input_prev_outpoint)?.script_public_key.clone());
+            }
+        };
+
+        None
```
biryukovmaxim, line 706:
```suggestion
 let matching_chain_block_hash = loop {
            let mid = low_index + (high_index - low_index) / 2;
            let hash = sc_read.get_by_index(mid).inspect_err(|_err| trace!("Did not find a hash at index {}", mid)).ok()?;
            let compact_header = self
                .headers_store
                .get_compact_header_data(hash)
                .inspect_err(|_err| trace!("Did not find a hash at index {}", mid))
                .ok()?;
            match compact_header.daa_score.cmp(&target_daa_score) {
                cmp::Ordering::Equal => {
                    // We found the chain block we need
                    break Some(hash);
                }
                cmp::Ordering::Greater => {
                    high_index = mid - 1;
                }
                cmp::Ordering::Less => {
                    low_index = mid + 1;
                }
            }
            if low_index > high_index {
                break None;
            }
        }?;

        let acceptance_data = self.acceptance_data_store.get(matching_chain_block_hash).ok()?;
        let (index, containing_acceptance) = acceptance_data.iter().find_map(|mbad| {
            let tx_arr_index =
                mbad.accepted_transactions.iter().enumerate().find_map(|(index, tx)| (tx.transaction_id == txid).then_some(index));
            tx_arr_index.map(|index| (index, mbad.clone()))
        })?;

        // Found Merged block containing the TXID
        let tx = &self.block_transactions_store.get(containing_acceptance.block_hash).unwrap()[index];

        if tx.id() != txid {
            // Should never happen, but do a sanity check. This would mean something went wrong with storing block transactions
            // Sanity check is necessary to guarantee that this function will never give back a wrong address (err on the side of None)
            warn!("Expected {} to match {} when checking block_transaction_store using array index of transaction", tx.id(), txid);
            return None;
        }

        if tx.inputs.is_empty() {
            // A transaction may have no inputs (like a coinbase transaction)
            return None;
        }

        let first_input_prev_outpoint = &tx.inputs[0].previous_outpoint;
        // Expected to never fail, since
        let utxo_diff = self.utxo_diffs_store.get(matching_chain_block_hash).unwrap();
        let removed_diffs = utxo_diff.removed();

        Some(removed_diffs.get(first_input_prev_outpoint)?.script_public_key.clone())
```

consensus/src/consensus/mod.rs
```diff
@@ -610,6 +615,97 @@ impl ConsensusApi for Consensus {
         sample_headers
     }
 
+    fn get_utxo_return_script_public_key(&self, txid: Hash, target_daa_score: u64) -> Option<ScriptPublicKey> {
+        // We need consistency between the past pruning points, selected chain and header store reads
+        let _guard = self.pruning_lock.blocking_read();
+
+        let sc_read = self.selected_chain_store.read();
+
+        let pp_hash = self.pruning_point_store.read().get().unwrap().pruning_point;
```
D-Stacks, line 624: this should be the `history_root` hash in `is_archival` cases. so i guess should do a quick `is_archival` check on the config, and choose accordingly here. you can also use the `get_source()` api from consensus, which handles this under the hood..

rpc/service/src/service.rs
```diff
@@ -620,6 +621,35 @@ NOTE: This error usually indicates an RPC conversion error between the node and
...
+    async fn get_utxo_return_address_call(&self, request: GetUtxoReturnAddressRequest) -> RpcResult<GetUtxoReturnAddressResponse> {
+        let session = self.consensus_manager.consensus().session().await;
+
+        let mut return_address = None;
+
+        // Convert a SPK to an Address
+        if let Some(spk) = session.async_get_utxo_return_script_public_key(request.txid, request.accepting_block_daa_score).await {
+            let script = spk.script();
+
+            // Standard Address scripts are only either 34 or 35 in length:
```
D-Stacks, line 633: in `kaspa_txscript` crate we have `ScriptClass::from_script()` which will return an enum `ScriptClass`:

```rs
pub enum ScriptClass {
    /// None of the recognized forms
    NonStandard = 0,
    /// Pay to pubkey
    PubKey,
    /// Pay to pubkey ECDSA
    PubKeyECDSA,
    /// Pay to script hash
    ScriptHash,
}
```

can this maybe be used to check if it is standard, by checking that it is not nonstandard?

consensus/src/consensus/mod.rs
D-Stacks, line 624: if we get `pruning_point` / `history_root` daa score and it is more then `target_daa_score`, we could return early with `None` here.

consensus/src/consensus/mod.rs
```diff
@@ -610,6 +615,97 @@ impl ConsensusApi for Consensus {
         sample_headers
     }
 
+    fn get_utxo_return_script_public_key(&self, txid: Hash, target_daa_score: u64) -> Option<ScriptPublicKey> {
```
D-Stacks, line 618: Instead of an Option<ScriptPublicKey>, i think an enum with background on the search result, makes more sense, or alternatively in "not found" cases differentiated consensus errors, which can be propagated back to rpc.  I see three distinct reasons why a script_public_key may not be found, amounting to:

1) Data is pruned, 
2) Txid is not found at the target_daa_score, 
3) Searched txid is a coinbase tx.

D-Stacks (commented):
‎

rpc/core/src/api/ops.rs
```diff
@@ -86,6 +86,8 @@ pub enum RpcApiOps {
     GetCoinSupply,
     /// Get DAA Score timestamp estimate
     GetDaaScoreTimestampEstimate,
+    /// Get UTXO Return Addresses
+    GetUtxoReturnAddress,
```
aspect, line 90: @coderofstuff  Please relocate the GetUtxoReturnAddress at the end of the enum as this currently breaks borsh serialization.

consensus/src/consensus/mod.rs
coderofstuff: Added in [8dc0874](https://github.com/kaspanet/rusty-kaspa/pull/436/commits/8dc0874523ce94707c2c294819f63291ac3dc9f4)

I couldn't apply as-is since `inspect_err` was unstable. So I replaced that with `map_err` but kept the rest.

Also added more comments.

rpc/core/src/api/ops.rs
coderofstuff: Fixed in [652ae15](https://github.com/kaspanet/rusty-kaspa/pull/436/commits/652ae15a84262660f8bfef8119073a1aa7869b87)

rpc/service/src/service.rs
coderofstuff: I did one better and refactored this to:

```
        // Convert a SPK to an Address
        let return_address = if let Some(spk) =
            session.async_get_utxo_return_script_public_key(request.txid, request.accepting_block_daa_score).await
        {
            extract_script_pub_key_address(&spk, self.config.prefix()).ok()
        } else {
            None
        };
```

In commit [56a4fcb](https://github.com/kaspanet/rusty-kaspa/pull/436/commits/56a4fcb07c9f4e2da42bde335f17e065b36c7a78)

consensus/src/consensus/mod.rs
coderofstuff: Updated to return an enum

KaffinPX:
Could we separate accepting block finder/iterator and return address finder on consensus to possibly allow ``getTransactionCurrentAcceptingBlock``in future to make transaction confirmation tracking simpler.

consensus/src/consensus/mod.rs
```diff
@@ -696,6 +700,120 @@ impl ConsensusApi for Consensus {
...
+        }
+
+        if tx.inputs.is_empty() {
+            // A transaction may have no inputs (like a coinbase transaction)
+            return ReturnAddress::TxFromCoinbase;
+        }
+
+        let first_input_prev_outpoint = &tx.inputs[0].previous_outpoint;
+        // Expected to never fail, since we found the acceptance data and therefore there must be matching diff
+        let utxo_diff = self.utxo_diffs_store.get(matching_chain_block_hash).unwrap();
```
michaelsutton, line 804: There might be a rare case where the tx spends an output of another tx which was accepted by this same chain block (contained within another merged block which proceeded it in the mergeset ordering). In which case the UTXO diff will not contain the created-and-immediately-spent UTXO.  The solution in that case is to get the UTXO address from the output of that tx which must appear in this chain block's acceptance data as well

consensus/src/consensus/mod.rs
```diff
@@ -696,6 +700,120 @@ impl ConsensusApi for Consensus {
         sample_headers
     }
 
+    fn get_utxo_return_address(&self, txid: Hash, target_daa_score: u64) -> ReturnAddress {
```
michaelsutton, line 703: I think we should consider moving this logic to a sub manager/processor or at least to another module file. Wdyt?

michaelsutton (changes requested):

consensus/src/consensus/mod.rs
michaelsutton: Actually the comment should belong on the line `removed_diffs.get(first_input_prev_outpoint).unwrap()` below

coderofstuff: > The solution in that case is to get the UTXO address from the output of that tx which must appear in this chain block's acceptance data as well

Let me check my understanding

Scenario:
- UTXO0 is spent by TX1 and creates UTXO1
- UTXO1 is spent by TX2 and creates UTXO2
- A chain block happens to accept both of these

In this scenario, I know the details of UTXO2 and it's my input to the function. Because both txs are accepted by the same chain block, `removed_diff` will not contain the details of UTXO1 (corresponds to `first_input_prev_outpoint`). So I need to find TX1 from `block_transactions_store` instead and look at the output that corresponds to UTXO1. Is this accurate?

consensus/src/consensus/mod.rs
coderofstuff: Did you have a specific location in mind?

coderofstuff: virtual processor probably makes sense

consensus/src/consensus/mod.rs
michaelsutton: one more comment: be defensive here and avoid unwrapping if possible (just in case we are missing more edge-case details like this). So prefer returning an RPC err + warn log with max details

consensus/src/consensus/mod.rs
coderofstuff: Moved in [b6b6ec1](https://github.com/kaspanet/rusty-kaspa/pull/436/commits/b6b6ec16c932f9abb734e5e8cfa015698213c9b1)

consensus/src/consensus/mod.rs
coderofstuff: The edge case has been addressed. Remaining to do is refactor further and handle unwraps safely

coderofstuff: Refactored and handled all the unwraps. I changed the return of this function to be a Result, instead of just the direct enum.

consensus/src/pipeline/virtual_processor/processor.rs
```diff
@@ -1188,6 +1191,190 @@ impl VirtualStateProcessor {
...
+            .map(|compact_header| compact_header.daa_score)
+            .map_err(|_| ReturnAddressError::NotFound(format!("Did not find compact header for source hash {}", source_hash)))?;
+
+        if target_daa_score < source_daa_score {
+            // Early exit if target daa score is lower than that of pruning point's daa score:
+            return Err(ReturnAddressError::AlreadyPruned);
+        }
+
+        let (matching_chain_block_hash, acceptance_data) =
+            self.find_accepting_chain_block_hash_at_daa_score(target_daa_score, source_hash, guard)?;
```
michaelsutton, line 1216: You are dropping `guard` here when `find_accepting_chain_block_hash_at_daa_score` returns (bcs you move it rather than pass it by ref). 
I think you should avoid passing it in the first place (even though the inner call kind of relies on it semantically as well). Instead you can document this assumption on the method doc and rely on it being used only locally.

consensus/src/pipeline/virtual_processor/processor.rs
```diff
@@ -1188,6 +1191,190 @@ impl VirtualStateProcessor {
         Ok(())
     }
 
+    pub fn get_utxo_return_address(
```
michaelsutton, line 1194: Maybe we should move this set of new methods to a new parallel file under `pipeline/virtual_processor` which will `impl VirtualStateProcessor` too.  We can call it `utxo_query.rs` or `utxo_query_manager.rs` and @D-Stacks will be able to extend it for his needs

consensus/src/pipeline/virtual_processor/processor.rs
```diff
@@ -1188,6 +1191,190 @@ impl VirtualStateProcessor {
...
+            .ok_or(ReturnAddressError::NotFound(format!("Did not find containing_acceptance for tx {}", txid)))?;
+
+        // Found Merged block containing the TXID
+        let tx = self
+            .block_transactions_store
+            .get(containing_acceptance.block_hash)
+            .map_err(|_| ReturnAddressError::NotFound(format!("Did not block {} at block tx store", containing_acceptance.block_hash)))
+            .and_then(|block_txs| {
+                block_txs.get(index).cloned().ok_or_else(|| {
+                    ReturnAddressError::NotFound(format!(
```
michaelsutton, line 1229: Since you already have a dedicated error type, I would make an err variant for each and every err instance in this method and completely avoid inline string descriptions. It will significantly improve readability

michaelsutton (changes requested):

consensus/src/pipeline/virtual_processor/processor.rs
michaelsutton: Maybe `utxo_inquirer.rs`

consensus/src/consensus/mod.rs
michaelsutton: @D-Stacks make sure you read this thread

consensus/src/pipeline/virtual_processor/processor.rs
coderofstuff: Addressed in [5fd8ac5](https://github.com/kaspanet/rusty-kaspa/pull/436/commits/5fd8ac5c80c4c53cb7c8401cc1b467a5146731f7)

michaelsutton (approved):

michaelsutton (approved):

michaelsutton (approved):

michaelsutton (approved):

Issue #437, opened 2024-03-23
error[E0412]: cannot find type `DeterministicHashData` in this scope


hyperji:
**Describe the bug**
error[E0412]: cannot find type `DeterministicHashData` in this scope

**To Reproduce**
1. Building on Linux
2. then  I run the test, it produces some error with the following code:
cd rusty-kaspa
cargo test --release

the error is in below: 

`error: Seal changed - was: 0xa7a4 now: 0x544d
  --> wallet/core/src/deterministic.rs:59:9
   |
59 | seal! { 0xa7a4, {
   |         ^^^^^^

error[E0412]: cannot find type `DeterministicHashData` in this scope
  --> wallet/core/src/deterministic.rs:78:45
   |
78 | fn make_hashes<T, const N: usize>(hashable: DeterministicHashData<T>) -> [Hash; N]
   |                                             ^^^^^^^^^^^^^^^^^^^^^ not found in this scope

error[E0422]: cannot find struct, variant or union type `DeterministicHashData` in this scope
  --> wallet/core/src/deterministic.rs:93:20
   |
93 |     let hashable = DeterministicHashData {
   |                    ^^^^^^^^^^^^^^^^^^^^^ not found in this scope

error[E0422]: cannot find struct, variant or union type `DeterministicHashData` in this scope
   --> wallet/core/src/deterministic.rs:106:20
    |
106 |     let hashable = DeterministicHashData {
    |                    ^^^^^^^^^^^^^^^^^^^^^ not found in this scope

error[E0422]: cannot find struct, variant or union type `DeterministicHashData` in this scope
   --> wallet/core/src/deterministic.rs:119:20
    |
119 |     let hashable = DeterministicHashData {
    |                    ^^^^^^^^^^^^^^^^^^^^^ not found in this scope

error[E0422]: cannot find struct, variant or union type `DeterministicHashData` in this scope
   --> wallet/core/src/deterministic.rs:132:20
    |
132 |     let hashable = DeterministicHashData {
    |                    ^^^^^^^^^^^^^^^^^^^^^ not found in this scope

error[E0412]: cannot find type `DeterministicHashData` in this scope
   --> wallet/core/src/deterministic.rs:145:19
    |
145 |     let hashable: DeterministicHashData<[PrvKeyDataId; 0]> = DeterministicHashData {
    |                   ^^^^^^^^^^^^^^^^^^^^^ not found in this scope

[...]

PR #438, opened 2024-03-23
fix deterministic.rs 'seal changed -was: 0xa7a4 now 0x544d'

hyperji:
fix issue: https://github.com/kaspanet/rusty-kaspa/issues/437 

reason:
seal changed -was: 0xa7a4 now 0x544d

### DIFF
--- a/wallet/core/src/deterministic.rs
+++ b/wallet/core/src/deterministic.rs
@@ -56,7 +56,7 @@ impl std::fmt::Display for AccountId {
     }
 }
 
-seal! { 0xa7a4, {
+seal! { 0x544d, {
     // IMPORTANT: This data structure is meant to be deterministic
     // so it can not contain any new fields or be changed.
     #[derive(BorshSerialize)]


D-Stacks (approved):
After a quick talk with @aspect I can confirm that changing the seal hash here is okay, as long as the fields did not change (which is the case).

coderofstuff (approved):
This allows compiling the entire repo (specifically the wallet part) with the latest rust version (1.77.0 as of this writing) but breaks older version.

kaspad and the rest of the components that do not depend on wallet are semantically the same and are not affected by this PR.

LGTM.

gvbgduh:
But what is this seal and its purpose and why does it suddenly got broken? Can we really just patch it with a new value?
It sure indicates that some hash of a data structure (?) got changed and there should be a reason it was sealed.

biryukovmaxim:
This [pr](https://github.com/rust-lang/rust/pull/120227) is included by 1.77 rustc version
After that `seal` proc macro gets different tokens. Without unnecessary whitespaces. That's why hash changed.

I suggest to @aspect to make it derive macro only depending on fields and types, he replied that this is general purpose macro and it can be applied to any code. In summary it will work fine until compiler change the logic again. 

Imho the proper solution is changing the hash and **pinning rust toolchain** in the repo. And maybe msrv

gvbgduh:
I see, thanks for clarification

coderofstuff:
Thank you for your contribution. As this was integrated into a more fundamental fix in https://github.com/kaspanet/rusty-kaspa/pull/442 this current PR has been superseded.

aspect:
Fixed in https://github.com/kaspanet/rusty-kaspa/pull/442

I couldn't merge this PR due to other ongoing issues with lints (caused by the new rustc version as well), so the change was applied in the above-mentioned PR that was just merged.

Issue #402, opened 2024-01-22
--rpclisten-borsh/json should support ip only like --grpclisten


supertypo:
kaspad supports specifying --rpclisten=0.0.0.0 for listening on all interfaces on default network port.

However, it doesn't support the same for --rpclisten-borsh and --rpclisten-json, even though the --help text specifies it as optional and is identical for the three:

```
      --rpclisten[=<IP[:PORT]>]
          Interface:port to listen for gRPC connections (default port: 16110, testnet: 16210).
      --rpclisten-borsh[=<IP[:PORT]>]
          Interface:port to listen for wRPC Borsh connections (default port: 17110, testnet: 17210).
      --rpclisten-json[=<IP[:PORT]>]
          Interface:port to listen for wRPC JSON connections (default port: 18110, testnet: 18210).
```

coderofstuff:
For `rpclisten-borsh` the input is parsed into a `WrpcNetAddress` which can be any of `default`, `public` or an `<ip>:<port>`. The goal here is to also support just `<ip>`.

The string parsing logic in `rpc/wrpc/server/src/addresses.rs` is:

```
impl FromStr for WrpcNetAddress {
    type Err = AddrParseError;
    fn from_str(s: &str) -> Result<Self, Self::Err> {
        match s {
            "default" => Ok(WrpcNetAddress::Default),
            "public" => Ok(WrpcNetAddress::Public),
            _ => {
                let addr: ContextualNetAddress = s.parse()?;
                Ok(Self::Custom(addr))
            }
        }
    }
}
```

The work is focused on the `_` arm of this match.

NOTE: Make sure that if only the IP is passed that you use the default port for the given network. Mainnet, testnet, simnet, devnet may have different default ports.

aspect:
This has been implemented in one of the merged PRs. Closing.

PR #439, opened 2024-03-23
Add support for IP only for --rpclisten-borsh/json (#402)

gvbgduh:
Adds support for IP only for the --rpclisten-borsh/json flags.
If a port is not provided a default port for a given network would be used.

### DIFF
--- a/kaspad/src/args.rs
+++ b/kaspad/src/args.rs
@@ -366,7 +366,7 @@ Setting to 0 prevents the preallocation and sets the maximum to {}, leading to 0
                 .long("ram-scale")
                 .require_equals(true)
                 .value_parser(clap::value_parser!(f64))
-                .help("Apply a scale factor to memory allocation bounds. Nodes with limited RAM (~4-8GB) should set this to ~0.3-0.5 respectively. Nodes with 
+                .help("Apply a scale factor to memory allocation bounds. Nodes with limited RAM (~4-8GB) should set this to ~0.3-0.5 respectively. Nodes with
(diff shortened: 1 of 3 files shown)


biryukovmaxim (commented):
Functional part looks good to me.
However tests look too repetitive. Wdyt about using array and for-loop/iterator? In addition, it is possible to make ip intanzation only once

D-Stacks:
> Functional part looks good to me. However tests look too repetitive. Wdyt about using array and for-loop/iterator? In addition, it is possible to make ip intanzation only once

One can call `.iter()` on both `NetworkType`, and `WrpcEncoding`, then just have a match on these cases, Personally i prefer this way because if new encodings, or network types, were to be added, the test can't compile without handling of new enum variants.

gvbgduh:
thanks @biryukovmaxim, good points, i updated the code

> One can call .iter() on both NetworkType, and WrpcEncoding, then just have a match on these cases
oh interesting, i actually did a quick docs lookup to see it's not standard functionality, but i somehow failed to see it's implemented for those enums already.
it'll sure make code neater, thanks

biryukovmaxim (approved):

aspect (dismissed):
@gvbgduh can you please fix lints (you can run `/check` to see them locally).  Otherwise LGTM.

aspect (approved):
LGTM, many thanks!

biryukovmaxim (approved):

coderofstuff (approved):

PR #442, opened 2024-03-26
Fix rustc 1.77 issues and lints

aspect:
- remove BorshSchema due to rustc 1.77 issues (BorshSchema is not used anywhere)
- fix seal macro hash error (hash changed due to different whitespace processing in rustc 1.77)
- fix all rustc 1.77 lints

This replicates #438 (unfortunately I can't merge this due to other lint errors, so if this merged #438 should be closed)

### DIFF
--- a/components/consensusmanager/src/session.rs
+++ b/components/consensusmanager/src/session.rs
@@ -23,9 +23,9 @@ pub use tokio::task::spawn_blocking;
 use crate::BlockProcessingBatch;
 
 #[derive(Clone)]
-pub struct SessionOwnedReadGuard(Arc<RfRwLockOwnedReadGuard>);
+pub struct SessionOwnedReadGuard(pub Arc<RfRwLockOwnedReadGuard>);
 
-pub struct SessionReadGuard<'a>(RfRwLockReadGuard<'a>);
+pub struct SessionReadGuard<'a>(pub RfRwLockReadGuard<'a>);
 
 pub struct SessionWriteGuard<'a>(RfRwLockWriteGuard<'a>);
 
--- a/components/consensusmanager/src/session.rs
+++ b/components/consensusmanager/src/session.rs
@@ -22,10 +22,12 @@ pub use tokio::task::spawn_blocking;
 
 use crate::BlockProcessingBatch;
 
+#[allow(dead_code)]
 #[derive(Clone)]
-pub struct SessionOwnedReadGuard(pub Arc<RfRwLockOwnedReadGuard>);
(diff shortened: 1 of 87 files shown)


wallet/core/src/deterministic.rs
```diff
@@ -56,7 +56,7 @@ impl std::fmt::Display for AccountId {
     }
 }
 
-seal! { 0xa7a4, {
+seal! { 0x544d, {
```
biryukovmaxim, line 59: Since it only works with version 1.77 we have to introduce msrv equals 1.77 in manifest files

components/consensusmanager/src/session.rs
```diff
@@ -23,9 +23,9 @@ pub use tokio::task::spawn_blocking;
 use crate::BlockProcessingBatch;
 
 #[derive(Clone)]
-pub struct SessionOwnedReadGuard(Arc<RfRwLockOwnedReadGuard>);
+pub struct SessionOwnedReadGuard(pub Arc<RfRwLockOwnedReadGuard>);
 
-pub struct SessionReadGuard<'a>(RfRwLockReadGuard<'a>);
+pub struct SessionReadGuard<'a>(pub RfRwLockReadGuard<'a>);
```
biryukovmaxim, line 28: Did this change occur  due to inter?
If so I would suggest to suppress it instead of making the field public

biryukovmaxim (changes requested):

michaelsutton (dismissed):

components/consensusmanager/src/session.rs
aspect: changed: https://github.com/kaspanet/rusty-kaspa/pull/442/commits/319a6247dda2546baf62fc4f8ed6aac705029fc1

wallet/core/src/deterministic.rs
aspect: MSRV added: https://github.com/kaspanet/rusty-kaspa/pull/442/commits/ad701e1c8135bb2846e3b7baad88dd6a64c2b03f

michaelsutton (dismissed):

coderofstuff (approved):

aspect:
> it doesn't work properly the line `rust-version.workspace.workspace = true` should be added to all crates in the workspace in case of old compiler

propagated `rust-version.workspace` to all crates in https://github.com/kaspanet/rusty-kaspa/commit/467e4ce0129be3f093a6965a39b26329a4359545

biryukovmaxim (approved):

coderofstuff (approved):

PR #448, opened 2024-04-08
Implement updated KIP9 mempool rules (excluding TN11)

michaelsutton:
As [suggested](https://github.com/kaspanet/kips/blob/master/kip-0009.md#current-status) by KIP9, this PR updates mass calculations to reflect the final proposal wherever possible. This includes:

- Applying a maximum over `stoarge_mass` and `compute_mass` rather than summing them
- Using the relaxed formula for the cases ` |O| = 1 OR |O| <= |I| <= 2`

The updates are applied on mainnet and TN10 on the mempool level. TN11 requires a hard-fork for this update and remains unchanged for now also on the mempool level (since the new rules are more permissive)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2149,7 +2149,7 @@ dependencies = [
 
 [[package]]
 name = "kaspa-addresses"
-version = "0.13.5"
+version = "0.13.6"
 dependencies = [
  "borsh",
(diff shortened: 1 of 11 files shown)


coderofstuff (approved):
Looks good to me

Issue #444, opened 2024-03-28
Rare chance of initial pruning proof building issues


coderofstuff:
**Describe the bug**

There is a very rare chance that the initial pruning proof sync is somehow corrupted. Example logs:

```
2024-03-28 18:13: 2024-03-28 18:13:23.283+01:00 [INFO ] Accepted block cbd53132098e712199b33968b34878edc09661baa9b1412ac1832c580693bd1b via relay
2024-03-28 18:13: thread 'pruning-processor' panicked at consensus/src/processes/pruning_proof/mod.rs:650:22:
2024-03-28 18:13: called Result::unwrap() on an Err value: "level: 32, err: block at depth error: high: 99fb3fdc4eb83defa53ffe908e58746875ba5a81d6b94fd082aa5db3c94d6945, depth: 2000, current: d07727509d0e36ec06ebff8fa843e92ab11f03e66b88cd6cedb1e32697ac397f, high blue score: 1801, current blue score: 793, key GhostdagCompact/32/935e0a789ea348d0d0d5edffaf2af58704260ac6e4105cb5aa58bb431bbf8698 not found in store"
2024-03-28 18:13: note: run with RUST_BACKTRACE=1 environment variable to display a backtrace
2024-03-28 18:13: Exiting...
2024-03-28 18:13: 2024-03-28 18:13:23.717+01:00 [INFO ] kaspad v0.13.4
```

**To Reproduce**
1. Run an initial sync.
2. Most of the times this will not occur, but sometimes it will

**Expected behavior**
The initial sync and building of pruning proof never fails

**Desktop (please complete the following information):**
 - Kaspad version: 0.13.4

**Additional context**
This is a **MUST FIX** bug for adoption of the rust node. The final release must contain the fix for this.

When this is encountered, the only recourse is to delete the datadir and do a resync. Either delete the datadir manually or run kaspad with `--reset-db` (once)

PR #449, opened 2024-04-10
Fix for the pruning proof rebuild issue (issue #444)

michaelsutton:
The following describes the pruning bug and the fix assuming sufficient prior knowledge. 

Some notation: 
 
- High levels: block levels at which there are less than 2000 blocks since genesis (so genesis is part of that proof level). Currently this is approx around level 39 and above. 
- Low levels: block levels close to current difficulty and hence it is likely that all 2000 proof blocks are within the last day before the corresponding pruning point.
- Intermediate levels: block levels which are not that rare but are still unlikely to have more than 2000 blocks all within a single day. Hence their level proof likely has blocks **both** before and after the previous pruning point.

The bug is mostly in the case of intermediate levels. It requires that such a block is mined in the anticone of the pruning point (so the bug is likely to occur ~once in a few days, and requires that someone synced a new node exactly at that day). 

The logic missing: we did not update relations per level for the pruning point anticone (aka "trusted" blocks). 

Scenario: a "lucky" intermediate-level block `R` happened to be in the pruning point anticone. A future block `B` on that level is likely to have `R` as its only parent (for that level). Since level relations are not updated for `R`, `B` computes Ghostdag for that level over `ORIGIN` -- essentially starting from blue score 0 instead of building over the previous 2000 blocks.  When building the new proof, this segment of blocks (`S` = `B` + its future) is topologically disconnected from block `N` which is the block at depth 1000 of the next level (which is likely below the current new pruning point). So `N` becomes `root`, and the traversal starting from it never reaches `S`. Leading to a possible prune of blocks in `S` (some may be preserved via other levels)

Result: following the prune, the proof rebuild fails bcs of blocks missing from `S`. This is only a symptom of the bug, the real problem is that the level proof is literally incomplete.  

Fix: update relations per level also for trusted blocks

### DIFF
--- a/consensus/src/processes/mass.rs
+++ b/consensus/src/processes/mass.rs
@@ -91,10 +91,13 @@ impl MassCalculator {
         let ins_len = tx.tx().inputs.len() as u64;
 
         /*
-           KIP-0009 relaxed formula for the cases |O| = 1 OR |O| <= |I| <= 2:
-               max( 0 , C·( |O|/H(O) - |I|/H(I) ) )
+          KIP-0009 relaxed formula for the cases |O| = 1 OR |O| <= |I| <= 2:
+              max( 0 , C·( |O|/H(O) - |I|/H(I) ) )
+
+           Note: in the case |I| = 1 both formulas are equal, yet the following code (harmonic_ins) is a bit more efficient.
+                 Hence, we transform the condition to |O| = 1 OR |I| = 1 OR |O| = |I| = 2 which is equivalent (and faster).
         */
-        if version == Kip9Version::Beta && (outs_len == 1 || (outs_len <= ins_len && ins_len <= 2)) {
+        if version == Kip9Version::Beta && (outs_len == 1 || ins_len == 1 || (outs_len == 2 && ins_len == 2)) {
             let harmonic_ins = tx
                 .populated_inputs()
                 .map(|(_, entry)| self.storage_mass_parameter / entry.amount)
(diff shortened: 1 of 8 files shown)


consensus/src/processes/mass.rs
```diff
@@ -91,10 +91,13 @@ impl MassCalculator {
...
-           KIP-0009 relaxed formula for the cases |O| = 1 OR |O| <= |I| <= 2:
-               max( 0 , C·( |O|/H(O) - |I|/H(I) ) )
+          KIP-0009 relaxed formula for the cases |O| = 1 OR |O| <= |I| <= 2:
+              max( 0 , C·( |O|/H(O) - |I|/H(I) ) )
+
+           Note: in the case |I| = 1 both formulas are equal, yet the following code (harmonic_ins) is a bit more efficient.
+                 Hence, we transform the condition to |O| = 1 OR |I| = 1 OR |O| = |I| = 2 which is equivalent (and faster).
         */
-        if version == Kip9Version::Beta && (outs_len == 1 || (outs_len <= ins_len && ins_len <= 2)) {
+        if version == Kip9Version::Beta && (outs_len == 1 || ins_len == 1 || (outs_len == 2 && ins_len == 2)) {
```
biryukovmaxim, line 100:
```suggestion
        if version == Kip9Version::Beta && matches!((outs_len, ins_len), (1, _) | (_, 1) | (2, 2)) {
```
not sure about that

tiram88 (changes requested):
I reviewed the last 3 commits.

My knowledge of the pruning system is not deep enough to bring a valuable review on this part of the PR.

consensus/src/processes/mass.rs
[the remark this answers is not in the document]
coderofstuff: It's covered by the first `OR` condition `outs_len == 1`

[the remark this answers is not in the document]
tiram88: of course!

consensus/src/processes/pruning_proof/mod.rs
```diff
@@ -691,11 +691,67 @@ impl PruningProofManager {
                     }
                 }
 
+                // Temp assertion for verifying a bug fix: assert that the full 2M chain is actually contained in the composed level proof
+                let set = BlockHashSet::from_iter(headers.iter().map(|h| h.hash));
```
coderofstuff, line 695: Since this uses the same logic as `block_at_depth` wouldn't this function panic before reaching this debug section, making the debug here unreachable for the error it's trying to capture?

consensus/src/processes/mass.rs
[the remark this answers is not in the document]
michaelsutton: It's included under `outs_len == 1` which doesn't care about number of inputs

michaelsutton: It's a cool comment that's for sure. I'll check the compiler output for both

coderofstuff (approved):
The fix is mainly the change in `header_processor`. There are 3 other sections unrelated to the pruning proof bug. Approving for these sections
- pruning proof bug fix
- kip9 change
- enable-mainnet-mining change

Please address Tiram's comments for the UTXO tracking section.

consensus/src/processes/pruning_proof/mod.rs
coderofstuff: The flow is (1) build_pruning_point_proof -> (2) prune -> (3) build_pruning_point_proof (sanity). It's at (1) where this is relevant.

notify/src/address/tracker.rs
```diff
@@ -568,7 +571,7 @@ impl Tracker {
     }
 
     pub fn max_addresses(&self) -> Option<usize> {
-        self.inner.read().max_addresses
+        Some(self.inner.read().max_addresses)
```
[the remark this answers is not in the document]
michaelsutton: Fixed

consensus/src/processes/mass.rs
coderofstuff: I have this pattern committed to memory, thanks @biryukovmaxim

coderofstuff (approved):

tiram88 (approved):

PR #453, opened 2024-04-15
Initialize pruning point depth store -- minor bug fix

michaelsutton:
Initialize the pruning point depth store for the rare case where the pruning point is the first sink during initial IBD (virtual processor calls the depth store for each new sink)

### DIFF
--- a/protocol/flows/src/v5/ibd/flow.rs
+++ b/protocol/flows/src/v5/ibd/flow.rs
@@ -294,6 +294,7 @@ impl IbdFlow {
         let mut trusted_set = pkg.build_trusted_subdag(entries)?;
 
         if self.ctx.config.enable_sanity_checks {
+            let con = self.ctx.consensus().unguarded_session_blocking();
             trusted_set = staging
                 .clone()
                 .spawn_blocking(move |c| {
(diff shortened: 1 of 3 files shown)


coderofstuff (approved):

0xtrossets:
Thanks @michaelsutton 
So this fixes https://github.com/kaspanet/rusty-kaspa/issues/455 ?

coderofstuff (dismissed):

protocol/flows/src/v5/ibd/flow.rs
```diff
@@ -294,6 +294,7 @@ impl IbdFlow {
         let mut trusted_set = pkg.build_trusted_subdag(entries)?;
 
         if self.ctx.config.enable_sanity_checks {
+            let con = self.ctx.consensus().unguarded_session_blocking();
```
coderofstuff, line 297: Since `unguarded_session_blocking()` now exists here, should that be used in the sync and validate way above for this line as well?

```
let proof = consensus.clone().spawn_blocking(move |c| c.validate_pruning_proof(&proof).map(|()| proof)).await?;
```

michaelsutton: No. The difference is that here we are already going into a blocking fn call for staging consensus, so we can access the current consensus via the blocking interface as well.

michaelsutton:
> Thanks @michaelsutton So this fixes #455 ?

Yes it should

PR #461, opened 2024-04-24
Add Dockerfile

timqi:


.dockerignore
```diff
@@ -0,0 +1,5 @@
+.git
+.github
+.gitignore
+
+Dockerfile
```
biryukovmaxim, line 5: could you add `target` folder to ignore

Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+FROM debian:bookworm
+
+COPY --from=builder /rusty-kaspa/target/release/kaspad /usr/bin/
+COPY --from=builder /rusty-kaspa/target/release/kaspa-wallet /usr/bin/
+COPY --from=builder /rusty-kaspa/target/release/kaspa_p2p_client /usr/bin/
+COPY --from=builder /rusty-kaspa/target/release/kaspa_p2p_server /usr/bin/
+COPY --from=builder /rusty-kaspa/target/release/kaspa_p2p_server /usr/bin/
+COPY --from=builder /rusty-kaspa/target/release/kaspa-wrpc-proxy /usr/bin/
+COPY --from=builder /rusty-kaspa/target/release/simpa /usr/bin/
+COPY --from=builder /rusty-kaspa/target/release/rothschild /usr/bin/
```
biryukovmaxim, line 31: could you split each binary to separated targets to make it possible to only build  specific stage/target/container with binary. I think for the most part everyone will only be looking for kaspad

Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+        libllvm-ocaml-dev libomp-dev libomp5 \
+        lld lldb llvm-dev llvm-runtime \
+        llvm python3-clang
+
+RUN cargo install wasm-pack
+RUN rustup target add wasm32-unknown-unknown
+
+COPY . /rusty-kaspa
+WORKDIR /rusty-kaspa
+RUN cargo build --release
```
biryukovmaxim, line 19: I think it would be better to put todo to cache rust dependencies and reuse them during build

biryukovmaxim (changes requested):

.dockerignore
timqi: sure

Dockerfile
timqi: Just add kaspad/kaspa-wallet, the most often used binaries to image

Dockerfile
timqi: Ok, added a TODO comment

biryukovmaxim (approved):
LGTM!

Dockerfile
```diff
@@ -0,0 +1,26 @@
+FROM rust:bookworm AS builder
+
+RUN apt-get update
+RUN apt-get install -y curl git build-essential libssl-dev pkg-config
```
starkbamse, line 4: For what purpose do we need git here? Since we are not cloning anything perhaps `git` could be removed. Same goes for `curl` I believe.

timqi: true, removed these two.

coderofstuff (changes requested):
Thank you for the contribution.

Content looks alright to me. I have some minor comments:
- [ ] Please add a section in the README.md, maybe under the `Installation` section as `Building on Docker`.
- [ ] The Dockerfile doesn't have an ENTRYPOINT or CMD (which is fine), so the documentation should also note examples on how to use the two binaries being built inside this Dockerfile

supertypo:
README.md: I think rusty-kaspa is a better tag

A few suggestions on changes to the Dockerfile based on Docker best practices:

* A VOLUME should be declared for the data folder (~/.rust-kaspa)
* EXPOSE could be used to declare the exposed ports (e.g. 16111, 16110, 17110, 18110 for mainnet)
* ENTRYPOINT should be kaspad, make sure it's on PATH or have an explicit path.
* CMD should be common sense default arguments to kaspad in array format (e.g. ["--yes", "--nologfiles", "--disable-upnp", "--utxoindex", "--rpclisten=0.0.0.0:16110", "--rpclisten-borsh=0.0.0.0:17110", "--rpclisten-json=0.0.0.0:18110"])
Docker containers runs on an internal network by default, upnp is not relevant and rpc ports must be exposed to all interfaces for it to be reachable using Docker port forwarding. The listen port (16111) is already public by default.
* A USER could ideally be declared to run with non-root permissions. It will complicate things a bit though as the user must be created beforehand. If a custom host folder is to be mounted in as the datafolder the folder must also be chowned to 16111 on the host. In case this is added, info on how to handle host mounts must be added to the README.md

someone235:
Docker support was added on #742

PR #468, opened 2024-05-04
Check block is not pruned

someone235:


someone235:
It's already guaranteed that any pruned block will be rejected because of missing parents (its parents should be pruned as well), so this extra check is redundant. I moved the tests to #752

PR #470, opened 2024-05-07
CLI watch-only account implementation

1bananagirl:
The new account type watch-only supports:

- bip32
- multisig

Changes:

- print the extended public key for the command "export mnemonic" like it is in the golang kaspwallet
- added handling of keyless accounts in _select_account_with_args_ and _list_ functions
- watch-only and xpub text hints added to the _list_ function to differentiate account type

Future:

- creating unsigned transactions with watch-only wallets

### DIFF
--- a/cli/src/cli.rs
+++ b/cli/src/cli.rs
@@ -552,6 +552,14 @@ impl KaspaCli {
             list_by_key.push((key.clone(), prv_key_accounts));
         }
 
+        let mut watchonly_accounts = Vec::<(usize, Arc<dyn Account>)>::new();
+        while let Some(account) = self.wallet.accounts(None).await?.try_next().await? {
+            if account.account_kind() == WATCH_ONLY_ACCOUNT_KIND {
+                watchonly_accounts.push((flat_list.len(), account.clone()));
+                flat_list.push(account.clone());
+            }
+        }
+
         if flat_list.is_empty() {
             return Err(Error::NoAccounts);
         } else if autoselect && flat_list.len() == 1 {
@@ -569,6 +577,16 @@ impl KaspaCli {
                     let ls_string = account.get_list_string().unwrap_or_else(|err| panic!("{err}"));
                     tprintln!(self, "    {seq}: {ls_string}");
                 })
+            });
+
+            if watchonly_accounts.len() > 0 {
+                tprintln!(self, "• Watch-only");
+            }
+
+            watchonly_accounts.iter().for_each(|(seq, account)| {
+                let seq = style(seq.to_string()).cyan();
+                let ls_string = account.get_list_string().unwrap_or_else(|err| panic!("{err}"));
+                tprintln!(self, "    {seq}: {ls_string}");
             });
 
             tprintln!(self);
--- a/cli/src/cli.rs
+++ b/cli/src/cli.rs
@@ -5,7 +5,9 @@ use crate::modules::miner::Miner;
 use crate::modules::node::Node;
 use crate::notifier::{Notification, Notifier};
 use crate::result::Result;
+use kaspa_bip32::Prefix;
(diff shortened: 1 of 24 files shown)


cli/src/wizards/account.rs
```diff
@@ -85,3 +85,49 @@ async fn create_multisig(ctx: &Arc<KaspaCli>, account_name: Option<String>, mnem
...
+        loop {
+            let xpub_key = term
+                .ask(false, "For a multisig watch-only account enter an additional extended public key: (empty to skip or stop) ")
+                .await?;
+            if xpub_key.is_empty() {
+                break;
+            }
+            additional_xpubs.push(xpub_key.trim().to_owned());
+        }
+    }
```
biryukovmaxim, line 115: should be part of multisig import functionality

wallet/core/src/account/variants/watchonly.rs
```diff
@@ -0,0 +1,300 @@
...
+    }
+}
+
+pub struct WatchOnly {
+    inner: Arc<Inner>,
+    xpub_keys: ExtendedPublicKeys,
+    minimum_signatures: u16,
+    ecdsa: bool,
+    derivation: Arc<AddressDerivationManager>,
+}
```
biryukovmaxim, line 91: make watch only acc only suitable for a single acc

biryukovmaxim (changes requested):

cli/src/cli.rs
```diff
@@ -655,6 +675,23 @@ impl KaspaCli {
...
             }
         }
+
+        let mut unfiltered_accounts = self.wallet.accounts(None).await?;
+        while let Some(account) = unfiltered_accounts.try_next().await? {
+            if account.account_kind() == WATCH_ONLY_ACCOUNT_KIND {
+                tprintln!(self, "• {}", account.get_list_string()?);
+                let watch_only = account.downcast_arc::<WatchOnly>()?;
+                let xpubs = watch_only.xpub_keys();
+                if xpubs.len() > 1 {
```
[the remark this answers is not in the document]
aspect: I actually gave that guidance as I think that it is better to be explicit about the purpose of the account. But you are right as multisig is technically a watch account by default.  Perhaps the account kind should then be explicit, as in "bip32-watch"?

1bananagirl:
Thanks for you review Maxim!

## Refactoring

I've renamed the watch-only account to bip32-watch and removed multisig polymorphism, thus multisig watch-only remaining solely in the multisig account.

I've followed aspect's advice to add a helper function in the wallet module to format XPUB export according to network.

The new helper function is triggered by providing a new type TaggedXpub that is a pair formed of an extended public key and the network ID.

Inside the bip32 prefix submodule resides a new converter to get the right prefix for a given network.

### Account feature trait

The notion of feature was added to the account trait as a fn, defaulting to Option None, and is implemented in Bip32Watch and Multisig/watch-only to return a string giving an indication on the special feature of an account - which is watch-only in our case - an indication used to highlight that feature in the list and details commands output.

That step was necessary for the iterator filtering in details and list commands to be simple stupid finding out which account without private keys to consider printing.

### Commands

CLI commands for importing watch-only accounts now look like this:

```
   account import bip32-watch
   account import multisig-watch
```

I'm open to switch this to "account create" if needed.

I've found it more efficient to make a new parameter "multisig-watch" instead of bypassing "account create multisig", but if you insist, I can make that work instead.


### Output

The display of xpubs for watch-only account has been moved from the CLI list command output to the details output in order to not clutter the list view.

According to the currently selected network, the xpubs are now formatted with the appropriate prefix.

aspect (approved):
This looks good.  We are going to need to rebase this against wRPC serialization PR (once it is merged) before merging this PR.

aspect:
Merged into #506

PR #476, opened 2024-05-16
TN11 bug fix: activate mass hashing when modifying a cached block template  

michaelsutton:
Fixes a TN11 bug where transaction mass field hashing was not activated when modifying a cached block template. 
This produced invalid blocks when two miners with different addresses were connected to the same node, frequently triggering this code path and resulting in many bad Merkle root blocks

### DIFF
--- a/consensus/core/src/api/mod.rs
+++ b/consensus/core/src/api/mod.rs
@@ -190,6 +190,10 @@ pub trait ConsensusApi: Send + Sync {
         unimplemented!()
     }
 
+    fn calc_transaction_hash_merkle_root(&self, txs: &[Transaction], pov_daa_score: u64) -> Hash {
+        unimplemented!()
+    }
+
     fn validate_pruning_proof(&self, proof: &PruningPointProof) -> PruningImportResult<()> {
         unimplemented!()
     }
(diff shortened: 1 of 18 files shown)


michaelsutton:
Perhaps a better way to handle this is to move more of `modify_block_template` logic to within consensus where storage mass activation is already accessible and can be checked. It feels slightly weird that consensus sensitive code is in the mempool block builder

coderofstuff:
> Perhaps a better way to handle this is to move more of `modify_block_template` logic to within consensus where storage mass activation is already accessible and can be checked. It feels slightly weird that consensus sensitive code is in the mempool block builder

I agree that it feels a bit off that consensus sensitive logic isn't in consensus. I would think a better way is for consensus to have something like `consensus.calc_hash_merkle_root` and from within consensus, decide whether to use the HF version or not.

michaelsutton:
> > Perhaps a better way to handle this is to move more of `modify_block_template` logic to within consensus where storage mass activation is already accessible and can be checked. It feels slightly weird that consensus sensitive code is in the mempool block builder
> 
> I agree that it feels a bit off that consensus sensitive logic isn't in consensus. I would think a better way is for consensus to have something like `consensus.calc_hash_merkle_root` and from within consensus, decide whether to use the HF version or not.

Done.

someone235 (dismissed):

coderofstuff (dismissed):

someone235 (approved):

coderofstuff (approved):

PR #479, opened 2024-05-29
Add input signature_script checking to submitTransaction RPC

### LINKED ISSUE #452: Signing with incomplete keys must produce easily understandable error
**Describe the bug**
When not all needed keys are passed to `sign`, it produces a transaction with an input missing a signature, eating up the error. This results in the error only appearing upon submitting the TX, which yields an error not easily linkable to missing keys during the signing process.

**To Reproduce**
1. Sign a tx but skip some needed keys
2. Submit the tx to the node
3. Get the error something like `stack needs 1 but has 0`

**Expected behavior**
1. If there are unsigned inputs, an error should be thrown during the `sign` process that clearly indicates so
2. Tx submit or script engine needs some sort of a pre-check for a completely empty script sig to produce an easy to understand error.

KaffinPX:
Closes #452


rpc/service/src/service.rs
```diff
@@ -493,6 +493,13 @@ NOTE: This error usually indicates an RPC conversion error between the node and
 
         let transaction: Transaction = (&request.transaction).try_into()?;
         let transaction_id = transaction.id();
+        let inputs = &transaction.inputs;
+
+        if let Some((index, _)) = inputs.iter().enumerate().find(|(_, input)| input.signature_script.is_empty()) {
+            let err = RpcError::EmptySignatureScript(transaction_id, index);
+            return Err(err)
```
aspect, line 500: Remove let, collapse declaration into `Err(RpcError::EmptySignatureScript(transaction_id, index))`
I think it would be better to turn this into a single if/else statement: `if Some() { Err() } else { submission handling; Ok() }`

aspect (changes requested):

rpc/core/src/error.rs
```diff
@@ -47,6 +47,9 @@ pub enum RpcError {
     #[error("Rejected transaction {0}: {1}")]
     RejectedTransaction(RpcTransactionId, String),
 
+    #[error("Transaction {0} has an input with an empty signature_script at index {1}.")]
+    EmptySignatureScript(RpcTransactionId, usize),
```
biryukovmaxim, line 51: it can be extended to find all indexes that include empty signatures, WDYT?

aspect: Yeah, I had this thought in the back of my head, but somehow didn't bring that up, perhaps in order to keep this KISS.  But yeah, @KaffinPX it might make sense to enumerate all inputs and return a concatenated string containing all indexes. U can use `.filter_map(|(index,input)| input.signature_script.is_empty().then(index)).collect::<Vec<_>>();` and react if the resulting vec is not empty... (or along those lines).  Then do `indexes.iter().map(String::from).join(",")` to get a single string that can be included in the error.

rpc/service/src/service.rs
```diff
@@ -493,17 +493,28 @@ NOTE: This error usually indicates an RPC conversion error between the node and
...
-            err
-        })?;
-        Ok(SubmitTransactionResponse::new(transaction_id))
+        let inputs = &transaction.inputs;
+
+        let empty_indices = inputs
+            .iter()
+            .enumerate()
+            .filter_map(|(index, input)| (input.signature_script.is_empty()).then_some(index.to_string()))
+            .collect::<Vec<_>>();
```
biryukovmaxim, line 502: Can be directly joined to the string by
https://docs.rs/itertools/latest/itertools/trait.Itertools.html#method.join

rpc/service/src/service.rs
```diff
@@ -493,17 +493,28 @@ NOTE: This error usually indicates an RPC conversion error between the node and
...
+        if !empty_indices.is_empty() {
+            Err(RpcError::EmptySignatureScript(transaction_id, empty_indices.join(", ")))
+        } else {
+            let session = self.consensus_manager.consensus().unguarded_session();
+            let orphan = match allow_orphan {
+                true => Orphan::Allowed,
+                false => Orphan::Forbidden,
+            };
+            self.flow_context.submit_rpc_transaction(&session, transaction, orphan).await.map_err(|err| {
```
biryukovmaxim, line 511: use inspect_err instead of map_err

biryukovmaxim (changes requested):

rpc/service/src/service.rs
aspect: It’s joined into the string when producing an error. I would rather test Vec and if not empty join as opposed to join and then test string. (saves one fn call).

This is how @KaffinPX had it, I asked to move join inside the if.

rpc/service/src/service.rs
aspect: inspect_err() was not stable when this was written :)

rpc/service/src/service.rs
biryukovmaxim: u can call next to understand if it has at least one element and then chain it back. Join method allocates one big string. Current solution allocates many small string, then allocates vector of them, then allocates one big string

aspect: @biryukovmaxim post an example impl.

biryukovmaxim:
```rust
 let mut empty_indices =
            inputs.iter().enumerate().filter_map(|(index, input)| (input.signature_script.is_empty()).then_some(index.to_string()));
        let first_empty = empty_indices.next();
        if let Some(first_empty) = first_empty {
            use itertools::Itertools;
            Err(RpcError::EmptySignatureScript(transaction_id, iter::once(first_empty).chain(empty_indices).join(", ")))
        } else { ...   
```

it requires adding itertools to cargo toml:
```
itertools.workspace = true
```

biryukovmaxim (approved):

aspect (approved):
LGTM!

michaelsutton:
I suggest the following instead of the above approach:

```diff
diff --git a/consensus/Cargo.toml b/consensus/Cargo.toml
index f151e404b..3f4a1b456 100644
--- a/consensus/Cargo.toml
+++ b/consensus/Cargo.toml
@@ -30,6 +30,7 @@ kaspa-muhash.workspace = true
 kaspa-notify.workspace = true
 kaspa-pow.workspace = true
 kaspa-txscript.workspace = true
+kaspa-txscript-errors.workspace = true
 kaspa-utils.workspace = true
 log.workspace = true
 once_cell.workspace = true
diff --git a/consensus/core/src/errors/tx.rs b/consensus/core/src/errors/tx.rs
index ec1899faa..1aeb5220f 100644
--- a/consensus/core/src/errors/tx.rs
+++ b/consensus/core/src/errors/tx.rs
@@ -80,6 +80,9 @@ pub enum TxRuleError {
     #[error("failed to verify the signature script: {0}")]
     SignatureInvalid(TxScriptError),
 
+    #[error("failed to verify empty signature script. Inner error: {0}")]
+    SignatureEmpty(TxScriptError),
+
     #[error("input {0} sig op count is {1}, but the calculated value is {2}")]
     WrongSigOpCount(usize, u64, u64),
 
diff --git a/consensus/src/processes/transaction_validator/transaction_validator_populated.rs b/consensus/src/processes/transaction_validator/transaction_validator_populated.rs
index 1835ba61e..586b23e57 100644
--- a/consensus/src/processes/transaction_validator/transaction_validator_populated.rs
+++ b/consensus/src/processes/transaction_validator/transaction_validator_populated.rs
@@ -1,7 +1,11 @@
 use crate::constants::{MAX_SOMPI, SEQUENCE_LOCK_TIME_DISABLED, SEQUENCE_LOCK_TIME_MASK};
-use kaspa_consensus_core::{hashing::sighash::SigHashReusedValues, tx::VerifiableTransaction};
+use kaspa_consensus_core::{
+    hashing::sighash::SigHashReusedValues,
+    tx::{TransactionInput, VerifiableTransaction},
+};
 use kaspa_core::warn;
 use kaspa_txscript::{get_sig_op_count, TxScriptEngine};
+use kaspa_txscript_errors::TxScriptError;
 
 use super::{
     errors::{TxResult, TxRuleError},
@@ -167,14 +171,21 @@ impl TransactionValidator {
         let mut reused_values = SigHashReusedValues::new();
         for (i, (input, entry)) in tx.populated_inputs().enumerate() {
             let mut engine = TxScriptEngine::from_transaction_input(tx, input, i, entry, &mut reused_values, &self.sig_cache)
-                .map_err(TxRuleError::SignatureInvalid)?;
-            engine.execute().map_err(TxRuleError::SignatureInvalid)?;
+                .map_err(|err| map_script_err(err, input))?;
+            engine.execute().map_err(|err| map_script_err(err, input))?;
         }
 
         Ok(())
     }
 }
 
+fn map_script_err(script_err: TxScriptError, input: &TransactionInput) -> TxRuleError {
+    match input.signature_script.is_empty() {
+        true => TxRuleError::SignatureEmpty(script_err),
+        false => TxRuleError::SignatureInvalid(script_err),
+    }
+}
+
 #[cfg(test)]
 mod tests {
     use super::super::errors::TxRuleError;
```

biryukovmaxim:
> ```diff
> match input.signature_script.is_empty() {
> +        true => TxRuleError::SignatureEmpty(script_err),
> +        false => TxRuleError::SignatureInvalid(script_err),
> +    }
> ```

match bool is always looks weird. Why not use simple if-else?

biryukovmaxim (approved):

michaelsutton (approved):

PR #494, opened 2024-06-22
On-demand calculation for Ghostdag for Higher Levels

coderofstuff:
## Introduction

Currently, for each block that we process header processor will calculate ghostdag for all levels. This preprocessing allows for the faster calculation of the pruning proof during pruning but imposes a significant overhead for the processing of each block. It is far better to defer this higher level GD calculation for when it's actually needed - during pruning proof building.

## Main Changes
- During header processing, calculate only the ghostdag for the header at level 0 (primary ghostdag). So instead of calculating for 255 levels, we only calculate for the one.
- Calculate the ghostdag for when it's needed (when `build_pruning_proof` is called):
  - during initial IBD
  - during pruning point movement
- Introduces 2 new prefixes:
  - `TempGhostdag`
  - `TempGhostdagCompact`
- Includes upgrade logic from version 3 to 4 that doesn't require a DB reset
  - Deletes all higher level GD records which are no longer necessary given that these will be calculated on-demand

## Not included
- minimization of proof size


consensus/src/processes/pruning_proof/mod.rs
```diff
@@ -822,19 +839,20 @@ impl PruningProofManager {
...
-                        || required_block_chain.contains(&current_required_chain_block.hash)
+                        || selected_chain.contains(&current_required_chain_block.hash)
                     {
                         intersected_with_required_block_chain = true;
                     }
                 }
 
                 if current_header.direct_parents().is_empty() // Stop at genesis
-                    || (pp_header.header.blue_score >= current_header.blue_score + required_level_0_depth
+                    // Need to ensure this does the same 2M+1 depth that block_at_depth does
```
someone235, line 849: Is it a TODO?

consensus/src/processes/pruning_proof/mod.rs
```diff
@@ -614,45 +712,312 @@ impl PruningProofManager {
...
+    }
+
+    fn find_sufficient_root(
+        &self,
+        pp_header: &HeaderWithBlockLevel,
+        level: BlockLevel,
+        required_block: Option<Hash>,
+        temp_db: Arc<DB>,
+    ) -> PruningProofManagerInternalResult<(Arc<DbGhostdagStore>, Hash, Hash)> {
+        let relations_service = MTRelationsService::new(self.relations_stores.clone(), level);
```
someone235, line 798: Maybe just build it in the constructor? It's easier this way to differentiate between temp stores and permanent ones

consensus/src/processes/pruning_proof/mod.rs
```diff
@@ -743,12 +744,26 @@ impl PruningProofManager {
...
     fn find_selected_parent_header_at_level(
         &self,
         header: &Header,
         level: BlockLevel,
+        relations_service: MTRelationsService<DbRelationsStore>,
     ) -> PruningProofManagerInternalResult<Arc<Header>> {
-        let parents = self.parents_manager.parents_at_level(header, level);
+        // Logic of apply_proof only inserts parent entries for a header from the proof
+        // into the relations store for a level if there was GD data in the old stores for that
+        // header. To mimic that logic here, we need to filter out parents that are NOT in the relations_service
```
someone235, line 757: Don't we add all blocks to `relations` in `apply_proof`?

consensus/src/processes/pruning_proof/mod.rs
```diff
@@ -922,18 +944,20 @@ impl PruningProofManager {
...
 
             tries += 1;
             if finished_headers {
-                panic!("Failed to find sufficient root for level {level} after {tries} tries. Headers below the current depth of {required_level_0_depth} are already pruned")
+                warn!("Failed to find sufficient root for level {level} after {tries} tries. Headers below the current depth of {required_level_0_depth} are already pruned. Trying anyway.");
+                break Ok((ghostdag_store, selected_tip, root));
             }
             required_level_0_depth <<= 1;
-            warn!("Failed to find sufficient root for level {level} after {tries} tries. Retrying again to find with depth {required_level_0_depth}");
+            trace!("Failed to find sufficient root for level {level} after {tries} tries. Retrying again to find with depth {required_level_0_depth}");
```
someone235, line 960: I think at least for now it should be `warn` since it should be pretty unlikely event that we should be aware of (at least with the safety margin of trying 5/4 of the required depth at first try)

consensus/src/processes/pruning_proof/mod.rs
```diff
@@ -922,18 +944,20 @@ impl PruningProofManager {
...
-                && (root == self.genesis_hash || ghostdag_store.get_blue_score(selected_tip).unwrap() >= required_level_depth)
+                && (root == self.genesis_hash || ghostdag_store.get_blue_score(selected_tip).unwrap() > required_level_depth)
             {
                 break Ok((ghostdag_store, selected_tip, root));
             }
 
             tries += 1;
             if finished_headers {
-                panic!("Failed to find sufficient root for level {level} after {tries} tries. Headers below the current depth of {required_level_0_depth} are already pruned")
+                warn!("Failed to find sufficient root for level {level} after {tries} tries. Headers below the current depth of {required_level_0_depth} are already pruned. Trying anyway.");
```
someone235, line 956: Why not panic?

consensus/src/processes/ghostdag/ordering.rs
```diff
@@ -44,8 +44,16 @@ impl Ord for SortableBlock {
 impl<T: GhostdagStoreReader, S: RelationsStoreReader, U: ReachabilityService, V: HeaderStoreReader> GhostdagManager<T, S, U, V> {
     pub fn sort_blocks(&self, blocks: impl IntoIterator<Item = Hash>) -> Vec<Hash> {
         let mut sorted_blocks: Vec<Hash> = blocks.into_iter().collect();
-        sorted_blocks
-            .sort_by_cached_key(|block| SortableBlock { hash: *block, blue_work: self.ghostdag_store.get_blue_work(*block).unwrap() });
+        sorted_blocks.sort_by_cached_key(|block| SortableBlock {
+            hash: *block,
+            // Since we're only calculating GD at all levels on-demand, we may get blocks from the relations
```
someone235, line 49: This comment is a little out of context because this function doesn't use `relations`. I guess you're referring to the GD calculation in the process of ordering the mergeset? In which part of this process do we rely on `relations`? I thought we use `parents_at_level` for that.

About `make relations store only return parents at the same or higher level`: can you specify what function do you refer that doesn't do it?

someone235 (changes requested):
It shouldn't be a blocker for merge, but it would be nice if each function will have a comment with a bit of explanation, because it can be hard to get the guiding logic for the whole process if you're not familiar with it

consensus/src/processes/ghostdag/ordering.rs
coderofstuff: I encounter a panic when when doing GD in `calc_gd_for_all_levels` during pruning point movement which happens `.unwrap()` that I switched to `.unwrap_or_default()`. I agree it's a little out of context here, but I placed it here so I can pinpoint where exactly the error occurs.

> About make relations store only return parents at the same or higher level: can you specify what function do you refer that doesn't do it?

This function `unordered_mergeset_without_selected_parent`, and specifically this line in `mergeset.rs`:

```
let current_parents = relations.get_parents(current).unwrap();
```

Is where I thought that the relations may be returning parents that are not in the level - but looking at it again, I think this assumption is wrong since `relations` passed to GD manager should already be relation service at a given level.

I need to revisit this change and dig deeper into what actually is causing this code to try to look for hash that isn't there yet (or maybe is no longer there?).

consensus/src/pipeline/pruning_processor/processor.rs
```diff
@@ -457,7 +457,8 @@ impl PruningProcessor {
         );
 
         if self.config.enable_sanity_checks {
-            self.assert_proof_rebuilding(proof, new_pruning_point);
+            // self.assert_proof_rebuilding(proof, new_pruning_point);
```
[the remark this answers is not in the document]
coderofstuff: No, I carried that change over from your original changes. I'm uncommenting it instead.

consensus/src/processes/pruning_proof/mod.rs
coderofstuff: No, it's to note that the +1 here is intentional since that's how `block_at_depth` behaves. The real TODO isn't written anywhere yet which is to update `block_at_depth` to actually find the block at depth `2M` instead of block at depth `2M+1`.

consensus/src/processes/pruning_proof/mod.rs
```diff
@@ -777,29 +792,31 @@ impl PruningProofManager {
...
-            required_level_depth
-        } else {
-            self.estimated_blue_depth_at_level_0(
-                level,
-                required_level_depth * 5 / 4, // We take a safety margin
-                current_dag_level,
-            )
-        };
+        let mut required_level_0_depth = required_level_depth;
+        // let mut required_level_0_depth = if level == 0 {
```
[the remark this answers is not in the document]
coderofstuff: I commented it out temporarily while waiting to figure out what the intended logic of `estimated_blue_depth_at_level_0` was. I'll be putting this back.

consensus/src/processes/pruning_proof/mod.rs
coderofstuff: Can you elaborate what you mean here? Do you mean relations_service is created outside of `find_sufficient_root`?

consensus/src/pipeline/pruning_processor/processor.rs
```diff
@@ -413,7 +411,9 @@ impl PruningProcessor {
                         let mut staging_level_relations = StagingRelationsStore::new(&mut level_relations_write[level]);
                         relations::delete_level_relations(MemoryWriter, &mut staging_level_relations, current).unwrap_option();
                         staging_level_relations.commit(&mut batch).unwrap();
-                        self.ghostdag_stores[level].delete_batch(&mut batch, current).unwrap_option();
+                        if level == 0 {
+                            self.ghostdag_primary_store.delete_batch(&mut batch, current).unwrap_option();
```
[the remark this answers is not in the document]
coderofstuff: Addressed in https://github.com/kaspanet/rusty-kaspa/pull/494/commits/f9b3fda63bab6d8b7e8c7bf62e72fda5f544041f

consensus/src/pipeline/pruning_processor/processor.rs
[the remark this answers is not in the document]
coderofstuff: Addressed in https://github.com/kaspanet/rusty-kaspa/pull/494/commits/f9b3fda63bab6d8b7e8c7bf62e72fda5f544041f

consensus/src/processes/pruning_proof/mod.rs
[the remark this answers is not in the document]
coderofstuff: Addressed in https://github.com/kaspanet/rusty-kaspa/pull/494/commits/f9b3fda63bab6d8b7e8c7bf62e72fda5f544041f

consensus/src/processes/pruning_proof/mod.rs
coderofstuff: During the first sync with another node that isn't on this version, you'd get something like:

```
2024-07-04 00:31:03.146-06:00 [INFO ] Validating level 24 from the pruning point proof (2002 headers)
2024-07-04 00:31:03.580-06:00 [INFO ] Validating level 23 from the pruning point proof (2002 headers)
2024-07-04 00:31:04.006-06:00 [INFO ] Validating level 22 from the pruning point proof (2002 headers)
2024-07-04 00:31:04.431-06:00 [INFO ] Validating level 21 from the pruning point proof (2002 headers)
2024-07-04 00:31:04.868-06:00 [INFO ] Validating level 20 from the pruning point proof (2054 headers)
2024-07-04 00:31:05.316-06:00 [INFO ] Validating level 19 from the pruning point proof (2101 headers)
2024-07-04 00:31:05.758-06:00 [INFO ] Validating level 18 from the pruning point proof (2108 headers)
2024-07-04 00:31:06.223-06:00 [INFO ] Validating level 17 from the pruning point proof (2002 headers)
2024-07-04 00:31:06.659-06:00 [INFO ] Validating level 16 from the pruning point proof (2002 headers)
2024-07-04 00:31:07.096-06:00 [INFO ] Validating level 15 from the pruning point proof (2001 headers)
2024-07-04 00:31:07.539-06:00 [INFO ] Validating level 14 from the pruning point proof (2020 headers)
2024-07-04 00:31:07.994-06:00 [INFO ] Validating level 13 from the pruning point proof (2001 headers)
2024-07-04 00:31:08.415-06:00 [INFO ] Validating level 12 from the pruning point proof (2001 headers)
2024-07-04 00:31:08.867-06:00 [INFO ] Validating level 11 from the pruning point proof (2001 headers)
2024-07-04 00:31:09.288-06:00 [INFO ] Validating level 10 from the pruning point proof (2001 headers)
2024-07-04 00:31:09.695-06:00 [INFO ] Validating level 9 from the pruning point proof (2001 headers)
2024-07-04 00:31:10.119-06:00 [INFO ] Validating level 8 from the pruning point proof (2001 headers)
2024-07-04 00:31:10.545-06:00 [INFO ] Validating level 7 from the pruning point proof (2001 headers)
2024-07-04 00:31:10.962-06:00 [INFO ] Validating level 6 from the pruning point proof (2001 headers)
2024-07-04 00:31:11.391-06:00 [INFO ] Validating level 5 from the pruning point proof (2001 headers)
2024-07-04 00:31:11.804-06:00 [INFO ] Validating level 4 from the pruning point proof (2001 headers)
2024-07-04 00:31:12.219-06:00 [INFO ] Validating level 3 from the pruning point proof (2001 headers)
2024-07-04 00:31:12.636-06:00 [INFO ] Validating level 2 from the pruning point proof (2001 headers)
2024-07-04 00:31:13.053-06:00 [INFO ] Validating level 1 from the pruning point proof (2001 headers)
2024-07-04 00:31:13.492-06:00 [INFO ] Validating level 0 from the pruning point proof (2001 headers)
```

And along with that you'd see these:

```
2024-07-04 00:31:27.716-06:00 [INFO ] Building the proof which was just applied (sanity test)
2024-07-04 00:31:27.986-06:00 [WARN ] Failed to find sufficient root for level 23 after 1 tries. Retrying again to find with depth 2560000
2024-07-04 00:31:28.231-06:00 [WARN ] Failed to find sufficient root for level 15 after 1 tries. Headers below the current depth of 5000 are already pruned. Trying anyway.
2024-07-04 00:31:28.293-06:00 [WARN ] Failed to find sufficient root for level 13 after 1 tries. Headers below the current depth of 2500 are already pruned. Trying anyway.
2024-07-04 00:31:28.321-06:00 [WARN ] Failed to find sufficient root for level 12 after 1 tries. Headers below the current depth of 2500 are already pruned. Trying anyway.
2024-07-04 00:31:28.350-06:00 [WARN ] Failed to find sufficient root for level 11 after 1 tries. Headers below the current depth of 2500 are already pruned. Trying anyway.
2024-07-04 00:31:28.379-06:00 [WARN ] Failed to find sufficient root for level 10 after 1 tries. Headers below the current depth of 2500 are already pruned. Trying anyway.
2024-07-04 00:31:28.409-06:00 [WARN ] Failed to find sufficient root for level 9 after 1 tries. Headers below the current depth of 2500 are already pruned. Trying anyway.
2024-07-04 00:31:28.441-06:00 [WARN ] Failed to find sufficient root for level 8 after 1 tries. Headers below the current depth of 2500 are already pruned. Trying anyway.
2024-07-04 00:31:28.473-06:00 [WARN ] Failed to find sufficient root for level 7 after 1 tries. Headers below the current depth of 2500 are already pruned. Trying anyway.
2024-07-04 00:31:28.502-06:00 [WARN ] Failed to find sufficient root for level 6 after 1 tries. Headers below the current depth of 2500 are already pruned. Trying anyway.
2024-07-04 00:31:28.533-06:00 [WARN ] Failed to find sufficient root for level 5 after 1 tries. Headers below the current depth of 2500 are already pruned. Trying anyway.
2024-07-04 00:31:28.565-06:00 [WARN ] Failed to find sufficient root for level 4 after 1 tries. Headers below the current depth of 2500 are already pruned. Trying anyway.
2024-07-04 00:31:28.596-06:00 [WARN ] Failed to find sufficient root for level 3 after 1 tries. Headers below the current depth of 2500 are already pruned. Trying anyway.
2024-07-04 00:31:28.628-06:00 [WARN ] Failed to find sufficient root for level 2 after 1 tries. Headers below the current depth of 2500 are already pruned. Trying anyway.
2024-07-04 00:31:28.658-06:00 [WARN ] Failed to find sufficient root for level 1 after 1 tries. Headers below the current depth of 2500 are already pruned. Trying anyway.
2024-07-04 00:31:28.849-06:00 [WARN ] Locally built proof for level 0 does not match the applied one
```

The new way of finding root takes up some safety margin, but the old nodes don't so we wouldn't otherwise be able to sync if we panic.

kaspad/src/daemon.rs
```diff
@@ -308,13 +309,92 @@ do you confirm? (answer y/n or pass --yes to the Kaspad command line to confirm
...
+                    let mut batch = rocksdb::WriteBatch::default();
+                    let mut gd_record_count: u32 = 0;
+                    let mut compact_record_count: u32 = 0;
+
+                    for level in 1..(config.max_block_level + 1) {
+                        let level_bytes = level.to_le_bytes();
+                        let ghostdag_prefix_vec = DatabaseStorePrefixes::Ghostdag.into_iter().chain(level_bytes).collect_vec();
+                        let ghostdag_prefix = ghostdag_prefix_vec.as_slice();
+
+                        for result in consensus_db.iterator(rocksdb::IteratorMode::From(ghostdag_prefix, rocksdb::Direction::Forward))
```
michaelsutton, line 341: It is possible to delete by range which is more efficient. See, e.g., here https://github.com/kaspanet/rusty-kaspa/blob/master/database/src/writer.rs#L53

consensus/src/processes/ghostdag/ordering.rs
coderofstuff: I reverted this specific change. It doesn't seem like I'm encountering the panic anymore, likely due to cleaning up a bunch of the ghostdag_store references everywhere.

consensus/src/processes/pruning_proof/mod.rs
coderofstuff: We apply all the blocks that are in the proof. Turns out, the logic being applied in the filtering in `apply_proof`. The filtering logic in `apply_proof` is this (in master):

```
            for header in headers.iter() {
                let parents = Arc::new(
                    self.parents_manager
                        .parents_at_level(header, level as BlockLevel)
                        .iter()
                        .copied()
                        .filter(|parent| self.ghostdag_stores[level].has(*parent).unwrap())
                        .collect_vec()
                        .push_if_empty(ORIGIN),
                );

                self.relations_stores.write()[level].insert(header.hash, parents.clone()).unwrap();
```

From reading the code around this in `apply_proof`, it looks like the only reason to do the filtering is to ensure that only the "relevant parents" (those that are included in the proof, and are consequently added to the ghostdag_store) of any block are considered.

I've updated the comment in this PR to read:

```
        // Parents manager parents_at_level may return parents that aren't in relations_service, so it's important
        // to filter to include only parents that are in relations_service.
        let parents = self
            .parents_manager
            .parents_at_level(header, level)
            .iter()
            .copied()
            .filter(|parent| relations_service.has(*parent).unwrap())
            .collect_vec()
            .push_if_empty(ORIGIN);
```

consensus/src/processes/pruning_proof/mod.rs
```diff
@@ -229,26 +251,29 @@ impl PruningProofManager {
...
                         .copied()
-                        .filter(|parent| self.ghostdag_stores[level].has(*parent).unwrap())
+                        .filter(|parent| level_ancestors.contains(parent))
                         .collect_vec()
                         .push_if_empty(ORIGIN),
                 );
 
                 self.relations_stores.write()[level].insert(header.hash, parents.clone()).unwrap();
-                let gd = if header.hash == self.genesis_hash {
-                    self.ghostdag_managers[level].genesis_ghostdag_data()
```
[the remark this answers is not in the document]
coderofstuff: The reason for removing this section is that we no longer need to populate any GD at level other than level 0. We only need to populate it with data needed from the current proof. If genesis is needed (such as in the case of a new network started) the proof would contain it. If not, then there's also no need to have genesis-specific handling.

consensus/src/pipeline/header_processor/processor.rs
```diff
@@ -451,10 +446,7 @@ impl HeaderProcessor {
         // Create a DB batch writer
         let mut batch = WriteBatch::default();
 
-        for (level, datum) in ghostdag_data.iter().enumerate() {
-            // This data might have been already written when applying the pruning proof.
-            self.ghostdag_stores[level].insert_batch(&mut batch, ctx.hash, datum).unwrap_or_exists();
-        }
+        self.ghostdag_primary_store.insert_batch(&mut batch, ctx.hash, ghostdag_data).unwrap_or_exists();
```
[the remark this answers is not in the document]
coderofstuff: Yes, it still applies. I've re-added this comment.

consensus/src/processes/pruning_proof/mod.rs
```diff
@@ -614,45 +712,312 @@ impl PruningProofManager {
...
+                    None
+                } else {
+                    Some((level - 1) as BlockLevel)
+                }
+            })
+            .unwrap_or(self.max_block_level)
+    }
+
+    fn estimated_blue_depth_at_level_0(&self, level: BlockLevel, level_depth: u64, current_dag_level: BlockLevel) -> u64 {
+        level_depth << current_dag_level.saturating_sub(level)
```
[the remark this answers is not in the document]
coderofstuff: Fixed. This is now:

```
    fn estimated_blue_depth_at_level_0(&self, level: BlockLevel, level_depth: u64, current_dag_level: BlockLevel) -> u64 {
        level_depth.checked_shl(level.saturating_sub(current_dag_level) as u32).unwrap_or(level_depth)
    }
```

kaspad/src/daemon.rs
coderofstuff: Updated to use `delete_range`:

```
// Apply delete of range from level 1 to max (+1) for Ghostdag and GhostdagCompact:
writer.delete_range(start_ghostdag_prefix_vec.clone(), end_ghostdag_prefix_vec.clone()).unwrap();
writer.delete_range(start_compact_prefix_vec.clone(), end_compact_prefix_vec.clone()).unwrap();
```

consensus/src/processes/ghostdag/ordering.rs
coderofstuff: Nvm, I had to put it back as I encountered this panic again when the node was pruning:

```
2024-07-06 12:16:01.923-06:00 [ERROR] thread 'pruning-processor' panicked at consensus/src/processes/ghostdag/ordering.rs:55:66: called `Result::unwrap()` on an `Err` value: KeyNotFound(TempGhostdagCompact/0e005a0e5eaac027ea36de5dc265694273ebeca9b012cb314fe6ea8cd19c87715b02)
thread 'pruning-processor' panicked at consensus/src/processes/ghostdag/ordering.rs:55:66:
called `Result::unwrap()` on an `Err` value: KeyNotFound(TempGhostdagCompact/0e005a0e5eaac027ea36de5dc265694273ebeca9b012cb314fe6ea8cd19c87715b02)
stack backtrace:
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
Exiting...
2024-07-06 12:16:01.954-06:00 [ERROR] thread 'virtual-processor' panicked at consensus/src/pipeline/virtual_processor/processor.rs:431:29: unexpected error rocksdb error Corruption: Mismatch in unique ID on table file 264926. Expected: {7800974614205984544,1475903054101850025} Actual: {6918025063187695999,149668763839083335}
thread 'virtual-processor' panicked at consensus/src/pipeline/virtual_processor/processor.rs:431:29:
unexpected error rocksdb error Corruption: Mismatch in unique ID on table file 264926. Expected: {7800974614205984544,1475903054101850025} Actual: {6918025063187695999,149668763839083335}
stack backtrace:
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
Exiting...
```

This happens during pruning, but I don't know what's causing it exactly yet. I think my original comment about getting parents not in the same level is wrong. Perhaps the issue is that this GD call is trying to get data that has just recently been pruned.

coderofstuff: In my original comment, this is where I was thinking that the call to `.ghostdag(parents)` eventually will try to get parents which may not be on the same level, and therefore will not exist in the `self.ghostdag_store` of that level: https://github.com/kaspanet/rusty-kaspa/blob/6a56461ff1249e11e3124090253ca0bf128f7189/consensus/src/processes/ghostdag/mergeset.rs#L20

Doing a bit more debugging, this is the error that I'm getting when trying to fill the ghostdag data of level 15 (during build_pruning_proof) in TN11:

```
2024-07-07 12:17:28.418-06:00 [ERROR] thread 'pruning-processor' panicked at consensus/src/processes/ghostdag/ordering.rs:56:17: Failed looking for this block: 97fc07b5fdd076c9c467a889a538d9f7d735371d0ecff1cd39b8cc34fcaf3ddf
thread 'pruning-processor' panicked at consensus/src/processes/ghostdag/ordering.rs:56:17:
Failed looking for this block: 97fc07b5fdd076c9c467a889a538d9f7d735371d0ecff1cd39b8cc34fcaf3ddf
```

(the log above is custom, but in master this corresponds to the unwrap at line https://github.com/kaspanet/rusty-kaspa/blob/6a56461ff1249e11e3124090253ca0bf128f7189/consensus/src/processes/ghostdag/ordering.rs#L48)

Now, I tried calculating the current DAG level of this specific hash `97fc07b5fdd076c9c467a889a538d9f7d735371d0ecff1cd39b8cc34fcaf3ddf` and I got this in the logs:

```
2024-07-07 12:17:28.404-06:00 [WARN ] 97fc07b5fdd076c9c467a889a538d9f7d735371d0ecff1cd39b8cc34fcaf3ddf => dag level 14
```

So it seems like even though we're processing level 15, we may got parents from different (lower) levels when doing `relations.get_parents()` as in `mergeset.rs` above. Is this expected behavior of `relations.get_parents()`?

coderofstuff: More debugging details:

Given we are processing level 15, and we note down the dag level of some interesting hashes:

```
2024-07-07 14:03:43.612-06:00 [WARN ] Level: 15 | Root: a5dd609448bcc80b8083194ace12baf512d4c1a6cf8bff3393d55968139eeb1b | Tip: a5efd2ce222f6f3e72316532768587923bdb9c96607ef5cc34bed1001ae658d2
2024-07-07 14:03:43.612-06:00 [WARN ] 2bdf6ef9066a4c5b980f6c53e80a122ea977a79157653b090291d83dc620dc50 => dag level 14
2024-07-07 14:03:43.612-06:00 [WARN ] 97fc07b5fdd076c9c467a889a538d9f7d735371d0ecff1cd39b8cc34fcaf3ddf => dag level 14
```

This is the BFS forward iteration from `Root` to `Tip`:
```
2024-07-07 14:03:43.612-06:00 [WARN ] Level: 15 | Hash: a5dd609448bcc80b8083194ace12baf512d4c1a6cf8bff3393d55968139eeb1b
2024-07-07 14:03:43.613-06:00 [WARN ] Level: 15 | Hash: a5dd609448bcc80b8083194ace12baf512d4c1a6cf8bff3393d55968139eeb1b | Child: 394ea60e79c41f5e2cc56638df23917a60d7a8d6d52058066a3def49d2e2c991
2024-07-07 14:03:43.613-06:00 [WARN ] Level: 15 | Hash: a5dd609448bcc80b8083194ace12baf512d4c1a6cf8bff3393d55968139eeb1b | Child: 3be8ced5bca8b9615068bf1e00ab49390598ddd20208870d42a73882dccdcfe9
2024-07-07 14:03:43.613-06:00 [WARN ] Level: 15 | Hash: 394ea60e79c41f5e2cc56638df23917a60d7a8d6d52058066a3def49d2e2c991
2024-07-07 14:03:43.623-06:00 [INFO ] mergeset.rs::unordered_mergeset_without_selected_parent | current: 394ea60e79c41f5e2cc56638df23917a60d7a8d6d52058066a3def49d2e2c991 | relations.get_parents(current): [a5dd609448bcc80b8083194ace12baf512d4c1a6cf8bff3393d55968139eeb1b]

2024-07-07 14:03:43.614-06:00 [WARN ] Level: 15 | Hash: 394ea60e79c41f5e2cc56638df23917a60d7a8d6d52058066a3def49d2e2c991 | Child: f7c78edec9c10204ed8b6827364edd04f835094750f1a123b23a7dede163c580
2024-07-07 14:03:43.614-06:00 [WARN ] Level: 15 | Hash: 394ea60e79c41f5e2cc56638df23917a60d7a8d6d52058066a3def49d2e2c991 | Child: 2bdf6ef9066a4c5b980f6c53e80a122ea977a79157653b090291d83dc620dc50
2024-07-07 14:03:43.614-06:00 [WARN ] Level: 15 | Hash: 394ea60e79c41f5e2cc56638df23917a60d7a8d6d52058066a3def49d2e2c991 | Child: bc162a2a937e30d5cedd84f8e898e6ac73b4a41d165c8695cf58ac876d86b1a9
2024-07-07 14:03:43.614-06:00 [WARN ] Level: 15 | Hash: 3be8ced5bca8b9615068bf1e00ab49390598ddd20208870d42a73882dccdcfe9
2024-07-07 14:03:43.621-06:00 [WARN ] Level: 15 | Hash: 3be8ced5bca8b9615068bf1e00ab49390598ddd20208870d42a73882dccdcfe9 | Child: 8cf27c51e35bd9227cd20a76d41b5bc3a7f504cdb7032dfca93daaf134303dad
2024-07-07 14:03:43.621-06:00 [WARN ] Level: 15 | Hash: 3be8ced5bca8b9615068bf1e00ab49390598ddd20208870d42a73882dccdcfe9 | Child: bc162a2a937e30d5cedd84f8e898e6ac73b4a41d165c8695cf58ac876d86b1a9
2024-07-07 14:03:43.622-06:00 [WARN ] Level: 15 | Hash: 2bdf6ef9066a4c5b980f6c53e80a122ea977a79157653b090291d83dc620dc50
2024-07-07 14:03:43.624-06:00 [INFO ] mergeset.rs::unordered_mergeset_without_selected_parent | current: 2bdf6ef9066a4c5b980f6c53e80a122ea977a79157653b090291d83dc620dc50 | relations.get_parents(current): [97fc07b5fdd076c9c467a889a538d9f7d735371d0ecff1cd39b8cc34fcaf3ddf, 394ea60e79c41f5e2cc56638df23917a60d7a8d6d52058066a3def49d2e2c991]
2024-07-07 14:03:43.626-06:00 [INFO ] mergeset.rs::unordered_mergeset_without_selected_parent | current: 2bdf6ef9066a4c5b980f6c53e80a122ea977a79157653b090291d83dc620dc50 | relations.get_parents(current): [97fc07b5fdd076c9c467a889a538d9f7d735371d0ecff1cd39b8cc34fcaf3ddf, 394ea60e79c41f5e2cc56638df23917a60d7a8d6d52058066a3def49d2e2c991]
2024-07-07 14:03:43.626-06:00 [INFO ] mergeset.rs::unordered_mergeset_without_selected_parent | current: f7c78edec9c10204ed8b6827364edd04f835094750f1a123b23a7dede163c580 | relations.get_parents(current): [97fc07b5fdd076c9c467a889a538d9f7d735371d0ecff1cd39b8cc34fcaf3ddf, 394ea60e79c41f5e2cc56638df23917a60d7a8d6d52058066a3def49d2e2c991]
2024-07-07 14:03:43.626-06:00 [INFO ] mergeset.rs::unordered_mergeset_without_selected_parent | current: 97fc07b5fdd076c9c467a889a538d9f7d735371d0ecff1cd39b8cc34fcaf3ddf | relations.get_parents(current): [927dd3dd32a9ceec0ee6b1561771185a14b003e1d97c50f2dc938b4405480056]
2024-07-07 14:03:43.627-06:00 [INFO ] mergeset.rs::unordered_mergeset_without_selected_parent | current: 927dd3dd32a9ceec0ee6b1561771185a14b003e1d97c50f2dc938b4405480056 | relations.get_parents(current): [40b48d2a3af40b10345c921b5e987132275337799bd4d51e6a188f7d6f25815c, 26fb1277ec995718517ac9bdc26d5f39fdbe36c0b7b75de424642144ecbc3f1f, 4c64959e29e632759a7a2b350d00c5086e70df786500295c94f10b6cea506593, 6f1a761ad755d81f76c43f023f6bad7c7775891232098b91e514b4200a26bc98]

2024-07-07 14:03:43.622-06:00 [WARN ] Level: 15 | Hash: 2bdf6ef9066a4c5b980f6c53e80a122ea977a79157653b090291d83dc620dc50 | Child: 1f11c4acbce3fd01bf786f04232ee4a9ed3b86b64b4576b58d9008951a1dd09c
2024-07-07 14:03:43.622-06:00 [WARN ] Level: 15 | Hash: 2bdf6ef9066a4c5b980f6c53e80a122ea977a79157653b090291d83dc620dc50 | Child: 0f857297b42847d04dd416af26917979653ef7203d071a28e6f61e7820f777d1
2024-07-07 14:03:43.622-06:00 [WARN ] Level: 15 | Hash: f7c78edec9c10204ed8b6827364edd04f835094750f1a123b23a7dede163c580
2024-07-07 14:03:43.623-06:00 [WARN ] Level: 15 | Hash: f7c78edec9c10204ed8b6827364edd04f835094750f1a123b23a7dede163c580 | Child: 1f11c4acbce3fd01bf786f04232ee4a9ed3b86b64b4576b58d9008951a1dd09c
2024-07-07 14:03:43.623-06:00 [WARN ] Level: 15 | Hash: f7c78edec9c10204ed8b6827364edd04f835094750f1a123b23a7dede163c580 | Child: 0f857297b42847d04dd416af26917979653ef7203d071a28e6f61e7820f777d1
2024-07-07 14:03:43.623-06:00 [WARN ] Level: 15 | Hash: bc162a2a937e30d5cedd84f8e898e6ac73b4a41d165c8695cf58ac876d86b1a9

2024-07-07 14:03:43.623-06:00 [WARN ] Level: 15 | Hash: bc162a2a937e30d5cedd84f8e898e6ac73b4a41d165c8695cf58ac876d86b1a9 | Child: 5e3eaf86fbdb217956c06e6339114599b89d88c3c376a584ce2646f0be29631f
2024-07-07 14:03:43.624-06:00 [WARN ] Level: 15 | Hash: 5e3eaf86fbdb217956c06e6339114599b89d88c3c376a584ce2646f0be29631f
2024-07-07 14:03:43.624-06:00 [WARN ] Level: 15 | Hash: 5e3eaf86fbdb217956c06e6339114599b89d88c3c376a584ce2646f0be29631f | Child: 0f857297b42847d04dd416af26917979653ef7203d071a28e6f61e7820f777d1
2024-07-07 14:03:43.624-06:00 [WARN ] Level: 15 | Hash: 1f11c4acbce3fd01bf786f04232ee4a9ed3b86b64b4576b58d9008951a1dd09c
2024-07-07 14:03:43.625-06:00 [WARN ] Level: 15 | Hash: 1f11c4acbce3fd01bf786f04232ee4a9ed3b86b64b4576b58d9008951a1dd09c | Child: 8cf27c51e35bd9227cd20a76d41b5bc3a7f504cdb7032dfca93daaf134303dad
2024-07-07 14:03:43.625-06:00 [WARN ] Level: 15 | Hash: 1f11c4acbce3fd01bf786f04232ee4a9ed3b86b64b4576b58d9008951a1dd09c | Child: 0cca5923823350455382dd5c451466676c2898d6ef501f23b61190a2548ca97a
2024-07-07 14:03:43.625-06:00 [WARN ] Level: 15 | Hash: 1f11c4acbce3fd01bf786f04232ee4a9ed3b86b64b4576b58d9008951a1dd09c | Child: 5e41a55408e4bfddbbed1cba56850d9b2655590c459814439782844e58411065
2024-07-07 14:03:43.625-06:00 [WARN ] Level: 15 | Hash: 8cf27c51e35bd9227cd20a76d41b5bc3a7f504cdb7032dfca93daaf134303dad
2024-07-07 14:03:43.625-06:00 [INFO ] mergeset.rs::unordered_mergeset_without_selected_parent | current: 3be8ced5bca8b9615068bf1e00ab49390598ddd20208870d42a73882dccdcfe9 | relations.get_parents(current): [a5dd609448bcc80b8083194ace12baf512d4c1a6cf8bff3393d55968139eeb1b]
2024-07-07 14:03:43.626-06:00 [WARN ] Level: 15 | Hash: 8cf27c51e35bd9227cd20a76d41b5bc3a7f504cdb7032dfca93daaf134303dad | Child: dcd74d6bd31f4fecbcda823d2f1ffa43c8d913dccea4d7e55a820f898ef51327
2024-07-07 14:03:43.626-06:00 [WARN ] Level: 15 | Hash: 0f857297b42847d04dd416af26917979653ef7203d071a28e6f61e7820f777d1

2024-07-07 14:03:43.627-06:00 [ERROR] thread 'pruning-processor' panicked at consensus/src/processes/ghostdag/ordering.rs:63:17: Failed to find block: 97fc07b5fdd076c9c467a889a538d9f7d735371d0ecff1cd39b8cc34fcaf3ddf
thread 'pruning-processor' panicked at consensus/src/processes/ghostdag/ordering.rs:63:17:
Failed to find block: 97fc07b5fdd076c9c467a889a538d9f7d735371d0ecff1cd39b8cc34fcaf3ddf
```
---

Notably, this happens because `relations.get_parents('2bdf6ef9066a4c5b980f6c53e80a122ea977a79157653b090291d83dc620dc50')` here returned a parent from level 14:

```
2024-07-07 14:03:43.624-06:00 [INFO ] mergeset.rs::unordered_mergeset_without_selected_parent | current: 2bdf6ef9066a4c5b980f6c53e80a122ea977a79157653b090291d83dc620dc50 | relations.get_parents(current): [97fc07b5fdd076c9c467a889a538d9f7d735371d0ecff1cd39b8cc34fcaf3ddf, 394ea60e79c41f5e2cc56638df23917a60d7a8d6d52058066a3def49d2e2c991]
```

So when `self.ghostdag_store.get_blue_work(*parent).unwrap()` is called where `parent = '97fc07b5fdd076c9c467a889a538d9f7d735371d0ecff1cd39b8cc34fcaf3ddf'` this panics because `97fc07b5fdd076c9c467a889a538d9f7d735371d0ecff1cd39b8cc34fcaf3ddf` is not in the current level's GD store as it's a hash from level 14.

coderofstuff: For now, I've fixed the issue as follows:
- unwrap_or_else -> default to 0 if the hash couldn't be found. I have a comment in the code explaining why this can be fine
- in `ghostdag`, skip hashes in the mergeset that are NOT in the GD store

Can discuss further when this PR is reviewed in detail

consensus/src/processes/pruning_proof/mod.rs
```diff
@@ -593,14 +710,25 @@ impl PruningProofManager {
...
+                continue;
+            }
+
             match relations_read[level_idx].get_parents(current_pp).unwrap_option() {
                 Some(parents) => {
                     if parents
                         .iter()
                         .copied()
-                        .any(|parent| self.ghostdag_stores[level_idx].get_blue_score(parent).unwrap() < 2 * self.pruning_proof_m)
+                        .any(|parent| proof_ghostdag_stores[level_idx].get_blue_score(parent).unwrap() < 2 * self.pruning_proof_m)
```
michaelsutton, line 731: This should be `current_consensus_ghostdag_stores`

consensus/src/processes/pruning_proof/mod.rs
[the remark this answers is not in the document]
someone235: Why `.unwrap_or(level_depth)`? Shouldn't it be `u64::MAX` or something like this?

consensus/src/processes/pruning_proof/mod.rs
someone235: I mean it should be part of `self`

consensus/src/processes/pruning_proof/mod.rs
someone235: So maybe add a comment that explains it?

consensus/src/processes/pruning_proof/mod.rs
```diff
@@ -614,45 +740,263 @@ impl PruningProofManager {
...
+        let parents = self
+            .parents_manager
+            .parents_at_level(header, level)
+            .iter()
+            .copied()
+            .filter(|parent| relations_service.has(*parent).unwrap())
+            .collect_vec()
+            .push_if_empty(ORIGIN);
+
+        let mut sp = SortableBlock { hash: parents[0], blue_work: self.headers_store.get_blue_score(parents[0]).unwrap_or(0).into() };
```
someone235, line 795: Why are we using here get_blue_score and not get_blue_work?

consensus/src/processes/pruning_proof/mod.rs
```diff
@@ -614,45 +740,263 @@ impl PruningProofManager {
...
+        for parent in parents.iter().copied().skip(1) {
+            let sblock = SortableBlock {
+                hash: parent,
+                blue_work: self
+                    .headers_store
+                    .get_blue_score(parent)
```
someone235, line 801: Here too

kaspad/src/daemon.rs
```diff
@@ -308,13 +312,106 @@ do you confirm? (answer y/n or pass --yes to the Kaspad command line to confirm
...
+                        }
+
+                        compact_record_count += 1;
+                    }
+
+                    trace!("Number of Ghostdag records to cleanup: {}", gd_record_count);
+                    trace!("Number of GhostdagCompact records to cleanup: {}", compact_record_count);
+                    info!("Number of deprecated records to cleanup: {}", gd_record_count + compact_record_count);
+
+                    let msg =
```
someone235, line 369: Why don't we ask before going over the db?

someone235 (changes requested):

kaspad/src/daemon.rs
coderofstuff: The code above this question is only there to show to the user that there are indeed some number of records to cleanup. We don't really need to go over the DB since the efficient way to delete it is done below (with ranges). We do so only because we want to show some stats to the user to inform their decision.

We can:
- move this question up, before the stats are shown. The UX is then that we ask the question first, then show the stats to the user, then (without asking again) proceed with the efficient delete
- remove the stats section altogether - the user is not informed of the number of records to be deleted, maybe it doesn't matter to them anyway and they would always have to upgrade if they want to use this version of the node and DB
- keep it as it is

consensus/src/processes/pruning_proof/mod.rs
```diff
@@ -632,37 +628,71 @@ impl PruningProofManager {
...
+            let proof_selected_tip_gd = proof_ghostdag_stores[level_idx].get_compact_data(selected_tip).unwrap();
+
+            // Next check is to see if this proof is "better" than what's in the current consensus
+            // Step 1 - look at only levels that have a full proof (least 2m blocks in the proof)
             if proof_selected_tip_gd.blue_score < 2 * self.pruning_proof_m {
                 continue;
             }
 
+            // Step 2 - if we can find a common ancestor between the proof and current consensus
+            // we can determine if the proof is better. The proof is better if the score difference between the
```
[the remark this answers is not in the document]
coderofstuff: I'll replace "score" with "blue work" here. The comment using "score" is a remnant of when this looked a little different but blue work is the correct wording here.

consensus/src/processes/pruning_proof/mod.rs
```diff
@@ -1036,6 +1015,8 @@ impl PruningProofManager {
                 } else {
                     block_at_depth_2m
                 };
+
+                // new root is expected to be always an ancestor of old root because new root takes a safety margin
```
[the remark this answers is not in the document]
coderofstuff: This block of code with `old_root` is kept as a form of sanity check. Perhaps `sanity_check_root` or `depth_based_root` is appropriate?

consensus/src/processes/pruning_proof/mod.rs
coderofstuff: Revisiting this idea maybe we should panic `finished_headers` is true and `has_required_block` is false. In the above example where we sync with an older node without margins `has_required_block` must still be true anyway.

consensus/src/processes/pruning_proof/mod.rs
```diff
@@ -632,37 +628,71 @@ impl PruningProofManager {
...
+        // This is expected to be fast because if a proof exists, it will be cached.
+        // If no proof exists, this is empty
+        let mut current_consensus_proof = self.get_pruning_point_proof();
+        if current_consensus_proof.is_empty() {
+            // An empty proof can only happen if we're at genesis. We're going to create a proof for this case that contains the genesis header only
+            let genesis_header = self.headers_store.get_header(self.genesis_hash).unwrap();
+            current_consensus_proof = Arc::new((0..=self.max_block_level).map(|_| vec![genesis_header.clone()]).collect_vec());
+        }
+        let mut current_consensus_stores_and_processes =
+            self.init_validate_pruning_point_proof_stores_and_processes(headers_estimate)?;
```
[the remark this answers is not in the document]
coderofstuff: Fixed in [fb3d1e9](https://github.com/kaspanet/rusty-kaspa/pull/494/commits/fb3d1e9b88f670555aaa89059b5dd86b3ac37235)

consensus/src/processes/pruning_proof/mod.rs
```diff
@@ -414,14 +412,8 @@ impl PruningProofManager {
 
     fn init_validate_pruning_point_proof_stores_and_processes(
         &self,
-        proof: &PruningPointProof,
+        headers_estimate: usize,
```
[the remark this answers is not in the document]
coderofstuff: Fixed in [fb3d1e9](https://github.com/kaspanet/rusty-kaspa/pull/494/commits/fb3d1e9b88f670555aaa89059b5dd86b3ac37235)

consensus/src/processes/pruning_proof/mod.rs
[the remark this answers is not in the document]
coderofstuff: I chose `depth_based_root` and added an explanation for why we're still calculating it.

consensus/src/processes/ghostdag/protocol.rs
```diff
@@ -106,6 +106,13 @@ impl<T: GhostdagStoreReader, S: RelationsStoreReader, U: ReachabilityService, V:
         let ordered_mergeset = self.ordered_mergeset_without_selected_parent(selected_parent, parents);
 
         for blue_candidate in ordered_mergeset.iter().cloned() {
+            if !self.ghostdag_store.has(blue_candidate).unwrap() {
+                // TODO: See comment in ordering.rs which gets called above at ordered_mergeset_without_selected_parent
+                // For now, ignore blocks hashes that aren't in this GD store.
+                // NOT expected to be relevant for calculations in GD level 0
+                continue;
+            }
+
```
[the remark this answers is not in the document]
coderofstuff: Your PR at [42a3030](https://github.com/kaspanet/rusty-kaspa/pull/494/commits/42a303026f33beb81e1ff3a280277234957d762a) fixes this

consensus/src/processes/ghostdag/ordering.rs
coderofstuff: Your PR at [42a3030](https://github.com/kaspanet/rusty-kaspa/pull/494/commits/42a303026f33beb81e1ff3a280277234957d762a) fixes this

coderofstuff: Just some logs to prove to myself the above fix was legit:

```
2024-08-27 13:07:00.167-06:00 [WARN ] Processing level: 16 | root: 5f5d89c3a0a89419f364e191986b6e3da2a769d75549401c00c00d471cddc6b6 | tip: 7fb7150cb83fe2135b58b71c4e8d536537d011739d72410cc3e3d73f498c6f54
2024-08-27 13:07:00.167-06:00 [WARN ] Is root ancestor of 7afb0a822d3939fca6ab4631a2d42f7a23da18506f9822dc000ae346fe5ef0ed: false
2024-08-27 13:07:00.168-06:00 [WARN ] Current: a4e1005961cd452069c16afb5849bb249ef4c832a72e9a6004a48d6c7196c317 | Parent: dbb318ffac3cda3b93aa5ecf61fd79fa8fe49d43a0979d152670c6147bb72f3c
2024-08-27 13:07:00.169-06:00 [WARN ] Current: dbb318ffac3cda3b93aa5ecf61fd79fa8fe49d43a0979d152670c6147bb72f3c | Parent: ee97a43b839817962a0a0a55f19e68fe3cd2a018ce3308ff3159263b53119c01
...
2024-08-27 13:07:00.169-06:00 [WARN ] Current: 474098b86913ce5dd1d019dd469c409ed720e67b8a874fed82cd536d18d79f0f | Parent: 7afb0a822d3939fca6ab4631a2d42f7a23da18506f9822dc000ae346fe5ef0ed
2024-08-27 13:07:00.169-06:00 [WARN ] Current: fb46ea0f151d8420ca87526d3a527543537ef3dcf2a309f2343d60d8e88defb4 | Parent: 850307ff2ced095744f656f20e474687de4d477c61f443622c481e9eda0bf3d4
2024-08-27 13:07:00.169-06:00 [WARN ] Current: fb46ea0f151d8420ca87526d3a527543537ef3dcf2a309f2343d60d8e88defb4 | Parent: a2ea43ecffca7d958f9d141190abc5cbdbc8459dd7898d1f60c03b72c0875064
...
2024-08-27 13:07:00.169-06:00 [WARN ] Current: 40de62455e96431302ca610871cb5344ec74ac44625bec240c2cbd3e3b70c1b4 | Parent: f30105d6ae65883fe72bb3f7da12fe02ef08d4581dddc4eb6beee239319916fd
2024-08-27 13:07:00.170-06:00 [ERROR] thread 'tokio-runtime-worker' panicked at consensus/src/processes/ghostdag/ordering.rs:68:66: called `Result::unwrap()` on an `Err` value: KeyNotFound(TempGhostdagCompact/16/007afb0a822d3939fca6ab4631a2d42f7a23da18506f9822dc000ae346fe5ef0ed)
thread 'tokio-runtime-worker' panicked at consensus/src/processes/ghostdag/ordering.rs:68:66:
called `Result::unwrap()` on an `Err` value: KeyNotFound(TempGhostdagCompact/16/007afb0a822d3939fca6ab4631a2d42f7a23da18506f9822dc000ae346fe5ef0ed)
Exiting...
```

consensus/src/processes/pruning_proof/mod.rs
```diff
@@ -1028,8 +1040,8 @@ impl PruningProofManager {
                     block_at_depth_2m
                 };
 
-                // new root is expected to be always an ancestor of old root because new root takes a safety margin
-                assert!(self.reachability_service.is_dag_ancestor_of(root, old_root));
+                // new root is expected to be always an ancestor of depth_based_root because new root takes a safety margin
+                assert!(self.reachability_service.is_dag_ancestor_of(root, depth_based_root));
```
someone235, line 1072: I don't think this assertion is guaranteed to be true. What if we get to [this](https://github.com/kaspanet/rusty-kaspa/pull/494/files/3e249f91cd3f5fadc1a1ee647a352cc5fa7e3d31..46dbac30a78fa48b503621d8970e019e9a45aa31#diff-593674986e16100c0c323e07c342dc26ddf962b1efce504ce245975d817f5922R952) branch?

someone235 (changes requested):

consensus/src/processes/pruning_proof/mod.rs
```diff
@@ -714,40 +721,280 @@ impl PruningProofManager {
...
+            };
+            if sblock > sp {
+                sp = sblock;
+            }
+        }
+        // TODO: For higher levels the chance of having more than two parents is very small, so it might make sense to fetch the whole header for the SortableBlock instead of blue_score (which will probably come from a compact header).
+        self.headers_store.get_header(sp.hash).unwrap_option().ok_or(PruningProofManagerInternalError::NotEnoughHeadersToBuildProof(
+            format!("find_selected_parent_header_at_level (level {level}) couldn't find the header for block {}", sp.hash,),
+        ))
+        // Ok(self.headers_store.get_header(sp.hash).unwrap_option().expect("already checked if compact header exists above"))
```
someone235, line 865: Remove comment

someone235 (changes requested):

consensus/src/processes/pruning_proof/mod.rs
```diff
@@ -225,18 +225,30 @@ impl PruningProofManager {
         let pruning_point_header = proof[0].last().unwrap().clone();
         let pruning_point = pruning_point_header.hash;
 
-        let proof_zero_set = BlockHashSet::from_iter(proof[0].iter().map(|header| header.hash));
+        // Create a copy of the proof, since we're going to be mutating the proof passed to us
+        let proof_sets: Vec<HashSet<Hash, kaspa_consensus_core::BlockHasher>> = (0..=self.max_block_level)
```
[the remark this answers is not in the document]
coderofstuff: Switching to `.collect_vec` and no explicit type is what I'll do

consensus/src/processes/pruning_proof/mod.rs
coderofstuff: Fixed in [56b439](https://github.com/kaspanet/rusty-kaspa/pull/494/commits/56b4392178bb5a751752cafc461c3b9aa9155250)

consensus/src/processes/pruning_proof/mod.rs
coderofstuff: Fixed in [66d4ebe](https://github.com/kaspanet/rusty-kaspa/pull/494/commits/66d4ebe674f798ec6a2ecc7c16f5e92f9bab8e8f)

consensus/src/processes/pruning_proof/mod.rs
coderofstuff: Fixed in [66d4ebe](https://github.com/kaspanet/rusty-kaspa/pull/494/commits/66d4ebe674f798ec6a2ecc7c16f5e92f9bab8e8f)

consensus/src/processes/pruning_proof/mod.rs
coderofstuff: Fixed in [18158e6](https://github.com/kaspanet/rusty-kaspa/pull/494/commits/18158e62794b8c6943fc234f58462268ce8af935)

consensus/src/processes/pruning_proof/mod.rs
```diff
@@ -691,8 +691,8 @@ impl PruningProofManager {
             }
 
             // Step 2 - if we can find a common ancestor between the proof and current consensus
-            // we can determine if the proof is better. The proof is better if the score difference between the
-            // old current consensus's tips and the common ancestor is less than the score difference between the
+            // we can determine if the proof is better. The proof is better if the blue work difference between the
+            // old current consensus's tips and the common ancestor is less than the blue work difference between the
```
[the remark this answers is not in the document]
coderofstuff: Fixed in [18158e6](https://github.com/kaspanet/rusty-kaspa/pull/494/commits/18158e62794b8c6943fc234f58462268ce8af935)

consensus/src/processes/pruning_proof/mod.rs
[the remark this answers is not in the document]
coderofstuff: Fixed in [18158e6](https://github.com/kaspanet/rusty-kaspa/pull/494/commits/18158e62794b8c6943fc234f58462268ce8af935)

consensus/src/processes/pruning_proof/mod.rs
coderofstuff: If it gets to that branch, then likely that calculated root == depth_based_root. The GD store `block_at_depth` uses is one that was constructed through `find_sufficient_root`.

The GD store is constructed by BFS forward traversing from calculated root to selected tip in `find_sufficient_root`, ignoring blocks that are not in the past of selected tip. So all blocks in this GD store must be in the future of root. More importantly, all selected parent chain from any block (other than root) in this store must lead back to root.

`depth_based_root` is calculated by traversing the selected parent chain from selected tip back to root. The furthest back that this backward chain traversal can go is root.

So at all times, root must be an ancestor of or equal to depth_based_root.

kaspad/src/daemon.rs
michaelsutton: I'm ok with it as is. It gives a solid picture to the user and it isn't really expensive

consensus/src/processes/pruning_proof/mod.rs
someone235: Ok, so IIUC this branch can only cause block_at_depth to fail, right?

michaelsutton (commented):
Initial pass -- minor details out of core changes

consensus/src/pipeline/header_processor/processor.rs
```diff
@@ -346,21 +346,17 @@ impl HeaderProcessor {
...
+            .get_data(ctx.hash)
+            .unwrap_option()
+            .unwrap_or_else(|| Arc::new(self.ghostdag_manager.ghostdag(&ctx.known_parents[0])));
+        self.counters.mergeset_counts.fetch_add(ghostdag_data.mergeset_size() as u64, Ordering::Relaxed);
         ctx.ghostdag_data = Some(ghostdag_data);
     }
 
     fn commit_header(&self, ctx: HeaderProcessingContext, header: &Header) {
-        let ghostdag_data = ctx.ghostdag_data.as_ref().unwrap();
+        let ghostdag_primary_data = ctx.ghostdag_data.as_ref().unwrap();
```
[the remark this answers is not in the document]
coderofstuff: Mostly due to this comment: https://github.com/kaspanet/rusty-kaspa/pull/494#discussion_r1649602486

But also everywhere except the pruning proof area, we only have a single GD store, manager and data we're talking about - the one at level 0

consensus/src/pipeline/header_processor/processor.rs
```diff
@@ -370,9 +366,9 @@ impl HeaderProcessor {
         // Append-only stores: these require no lock and hence done first in order to reduce locking time
         //
 
-        for (level, datum) in ghostdag_data.iter().enumerate() {
-            self.ghostdag_stores[level].insert_batch(&mut batch, ctx.hash, datum).unwrap();
-        }
+        // This data might have been already written when applying the pruning proof.
```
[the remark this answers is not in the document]
coderofstuff: Removed in [2e8deab](https://github.com/kaspanet/rusty-kaspa/pull/494/commits/2e8deabd2526cee8c0535b35a61aebaa25545d74)

consensus/src/pipeline/header_processor/processor.rs
```diff
@@ -346,21 +346,17 @@ impl HeaderProcessor {
 
     /// Runs the GHOSTDAG algorithm for all block levels and writes the data into the context (if hasn't run already)
```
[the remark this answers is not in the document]
coderofstuff: Removed in [2e8deab](https://github.com/kaspanet/rusty-kaspa/pull/494/commits/2e8deabd2526cee8c0535b35a61aebaa25545d74)

consensus/src/pipeline/header_processor/processor.rs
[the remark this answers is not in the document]
michaelsutton: I know, that's my point.. here you actually added the primary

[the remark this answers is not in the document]
coderofstuff: Fixed in [4168456](https://github.com/kaspanet/rusty-kaspa/pull/494/commits/41684567ec17c4c0f4faf7cbe7ef9010b3e75135)

consensus/src/processes/pruning_proof/mod.rs
```diff
@@ -614,45 +791,267 @@ impl PruningProofManager {
...
+            })
+            .unwrap_or(self.max_block_level)
+    }
+
+    fn estimated_blue_depth_at_level_0(&self, level: BlockLevel, level_depth: u64, current_dag_level: BlockLevel) -> u64 {
+        level_depth.checked_shl(level.saturating_sub(current_dag_level) as u32).unwrap_or(level_depth)
+    }
+
+    /// selected parent at level = the parent of the header at the level
+    /// with the highest blue_work (using score as work in this case)
```
michaelsutton, line 828: ~~(using score as work in this case)~~ 

we're actually using the header blue work here

consensus/src/processes/pruning_proof/mod.rs
```diff
@@ -614,45 +791,267 @@ impl PruningProofManager {
...
+            };
+            if sblock > sp {
+                sp = sblock;
+            }
+        }
+        // TODO: For higher levels the chance of having more than two parents is very small, so it might make sense to fetch the whole header for the SortableBlock instead of blue_score (which will probably come from a compact header).
+        self.headers_store.get_header(sp.hash).unwrap_option().ok_or(PruningProofManagerInternalError::NotEnoughHeadersToBuildProof(
+            format!("find_selected_parent_header_at_level (level {level}) couldn't find the header for block {}", sp.hash,),
+        ))
+    }
```
michaelsutton, line 869: I believe the following captures the same logic precisely 

```suggestion
    fn find_selected_parent_header_at_level(
        &self,
        header: &Header,
        level: BlockLevel,
    ) -> PruningProofManagerInternalResult<Arc<Header>> {
        // Parents manager parents_at_level may return parents that aren't in relations_service, so it's important
        // to filter to include only parents that are in relations_service.
        let sp = self
            .parents_manager
            .parents_at_level(header, level)
            .iter()
            .copied()
            .filter(|p| self.level_relations_services[level as usize].has(*p).unwrap())
            .filter_map(|p| self.headers_store.get_header(p).unwrap_option().map(|h| SortableBlock::new(p, h.blue_work)))
            .max()
            .ok_or(PruningProofManagerInternalError::NotEnoughHeadersToBuildProof("no parents with header".to_string()))?;

        Ok(self.headers_store.get_header(sp.hash).expect("unwrapped above"))
    }
```

michaelsutton: it's a bit more forgiving in the sense that it allows a parent with no header if others have a header, but I see no reason for this code to no allow that (and it's an unexpected scenario anyway: only origin has relations w/o header and origin would be a single parent by def)

consensus/src/processes/pruning_proof/mod.rs
```diff
@@ -723,6 +1093,87 @@ impl PruningProofManager {
             .collect_vec()
     }
 
+    /// BFS forward iterates from genesis_hash until selected tip, ignoring blocks in the antipast of selected_tip.
+    /// For each block along the way, insert that hash into the ghostdag_store
+    /// If we have a required_block to find, this will return true if that block was found along the way
+    fn fill_level_proof_ghostdag_data(
+        &self,
+        genesis_hash: Hash,
```
michaelsutton, line 1101: The name `genesis_hash` is a bit confusing, bcs we usually only use it for the real genesis. Should this be simply `root` ?

consensus/src/processes/pruning_proof/mod.rs
michaelsutton: the text `(using score as work in this case)` still needs to be removed

consensus/src/processes/pruning_proof/mod.rs
```diff
@@ -614,45 +791,233 @@ impl PruningProofManager {
...
+        temp_db: Arc<DB>,
+    ) -> PruningProofManagerInternalResult<(Arc<DbGhostdagStore>, Hash, Hash)> {
+        // Step 1: Determine which selected tip to use
+        let selected_tip = if pp_header.block_level >= level {
+            pp_header.header.hash
+        } else {
+            self.find_selected_parent_header_at_level(&pp_header.header, level)?.hash
+        };
+
+        let cache_policy = CachePolicy::Count(2 * self.pruning_proof_m as usize); // TODO: We can probably reduce cache size
```
michaelsutton, line 871: I think this TODO can be removed. 2M is what most levels need, and it's not a large number in the first place even if not used by some higher levels

consensus/src/processes/pruning_proof/mod.rs
```diff
@@ -723,6 +1093,87 @@ impl PruningProofManager {
...
+        ghostdag_store.insert(genesis_hash, Arc::new(gd_manager.genesis_ghostdag_data())).unwrap();
+        ghostdag_store.insert(ORIGIN, gd_manager.origin_ghostdag_data()).unwrap();
+
+        let mut topological_heap: BinaryHeap<_> = Default::default();
+        let mut visited = BlockHashSet::new();
+        for child in relations_service.get_children(genesis_hash).unwrap().read().iter().copied() {
+            topological_heap.push(Reverse(SortableBlock {
+                hash: child,
+                // It's important to use here blue work and not score so we can iterate the heap in a way that respects the topology
+                blue_work: self.headers_store.get_header(child).unwrap().blue_work, // TODO: Maybe add to compact data?
```
michaelsutton, line 1131: I think this TODO can be removed. Adding blue work to compact data would be a DB version change and there's no need to specify this TODO here

consensus/src/processes/pruning_proof/mod.rs
```diff
@@ -723,6 +1093,87 @@ impl PruningProofManager {
...
+                .collect();
+            let current_gd = gd_manager.ghostdag(&relevant_parents);
+
+            ghostdag_store.insert(current_hash, Arc::new(current_gd)).unwrap_or_exists();
+
+            for child in relations_service.get_children(current_hash).unwrap().read().iter().copied() {
+                topological_heap.push(Reverse(SortableBlock {
+                    hash: child,
+                    // It's important to use here blue work and not score so we can iterate the heap in a way that respects the topology
+                    blue_work: self.headers_store.get_header(child).unwrap().blue_work, // TODO: Maybe add to compact data?
```
michaelsutton, line 1169: Same. TODO can be removed.

consensus/src/processes/pruning_proof/mod.rs
```diff
@@ -88,15 +100,59 @@ impl<T> Clone for CachedPruningPointData<T> {
     }
 }
 
+struct TempProofContext {
+    headers_store: Arc<DbHeadersStore>,
+    ghostdag_stores: Vec<Arc<DbGhostdagStore>>,
+    relations_stores: Vec<DbRelationsStore>,
+    reachability_stores: Vec<Arc<parking_lot::lock_api::RwLock<parking_lot::RawRwLock, DbReachabilityStore>>>,
```
michaelsutton, line 107:
```suggestion
    reachability_stores: Vec<Arc<RwLock<DbReachabilityStore>>>,
```

consensus/src/processes/pruning_proof/mod.rs
```diff
@@ -88,15 +100,59 @@ impl<T> Clone for CachedPruningPointData<T> {
...
+    fn has(&self, hash: Hash) -> Result<bool, StoreError> {
+        if self.reachability_service.is_dag_ancestor_of(self.root, hash) {
+            Ok(false)
+        } else {
+            self.relations_store.has(hash)
+        }
+    }
+
+    fn counts(&self) -> Result<(usize, usize), kaspa_database::prelude::StoreError> {
+        panic!("unimplemented")
```
michaelsutton, line 142:
```suggestion
        unimplemented!()
```

consensus/src/processes/pruning_proof/mod.rs
```diff
@@ -258,18 +331,18 @@ impl PruningProofManager {
                             mergeset_reds: calculated_gd.mergeset_reds.clone(),
                             blues_anticone_sizes: calculated_gd.blues_anticone_sizes.clone(),
```
michaelsutton, line 332:
```suggestion
                            mergeset_reds: calculated_gd.mergeset_reds,
                            blues_anticone_sizes: calculated_gd.blues_anticone_sizes,
```

same goes for the line above `mergeset_blues: calculated_gd.mergeset_blues,`

consensus/src/processes/pruning_proof/mod.rs
```diff
@@ -614,45 +791,233 @@ impl PruningProofManager {
...
+                    // with such an older node. As long as we found the required block, we can still proceed.
+                    warn!("Failed to find sufficient root for level {level} after {tries} tries. Headers below the current depth of {required_base_level_depth} are already pruned. Required block found so trying anyway.");
+                    break Ok((ghostdag_store, selected_tip, root));
+                } else {
+                    panic!("Failed to find sufficient root for level {level} after {tries} tries. Headers below the current depth of {required_base_level_depth} are already pruned");
+                }
+            }
+
+            // If we don't have enough depth now, we need to look deeper
+            required_base_level_depth <<= 1;
```
michaelsutton, line 954: Maybe we should do a more subtle increase? 
```suggestion
            required_base_level_depth = (required_base_level_depth as f64 * 1.1) as u64;
```

It's still gonna be a logarithmic number of tries, although with a smaller base. Another option is to increase the multiplier (1.1) at each iteration by 0.05 and bound it by 2.0

consensus/src/processes/pruning_proof/mod.rs
```diff
@@ -723,6 +1093,87 @@ impl PruningProofManager {
...
+            if !has_required_block && required_block.is_some_and(|required_block| current_hash == required_block) {
+                has_required_block = true;
+            }
+
+            let relevant_parents: Box<[Hash]> = relations_service
+                .get_parents(current_hash)
+                .unwrap()
+                .iter()
+                .copied()
+                .filter(|parent| self.reachability_service.is_dag_ancestor_of(genesis_hash, *parent))
```
michaelsutton, line 1159: this filtering seems redundant now that `relations_service` is actually `RelationsStoreInFutureOfRoot`

consensus/src/processes/pruning_proof/mod.rs
```diff
@@ -672,30 +1037,35 @@ impl PruningProofManager {
...
+                queue.push(Reverse(SortableBlock::new(root, self.headers_store.get_header(root).unwrap().blue_work)));
                 while let Some(current) = queue.pop() {
                     let current = current.0.hash;
                     if !visited.insert(current) {
                         continue;
                     }
 
-                    if !self.reachability_service.is_dag_ancestor_of(current, selected_tip) {
+                    if !self.reachability_service.is_dag_ancestor_of(current, selected_tip)
+                        || !ghostdag_stores[level].has(current).is_ok_and(|found| found)
```
michaelsutton, line 1054: we are iterating from `root` up in the exact same way `fill_level_proof_ghostdag_data` iterates up when filing `ghostdag_stores[level]`. So it seems this `has` check will always return true

consensus/src/processes/pruning_proof/mod.rs
```diff
@@ -88,15 +100,59 @@ impl<T> Clone for CachedPruningPointData<T> {
...
+}
+
+#[derive(Clone)]
+struct RelationsStoreInFutureOfRoot<T: RelationsStoreReader, U: ReachabilityService> {
+    relations_store: T,
+    reachability_service: U,
+    root: Hash,
+}
+
+impl<T: RelationsStoreReader, U: ReachabilityService> RelationsStoreReader for RelationsStoreInFutureOfRoot<T, U> {
```
michaelsutton, line 120: A more fancy approach would (move out of this file and) do a composition where filtering  can be applied in layers and then using extensions you would do something like 

```
filtered_relations = base_relations
    .intersect_with_future_of(root, reachability_service)
    .intersect_with_past_of(selected_tip, reachability_service) // some places in the code below do this step as well
``` 

Can be done in a style like `UtxoView` instances can be composed. Probably better if done in a separate PR

consensus/src/processes/pruning_proof/mod.rs
coderofstuff: I think it's important to retain this check even though it seems redundant since it is the only thing that ties the work done in `fill_level_proof_ghostdag_data` to the actual set of headers that goes into the proof (which is constructed here). It serves as an assertion that it must be the same headers that we arrive at.

consensus/src/processes/pruning_proof/mod.rs
coderofstuff: It is. Will change it to just `root`

consensus/src/processes/pruning_proof/mod.rs
coderofstuff: Fixed in [5cd27f3](https://github.com/kaspanet/rusty-kaspa/pull/494/commits/5cd27f3b71d60f0f9b646a67d75a5a3d50d6b2c2)

consensus/src/processes/pruning_proof/mod.rs
coderofstuff: Fixed in [5cd27f3](https://github.com/kaspanet/rusty-kaspa/pull/494/commits/5cd27f3b71d60f0f9b646a67d75a5a3d50d6b2c2)

consensus/src/processes/pruning_proof/mod.rs
coderofstuff: Created an issue to track it in: https://github.com/kaspanet/rusty-kaspa/issues/585

consensus/src/processes/pruning_proof/mod.rs
michaelsutton: See the latest comment I added

michaelsutton (approved):

michaelsutton (approved):

PR #498, opened 2024-06-30
Update params.rs - new DNSSeeder

gerri-xa:
added new DNS-Seeder for Mainnet, Testnet10, Testnet11


michaelsutton:
Merged via #506

PR #503, opened 2024-07-03
implement rpc method that fetches fee per mass based on block template

### LINKED ISSUE #501: Implement Dynamic Network fee estimation
The recent congestion has demonstrated that there are opportunities for improvement related to making wallet adapt to fee markets. Existing wallet implementations only attempt to pay the lowest possible fees and don't take into consideration the traffic in the network. This feature allows wallets to have a view of that traffic by having a way to estimate fees that achieves a certain SLA

Provide wallets an RPC method to return a dynamic fee that meets some SLA. Specific SLA to be met is TBD, and can vary. There may be multiple.

biryukovmaxim:
handler builds block template and calculates max/min/median values
closes #501


aspect:
This API call should track the last request timestamp and return a cached value if this timestamp is within some threshold range (like 1 second).  Due to the potential high frequency of this call invocation (easily a few thousand requests per second), we want to save compute and return the last calculated value.

rpc/grpc/core/proto/rpc.proto
```diff
@@ -851,3 +851,19 @@ message GetDaaScoreTimestampEstimateResponseMessage{
...
         RPCError error = 1000;
 }
+
+message GetPriorityFeeEstimateRequestMessage {}
+
+message GetPriorityFeeEstimateResponseMessage {
+  oneof fee_per_mass {
+    Virtual virtual = 2; // 1 is reserved for `All`
+  }
+  uint64 mempool_total_mass = 11;
```
michaelsutton, line 861: What would be the usage of total mass from the client side? I'm not sure we want to commit ourselves to such an API. I have another suggestion for a more abstracted response

rpc/service/src/service.rs
```diff
@@ -646,6 +654,91 @@ NOTE: This error usually indicates an RPC conversion error between the node and
...
+                .sort_unstable_by(|(lhs_fee, lhs_mass), (rhs_fee, rhs_mass)| lhs_fee.mul(rhs_mass).cmp(&rhs_fee.mul(lhs_mass)));
+            let max = {
+                let (fee, mass) = fees_and_masses.last().unwrap();
+                *fee as f64 / *mass as f64
+            };
+            let min = {
+                let (fee, mass) = fees_and_masses.first().unwrap();
+                *fee as f64 / *mass as f64
+            };
+            let (div, rem) = fees_and_masses.len().div_rem(2);
```
michaelsutton, line 717: I think that in the even case we can simply take the first entry (so simply using `fees_and_masses[fees_and_masses.len()/2]`). It seems like averaging the two middle values is an overkill, and to me it looses the property of the median of being an actually used fee/mass ratio value (unlike the avg).

rpc/service/src/service.rs
```diff
@@ -977,3 +1070,12 @@ impl AsyncService for RpcCoreService {
         })
     }
 }
+
+#[derive(Debug, Default)]
+struct EstimatedFeeCache {
+    min: portable_atomic::AtomicF64,
```
michaelsutton, line 1076: Can you explain the reasoning behind using multiple float atomics rather than an atomic arc holding the final response object? (also - adding another lib dep is another negative point for using atomic floats)

michaelsutton (changes requested):

rpc/grpc/core/proto/rpc.proto
biryukovmaxim: based on this value we can track both: absolute values/relative to bps and block limit. that's more flexible

rpc/service/src/service.rs
biryukovmaxim: > Can you explain the reasoning behind using multiple float atomics rather than an atomic arc holding the final response object? (also - adding another lib dep is another negative point for using atomic floats)

1) arc will allocate it on heap
2.1) that's not a new dep in codebase, it's new to rpc crate
2.2) ofc I can make it u64 and convert it to/from float by `from_bits`, `to_bits`. but what is the reason to have boilerplate code like that in our codebase when `portable-atomic` already has it

michaelsutton: I don't think such an heap allocation (very small, once per second) is a real consideration. And imho it will simplify the code

rpc/grpc/core/proto/rpc.proto
michaelsutton: Wdyt @aspect?

wasm/build/docs/typedoc.json
```diff
@@ -1,7 +1,7 @@
 {
     "$schema": "https://typedoc.org/schema.json",
-    "treatWarningsAsErrors": true,
+    "treatWarningsAsErrors": false,
```
coderofstuff, line 3: Why was this change necessary?

rpc/grpc/core/proto/messages.proto
```diff
@@ -59,6 +59,7 @@ message KaspadRequest {
     GetServerInfoRequestMessage getServerInfoRequest = 1092;
     GetSyncStatusRequestMessage getSyncStatusRequest = 1094;
     GetDaaScoreTimestampEstimateRequestMessage GetDaaScoreTimestampEstimateRequest = 1096;
+    GetPriorityFeeEstimateRequestMessage getPriorityFeeEstimateRequest = 1097;
```
coderofstuff, line 62:
```suggestion
    GetPriorityFeeEstimateRequestMessage getPriorityFeeEstimateRequest = 1098;
```

KaspadRequest uses even numbers

rpc/grpc/core/proto/messages.proto
```diff
@@ -118,6 +119,7 @@ message KaspadResponse {
     GetServerInfoResponseMessage getServerInfoResponse = 1093;
     GetSyncStatusResponseMessage getSyncStatusResponse = 1095;
     GetDaaScoreTimestampEstimateResponseMessage GetDaaScoreTimestampEstimateResponse = 1097;
+    GetPriorityFeeEstimateResponseMessage getPriorityFeeEstimateResponse = 1098;
```
coderofstuff, line 122:
```suggestion
    GetPriorityFeeEstimateResponseMessage getPriorityFeeEstimateResponse = 1099;
```

KaspadResponse uses odd numbers

rpc/core/src/model/message.rs
```diff
@@ -825,6 +825,37 @@ impl GetDaaScoreTimestampEstimateResponse {
...
 }
 
+#[derive(Clone, Debug, Serialize, Deserialize, BorshSerialize, BorshDeserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct GetPriorityFeeEstimateRequest {}
+
+#[derive(Clone, Copy, Debug, Default, Serialize, Deserialize, BorshSerialize, BorshDeserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct VirtualFeePerMass {
+    pub max: f64,
```
coderofstuff, line 835: Do these still represent any kind of SLA? If so, please add comments as to what these would mean in terms of time-to-inclusion

coderofstuff (changes requested):
Tested via wrpc and looks to be working fine

Not a blocker, but I think needs to be thought about in terms of dev UX: The term used here is "Priority" fee, but the actual calculation is done over real fees in the mempool.

This can be confusing because:
- when the mempool is empty, all the values are zeros - which makes sense, there is nothing to add on top of your regular tx fees to make it be included immediately
- when the mempool is congested, the values will be non-zeros. as a developer using an api that says "Priority" fee, this can be taken as fee I have to pay on top of the regular tx fees.

wasm/build/docs/typedoc.json
biryukovmaxim: In short: @aspect broke cicd wasm build

rpc/core/src/model/message.rs
biryukovmaxim: I renamed method.
this value represents  max value fee/mass in the current mempool.
I will add some description from internal discussion:

What I came to this:
When the network is overloaded, we need to know, and it will have meaning like:
min from bbt - there's a chance my tx will be included 
Something in the middle - likely tx will be included 
Max - very likely 
```
In case of normal load:
Max - it's almost included, check Ur balance 
Middle - very likely 
Min - likely 
ComputeFeeOnly - there's a chance 

For that we only  need fee_per_mass based on bbt - already in pr

Flag if mempool is overloaded - I want to calculate  and expose  mempool_total_mass.

mempool_total_mass/(block_mass_limit*bps) let's call it mempool load factor - Mlf.

If Mlf < 1 (mempool is underload) then even tx with compute fee only will be included

Mlf in [1,1.5) - normal mempool load
Mlf > 1.5 - overload
```

biryukovmaxim:
> The term used here is "Priority" fee, but the actual calculation is done over real fees in the mempool.

I renamed the method name. should be more clear now

coderofstuff (commented):
My initial comments have been addressed. I did some basic sanity tests and RPC api appears to be working:

```
$ connect 127.0.0.1
$ rpc get-fee-info
Ok(
    GetFeeInfoResponse {
        fee_per_mass: VirtualFeePerMass(
            VirtualFeePerMass {
                max: 0.0,
                median: 0.0,
                min: 0.0,
            },
        ),
        mempool_total_mass: 0,
    },
)
```

And I get non-zero values when mempool has entries (like in TN11):
```
$ rpc get-fee-info
Ok(
    GetFeeInfoResponse {
        fee_per_mass: VirtualFeePerMass(
            VirtualFeePerMass {
                max: 6.21922511034821,
                median: 6.21922511034821,
                min: 6.21922511034821,
            },
        ),
        mempool_total_mass: 34663,
    },
)
```

---

I have a few more questions related to dev UX:
1. How does a developer use these figures? Create their TX, calculate its mass and then calculate their fee by multiplying the mass and the `min`, `mid`, `max` figures they get here?
2. Is the figure returned in KAS/mass or sompi/mass? If it's in KAS, would it be better to return it in sompi and then truncate to integers? Reason being is that the eventual actual fee will be in integers, not floating numbers.
    - If we must use floats, we need to communicate clearly to the API user that they should truncate their decimals when presenting fees (or if they present fees in kas, then use only up to 8 decimal places).

biryukovmaxim:
> 1. How does a developer use these figures? Create their TX, calculate its mass and then calculate their fee by multiplying the mass and the `min`, `mid`, `max` figures they get here?

1) mempool total mass represents network load. Based on it dev will pick min/mid/max/0 value. Multiply it by tx mass
2) it's in sompi. Could be micro sompi if we want to not use floats or ratio: numerator and denominator

PR #489, opened 2024-06-18
kaspa-wallet (cli) updates (using latest resolver APIs + guide cleanup)

aspect:
Minor updates to the `kaspa-cli` crate (used by `kaspa-wallet`):
- Update CLI to use the latest `Resolver` API (enables automatic reconnect and node hopping on connection failure).
- Clean up the CLI `guide` command to remove old references to the `KOS` functionality.

This improves functionality when the `server` is set to `public` (default) allowing `kaspa-wallet` to connect to a random public node.

PR #486, opened 2024-06-13
WIP - WASM SDK TypeScript bindings (update)

aspect:
- Fix `PublicKeyGenerator::change_address_as_string()` that was returning the receive address.
- WASM SDK now builds as a GitHub artifact during the CI process.
- `State` renamed to `PoW`
- Docs now have a PoW section that unifies all PoW-related classes and functions.
- `TransactionRecord.data` (`TransactionData`) now has correct TypeScript bindings.

PR #506, opened 2024-07-22
Cumulative PR - omega branch (wRPC v2, integrations, typescript-v2, resolver v2)

aspect:
This is a cumulative PR that merges the following PRs:

On `kaspanet/rusty-kaspa`:
- #480 - wRPC v2 (refactored Borsh serialization)
- #470 - CLI wallet watch-only account implementation by @1bananagirl
- #489 - CLI updates
- #486 - Updated typescript bindings (a.k.a. typescript-v2)

On `aspectron/rusty-kaspa`:
- https://github.com/aspectron/rusty-kaspa/pull/73 - Fix Wallet API guards (related to Wallet API producing "insufficient-funds" error in async clients like KNG)
- https://github.com/aspectron/rusty-kaspa/pull/69 - Custom input signature script API enhancements by @KaffinPX (extends `kaspa-wallet-core/PendingTransaction` APIs allowing for non-standard transactions such as KRC-20).
- https://github.com/aspectron/rusty-kaspa/pull/68 - Enhancements to WASM API for ScriptBuilder by @surinder83singh and @KaffinPX
- https://github.com/aspectron/rusty-kaspa/pull/67 - HexView interface for ScriptBuilder (formatted hex view for scripts)
- https://github.com/aspectron/rusty-kaspa/pull/66 - client-side framework improvements to Mass Calculator (Kip9::Beta variant)
- https://github.com/aspectron/rusty-kaspa/pull/59 - rebased #470 

***

This branch also implements the following RPC methods or updates (all of them are related to `kaspa-resolver` refactoring): 
- `get_sync_status()` - provides node sync status (can be used for low-bandwidth node sync status polling)
- `get_connection_count()` - provides the total number of TCP sockets in use by the node (sum of gRPC clients, wRPC clients, and connected peers)
- `get_sysem_info()` - provides a unique ID of the system (a.k.a. machine-id in the Linux universe, with fallback on the mac address on other platforms), git commit hash of the rusty-kaspa repository (if available), total system RAM, number of CPU cores, and the internal FD limit of the kaspad process. These metrics are used by `kaspa-resolver` to profile the node for its socket (RPC client) connection capacity.
- `get_metrics()` RPC call introduces a provisional dictionary-based map for custom metrics.

IMPORTANT: `get_server_info()` used by the client-side framework is broken in this update due to the change in the data structure (this includes wRPC and gRPC protocols).

wRPC v2 Borsh encoding introduces breaking protocol changes. wRPC resolver has been reconfigured to operate against the new public node infrastructure that uses wRPC v2 protocol. (ecosystem contributors are currently deploying this infrastructure). Rust and WASM SDK users will not experience any issues as when they update to the new version of the Rust or WASM framework based on this branch, they will switch over to already operational resolvers supporting the new protocol.

wRPC v2 now introduces custom serializers (located beside their respective associated primitives) where each struct (as well as structs serialized with `serialize!()` macro are serialized into a `Vec<u8>`, which is then serialized into the wrapping structure.  This allows each data structure serialization to become "elastic" and introduce new data into payloads, while older clients will ignore it.

***

Other updates:
- `kaspa-resolver` crate has been removed from the workspace and relocated to [aspectron/kaspa-resolver](https://github.com/aspectron/kaspa-resolver). This was done as a part of the v2 refactoring that uses an improved methodology for public node tracking.
- Implement paginated access to the Transaction store backed by the browser `IndexDb` WASM interface inside of the `kaspa-wallet-core` crate (by @surinder83singh).
- Fix conditions in `kaspa-wallet-core/UtxoProcessor` causing balance fluctuations during batch transaction processing (by @surinder83singh).
- Introduced `kaspa-utils/git` now contains Git commit hash (long and short), which is now appended to `kaspad` version (a similar functionality was previously PRed in - #338, but in this update, @biryukovmaxim and I have created a custom solution that has a better security profile as it does not require any external `build.rs` dependencies).


Cargo.toml
```diff
@@ -195,9 +192,11 @@ ipnet = "2.9.0"
 itertools = "0.11.0"
 js-sys = "0.3.67"
 keccak = "0.1.4"
+lazy_static = "1.4.0"
```
biryukovmaxim, line 195: afair once_cell has all features of lazy_static. No need to have a new dep

rpc/grpc/core/proto/rpc.proto
```diff
@@ -819,19 +846,21 @@ message GetMetricsResponseMessage{
...
 }
 
 message GetServerInfoResponseMessage{
-  repeated uint32 rpcApiVersion = 1; // Expecting exactly 4 elements
-  string serverVersion = 2;
-  string networkId = 3;
-  bool hasUtxoIndex = 4;
-  bool isSynced = 5;
-  uint64 virtualDaaScore = 6;
+  uint32 rpcApiVersion = 1;
```
biryukovmaxim, line 857: It's a breaking change

rpc/grpc/core/proto/rpc.proto
```diff
@@ -819,19 +846,21 @@ message GetMetricsResponseMessage{
...
   BandwidthMetrics bandwidthMetrics = 13;
   ConsensusMetrics consensusMetrics = 14;
+  StorageMetrics storageMetrics = 15;
   RPCError error = 1000;
 }
 
 message GetServerInfoRequestMessage{
 }
 
 message GetServerInfoResponseMessage{
```
biryukovmaxim, line 856: It's a breaking changes. Why not adding fields to the end?

rpc/grpc/core/proto/rpc.proto
```diff
@@ -819,19 +846,21 @@ message GetMetricsResponseMessage{
...
-  repeated uint32 rpcApiVersion = 1; // Expecting exactly 4 elements
-  string serverVersion = 2;
-  string networkId = 3;
-  bool hasUtxoIndex = 4;
-  bool isSynced = 5;
-  uint64 virtualDaaScore = 6;
+  uint32 rpcApiVersion = 1;
+  uint32 rpcApiRevision = 2;
```
biryukovmaxim, line 858: It's a breaking changes. Why not adding them to the end?

aspect: Yes, it is a breaking change for gRPC (for a method that is not present in Golang, as such its exposure surface should be reduced). It replaces existing variables that are at the front.  (it's a method that was introduced not so long ago to accommodate `UtxoProcessor` in `kaspa-wallet-core`).  I have changed the way the versioning works in the RPC as such the previous data structure for this method no longer makes sense.

Cargo.toml
aspect: Replaced in this side PR: https://github.com/aspectron/rusty-kaspa/pull/78

@biryukovmaxim  Please review.

rpc/grpc/core/proto/messages.proto
```diff
@@ -118,6 +120,8 @@ message KaspadResponse {
     GetServerInfoResponseMessage getServerInfoResponse = 1093;
     GetSyncStatusResponseMessage getSyncStatusResponse = 1095;
     GetDaaScoreTimestampEstimateResponseMessage GetDaaScoreTimestampEstimateResponse = 1097;
+    GetConnectionsResponseMessage getConnectionsResponse= 1099;
```
coderofstuff, line 123: For these two new responses, please use 1101 and 1103

rpc/grpc/core/proto/messages.proto
```diff
@@ -59,6 +59,8 @@ message KaspadRequest {
     GetServerInfoRequestMessage getServerInfoRequest = 1092;
     GetSyncStatusRequestMessage getSyncStatusRequest = 1094;
     GetDaaScoreTimestampEstimateRequestMessage GetDaaScoreTimestampEstimateRequest = 1096;
+    GetConnectionsRequestMessage getConnectionsRequest = 1098;
+    GetSystemInfoRequestMessage getSystemInfoRequest = 1100;
```
coderofstuff, line 63: For these two new requests, please use 1100 and 1102

coderofstuff (changes requested):

wasm/examples/nodejs/typescript/src/scriptBuilder.ts
```diff
@@ -0,0 +1,13 @@
...
+const myScript = new ScriptBuilder()
+                .addOp(Opcodes.OpTrue)
+
+const P2SHScript = myScript.createPayToScriptHashScript()
+const address = addressFromScriptPublicKey(P2SHScript, NetworkType.Mainnet)
+
+// Payable address
+console.log(address!.toString())
+// Unlock signature script
+console.log(myScript.createPayToScriptHashSignatureScript(""))
```
coderofstuff, line 13: Running `npx ts-node scriptBuilder.ts` runs into this error:

```
/Users/dev/.npm/_npx/1bf7c3c15bf47d04/node_modules/ts-node/src/index.ts:859
    return new TSError(diagnosticText, diagnosticCodes, diagnostics);
           ^
TSError: ⨯ Unable to compile TypeScript:
scriptBuilder.ts:13:22 - error TS2551: Property 'createPayToScriptHashSignatureScript' does not exist on type 'ScriptBuilder'. Did you mean 'encodePayToScriptHashSignatureScript'?

13 console.log(myScript.createPayToScriptHashSignatureScript(""))
                        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

  ../../../../nodejs/kaspa/kaspa.d.ts:6832:3
    6832   encodePayToScriptHashSignatureScript(signature: HexString | Uint8Array): HexString;
           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    'encodePayToScriptHashSignatureScript' is declared here.
```

KaffinPX: Looks like an outdated example, must be replaced w encodePayToScriptHashSignatureScript

aspect: Fixed this here: https://github.com/aspectron/rusty-kaspa/commit/98ba65271cd434f13b3cf98589331ddf5d703201

what carries a better meaning here... `encode` or `create`?

KaffinPX: technically its encoding a sigScript

consensus/client/src/transaction.rs
```diff
@@ -383,7 +389,37 @@ impl Transaction {
...
+    }
+
+    pub fn inputs(&self) -> Vec<cctx::TransactionInput> {
+        let inner = self.inner();
+        let inputs = inner.inputs.iter().map(Into::into).collect::<Vec<cctx::TransactionInput>>();
+        inputs
+    }
+
+    pub fn inputs_outputs(&self) -> (Vec<cctx::TransactionInput>, Vec<cctx::TransactionOutput>) {
+        let inner = self.inner();
```
coderofstuff, line 419: I assume that avoiding having to lock twice is the reason this doesn't reuse the existing `inputs` and `outputs` fn right above it. Correct?

cli/src/modules/account.rs
```diff
@@ -84,6 +84,8 @@ impl Account {
                                 "account import mnemonic multisig [additional keys]",
                                 "Import mnemonic and additional keys for a multisig account",
                             ),
+                            ("account import bip32-watch", "Import a extended public key for a watch-only bip32 account"),
```
coderofstuff, line 87: Why not use `account watch bip32` and `account watch multisig` here?

cli/src/modules/pskb.rs
```diff
@@ -0,0 +1,259 @@
...
+                } else {
+                    let pskb = Self::parse_input_pskb(argv.first().unwrap().as_str())?;
+
+                    match account.pskb_broadcast(&pskb).await {
+                        Ok(sent) => tprintln!(ctx, "Sent transactions {:?}", sent),
+                        Err(e) => terrorln!(ctx, "Send error {:?}", e),
+                    }
+                }
+            }
+            "debug" => {
```
coderofstuff, line 196: Rather than `debug` this probably should be `parse`. My thinking is that parsing a PSKB is a normal thing that someone would do to check out it's contents before trying to sign it or do anything else with it (if they receive it from someone else). So in that sense, they're not really trying to debug anything - they're just doing due diligence. `debug` implies you're trying to investigate something.

consensus/core/src/config/params.rs
```diff
@@ -309,6 +309,10 @@ pub const MAINNET_PARAMS: Params = Params {
         "kaspadns.kaspacalc.net",
         // This DNS seeder is run by supertypo
         "n-mainnet.kaspa.ws",
+        // This DNS seeder is run by -gerri-
+        "dnsseeder-kaspa-mainnet.x-con.at",
```
coderofstuff, line 313: I haven't verified but I presume these have been checked and that they're legit. Right?

rpc/wrpc/wasm/src/client.rs
```diff
@@ -957,6 +945,8 @@ build_wrpc_wasm_bindgen_interface!(
         /// performance and status of the Kaspa node.
         /// Returned information: Memory usage, CPU usage, network activity.
         GetMetrics,
+        /// Retrieves current number of network connections
+        GetConnections,
```
coderofstuff, line 949: Should `GetSystemInfo` be added somewhere here too?

wasm/examples/nodejs/javascript/transactions/simple-transaction.js
coderofstuff, line 1: This file is pretty old and I think can be completely removed. It's contents are covered in `single-transaction-demo.js`

wallet/pskt/src/wasm/bundle.rs
coderofstuff, line 1: What's this empty file for?

coderofstuff (changes requested):
I'm halfway through the PR. Initial comments and questions.

## For fixing
- [ ] Running this:
```
cargo run --bin kaspa-cli

connect public
wallet create main
```

Results in the terminal hanging. Note: I have no running node locally.  If I try something else though like `rpc get-system-info` I get a response, so I must be connected to somewhere.

wallet/pskt/src/wasm/bundle.rs
aspect: placeholder. we don't have WASM bindings for Bundle yet. Will try to get to it in the coming days.

wasm/examples/nodejs/javascript/transactions/simple-transaction.js
aspect: Negatory on that:  `simple-transaction.js` uses `Generator` by using `createTransactions()` (wrapper) whereas `single-transaction-demo.js` uses `createTransaction()` (raw transaction creation)

rpc/wrpc/wasm/src/client.rs
aspect: I suppose.  This was added specifically for `kaspa-resolver` machine profiling, which is why it was missed.

consensus/core/src/config/params.rs
aspect: Gerri (x-con.at) is a contributor from Discord on `#testnet`. I don't know more. I added this from https://github.com/kaspanet/rusty-kaspa/pull/498

`kaspa-dnsseeder.net` is H@H who I can vouch for.

tn-10 became completely dysfunctional due to the lack of any DNS seeders.  Gerri brought up one on his own, I asked H@H to set up an instance.

cli/src/modules/pskb.rs
aspect: @1bananagirl ^ WDYT?  

I think if `parse` is introduced, it should present a formatted view (nice layout) while debug can remain as a debug view.

cli/src/modules/account.rs
aspect: `bip32-watch` is the account type (it is a `FromStr` - not a command).  If I remember correctly, multisig doesn't need watch accounts. @biryukovmaxim @1bananagirl

consensus/client/src/transaction.rs
aspect: Correct.  All functions here are modular extractors designed to de-compose tx into components as-needed.

cli/src/modules/rpc.rs
```diff
@@ -38,19 +38,27 @@ impl Rpc {
                 tprintln!(ctx, "ok");
             }
             RpcApiOps::GetMetrics => {
-                let result = rpc.get_metrics(true, true, true, true).await?;
+                let result = rpc.get_metrics(true, true, true, true, true, true).await?;
+                self.println(&ctx, result);
+            }
+            RpcApiOps::GetSystemInfo => {
+                let result = rpc.get_system_info().await?;
```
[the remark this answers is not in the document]
aspect: k.  I'll add a custom debug output trait for it.

cli/src/modules/account.rs
1bananagirl: Discussion source of the reason here: https://github.com/kaspanet/rusty-kaspa/pull/470

cli/src/modules/pskb.rs
1bananagirl: Good idea. Does the parse view also have all fields displayed, or a limited range of ?

cli/src/modules/account.rs
1bananagirl: @coderofstuff yes we can rename import to watch !

aspect:
@coderofstuff @surinder83singh fixed this yesterday. It's the API guard. An "async gift that keeps giving". Should be back normal now (albeit i haven't yet had a chance to test myself).

cli/src/modules/pskb.rs
1bananagirl: Added in PR  https://github.com/aspectron/rusty-kaspa/pull/92 @coderofstuff @aspect

cli/src/modules/account.rs
1bananagirl: Added in PR  https://github.com/aspectron/rusty-kaspa/pull/92 @coderofstuff @aspect

ZenVoich:
I found some inconsistencies in TS types and actual data. Using latest node build and wasm sdk from this branch.

In `ITransactionOutpoint` field name `transactionId`, in data `transaction_id`

In `ITransactionOutput` field `scriptPublicKey` has type `IScriptPublicKey { version : number; script: HexString; } `, in data - hex string

aspect:
> I found some inconsistencies in TS types and actual data. Using latest node build and wasm sdk from this branch.
> 
> In `ITransactionOutpoint` field name `transactionId`, in data `transaction_id`
> 
> In `ITransactionOutput` field `scriptPublicKey` has type `IScriptPublicKey { version : number; script: HexString; } `, in data - hex string

I unfortunately can not find what you are referring to.  If you can give an actual description of what "data" means, I would appreciate.

The issue that we have is that data comes from different directions - there is data structures that come from consensus made available via RPC, which are serialized using serde (so property names are auto-translated using serde "camelCase" attribute), there is a WASM implementation of all primitives (that mostly present data as getters)..   I've updated `ITransactionOutput::scriptPublicKey` to declare as `IScriptPublicKey | HexString`, however I can't see where `transaction_id` could be coming from.

If you could give an example in a form of "I am calling this and in response I get this data" would be helpful.

cli/src/modules/pskb.rs
aspect: This has been merged ^.

cli/src/modules/rpc.rs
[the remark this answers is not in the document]
aspect: @coderofstuff I've added a custom `Debug` `impl` for `GetSystemInfoResource` here https://github.com/aspectron/rusty-kaspa/commit/7d0016974baf9028ccca9dda628ea9f501f4b5b5

ZenVoich:
> I unfortunately can not find what you are referring to. If you can give an actual description of what "data" means, I would appreciate.

Data is what actual data I get from RPC call using wRPC borsh



> If you could give an example in a form of "I am calling this and in response I get this data" would be helpful.

I get blocks like this
```js
let blocks = await rpc.getBlocks({
  lowHash,
  includeBlocks: true,
  includeTransactions: true,
});
```

and for example
```js
blocks[0].transactions[0].inputs[1].previousOutpoint.transaction_id // tx id is here
blocks[0].transactions[0].inputs[1].previousOutpoint.transactionId // no such field
```
but there is a type error "Property 'transaction_id' does not exist on type 'ITransactionOutpoint'. Did you mean 'transactionId'?"

Same with `blocks[0].transactions[0].outputs[1].scriptPublicKey`

Probably it's already fixed somewhere, I will wait for the release and check it again.

aspect:
@ZenVoich my bad.  Thanks for pointing this out.  One of the type aliases was converted to an actual struct, but I didn't realize that and I kept looking in the wrong place.  Fixed in https://github.com/aspectron/rusty-kaspa/commit/9cec07787edf473a68f5a300842fab81747ba67e

michaelsutton (approved):

coderofstuff (approved):

consensus/core/src/config/params.rs
gerri-xa: I am legit yes :)
Kaspa supporter since Nov 2021, Archival Node Maintainer, Community Tester
4 redundant DNS-Nodes are located in Austria, Germany, Iceland, Netherlands

6975499756 (commented):
********

PR #504, opened 2024-07-06
chore(docker): init

IzioDev:
next step would be helm chart or kustomize, probably.

sad that image size couldn't be reduced

*question*: interested in auto ci build&push on docker hub or any other registries?

PR #511, opened 2024-07-30
Add support for environment variable configuration, and Docker container images via ghcr

based64god:
This addresses a long standing issue with difficulty of deploying `kaspad` in a way that makes sense in a modern datacenter. This PR adds sensible default docker container build pipelines for the most common server microarchitectures, amd64 and arm64, as well as the ability to configure kaspad with environment variables, which is generally easier in a containerized environment. arm builds are likely to be quite slow until better support for arm matrix runners is added to github runners since everything needs to happen through qemu.


based64god:
This PR likely supercedes #504

someone235:
Handled by #762

PR #513, opened 2024-08-05
O(k log n) mempool transaction sampler + Fee estimation API 

michaelsutton:
The following PR includes two major components: 
1. A new efficient `O(k log n)` mempool transaction samplig mechanism; and a
2. Fee Estimator which utilzes some of the same underlying structures.

_Background points_: 
-  Kaspa's high BPS implies that miners must use probabilistic weighted sampling in order to reduce collisions while maximizing fees.  To understand why, observe that on the one hand choosing the transactions which maximze fees determinitically will result in many duplicates across parallel mined blocks (of which only a single miner will earn a fee). On the other hand, uniform sampling is definetly not a beneficial strategy. The soultion is weighted sampling with sufficient bias towards higher fee transactions, but yet with reduced collision probability. 
- Naively (as in the current implementation), sampling from a weighted probability space requies maintaining a weight interval vector which requires `O(n)` runtime cost for each item insertion/removal. Since such a cost is unacceptable (per transaction update), we currently compose the vector for each new block template, which is an operation with BPS frequency. Running this sampling process over a mempool with 1M transactions would take ~350ms, which is way too high for TN11 which has 100ms block time (10BPS), hence the current code uniformly samples 50K transactions from the full mempool and then runs the weighted sampler only over these 50K transactions. This uniform sampling step is ok if the mempool fee distribution is not too unbalanced, but on a full mempool with a few high priority transactions (outliers), this uniform sampling step will be unjustified. 

_Weighted sampling via trees_ (with log time insertions/removals and log time sampling time):
- The PR includes a `SearchTree` data structue, based on the great highly customizable B+ tree [library ](https://crates.io/crates/sweep-bptree/0.4.0) by @shuoli84 ([github](https://github.com/shuoli84/sweep-bptree)). See [here](https://github.com/michaelsutton/rusty-kaspa/blob/feerate-est-ng/mining/src/mempool/model/frontier/search_tree.rs) how it is used. 
- The `SearchTree` is a transaction search tree sorted by feerate order and searchable for probabilistic weighted sampling.
The tree has the following properties:
    1. Linear time ordered access (ascending / descending)
    2. Insertions/removals in `log(n)` time
    3. Search for a weight point `p ∈ [0, total_weight)` in `log(n)` time
    4. Compute the prefix weight of a key, i.e., the sum of weights up to that key (inclusive) according to key order, in `log(n)` time
    5. Access the total weight in `O(1)` time. The total weight has numerical stability since it is recomputed from subtree weights for each item insertion/removal
    6. Computing the prefix weight is a crucial operation if the tree is used for random sampling and is highly imbalanced in terms of weight variance. See `Frontier::sample_inplace` for more details.  

_Sampling performance_:

Based on this search tree, we implemented efficient in-place mempool sampling. We then select the most efficient sampling method dynamically based on mempool size. Overall, this dynamic policy provides full transaction selection in less than 150 µs even if the mempool has 1M entries (!!). See `mining/benches` for more details.     

_Fee estimator API_:

The PR includes fee estimation logic based on mathemtical analysis in a joint effort with @ShaiW and @hashdag. For full details see the following [Jupyter Notebook](https://github.com/michaelsutton/rusty-kaspa/blob/feerate-est-ng/mining/src/feerate/fee_estimation.ipynb).


utils/src/expiring_cache.rs
```diff
@@ -0,0 +1,152 @@
...
+        }
+
+        // We reach here if either we are the refetching thread or the current data has fully expired
+        let new_item = refetch_future.await;
+        let timestamp = Instant::now();
+        // Update the store even if we were not in charge of refetching - let the last thread make the final update
+        self.store.store(Some(Arc::new(Entry { item: new_item.clone(), timestamp })));
+
+        if fetching {
+            let result = self.fetching.compare_exchange(true, false, Ordering::SeqCst, Ordering::SeqCst);
```
biryukovmaxim, line 69: why not `.store(false, Release)`

rpc/core/src/model/feerate_estimate.rs
```diff
@@ -0,0 +1,51 @@
...
+    /// A vector of *normal* priority feerate values. The first value of this vector is guaranteed to exist and
+    /// provide an estimation for sub-*minute* DAG inclusion. All other values will have shorter estimation
+    /// times than all `low_bucket` values. Therefor by chaining `[priority] | normal | low` and interpolating
+    /// between them, one can compose a complete feerate function on the client side. The API makes an effort
+    /// to sample enough "interesting" points on the feerate-to-time curve, so that the interpolation is meaningful.
+    pub normal_buckets: Vec<FeerateBucket>,
+
+    /// A vector of *low* priority feerate values. The first value of this vector is guaranteed to
+    /// exist and provide an estimation for sub-*hour* DAG inclusion.
+    pub low_buckets: Vec<FeerateBucket>,
```
biryukovmaxim, line 29: why not male it depending on request params to only transmit requested data?

coderofstuff (commented):
I'm most of the way through this PR (39/50). Still reviewing the crucial parts.

Initial question: When the mempool is empty, what should fee estimate return? The minimum amount of fee that a node will accept for the tx? Or just 0?

consensus/core/src/config/params.rs
```diff
@@ -501,7 +501,8 @@ pub const SIMNET_PARAMS: Params = Params {
     target_time_per_block: Testnet11Bps::target_time_per_block(),
     past_median_time_sample_rate: Testnet11Bps::past_median_time_sample_rate(),
     difficulty_sample_rate: Testnet11Bps::difficulty_adjustment_sample_rate(),
-    max_block_parents: Testnet11Bps::max_block_parents(),
+    // For simnet, we deviate from TN11 configuration and allow at least 64 parents in order to support mempool benchmarks out of the box
+    max_block_parents: if Testnet11Bps::max_block_parents() > 64 { Testnet11Bps::max_block_parents() } else { 64 },
```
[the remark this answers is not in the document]
michaelsutton: The `max` function (coming from `Ord` trait) is non-const so can't be used here within a const params decleration

rpc/core/src/model/feerate_estimate.rs
michaelsutton: Wer'e talking about 3-10 data points here at most, so I don't see why this shoud be optimized in any way

utils/src/expiring_cache.rs
michaelsutton: I personally like the compare exchange bcs it allows the assert below, and is the more "accurate" operation here

michaelsutton:
> I'm most of the way through this PR (39/50). Still reviewing the crucial parts.
> 
> Some Questions:
> 
> 1. When the mempool is empty, what should fee estimate return? The minimum amount of fee that a node will accept for the tx? Or just 0?
> 2. `GetFeeEstimate` is missing in `rpc/wrpc/wasm/src/router.rs` and `rpc/wrpc/wasm/src/client.rs`. Is this intentional?

1. We return the minimum standard feerate (currently 1.0) 
2. No. I'll check it out

coderofstuff:
Diff to add the calls to kaspa-cli:

```
diff --git a/cli/src/modules/rpc.rs b/cli/src/modules/rpc.rs
index c84915480..53478c174 100644
--- a/cli/src/modules/rpc.rs
+++ b/cli/src/modules/rpc.rs
@@ -229,6 +229,15 @@ impl Rpc {
                     }
                 }
             }
+            RpcApiOps::GetFeeEstimate => {
+                let result = rpc.get_fee_estimate_call(GetFeeEstimateRequest {}).await?;
+                self.println(&ctx, result);
+            }
+            RpcApiOps::GetFeeEstimateExperimental => {
+                let verbose = if argv.is_empty() { false } else { argv.remove(0).parse().unwrap_or(false) };
+                let result = rpc.get_fee_estimate_experimental_call(GetFeeEstimateExperimentalRequest { verbose }).await?;
+                self.println(&ctx, result);
+            }
             _ => {
                 tprintln!(ctx, "rpc method exists but is not supported by the cli: '{op_str}'\r\n");
                 return Ok(());
diff --git a/rpc/wrpc/server/src/router.rs b/rpc/wrpc/server/src/router.rs
index 4c0d2b3aa..024f6a9c6 100644
--- a/rpc/wrpc/server/src/router.rs
+++ b/rpc/wrpc/server/src/router.rs
@@ -68,6 +68,8 @@ impl Router {
                 SubmitTransaction,
                 SubmitTransactionReplacement,
                 Unban,
+                GetFeeEstimate,
+                GetFeeEstimateExperimental,
             ]
         );
 

```

mining/src/mempool/model/frontier.rs
```diff
@@ -0,0 +1,391 @@
...
+                    }
+                    let query = distr.sample(rng);
+                    item = self.search_tree.search(query);
+                }
+                item
+            };
+            sequence.push(item.tx.clone(), item.mass);
+            total_selected_mass += item.mass; // Max standard mass + Mempool capacity bound imply this will not overflow
+        }
+        trace!("[mempool frontier sample inplace] collisions: {_collisions}, cache: {}", cache.len());
```
coderofstuff, line 151: Do you have metrics on the average number of collisions when the mempool size is something like 1M?

mining/src/mempool/model/frontier.rs
```diff
@@ -0,0 +1,391 @@
...
+        let mut _collisions = 0;
+
+        // The sampling process is converging so the cache will eventually hold all entries, which guarantees loop exit
+        'outer: while cache.len() < self.search_tree.len() && total_selected_mass <= desired_mass {
+            let query = distr.sample(rng);
+            let item = {
+                let mut item = self.search_tree.search(query);
+                while !cache.insert(item.tx.id()) {
+                    _collisions += 1;
+                    if top == item {
```
coderofstuff, line 134: I think the reason you chose this as a method for reducing the sampling collisions (by reducing the search space) is because the `top` is the item that has the likeliest probability of being chosen, given the weight. Is that right?

mining/src/mempool/model/frontier.rs
```diff
@@ -0,0 +1,391 @@
...
+            Box::new(TakeAllSelector::new(self.search_tree.ascending_iter().map(|k| k.tx.clone()).collect()))
+        } else if self.total_mass > policy.max_block_mass * COLLISION_FACTOR {
+            let mut rng = rand::thread_rng();
+            Box::new(SequenceSelector::new(self.sample_inplace(&mut rng, policy), policy.clone()))
+        } else {
+            Box::new(RebalancingWeightedTransactionSelector::new(
+                policy.clone(),
+                self.search_tree.ascending_iter().cloned().map(CandidateTransaction::from_key).collect(),
+            ))
+        }
```
coderofstuff, line 180: The ordering here threw me off a bit, and I was a bit confused why the last one had an `ascending_iter`. Reading more into it, the second if-statement with the `SequenceSelector` is where the real magic of this PR happens (and not at the 3rd one).

coderofstuff (changes requested):
This is outstanding work. The code itself is solid and looks fundamentally sound.

Running it over TN11 for a few days and it seems to be working fine.

My comments are more on along the lines of documentation, questions or thoughts

## Requested changes
- [ ] `Feerate` needs to be defined in the jupyter notebook. I know it's `fee / mass` but I don't see it being defined anywhere. It might not be immediately clear to other devs.
- [ ] The API is different from how typical fee estimation APIs work. Provide an example usage of how the Fee rate api should be used by a wallet dev. It returns feerates, not direct fees. I only assume that I need to multiply my TX's mass by the fee rate to get an actual fee to use
- [ ] In line with the above, the API uses `GetFeeEstimate` and `GetFeeEstimateExperimental` but instead of actual fees, it returns fee rates. Probably should change those names to `GetFeeRateEstimate` (or `GetFeerateEstimate`?)

## Other thoughts
- "feerate" is being used here like it's a single word but it's actually "fee rate". Not something that needs to be changed but it seems weird to me.

mining/src/mempool/model/frontier.rs
michaelsutton: If the feerate distribution is more or less linear (i.e., linearly distributed over some range), you get ~0 collisions with 1M entries. You start getting collisions only below 100K entries (100K: ~1 collisions, 10K: ~16, 5000: ~33, 2000: ~100; below ~2000 we use the other selector).

The interesting part is that following this question I benchmarked the worst-case distribution and noticed an easy optimization which made a dramatic improvement ([commit](https://github.com/kaspanet/rusty-kaspa/pull/513/commits/18a8fc45349f7cb0e919fd314a67a6ff3eb768f6)) 

The change was to eagerly iterate the down iterator until finding a top item which was not sampled yet. After this, also with worst-case scenarios, the results look good:

```
mempool inplace sampling/inplace sampling worst case (bucket size: 300)
                        time:   [206.51 µs 207.48 µs 208.37 µs]
Found 1 outliers among 100 measurements (1.00%)
  1 (1.00%) high mild
---------------------- 
  Avg collisions: 1293
```

mining/src/mempool/model/frontier.rs
michaelsutton: Correct. See my other [comment](https://github.com/kaspanet/rusty-kaspa/pull/513#discussion_r1712984709) form more on the subject

michaelsutton: Also, it would be challenging to reduce the space by dropping items from the middle of the tree (higher complexity + not supported by the visitor pattern I used).

mining/src/mempool/model/frontier.rs
michaelsutton: Right, the 2nd case is the interesting one. The `ascending_iter()` used in the 1rd and 3rd is just the default `iter()` method of the tree. The rebalancing selector is not sensitive to the order of input txs

mining/src/mempool/model/frontier/feerate_key.rs
```diff
@@ -0,0 +1,107 @@
...
+    fn cmp(&self, other: &Self) -> std::cmp::Ordering {
+        // Our first priority is the feerate.
+        // The weight function is monotonic in feerate so we prefer using it
+        // since it is cached
+        match self.weight().total_cmp(&other.weight()) {
+            core::cmp::Ordering::Equal => {}
+            ord => return ord,
+        }
+
+        // If feerates (and thus weights) are equal, prefer the higher fee in absolute value
```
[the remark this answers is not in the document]
michaelsutton: prefer in the sense that you return `Ordering::Greater` for the higher fee

tiram88 (changes requested):
**First and foremost, this is an amazing PR!**

The new selector over the full mempool frontier is a masterpiece both conceptually and at engeniring level.

I just added some comments, essentially formal and nit things. The feerate estimator though revealed a little problem with NaN being returned on some edge cases.

IMHO `Feerate` should be defined precisely (dimensions and units) somewhere in the codebase, either as a documented type or in a glossary or any other meaningful way.

mining/src/mempool/model/frontier.rs
```diff
@@ -0,0 +1,391 @@
...
+            true
+        } else {
+            false
+        }
+    }
+
+    pub fn remove(&mut self, key: &FeerateTransactionKey) -> bool {
+        let mass = key.mass;
+        if self.search_tree.remove(key) {
+            self.total_mass -= mass;
```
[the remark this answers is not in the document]
michaelsutton: You mean below zero? if the search tree is consistent this should never happen bcs we track total_mass precisely (and if it does happen, I think we want to crash).

mining/src/mempool/model/transactions_pool.rs
```diff
@@ -133,18 +137,20 @@ impl TransactionsPool {
...
             }
         }
         self.parent_transactions.remove(transaction_id);
         self.chained_transactions.remove(transaction_id);
-        self.ready_transactions.remove(transaction_id);
 
         // Remove the transaction itself
         let removed_tx = self.all_transactions.remove(transaction_id).ok_or(RuleError::RejectMissingTransaction(*transaction_id))?;
 
+        self.ready_transactions.remove(&(&removed_tx).into());
```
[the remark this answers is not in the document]
michaelsutton: `Frontier::remove` requires the feerate key which includes the fee and the mass. We can only get this info from the full tx once we find it in `all_txs` (note that `into`). 
I've verified multiple times that all collections are consistent and updated at once (insert/remove), so if the tx is not in `all_txs` there will be no need to remove it from `ready_txs`

mining/src/mempool/config.rs
```diff
@@ -29,8 +28,7 @@ pub(crate) const DEFAULT_MAXIMUM_STANDARD_TRANSACTION_VERSION: u16 = TX_VERSION;
 
 #[derive(Clone, Debug)]
 pub struct Config {
-    pub maximum_transaction_count: u64,
-    pub maximum_ready_transaction_count: u64,
+    pub maximum_transaction_count: u32,
```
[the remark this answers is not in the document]
michaelsutton: I think my reasoning was that having this limit under `u32::MAX` + standard tx mass limit, implies that `Frontier::total_mass` will never overflow

mining/src/mempool/model/frontier.rs
```diff
@@ -0,0 +1,391 @@
...
+        let mut mass_per_block = args.maximum_mass_per_block as f64;
+        let mut inclusion_interval = average_transaction_mass / (mass_per_block * bps);
+        let mut estimator = FeerateEstimator::new(self.total_weight(), inclusion_interval);
+
+        // Corresponds to the removal of the top item, hence the skip(1) below
+        mass_per_block -= average_transaction_mass;
+        inclusion_interval = average_transaction_mass / (mass_per_block * bps);
+
+        // Search for better estimators by possibly removing extremely high outliers
+        for key in self.search_tree.descending_iter().skip(1) {
```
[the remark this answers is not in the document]
michaelsutton: I think I disagree. For instance, if there are only 10 txs overall, each with a huge feerate. I'd want the estimator to ignore all of them and consider blocks as if they have 10 captured slots -- leading to estimations which do not need to compete with these 10 huge fee txs

mining/src/feerate/mod.rs
```diff
@@ -0,0 +1,146 @@
...
+#[derive(Clone, Debug)]
+pub struct FeerateEstimations {
+    /// *Top-priority* feerate bucket. Provides an estimation of the feerate required for sub-second DAG inclusion.
+    pub priority_bucket: FeerateBucket,
+
+    /// A vector of *normal* priority feerate values. The first value of this vector is guaranteed to
+    /// provide an estimation for sub-*minute* DAG inclusion. All other values will have shorter estimation
+    /// times than all `low_bucket` values. Therefor by chaining `[priority] | normal | low` and interpolating
+    /// between them, one can compose a complete feerate function on the client side. The API makes an effort
+    /// to sample enough "interesting" points on the feerate-to-time curve, so that the interpolation is meaningful.
```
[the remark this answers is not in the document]
michaelsutton: The description is on the API level, meant to describe the contract we are committing to

mining/src/feerate/mod.rs
```diff
@@ -0,0 +1,146 @@
...
+        c1 * c2 / (-2f64 * feerate.powi(ALPHA - 1))
+    }
+
+    fn quantile(&self, lower: f64, upper: f64, frac: f64) -> f64 {
+        assert!((0f64..=1f64).contains(&frac));
+        let (c1, c2) = (self.inclusion_interval, self.total_weight);
+        let z1 = self.feerate_to_time_antiderivative(lower);
+        let z2 = self.feerate_to_time_antiderivative(upper);
+        let z = frac * z2 + (1f64 - frac) * z1;
+        ((c1 * c2) / (-2f64 * z)).powf(1f64 / (ALPHA - 1) as f64)
```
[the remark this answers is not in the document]
michaelsutton: afaiu the case `z == 0` requires that `c1 · c2 == 0`, so I handled that in a condition above

tiram88 (changes requested):

mining/src/mempool/model/transactions_pool.rs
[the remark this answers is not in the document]
tiram88: Lgtm

michaelsutton:
> 2\. Is this intentional?

@aspect please note that I'm not updating the wasm API bindings and leaving that to you. Also note that RBF API was not added yet to wasm (i.e., in rpc/wrpc/wasm/src/client.rs)

michaelsutton:
> This is outstanding work. The code itself is solid and looks fundamentally sound.
> 
> Running it over TN11 for a few days and it seems to be working fine.
> 
> My comments are more on along the lines of documentation, questions or thoughts
> 
> ## Requested changes
> * [ ]  `Feerate` needs to be defined in the jupyter notebook. I know it's `fee / mass` but I don't see it being defined anywhere. It might not be immediately clear to other devs.
> * [ ]  The API is different from how typical fee estimation APIs work. Provide an example usage of how the Fee rate api should be used by a wallet dev. It returns feerates, not direct fees. I only assume that I need to multiply my TX's mass by the fee rate to get an actual fee to use
> * [ ]  In line with the above, the API uses `GetFeeEstimate` and `GetFeeEstimateExperimental` but instead of actual fees, it returns fee rates. Probably should change those names to `GetFeeRateEstimate` (or `GetFeerateEstimate`?)
> 
> ## Other thoughts
> * "feerate" is being used here like it's a single word but it's actually "fee rate". Not something that needs to be changed but it seems weird to me.

The feerate is now defined in multiple places, including the notebook. Usage examples were provided as well. I think GetFeeEstimate is fine, I see it as "the data required for making fee estimates".

mining/src/mempool/model/frontier.rs
```diff
@@ -89,10 +89,11 @@ impl Frontier {
...
     ///        that the probability of collisions is low enough and the sampling process will converge in `O(k log(n))` w.h.p.
     ///     3. It remains to deal with the case where the weight distribution is highly biased. The process implemented below
-    ///        keeps track of the top-weight element. If the distribution is highly biased, this element will be sampled twice
-    ///        with sufficient probability (in constant time), in which case we narrow the sampling space to exclude it. We do
-    ///        this by computing the prefix weight up to this top item (exclusive) and then continue the sampling process over
-    ///        the narrowed space. This process is repeated until acquiring the desired mass.  
+    ///        keeps track of the top-weight element. If the distribution is highly biased, this element will be sampled with
+    ///        sufficient probability (in constant time). Following each sampling collision we search for a consecutive range of
+    ///        top elements which were already sampled and narrow the sampling space to exclude them all . We do this by computing
```
tiram88, line 94: "all."

tiram88 (changes requested):

coderofstuff (approved):

PR #520, opened 2024-08-15
WASM SDK: createTransaction produces "The address is invalid" error

IgorKhomenko:
Having the following code from example https://github.com/kaspanet/rusty-kaspa/blob/master/wasm/examples/nodejs/javascript/transactions/single-transaction-demo.js :

```
public async buildTransaction(
    api: KaspaAPI,
    to: string,
    fromAddress: string,
    amount: string,
    network: KaspaNetwork
  ): Promise<kaspa.Transaction> {
    const { kaspaToSompi, createTransaction, Transaction } = kaspa;

    const { entries: utxos } = await api.rpc.getUtxosByAddresses([fromAddress]);

    if (utxos.length === 0) {
      throw `Send some kaspa to ${to} before proceeding with the demo`;
    }

    const outputs = [
      {
        address: to,
        amount: kaspaToSompi(amount) as bigint,
      },
    ];

    return createTransaction(utxos, outputs, fromAddress, 0n, 0, 1, 1);
 }
```

This code produces "The address is invalid" error as a part of `createTransaction` call.

After deep investigation, I found the `entries` returned from `rpc.getUtxosByAddresses` have diff format from what `UtxoEntryReference.try_cast_from` expects.

This is what is returned after `rpc.getUtxosByAddresses` call:
Screenshot 2024-08-15 at 18 21 33

This PR address this issue

PR #521, opened 2024-08-16
WASM SDK: rpc.submitTransaction produces "missing field `transaction`" error  

IgorKhomenko:
Having the following code from example https://github.com/kaspanet/rusty-kaspa/blob/master/wasm/examples/nodejs/javascript/transactions/single-transaction-demo.js :

This code produces "missing field `transaction`" error as a part of `let result = await rpc.submitTransaction({transaction});` call.

It turned out the issue is with this code, `else` branch:

```
    let request = if let Ok(transaction) = Transaction::try_owned_from(&transaction) {
        SubmitTransactionRequest {
            transaction : transaction.into(),
            allow_orphan,
        }
    } else {
      from_value(transaction)? 
    };
```
means it return a transaction object instead of `SubmitTransactionRequest` object.
Once this is fixed, there were some more incompatibility errors, that are now addressed with this PR

Related to https://github.com/kaspanet/rusty-kaspa/pull/520

### DIFF
--- a/consensus/client/src/transaction.rs
+++ b/consensus/client/src/transaction.rs
@@ -214,12 +214,12 @@ impl Transaction {
         self.inner().version = v;
     }
 
-    #[wasm_bindgen(getter, js_name = lock_time)]
+    #[wasm_bindgen(getter, js_name = lockTime)]
     pub fn get_lock_time(&self) -> u64 {
         self.inner().lock_time
     }
 
-    #[wasm_bindgen(setter, js_name = lock_time)]
+    #[wasm_bindgen(setter, js_name = lockTime)]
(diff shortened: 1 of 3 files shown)


aspect:
Can you please review the `omega` branch and make a corresponding PR in aspectron/omega-dev -  https://github.com/aspectron/rusty-kaspa/tree/omega-dev (if you see that it is still required).  The kaspanet/master is outdated (+ PRs should go against `dev`). Please check on Discord#development channel for any updates related to WASM SDK.

IgorKhomenko:
Still face it in `omega` branch https://github.com/aspectron/rusty-kaspa/tree/omega. Gonna move the PR towards `omega-dev`

PR #523, opened 2024-08-18
feat: implement next_block_template_feerate

biryukovmaxim:


mining/src/manager.rs
```diff
@@ -210,24 +210,66 @@ impl MiningManager {
...
+                &[0u8; 32],
+            ));
+            let miner_data: MinerData = MinerData::new(script_public_key, vec![]);
+
+            let BlockTemplate { block: kaspa_consensus_core::block::MutableBlock { transactions, .. }, calculated_fees, .. } =
+                self.get_block_template(consensus, &miner_data)?;
+            // don't consider coinbase tx
+            if transactions.len() < 2 {
+                return Ok(resp);
+            }
```
michaelsutton, line 249: wdyt about failing gracefully if fees len is not as expected? i.e., 

```rust
if transactions.len() - 1 != calculated_fees.len() {
       log warn...
       return Ok(resp);
}
```

mining/src/manager.rs
```diff
@@ -210,24 +210,66 @@ impl MiningManager {
...
+            let mut fees_and_masses = transactions
+                .iter()
+                // skip coinbase tx
+                .skip(1)
+                .map(Transaction::mass)
+                .zip(calculated_fees)
```
michaelsutton, line 255: you zip `fee` to `mass` (`mass` first), but then use `(fee, mass)` tuples..

mining/src/manager.rs
```diff
@@ -210,24 +210,66 @@ impl MiningManager {
...
+                .collect_vec();
+            fees_and_masses
+                .sort_unstable_by(|(lhs_fee, lhs_mass), (rhs_fee, rhs_mass)| lhs_fee.mul(rhs_mass).cmp(&rhs_fee.mul(lhs_mass)));
```
michaelsutton, line 258: Can you add a comment explaining that mul is faster than div and this way we only div the min/max/median?

consensus/core/src/block.rs
```diff
@@ -105,6 +105,8 @@ pub struct BlockTemplate {
     pub selected_parent_timestamp: u64,
     pub selected_parent_daa_score: u64,
     pub selected_parent_hash: Hash,
+    /// length one less than txs length due to lack of coinbase transaction
```
michaelsutton, line 108:
```suggestion
    /// Expected length is one less than txs length due to lack of coinbase transaction
```

mining/src/manager.rs
michaelsutton: maybe move this piece of logic (i.e., the logic of getting min/max/median from a vec of txs and a vec of fees) to a separate method and add a short test verifying it

michaelsutton (changes requested):

mining/src/manager.rs
biryukovmaxim: [8f5f802](https://github.com/kaspanet/rusty-kaspa/pull/523/commits/8f5f802fdaee63737de3435be5dc81c8c66ce84a)

mining/src/manager.rs
biryukovmaxim: [8f5f802](https://github.com/kaspanet/rusty-kaspa/pull/523/commits/8f5f802fdaee63737de3435be5dc81c8c66ce84a)

mining/src/manager.rs
biryukovmaxim: [8f5f802](https://github.com/kaspanet/rusty-kaspa/pull/523/commits/8f5f802fdaee63737de3435be5dc81c8c66ce84a)

michaelsutton (approved):

PR #524, opened 2024-08-20
Mempool tweaks

someone235:
This PR:

* Increases mempool expiry time to 24h
* Increases orphan tx limit to 500
* Adds a compute-mass limit to the mempool (as a proxy to mempool memory size) and changing the eviction mechanism to prefer transactions with lower fee rate.

### DIFF
--- a/mining/src/manager_tests.rs
+++ b/mining/src/manager_tests.rs
@@ -1121,14 +1121,14 @@ mod tests {
         let consensus = Arc::new(ConsensusMock::new());
         let counters = Arc::new(MiningCounters::default());
         const TX_COUNT: u32 = 10;
-        const MASS_LIMIT: u64 = 10_000;
-        let mut config = Config::build_default(TARGET_TIME_PER_BLOCK, false, MASS_LIMIT);
-        config.mempool_compute_mass_limit = MASS_LIMIT;
+        const SIZE_LIMIT: u64 = 10_000;
+        let mut config = Config::build_default(TARGET_TIME_PER_BLOCK, false, MAX_BLOCK_MASS);
+        config.mempool_size_limit = SIZE_LIMIT;
         let mining_manager = MiningManager::with_config(config, None, counters);
         let txs = (0..TX_COUNT)
             .map(|i| {
                 let mut tx = create_transaction_with_utxo_entry(i, 0);
-                tx.calculated_compute_mass = Some(1000);
+                tx.estimated_size = Some(1000);
                 tx
             })
             .collect_vec();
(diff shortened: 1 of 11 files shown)


mining/src/mempool/validate_and_insert_transaction.rs
```diff
@@ -73,9 +74,30 @@ impl Mempool {
...
-        })?;
+        let transaction_size = transaction.tx.estimate_mem_bytes();
+        let txs_to_remove = self.transaction_pool.limit_transaction_count(&transaction)?;
+        for x in txs_to_remove.iter() {
+            self.remove_transaction(x, true, TxRemovalReason::MakingRoom, format!(" for {}", transaction_id).as_str())?;
+            // self.transaction_pool.limit_transaction_count(&transaction) returns the
+            // smallest prefix of `ready_transactions` (sorted by ascending fee-rate)
+            // that makes enough room for `transaction`, but since each call to `self.remove_transaction`
+            // also removes all transactions dependant on `x` we might already have enough room.
+            if self.transaction_pool.len() < self.config.maximum_transaction_count
```
coderofstuff, line 85: Same question here. Shouldn't this be `self.transaction_pool.len() <= self.config.maximum_transaction_count` ?

mining/src/manager_tests.rs
```diff
@@ -1116,6 +1117,47 @@ mod tests {
         // TODO: extend the test according to the golang scenario
     }
 
+    #[test]
+    fn test_evict() {
```
coderofstuff, line 1121: Please add comments to explain what's happening in this test. Maybe add inline comments.

From reading it, it creates 10 txs and a heavy tx and the expectation is that to accommodate the heavy tx, 6 of the lighter txs are evicted.

mining/src/mempool/model/transactions_pool.rs
```diff
@@ -181,70 +187,49 @@ impl TransactionsPool {
...
-            } else {
-                let mut low_priority_txs = low_priority_txs.collect::<Vec<_>>();
-                if low_priority_txs.len() > trim_size {
-                    low_priority_txs.sort_by(|a, b| a.fee_rate().partial_cmp(&b.fee_rate()).unwrap());
-                    transactions_to_remove.extend_from_slice(&low_priority_txs[0..usize::min(trim_size, low_priority_txs.len())]);
-                } else {
-                    transactions_to_remove = low_priority_txs;
-                }
+        let transaction_size = transaction.tx.estimate_mem_bytes();
+        let mut txs_to_remove = Vec::with_capacity(1);
```
coderofstuff, line 202: Why `Vec::with_capacity(1)`? Add a comment as to why. My guess is that the expected number of transactions to remove is almost always exactly `1`.

mining/src/mempool/model/transactions_pool.rs
```diff
@@ -181,70 +187,49 @@ impl TransactionsPool {
...
     ///
-    /// An error is returned if the mempool is filled with high priority transactions.
-    pub(crate) fn limit_transaction_count(
-        &self,
-        free_slots: usize,
-        transaction: &MutableTransaction,
-    ) -> RuleResult<Vec<TransactionId>> {
-        assert!(free_slots > 0);
+    /// An error is returned if the mempool is filled with high priority transactions, or
+    /// at least one candidate for removal has a higher fee rate than `transaction`.
```
coderofstuff, line 196:
```
at least one candidate for removal has a higher fee rate than `transaction`
```

is probably better stated as

```
there are not enough lower fee rate transactions that can be removed to accommodate `transaction`
```

coderofstuff (changes requested):
Changes look sound, I have a few comments to be fixed.

coderofstuff (approved):

michaelsutton (approved):

PR #527, opened 2024-08-21
A few optimizations related to multi-level relations

michaelsutton:
1. In `ParentsBuilder` optimize the execution path of the common case for high levels
2. Prune relations for all levels below the affiliated proof level for block
3. Keep relations for multi-level parents of pruning point and its anticone  

(each item is in its own commit, by order)

### DIFF
--- a/consensus/src/model/services/reachability.rs
+++ b/consensus/src/model/services/reachability.rs
@@ -1,7 +1,7 @@
 use std::ops::Deref;
 use std::sync::Arc;
 
-use kaspa_consensus_core::blockhash;
+use kaspa_consensus_core::{blockhash, BlockHashSet};
 use parking_lot::RwLock;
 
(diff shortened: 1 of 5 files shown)


consensus/src/model/services/reachability.rs
```diff
@@ -18,6 +18,18 @@ pub trait ReachabilityService {
     fn get_next_chain_ancestor(&self, descendant: Hash, ancestor: Hash) -> Hash;
     fn get_chain_parent(&self, this: Hash) -> Hash;
     fn has_reachability_data(&self, this: Hash) -> bool;
+
+    /// Returns the topological roots of this set, i.e., the minimal subset of blocks which has all
```
someone235, line 22: This won't work for a descending chain

someone235 (changes requested):

someone235 (approved):

someone235 (approved):

PR #528, opened 2024-08-22
get_current_block_color RPC utility

KaffinPX:


consensus/src/consensus/mod.rs
```diff
@@ -504,6 +509,52 @@ impl ConsensusApi for Consensus {
...
+            if visited.insert(child) {
+                queue.push(Reverse(SortableBlock::new(child, block.header.blue_work)));
+            }
+        }
+
+        while let Some(Reverse(current_block)) = queue.pop() {
+            let current_block_hash = current_block.hash;
+            let current_block_data = self.get_ghostdag_data(current_block_hash).unwrap();
+
+            if self.services.reachability_service.is_chain_ancestor_of(current_block_hash, self.get_sink()) {
```
michaelsutton, line 533: Keep sink in a variable out of the loop. 
a. It is more efficient 
b. For consistency. Sink might change and might render a block you already visited to be the new merging chain block.

consensus/src/consensus/mod.rs
```diff
@@ -504,6 +509,52 @@ impl ConsensusApi for Consensus {
...
+            let block = self.get_block(child).unwrap();
+
+            if visited.insert(child) {
+                queue.push(Reverse(SortableBlock::new(child, block.header.blue_work)));
+            }
+        }
+
+        while let Some(Reverse(current_block)) = queue.pop() {
+            let current_block_hash = current_block.hash;
+            let current_block_data = self.get_ghostdag_data(current_block_hash).unwrap();
```
michaelsutton, line 531: Efficiency: get ghostdag data only if indeed `is_chain_ancestor_of` returns true

consensus/src/consensus/mod.rs
```diff
@@ -504,6 +509,52 @@ impl ConsensusApi for Consensus {
         self.headers_store.get_timestamp(self.get_sink()).unwrap()
     }
 
+    fn get_current_block_color(&self, hash: Hash) -> Option<bool> {
```
michaelsutton, line 512: 1. prune guard is needed in order to make sure data is consistent and does not get pruned during the loop
2. pre condition: check that hash exists (see other methods in this file)

```suggestion
    fn get_current_block_color(&self, hash: Hash) -> Option<bool> {
        let _guard = self.pruning_lock.blocking_read();
        self.validate_block_exists(hash)?;
```

consensus/src/consensus/mod.rs
```diff
@@ -504,6 +509,52 @@ impl ConsensusApi for Consensus {
...
+        let mut queue: BinaryHeap<Reverse<SortableBlock>> = BinaryHeap::new();
+        let mut visited = BlockHashSet::new();
+
+        let initial_childrens = match self.get_block_children(hash) {
+            Some(children) => children,
+            None => return None,
+        };
+
+        for child in initial_childrens {
+            let block = self.get_block(child).unwrap();
```
michaelsutton, line 522: Ghostdag store has compact data access 

```suggestion
            let blue_work = self.ghostdag_primary_store.get_blue_work(child).unwrap();
```

The same applies below within the main loop

consensus/src/consensus/mod.rs
```diff
@@ -504,6 +509,52 @@ impl ConsensusApi for Consensus {
...
+
+            if self.services.reachability_service.is_chain_ancestor_of(current_block_hash, self.get_sink()) {
+                if current_block_data.mergeset_blues.contains(&hash) {
+                    return Some(true);
+                } else if current_block_data.mergeset_reds.contains(&hash) {
+                    return Some(false);
+                }
+            }
+
+            let childrens = match self.get_block_children(current_block_hash) {
```
michaelsutton, line 541: childrens -> children

consensus/src/consensus/mod.rs
```diff
@@ -504,6 +509,52 @@ impl ConsensusApi for Consensus {
...
+                Some(childrens) => childrens,
+                None => continue,
+            };
+
+            for child in childrens {
+                let block = self.get_block(child).unwrap();
```
michaelsutton, line 547:
```suggestion
                let blue_work = self.ghostdag_primary_store.get_blue_work(child).unwrap();
```

rpc/grpc/core/proto/rpc.proto
```diff
@@ -922,3 +922,13 @@ message GetFeeEstimateExperimentalResponseMessage {
...
+
+message GetCurrentBlockColorRequestMessage {
+  string hash = 1;
+}
+
+message GetCurrentBlockColorResponseMessage {
+  bool blue = 1;
+
+  RPCError error = 1000;
+}
```
michaelsutton, line 934: there should be a new empty line at end of the file

rpc/service/src/service.rs
```diff
@@ -373,6 +373,15 @@ NOTE: This error usually indicates an RPC conversion error between the node and
         })
     }
 
+    async fn get_current_block_color_call(&self, _request: GetCurrentBlockColorRequest) -> RpcResult<GetCurrentBlockColorResponse> {
+        let session = self.consensus_manager.consensus().session().await;
```
michaelsutton, line 377: Since consensus will apply the prune guard internally (see other comments), you can use an `unguarded_session` here. You only need a guarded session when you need consistency *between* consensus calls, otherwise it's consensus's responsibility

michaelsutton (changes requested):

aspect (approved):

aspect (dismissed):

michaelsutton (dismissed):

michaelsutton (approved):

aspect (approved):

PR #530, opened 2024-08-26
Query all DNS seeders if missing many connections 

michaelsutton:


components/connectionmanager/src/lib.rs
```diff
@@ -227,12 +227,14 @@ impl ConnectionManager {
...
-                cmgr.dns_seed(missing_connections); //TODO: Consider putting a number higher than `missing_connections`.
-            })
-            .await;
+            if missing_connections > self.outbound_target / 2 {
+                // If we are missing more than half of our target, query all in parallel.
+                // This will always be the case on new node start-up and is the most resilient strategy in such a case.
+                self.dns_seed_many(self.dns_seeders.len()).await;
+            } else {
+                // Try to obtain at least twice the number of missing connections
+                self.dns_seed_sequential(2 * missing_connections).await;
```
biryukovmaxim, line 236: naming is not clear, both variants adds many connections. 
my suggestion is `dns_seed_many` -> `dns_seed_in_parallel`

components/connectionmanager/src/lib.rs
```diff
@@ -279,6 +271,42 @@ impl ConnectionManager {
...
+    }
+
+    /// Query a single DNS seeder and add the obtained addresses to the address manager.
+    ///
+    /// DNS lookup is a blocking i/o operation so this function is assumed to be called
+    /// from a blocking execution context.
+    fn dns_seed_single(self: &Arc<Self>, seeder: &str) -> usize {
+        info!("Querying DNS seeder {}", seeder);
+        // Since the DNS lookup protocol doesn't come with a port, we must assume that the default port is used.
+        let addrs = match (seeder, self.default_port).to_socket_addrs() {
```
biryukovmaxim, line 292: what about async call her with: 
https://docs.rs/tokio/latest/tokio/net/fn.lookup_host.html https://docs.rs/tokio/latest/tokio/net/trait.ToSocketAddrs.html


maybe unrelated change, does it worth TODO, wdyt?

michaelsutton: Cool. I think we were looking for something like this. I'll give it a try now

biryukovmaxim: I guess it doesn't work the way I thought about

michaelsutton: You tried?

components/connectionmanager/src/lib.rs
michaelsutton: They are also different in the way they decide to break (min addresses or num seeders), I think parallel vs. seq is misleading. Any other better suggestion?

biryukovmaxim: `_by_seeders[_count]` and `_by_connections[_count]`

components/connectionmanager/src/lib.rs
biryukovmaxim: In rust playground. I only got errors.

That's what we can use:
https://docs.rs/trust-dns-resolver/0.23.2/trust_dns_resolver/struct.Resolver.html#method.lookup_ip

michaelsutton: That's for another day

components/connectionmanager/src/lib.rs
michaelsutton: Done

PR #535, opened 2024-09-01
Transaction pool -- fix rare cases where byte size tracking was inaccurate

michaelsutton:
**_Problem_**: when revalidating high priority txs, the revalidated tx was updated within the inner tx pool.  If the revalidation resulted in missing inputs, this update would result in a change in the size of the updated tx which wasn't tracked by `transaction_pool.estimated_size`. The first 2 commits in this PR contain 2 orthogonal fixes to this problem, which are both correct independently. 

**_Fix 1_**: remove crate level mut access to inner tx structure, and perform correct byte tracking whenever the inner collection is accessed.

**_Fix 2_**: in `revalidate_high_priority_transactions`, update the revalidated tx only if it validated correctly, otherwise we remove it anyway so there's no need to update. Technically, this fixes the problem as well, since fully populated txs never change in byte size. This change is also the right way to go conceptually -- the tx pool should never contain a tx which is not fully populated, even for a brief moment.

### DIFF
--- a/mining/src/manager.rs
+++ b/mining/src/manager.rs
@@ -738,15 +738,7 @@ impl MiningManager {
                             .entries
                             .iter()
                             .zip(transaction.tx.inputs.iter())
-                            .filter_map(
-                                |(entry, input)| {
-                                    if entry.is_none() {
-                                        Some(input.previous_outpoint.transaction_id)
-                                    } else {
-                                        None
-                                    }
-                                },
-                            )
+                            .filter_map(|(entry, input)| entry.is_none().then_some(input.previous_outpoint.transaction_id))
                             .collect::<Vec<_>>();
 
                         // A transaction may have missing outpoints for legitimate reasons related to concurrency, like a race condition between
@@ -762,15 +754,14 @@ impl MiningManager {
                         };
 
                         // This call cleanly removes the invalid transaction.
-                        let result = mempool.remove_transaction(
-                            &transaction_id,
-                            false,
-                            TxRemovalReason::RevalidationWithMissingOutpoints,
-                            extra_info.as_str(),
-                        );
-                        if let Err(err) = result {
-                            warn!("Failed to remove transaction {} from mempool: {}", transaction_id, err);
-                        }
+                        _ = mempool
+                            .remove_transaction(
+                                &transaction_id,
+                                false,
+                                TxRemovalReason::RevalidationWithMissingOutpoints,
+                                extra_info.as_str(),
+                            )
+                            .inspect_err(|err| warn!("Failed to remove transaction {} from mempool: {}", transaction_id, err));
                         missing_outpoint += 1;
                     }
                     Err(err) => {
@@ -782,10 +773,9 @@ impl MiningManager {
                             transaction_id, err
                         );
                         // This call cleanly removes the invalid transaction and its redeemers.
-                        let result = mempool.remove_transaction(&transaction_id, true, TxRemovalReason::Muted, "");
-                        if let Err(err) = result {
-                            warn!("Failed to remove transaction {} from mempool: {}", transaction_id, err);
-                        }
+                        _ = mempool
+                            .remove_transaction(&transaction_id, true, TxRemovalReason::Muted, "")
+                            .inspect_err(|err| warn!("Failed to remove transaction {} from mempool: {}", transaction_id, err));
                         invalid += 1;
                     }
                 }
--- a/mining/src/manager.rs
+++ b/mining/src/manager.rs
@@ -799,6 +799,12 @@ impl MiningManager {
                     missing_outpoint,
                     invalid,
                 );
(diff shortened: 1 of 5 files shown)


mining/src/manager.rs
```diff
@@ -717,10 +717,10 @@ impl MiningManager {
...
             for (transaction, validation_result) in chunk {
                 let transaction_id = transaction.id();
-                // Only consider transactions still being in the mempool since during the validation some might have been removed.
-                if mempool.update_revalidated_transaction(transaction) {
-                    match validation_result {
-                        Ok(()) => {
+                match validation_result {
+                    Ok(()) => {
+                        // Only consider transactions still being in the mempool since during the validation some might have been removed.
+                        if mempool.update_revalidated_transaction(transaction) {
```
biryukovmaxim, line 723: style suggestion
```rust
                    // Only consider transactions still being in the mempool since during the validation some might have been removed.
                    Ok(()) if mempool.update_revalidated_transaction(transaction) => {
                        // A following transaction should not remove this one from the pool since we process in a topological order.
                        // Still, considering the (very unlikely) scenario of two high priority txs sandwiching a low one, where
                        // in this case topological order is not guaranteed since we only considered chained dependencies of
                        // high-priority transactions, we might wrongfully return as valid the id of a removed transaction.
                        // However, as only consequence, said transaction would then be advertised to registered peers and not be
                        // provided upon request.
                        valid_ids.push(transaction_id);
                        valid += 1;
                    }
                    Ok(()) => {
                        other += 1;
                    }
```

mining/src/manager.rs
```diff
@@ -729,66 +729,65 @@ impl MiningManager {
...
+                            .zip(transaction.tx.inputs.iter())
+                            .filter_map(
+                                |(entry, input)| {
+                                    if entry.is_none() {
+                                        Some(input.previous_outpoint.transaction_id)
+                                    } else {
+                                        None
+                                    }
+                                },
+                            )
```
biryukovmaxim, line 749:
```suggestion
                              .filter_map(|(entry, input)| entry.is_none().then_some(input.previous_outpoint.transaction_id))
```

mining/src/manager.rs
```diff
@@ -729,66 +729,65 @@ impl MiningManager {
...
+                        // This call cleanly removes the invalid transaction.
+                        let result = mempool.remove_transaction(
+                            &transaction_id,
+                            false,
+                            TxRemovalReason::RevalidationWithMissingOutpoints,
+                            extra_info.as_str(),
+                        );
+                        if let Err(err) = result {
+                            warn!("Failed to remove transaction {} from mempool: {}", transaction_id, err);
                         }
```
biryukovmaxim, line 773:
```suggestion
_ = mempool
                            .remove_transaction(
                                &transaction_id,
                                false,
                                TxRemovalReason::RevalidationWithMissingOutpoints,
                                extra_info.as_str(),
                            )
                            .inspect_err(|err| warn!("Failed to remove transaction {} from mempool: {}", transaction_id, err));
```

mining/src/manager.rs
```diff
@@ -729,66 +729,65 @@ impl MiningManager {
...
+                        // The failed revalidation is simply logged and the process continues.
+                        warn!(
+                            "Removing high priority transaction {0} and its redeemers, it failed revalidation with {1}",
+                            transaction_id, err
+                        );
+                        // This call cleanly removes the invalid transaction and its redeemers.
+                        let result = mempool.remove_transaction(&transaction_id, true, TxRemovalReason::Muted, "");
+                        if let Err(err) = result {
+                            warn!("Failed to remove transaction {} from mempool: {}", transaction_id, err);
                         }
```
biryukovmaxim, line 788:
```suggestion
                        _ = mempool
                            .remove_transaction(&transaction_id, true, TxRemovalReason::Muted, "")
                            .inspect_err(|err| warn!("Failed to remove transaction {} from mempool: {}", transaction_id, err));
```

biryukovmaxim (approved):
Just some style-related suggestions

mining/src/manager.rs
```diff
@@ -729,66 +729,65 @@ impl MiningManager {
                             // provided upon request.
                             valid_ids.push(transaction_id);
                             valid += 1;
+                        } else {
+                            other += 1;
```
someone235, line 733: This field is not used

someone235 (changes requested):

tiram88 (dismissed):

mining/src/manager.rs
michaelsutton: This can't be done because it consumes `transaction` which is used in the err branches

mining/src/manager.rs
michaelsutton: Added a debug log

someone235 (approved):

tiram88 (approved):

PR #537, opened 2024-09-01
Feerate estimator outliers: capture block slot by the specific mass of the outlier transaction

michaelsutton:
The feerate estimator deals with outliers (extreme high feerate transactions which highly deviate from the average) by considering them as if they capture block space, and then estimating the feerate required for being included in the remaining block space. See 
the [python notebook](https://github.com/kaspanet/rusty-kaspa/blob/dev/mining/src/feerate/fee_estimation.ipynb) for full details. 

Up until now, the captured block space for each outlier was the _average_ tx mass. However this is inaccurate if the outliers happen to have mass << average mass. In which case they capture too much of the block space and the predicted feerate estimations turn out to be much higher than needed. 

In this PR we fix this by capturing block space by the specific mass of the outlier transaction. The PR also includes a test for verifying this scenario is indeed addressed by the changes.

### DIFF
--- a/mining/src/mempool/model/frontier.rs
+++ b/mining/src/mempool/model/frontier.rs
@@ -220,11 +220,13 @@ impl Frontier {
         let mut estimator = FeerateEstimator::new(self.total_weight(), inclusion_interval);
 
         // Search for better estimators by possibly removing extremely high outliers
-        let mut down_iter = self.search_tree.descending_iter().skip(1);
-        loop {
-            // Update values for the next iteration. In order to remove the outlier from the
-            // total weight, we must compensate by capturing a block slot.
-            mass_per_block -= average_transaction_mass;
+        let mut down_iter = self.search_tree.descending_iter().peekable();
+        while let Some(current) = down_iter.next() {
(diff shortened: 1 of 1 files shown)


coderofstuff (approved):

mining/src/mempool/model/frontier.rs
```diff
@@ -220,11 +220,13 @@ impl Frontier {
...
-            // total weight, we must compensate by capturing a block slot.
-            mass_per_block -= average_transaction_mass;
+        let mut down_iter = self.search_tree.descending_iter().peekable();
+        while let Some(current) = down_iter.next() {
+            // Update values for the coming iteration. In order to remove the outlier from the
+            // total weight, we must compensate by capturing a block slot. Note we capture the
+            // slot with correspondence to the outlier actual mass. This is important in cases
+            // where the high-feerate txs have mass which deviates from the average.
+            mass_per_block -= current.mass as f64;
             if mass_per_block <= average_transaction_mass {
```
coderofstuff, line 230: Should the `average_transaction_mass` also be updated?

Issue #536, opened 2024-09-01
wasm can't get address from ECDSA public key


drunkirishcoder:
**Describe the bug**
Unable to get the ECDSA address from a public key. works with a private key, but does not work if you direct get it by using a public key.

**To Reproduce**

```
const publicKey = new PublicKey('...'); 
publicKey.toAddressECDSA(NetworkType.Mainnet).toString()
```

the code will panic

problem appears to be [here](https://github.com/kaspanet/rusty-kaspa/blob/6a56461ff1249e11e3124090253ca0bf128f7189/wallet/keys/src/publickey.rs#L84). instead using `xonly_public_key`, it should try use `self.public_key`.

aspect:
@drunkirishcoder thanks for reporting this. The next drink is on me!

- Fixed in https://github.com/kaspanet/rusty-kaspa/pull/538/commits/8eb7500f9517f456df7d5c44673b82633ff35e31
- Merged into https://github.com/kaspanet/rusty-kaspa/tree/dev
- WASM SDK artifact can be obtained from https://github.com/kaspanet/rusty-kaspa/actions/runs/10672898810

Cheers!

drunkirishcoder:
thank you

PR #538, opened 2024-09-02
Refactoring of mass calculator utilities

aspect:
This PR contains refactoring and various improvements:
- Relocate consensus `MassCalculator` from `kaspa_consensus` to `kaspa_consensus_core` (`kaspa_consensus` is not accessible from broader framework due to dependencies, so this makes the consensus `MassCalculator` accessible from all crates)
- Rename client `MassCalculator` functions with suffixes denoting transaction types (`client` or `consensus`) i.e. `calculate_something_for_transaction()` -> `calculate_something_for_consensus_transaction()`
- Change client `MassCalculator` function names to `unsigned_transaction`
- Introduce dedicated functions for `consensus` and `client` transaction mass estimates
- Propagate `mass` field across all transaction data structures
- Add missing `mass` to `ITransaction` TypeScript interface (so that it is visible when it comes in via `RpcTransaction` data type)
- Fix incorrect use of `XPublicOnlyKey` in ECDSA address creation (https://github.com/kaspanet/rusty-kaspa/issues/536).


consensus/core/src/mass/mod.rs
```diff
@@ -1,8 +1,14 @@
 use crate::{
+    config::params::Params,
     subnets::SUBNETWORK_ID_SIZE,
-    tx::{Transaction, TransactionInput, TransactionOutput},
+    tx::{Transaction, TransactionInput, TransactionOutput, VerifiableTransaction},
 };
 use kaspa_hashes::HASH_SIZE;
+// pub use crate::mass::Kip9Version;
```
michaelsutton, line 7: Remove the commented code?

consensus/core/src/mass/mod.rs
```diff
@@ -66,3 +72,253 @@ pub fn transaction_output_estimated_serialized_size(output: &TransactionOutput)
     size += output.script_public_key.script().len() as u64;
     size
 }
+
+// TODO (aspect) - review and potentially merge this with the new MassCalculator currently located in the wallet core
```
michaelsutton, line 76: Can this TODO be updated or removed?

consensus/src/processes/mass.rs
```diff
@@ -1,246 +1 @@
-pub use kaspa_consensus_core::mass::Kip9Version;
```
michaelsutton, line 1: Remove this file?

rpc/core/src/model/tx.rs
```diff
@@ -359,7 +359,7 @@ impl Deserializer for RpcTransactionVerboseData {
         let block_hash = load!(RpcHash, reader)?;
         let block_time = load!(u64, reader)?;
 
-        Ok(Self { transaction_id, hash, mass, block_hash, block_time })
+        Ok(Self { transaction_id, hash, compute_mass: mass, block_hash, block_time })
```
michaelsutton, line 362: Rename `mass` to `compute_mass`

michaelsutton:
I couldn't find this line in the changes https://github.com/kaspanet/rusty-kaspa/blob/dev/rpc/core/src/wasm/convert.rs#L70

consensus/core/src/mass/mod.rs
aspect: done

consensus/core/src/mass/mod.rs
aspect: changed the comment noting that consensus `MassCalculator` operates on signed transactions and to use `kaspa_wallet_core::tx::mass::MassCalculator` if the developer needs to calculate mass on unsigned transactions.

consensus/src/processes/mass.rs
aspect: removed

rpc/core/src/model/tx.rs
aspect: strange... I already did this...

coderofstuff (approved):

rpc/core/src/model/block.rs
```diff
@@ -158,6 +158,21 @@ cfg_if::cfg_if! {
             mergeSetRedsHashes: HexString[];
             isChainBlock: boolean;
         }
+
+        /**
+         * Interface defining the structure of a raw block.
+         * 
+         * Raw block is a structure used by GetBlockTemplate and SubmitBlock RPCs
+         * and differs from `IBlock` in that it does not verbose data and carries
```
michaelsutton, line 166: does not *include* verbose data

rpc/core/src/model/block.rs
```diff
@@ -158,6 +158,21 @@ cfg_if::cfg_if! {
...
+         * `IRawHeader` that does not include a PoW hash.
```
michaelsutton, line 167: a PoW hash -> a cached block hash

rpc/core/src/model/block.rs
aspect: fixed

rpc/core/src/model/block.rs
aspect: fixed

consensus/core/src/mass/mod.rs
```diff
@@ -66,3 +67,271 @@ pub fn transaction_output_estimated_serialized_size(output: &TransactionOutput)
...
+        version: Kip9Version,
+    ) -> Option<u64> {
+        match version {
+            Kip9Version::Alpha => self
+                .calc_tx_storage_mass(tx, version)
+                .and_then(|mass| mass.checked_add(cached_compute_mass.unwrap_or_else(|| self.calc_tx_compute_mass(tx.tx())))),
+            Kip9Version::Beta => self
+                .calc_tx_storage_mass(tx, version)
+                .map(|mass| mass.max(cached_compute_mass.unwrap_or_else(|| self.calc_tx_compute_mass(tx.tx())))),
+        }
```
coderofstuff, line 149: Consider:

```
        let storage_mass_result = self.calc_tx_storage_mass(tx, version);
        let tx_mass = cached_compute_mass.unwrap_or_else(|| self.calc_tx_compute_mass(tx.tx()));
        match version {
            Kip9Version::Alpha => storage_mass_result.and_then(|mass| mass.checked_add(tx_mass)),
            Kip9Version::Beta => storage_mass_result.map(|mass| mass.max(tx_mass)),
        }
```

Also, why is one `.and_then` and the other `.map`?

aspect: `map` transforms the value by applying a transform function on the underlying value if it is `Some`, retaining the transformed value afterwards.  whereas `and_then` pretty much does the same thing, but expects the transform output to be an option as well, allowing the transform fn to also transform the output to `None`.

since using `checked_add` here, which returns `Option` `and_then` is more fitting.  if a `saturating_add` was used, then a `map` would be used.  In this context both are sort of the same. They are just more "fitted" to the data pipeline.

coderofstuff (approved):

PR #541, opened 2024-09-03
Deprecate zigbuild and glibc in favor of static musl binaries for Linux builds.

starkbamse:
**What is this**
This PR replaces current ways to build linux binaries (zigbuild) with a musl build using custom built OpenSSL and CTNG musl toolchains.

**How**
We utilize [ctng](https://crosstool-ng.github.io/) to generate a musl toolchain using a configuration which includes libstdc++, a requirement by RocksDB. This allows us to compile OpenSSL and rusty-kaspa against musl.

**Tests performed**
Ran by two testers in #qa for two days.

**Why do we need it**
We need it because on certain machines (Ubuntu 22.04.1 LTS) when running TN10 we get a segmentation fault with the following backtrace snippet:

```sh
Core was generated by `./kaspad17800141 --testnet --utxoindex --disable-upnp --ram-scale=0.8 --appdir'.
Program terminated with signal SIGSEGV, Segmentation fault.
#0  rocksdb::CoreLocalArray<rocksdb::ConcurrentArena::Shard>::AccessElementAndIndex (this=0x48) at rocksdb/util/core_local.h:73
73  rocksdb/util/core_local.h: No such file or directory.
[Current thread is 1 (LWP 710665)]
warning: Missing auto-load script at offset 0 in section .debug_gdb_scripts
of file /home/starkbamse/coredump-old-rust/kaspad1780.
Use `info auto-load python-scripts [REGEXP]' to list them.
(gdb) bt
#0  rocksdb::CoreLocalArray<rocksdb::ConcurrentArena::Shard>::AccessElementAndIndex (this=0x48) at rocksdb/util/core_local.h:73
#1  rocksdb::ConcurrentArena::Repick (this=0x0) at rocksdb/memory/concurrent_arena.cc:38
#2  0x00005c0fbe499273 in rocksdb::ConcurrentArena::AllocateImpl<rocksdb::ConcurrentArena::AllocateAligned(unsigned long, unsigned long, rocksdb::Logger*)::{lambda()#1}>(unsigned long, bool, rocksdb::ConcurrentArena::AllocateAligned(unsigned long, unsigned long, rocksdb::Logger*)::{lambda()#1} const&) (this=0x471e51638a0, bytes=56, force_arena=<optimized out>, func=...)
```

**Cache updates**
You can trigger a cache update by updating the preset.sh. Both Actions, and the build.sh script keep track of the latest hash and rebuild in the case of a mismatch.

Considerations:
Performance, could be worse if areas where performance is crucial utilize C memory allocators. It is planned to maybe replace the allocator with mimalloc to solve this problem.


starkbamse:
@elichai @supertypo

michaelsutton:
How does one generate the musl .config file? Can you add instructions for that somewhere?

starkbamse:
> How does one generate the musl .config file? Can you add instructions for that somewhere?
You use `ct-ng menuconfig` for that. But yes, I'll write instructions.

musl-toolchain/build.sh
```diff
@@ -0,0 +1,94 @@
...
+  ./Configure no-shared --static linux-x86_64 -fPIC --prefix=$HOME/openssl
+  make depend
+  make
+  make install
+  # Check if OpenSSL was installed successfully
+  ls -la $HOME/openssl
+
+  # Store the current hash of preset.sh after successful build
+  echo "$CURRENT_PRESET_HASH" > "$PRESET_HASH_FILE"
+fi
```
elichai, line 76: This can probably be fixed with: https://github.com/kaspanet/rusty-kaspa/blob/6a56461ff1249e11e3124090253ca0bf128f7189/rpc/wrpc/server/Cargo.toml#L38-L41

musl-toolchain/build.sh
```diff
@@ -0,0 +1,94 @@
...
+    cat .config
+    cat build.log
+  fi
+fi
+
+# Update toolchain variables: C compiler, C++ compiler, linker, and archiver
+export CC=$HOME/x-tools/$CTNG_PRESET/bin/$CTNG_PRESET-gcc
+export CXX=$HOME/x-tools/$CTNG_PRESET/bin/$CTNG_PRESET-g++
+export LD=$HOME/x-tools/$CTNG_PRESET/bin/$CTNG_PRESET-ld
+export AR=$HOME/x-tools/$CTNG_PRESET/bin/$CTNG_PRESET-ar
```
elichai, line 59: Should probably instead/also define:
```
CC_x86_64_unknown_linux_musl
CXX_x86_64_unknown_linux_musl
AR_x86_64_unknown_linux_musl
LD_x86_64_unknown_linux_musl
RANLIB_x86_64_unknown_linux_musl
```

These are for the `cc` crate used in build scripts: https://docs.rs/cc/latest/cc/#external-configuration-via-environment-variables

musl-toolchain/build.sh
```diff
@@ -0,0 +1,94 @@
...
+# Add target
+rustup target add x86_64-unknown-linux-musl
+# Install missing dependencies
+cargo fetch --target x86_64-unknown-linux-musl
+# Patch missing include in librocksdb-sys-0.16.0+8.10.0. Credit: @supertypo
+FILE_PATH=$(find $HOME/.cargo/registry/src/ -path "*/librocksdb-sys-0.16.0+8.10.0/*/offpeak_time_info.h")
+if [ -n "$FILE_PATH" ]; then
+  sed -i '1i #include <cstdint>' "$FILE_PATH"
+else
+  echo "No such file for sed modification."
```
elichai, line 93: Can you expand on this a bit?

starkbamse: For one reason or another, this
```c
#include <cstdint>
```
Is missing in that file for rocksdb. This sed command inserts that line into the file.

starkbamse: @supertypo Englightened me about this, as it seemed to be an upstream issue.

musl-toolchain/build.sh
starkbamse: Yep, doing this.

musl-toolchain/build.sh
starkbamse: Will add after this build.

musl-toolchain/build.sh
michaelsutton: My guess is that it must be included indirectly by some other ifdef or include which musl doesn't trigger

supertypo: I believe cstdint was transitively included before. Anyway, the compiler error message is pretty clear in there being a missing include, so I just patched it in and didn't think more of it.

starkbamse: Yeah, we can also see clearly that it is missing from here: https://github.com/facebook/rocksdb/blob/54d628602706c0c718cf81f87202e0b8f6615faf/options/offpeak_time_info.h

starkbamse: Do note latest master in RocksDB has the include.

starkbamse: https://github.com/facebook/rocksdb/blob/main/options/offpeak_time_info.h

starkbamse: Another reference: https://github.com/facebook/rocksdb/commit/395d24f0fa2d68420cd2b72bdf68ed533b4b83b0

musl-toolchain/build.sh
starkbamse: Added

rpc/wrpc/server/Cargo.toml
```diff
@@ -35,8 +35,12 @@ workflow-log.workspace = true
 workflow-rpc.workspace = true
 workflow-serializer.workspace = true
 
-[target.x86_64-unknown-linux-gnu.dependencies]
-# Adding explicitely the openssl dependency here is needed for a successful build with zigbuild
+# Adding explicitely the openssl dependency here is needed for a successful build with zigbuild and musl
```
KaffinPX, line 38: "zigbuild" seems to be forgotten

starkbamse: What do you mean by forgotten

KaffinPX: I mean not used anymore so doesnt belong to comment -- just a small detail anyway

starkbamse: For the CI this may be true but we figure that this should be kept in the case that one builds with zigbuild anyway. And for clarification, we add the comment explaining why and when the openssl dependency is needed.

michaelsutton (approved):

elichai (approved):
Nice! THis looks great now :)

PR #551, opened 2024-09-09
Optimize window caches for ibd

D-Stacks:
### Background

Currently we Cache block windows during header processing, but rely on the cache in both block body processing, and virtual processing. A cache miss on the window cache results in a large overhead in re-constructing the windows. Specifically this results in significant processing overhead during IBD, where header, and body / virtual processing is not streamlined against each other.

This PR addresses, and patches this issue (thus far, only for tn11) by:

1) caching windows in the virtual processor on new sinks, if it's window is not already cached. 
2) applying an optimization whereby the window_manager will try to not reconstruct a whole window from scratch when the cache is missed, but abort early by merging with an ancestor cache (if found). 
3) applying a lazy load for the past median time cache, which only triggers after a 'quick-and-dirty' pre-check that needs to find a tx where  tx.locktime != 0 first, before validating it in full in `check_block_transactions_in_context`

### Tests and Benching

I currently have created a branch, [here](https://github.com/D-Stacks/rusty-kaspa/tree/25eb1e2) which checks the new and old build method against each other, and logs time spent building the block windows. 

all window builds are passing asserts between old and new.

regarding benching, here is an excerpt, taken during IBD:

```
2024-09-10 00:45:11.731+02:00 [INFO ] IBD: Processed 396000 blocks (26%) last block timestamp: 2024-09-08 16:04:14.000:+0200
2024-09-10 00:45:11.895+02:00 [INFO ] Processed 6280 blocks and 0 headers in the last 10.00s (296652 transactions; 2594 UTXO-validated blocks; 0.00 parents; 0.00 mergeset; 47.24 TPB; 94187.4 mass)
2024-09-10 00:45:21.318+02:00 [INFO ] New Total Time spent in build_block_window: 36.989105s 

            New iterations difficulty: 426446 

            New iterations pmt: 18802163 

            Old Total Time spent in build_block_window: 751.755591s 

            Old iterations difficulty: 150506854 

            Old iterations pmt: 455390110 
```
Note: 

1) most of this performance increase is during block body IBD syncing, where the difficulty window is built, or more  specifically on virtual resolve during IBD. but it also reduces time spent in `past_median_time` window by about 75% (from 16 secs to about 4 secs, on my machine). 

2) it doesn't factor in the parallelization and real-time gain during syncing.


consensus/src/processes/window.rs
```diff
@@ -387,50 +365,96 @@ impl<T: GhostdagStoreReader, U: BlockWindowCacheReader, V: HeaderStoreReader, W:
...
+        sample_rate: u64,
+        cache: Option<&Arc<U>>,
+        ghostdag_data: &GhostdagData,
+        selected_parent_blue_work: BlueWorkType,
+        mut mergeset_non_daa_inserter: impl FnMut(Hash),
+    ) -> Option<Arc<BlockWindowHeap>> {
+        cache.and_then(|cache| {
+            cache.get(&ghostdag_data.selected_parent).map(|selected_parent_window| {
+                let mut heap = Lazy::new(|| BoundedSizeBlockHeap::from_binary_heap(window_size, (*selected_parent_window).clone()));
+                for block in self.sampled_mergeset_iterator(sample_rate, ghostdag_data, selected_parent_blue_work) {
```
coderofstuff, line 425: Can this block be replaced by `self.push_mergeset`?

consensus/src/processes/window.rs
```diff
@@ -465,7 +489,8 @@ impl<T: GhostdagStoreReader, U: BlockWindowCacheReader, V: HeaderStoreReader, W:
     for SampledWindowManager<T, U, V, W>
 {
     fn block_window(&self, ghostdag_data: &GhostdagData, window_type: WindowType) -> Result<Arc<BlockWindowHeap>, RuleError> {
-        self.build_block_window(ghostdag_data, window_type, |_| {})
+        let new_window = self.build_block_window(ghostdag_data, window_type, &|_| {})?;
```
coderofstuff, line 492: Why is the assignment here necessary?

consensus/src/processes/window.rs
```diff
@@ -686,4 +711,24 @@ impl BoundedSizeBlockHeap {
         self.binary_heap.push(r_sortable_block);
         true
     }
+
+    // This method is intended to be used to merge the ancestor heap with the current heap.
+    // It is expected that any ancestor heap will have lower blue work blocks than the current heap.
+    fn merge_ancestor_heap(&mut self, ancestor_heap: &mut BlockWindowHeap) {
+        // because popping has O(log(n)) complexity, we pop the excess amount first, then append the two.
+        let overflow_amount = (self.len() + ancestor_heap.len()) - self.size_bound;
```
coderofstuff, line 719: `self.size_bound` can be greater than `self.len() + ancestor_heap.len()` here so probably better to do:

```
let overflow_amount = (self.len() + ancestor_heap.len()).saturating_sub(self.size_bound);
```

consensus/src/processes/window.rs
```diff
@@ -686,4 +711,24 @@ impl BoundedSizeBlockHeap {
...
+        if overflow_amount > 0 {
```
coderofstuff, line 720: You don't need this `if` condition. `overflow_amount` will always be either 0 or exactly the overflow amount. (Also if it were negative, the above would panic without the saturating sub change).

coderofstuff (changes requested):
Functionality-wise, this works great. I tested this with IBD and flamegraphs and I observed that it removed the hotspots related to PMT window during block IBD.

consensus/src/processes/window.rs
D-Stacks: I think not, because self.push_mergeset takes an initialized heap it would void the lazy init optimization (also this is still original code from tiram, so i didn't actually change anything here).

consensus/src/processes/window.rs
D-Stacks: Was under the impression that `self.size_bound` will never be larger then the then merging with the ancestor heap, plus what was already added to the current heap.. but I am now thinking that this might be a condition if we are close to origin, or genesis. but I guess there is no harm in saturating as it will cover this potential logic and shouldn't do any harm otherwise.

consensus/src/processes/window.rs
D-Stacks: I do find the explicit > 0 check to be more readable, but yes iterating form 0 to 0 should be a no-op i guess, might just add a comment in then.

PR #553, opened 2024-09-10
Bump tonic and prost versions, adapt middlewares

biryukovmaxim:


utils/tower/src/middleware.rs
```diff
@@ -68,40 +65,69 @@ where
...
+pin_project! {
+    pub struct ChannelBody<T> {
+        #[pin]
+        rx: tokio::sync::mpsc::UnboundedReceiver<Frame<T>>,
+    }
+}
+
+impl<T> ChannelBody<T> {
+    pub fn new() -> (tokio::sync::mpsc::UnboundedSender<Frame<T>>, Self) {
+        let (tx, rx) = tokio::sync::mpsc::unbounded_channel();
```
michaelsutton, line 84: Can you explain the possible implications of using an unbounded channel here?

biryukovmaxim: In the original implementation it was bounded channel with 0-sized buffer, meaning it was a blocking operation. It had inconsistent api between hyper channel and tokio::mpsc channel, where 0 is invalid value. I believe it should be bounded channel of size equal to 1.


Answering the question: unbounded channel can lead to bugger memory allocations if we don't have enough tokio workers to read from the channel.

I will change it to bounded(1), wdyt?

michaelsutton (approved):

Issue #417, opened 2024-01-27
Windows: Missing VC runtime issue


coderofstuff:
**Impact:** All windows users without VCRedist installed

Users on windows encounter this upon running kaspad

![Image](https://github.com/kaspanet/rusty-kaspa/assets/114628839/8b312a71-d576-4c37-b288-5bde8a9fa876)

Workaround for now is to install VCRedist following the doc at: https://learn.microsoft.com/en-us/cpp/windows/latest-supported-vc-redist?view=msvc-170#visual-studio-2015-2017-2019-and-2022

Possible solutions:
- Update the GitHub action that builds kaspad in windows to ensure it builds a fully standalone app
- Update README.md to include the VCRedist installation step

PR #559, opened 2024-09-13
Release builds will use `lto = false` instead of thin

x100111010:
* breaks x86_64-pc-windows-gnullvm target on Windows as with llvm optimization it is impossible to link clang libc++ statically. The gnullvm target is needed as it will link ucrt instead of msvc and allow running it without VC runtime. In theory it should make a fix possible for: kaspanet/rusty-kaspa#417

* occasionally lead to node crashes in futures crate, especially when mixed with unsafe and C/C++ code. It is hard to reproduce, but at least with the from scratch reproduces the crash relatively often.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -326,7 +326,6 @@ workflow-wasm = { version = "0.18.0" }
 # nw-sys = { path = "../nw-sys" }
 
 [profile.release]
-lto = "thin"
 strip = true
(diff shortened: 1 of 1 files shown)


x100111010:
Any comment?

freshair18:
A reference discussion here:

https://t.me/kasparnd/1/5544

PR #570, opened 2024-09-26
Documentation updates

aspect:
This PR includes documentation updates to RK modules (wallet, consensus client, RPC/wRPC).

Also adding CI pass for `cargo doc` (as it appears we have not been running this in CI) and fixing various documentation link issues.


wallet/keys/src/prelude.rs
```diff
@@ -1,3 +1,7 @@
+//!
+//! Glob imports for the `kaspa_wallet_keys` crate.
```
coderofstuff, line 2: What is `Glob`?

consensus/client/src/lib.rs
```diff
@@ -1,3 +1,17 @@
+//!
+//! # Client-side consensus primitives.
+//!
+//! This crate offers client-side primitives mirroring the consensus layer of the Kaspa p2p node.
```
coderofstuff, line 4: nit: `p2p` -> `P2P`

consensus/core/src/config/bps.rs
```diff
@@ -33,7 +33,7 @@ impl<const BPS: u64> Bps<BPS> {
     }
 
     /// Returns the GHOSTDAG K value which was pre-computed for this BPS
-    /// (see [`calculate_ghostdag_k`] and [`gen_ghostdag_table`] for the full calculation)
+    /// (see [`calculate_ghostdag_k`] and `gen_ghostdag_table` for the full calculation)
```
coderofstuff, line 36: What's the logic for which text is surrounded by square brackets (`[]`)?

mining/src/feerate/fee_estimation.ipynb
coderofstuff, line 1: What were the changes done to this file? If it wasn't intentional, please revert this change.

rpc/core/src/lib.rs
```diff
@@ -9,6 +22,7 @@ pub mod notify;
 pub mod wasm;
 
 pub mod prelude {
+    //! Glob imports of the most commonly used types and traits in this crate.
```
coderofstuff, line 25: `Glob` appears here again

wallet/core/src/lib.rs
```diff
@@ -28,15 +33,28 @@
...
-//! For JavaScript / TypeScript environments, there are two
+//! WASM32 SDK documentation is available at:
+//! <https://kaspa.aspectron.org/docs/>
+//!
+//! For NodeJS JavaScript and TypeScript environments, there are two
 //! available NPM modules:
 //! - <https://www.npmjs.com/package/kaspa>
 //! - <https://www.npmjs.com/package/kaspa-wasm>
 //!
+//! NOTE: for security reasons (to mitigate potential upstream vendor
```
coderofstuff, line 51: The recommendation here is to **build** from source or **download** from pre-built redistributables as the mitigation. What is the insecure alternative this is trying to warn against?

wallet/core/src/utxo/processor.rs
```diff
@@ -604,9 +604,7 @@ impl UtxoProcessor {
             match kind {
                 MetricsUpdateKind::WalletMetrics => {
                     let mempool_size = snapshot.get(&Metric::NetworkMempoolSize) as u64;
-                    let node_peers = snapshot.get(&Metric::NodeActivePeers) as u32;
-                    let network_tps = snapshot.get(&Metric::NetworkTransactionsPerSecond);
-                    let metrics = MetricsUpdate::WalletMetrics { mempool_size, node_peers, network_tps };
+                    let metrics = MetricsUpdate::WalletMetrics { mempool_size };
```
coderofstuff, line 607: Was the removal of the node_peers and network_tps here intentional? And why?

wallet/core/src/wallet/mod.rs
```diff
@@ -59,6 +67,7 @@ pub struct MultisigWalletFileV0<'a, T: AsRef<[u8]>> {
     pub encrypted_mnemonics: Vec<EncryptedMnemonic<T>>,
     pub xpublic_keys: Vec<&'a str>, // includes pub keys from encrypted
     pub required_signatures: u16,
+    #[allow(dead_code)]
```
coderofstuff, line 70: Why is this necessary here?

coderofstuff (changes requested):
I just have a bunch of questions here first

consensus/core/src/config/bps.rs
KaffinPX: Typescript Type/Interface forwarding

consensus/client/src/lib.rs
aspect: I use p2p everywhere.

consensus/core/src/config/bps.rs
aspect: This is a RustDoc syntax that creates intra-doc links to identifiers.  Identifiers must be visible within the Rust scope of the documentation or referred to via absolute paths (just like you would do when performing regular imports).

In this specific case, `gen_ghostdag_table` can not be referenced to as it is not in the scope accessible by this documentation block (as this is a reference to a unit test function that does not exist during normal compilation).

mining/src/feerate/fee_estimation.ipynb
aspect: No reversal necessary - changes were _intentional and are correct_.  This diff fixes multiple spelling mistakes in the file.  I don't know why VSC has re-arranged related image blocks, if you look at them, they haven't changed.

wallet/core/src/lib.rs
aspect: This note is in the NPM module section...  should we rephrase with more focus on NPM?   If you look at this in the actual docs (i.e. clone the repo and run `cd wallet/core && cargo doc --open`), each section of this documentation will appear strictly separated due to H1 headings, so it is much more apparent that this note applies to the NPM block.  I don't think this should be changed.

wallet/core/src/utxo/processor.rs
aspect: cleanup. this code is not currently active and the APIs that use similar code in KNG, only utilize mempool_size.

This change is a victim of documentation pass (couldn't document what shouldn't be there).

wallet/core/src/wallet/mod.rs
aspect: Removed.

I accidentally left this there as a part of the effort to remove `pub` from various internal data structures, which in this case didn't work out as these data structures were used cross-crate in the workspace.  We need `pub (workspace)` :)

rpc/core/src/lib.rs
aspect: Glob is sometimes used as a reference to wildcards.  

Replaced wording with "Re-export".

wallet/keys/src/prelude.rs
aspect: Replaced wording with "Re-export".

coderofstuff (approved):

PR #575, opened 2024-10-03
Parallelize MuHash calculations

coderofstuff:
Parallelize parts of the muhash calculations (`add_utxo` and `add_transaction`) then later combine results so we can delegate the heavier work to multiple threads.

### DIFF
--- a/consensus/src/pipeline/virtual_processor/utxo_validation.rs
+++ b/consensus/src/pipeline/virtual_processor/utxo_validation.rs
@@ -31,6 +31,7 @@ use kaspa_muhash::MuHash;
 use kaspa_utils::refs::Refs;
 
 use rayon::prelude::*;
+use smallvec::{smallvec, SmallVec};
 use std::{iter::once, ops::Deref};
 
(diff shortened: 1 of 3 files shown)


consensus/src/pipeline/virtual_processor/utxo_validation.rs
```diff
@@ -177,7 +179,7 @@ impl VirtualStateProcessor {
 
         // Verify all transactions are valid in context
         let current_utxo_view = selected_parent_utxo_view.compose(&ctx.mergeset_diff);
-        let validated_transactions =
+        let (validated_transactions, _) =
```
michaelsutton, line 182: This is a frequent call (called per chain block), which does not use the combined muhash (thus adding significant cost not present beforehand). 
I think we should keep `validate_transactions_in_parallel` like it was, and simply apply a new parallel operation for the muhash reduce to be applied only within `calculate_utxo_state`

michaelsutton: or create two different `validate_transactions_in_parallel` impls

michaelsutton (approved):

PR #576, opened 2024-10-03
Optimize window cache building for ibd

D-Stacks:
### Background

Currently we Cache block windows during header processing, but rely on the cache in both block body processing, and virtual processing. A cache miss on the window cache results in a large overhead in re-constructing the windows. Specifically this results in significant processing overhead during IBD, where header, and body / virtual processing is not streamlined against each other.

This PR addresses, and patches this issue (thus far, only for tn11) by:

1) caching windows in the virtual processor on new sinks, if it's window is not already cached. 
2) applying an optimization whereby the window_manager will try to not reconstruct a whole window from scratch when the cache is missed, but abort early by merging with an ancestor cache (if found). 
3) applying a lazy load for the past median time cache, which only triggers after a 'quick-and-dirty' pre-check that needs to find a tx where  tx.locktime != 0 first, before validating it in full in `check_block_transactions_in_context`

### Tests and Benching

I currently have created a branch, [here](https://github.com/D-Stacks/rusty-kaspa/tree/25eb1e2) which checks the new and old build method against each other, and logs time spent building the block windows. 

all window builds are passing asserts between old and new.

regarding benching, here is an excerpt, taken during IBD:

```
2024-09-10 00:45:11.731+02:00 [INFO ] IBD: Processed 396000 blocks (26%) last block timestamp: 2024-09-08 16:04:14.000:+0200
2024-09-10 00:45:11.895+02:00 [INFO ] Processed 6280 blocks and 0 headers in the last 10.00s (296652 transactions; 2594 UTXO-validated blocks; 0.00 parents; 0.00 mergeset; 47.24 TPB; 94187.4 mass)
2024-09-10 00:45:21.318+02:00 [INFO ] New Total Time spent in build_block_window: 36.989105s 

            New iterations difficulty: 426446 

            New iterations pmt: 18802163 

            Old Total Time spent in build_block_window: 751.755591s 

            Old iterations difficulty: 150506854 

            Old iterations pmt: 455390110 
```
Note: 

1) most of this performance increase is during block body IBD syncing, where the difficulty window is built, or more  specifically on virtual resolve during IBD. but it also reduces time spent in `past_median_time` window by about 75% (from 16 secs to about 4 secs, on my machine). 

2) it doesn't factor in the parallelization and real-time gain during syncing.


coderofstuff (dismissed):
In the other PR #551 I noted that I've tested this to completion and observed hotspots related to window calc removed.

Minor comment left

consensus/src/pipeline/body_processor/body_validation_in_context.rs
```diff
@@ -18,13 +19,25 @@ impl BlockBodyProcessor {
     }
 
     fn check_block_transactions_in_context(self: &Arc<Self>, block: &Block) -> BlockProcessResult<()> {
-        let (pmt, _) = self.window_manager.calc_past_median_time(&self.ghostdag_store.get_data(block.hash()).unwrap())?;
+        // TODO: this is somewhat expensive during ibd, as it incurs cache misses.
+        let pmt = Lazy::new(|| {
+            let (pmt, _) = self
+                .window_manager
+                .calc_past_median_time(&self.ghostdag_store.get_data(block.hash()).unwrap())
+                .expect("expected header processor to have checked error case");
```
michaelsutton, line 27: Actually not 100% about this `expect`

michaelsutton (changes requested):

consensus/src/pipeline/body_processor/body_validation_in_context.rs
D-Stacks: I can also init after the quick check, and propagate error back in the function.

consensus/src/processes/window.rs
```diff
@@ -387,50 +365,96 @@ impl<T: GhostdagStoreReader, U: BlockWindowCacheReader, V: HeaderStoreReader, W:
...
+        sample_rate: u64,
+        cache: Option<&Arc<U>>,
+        ghostdag_data: &GhostdagData,
+        selected_parent_blue_work: BlueWorkType,
+        mut mergeset_non_daa_inserter: impl FnMut(Hash),
+    ) -> Option<Arc<BlockWindowHeap>> {
+        cache.and_then(|cache| {
+            cache.get(&ghostdag_data.selected_parent).map(|selected_parent_window| {
+                let mut heap = Lazy::new(|| BoundedSizeBlockHeap::from_binary_heap(window_size, (*selected_parent_window).clone()));
+                for block in self.sampled_mergeset_iterator(sample_rate, ghostdag_data, selected_parent_blue_work) {
```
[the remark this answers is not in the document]
D-Stacks: will add comment.

coderofstuff:
You have the test and benches in another branch. Any way we can get those into here in a form that makes sense to run regularly?

D-Stacks:
> You have the test and benches in another branch. Any way we can get those into here in a form that makes sense to run regularly?

I don't have unit tests per se, Just ones that compare and bench during runtime. The runtime cost is not trivial. Maybe it can be done behind the sanity tests flag? But even then it is a bit expensive.

coderofstuff (approved):
I've tested the latest commit in TN11 and mainnet and both IBD and run fine

consensus/src/pipeline/virtual_processor/processor.rs
```diff
@@ -303,11 +311,19 @@ impl VirtualStateProcessor {
             .expect("all possible rule errors are unexpected here");
 
         // Update the pruning processor about the virtual state change
-        let sink_ghostdag_data = self.ghostdag_store.get_compact_data(new_sink).unwrap();
+        let compact_sink_ghostdag_data = if prev_sink != new_sink {
+            // we need to check with full data here, since we may need to update the window caches
+            let sink_ghostdag_data = self.ghostdag_store.get_data(new_sink).unwrap();
+            // update window caches - for ibd performance. see method comment for more details.
+            self.maybe_commit_windows(new_sink, &sink_ghostdag_data);
```
michaelsutton, line 318: 1. It seems like the call to `maybe_commit_windows` should take place before the call `calculate_and_commit_virtual_state` where virtual DAA and median time windows are built (so we'll get 100% cache hits for virtual's selected parent)
2. Let's simply call it `cache_windows`. Then it will be obvious that if these windows are already cached it's a no op.

consensus/src/pipeline/virtual_processor/processor.rs
```diff
@@ -540,6 +556,24 @@ impl VirtualStateProcessor {
         drop(selected_chain_write);
     }
 
+    fn maybe_commit_windows(&self, new_sink: Hash, sink_ghostdag_data: &GhostdagData) {
+        // this is only important for ibd performance, as we incur expensive cache misses otherwise.
+        // this occurs because we cannot rely on header processing to pre-cache in this scenario.
+
+        // TODO: We could optimize this by only committing the windows if virtual processor where to have explicit knowledge of being in ibd.
```
michaelsutton, line 563: I don't think this TODO is required. In non-IBD mode these keys will almost always already exist.

consensus/src/pipeline/body_processor/body_validation_in_context.rs
```diff
@@ -18,13 +19,21 @@ impl BlockBodyProcessor {
     }
 
     fn check_block_transactions_in_context(self: &Arc<Self>, block: &Block) -> BlockProcessResult<()> {
-        let (pmt, _) = self.window_manager.calc_past_median_time(&self.ghostdag_store.get_data(block.hash()).unwrap())?;
+        // TODO: this is somewhat expensive during ibd, as it incurs cache misses.
+        let pmt_res =
+            Lazy::new(|| match self.window_manager.calc_past_median_time(&self.ghostdag_store.get_data(block.hash()).unwrap()) {
```
michaelsutton, line 24: here too we should check the cache first, and if we had to build ourselves -- save the result in the cache

tiram88 (changes requested):
Here are a few first comments. I carry on tomorrow.

consensus/src/processes/window.rs
```diff
@@ -387,50 +370,89 @@ impl<T: GhostdagStoreReader, U: BlockWindowCacheReader, V: HeaderStoreReader, W:
...
-            return true;
-        }
-
+        mut mergeset_non_daa_inserter: impl FnMut(Hash),
+    ) {
         for block in self.sampled_mergeset_iterator(sample_rate, ghostdag_data, selected_parent_blue_work) {
             match block {
                 SampledBlock::Sampled(block) => {
-                    if !heap.try_push(block.hash, block.blue_work) {
-                        break;
```
michaelsutton, line 427: Any particular reason why was this `break` removed? `try_push` failing means we reached a block with blue work which is too low for entering the window. Since `sampled_mergeset_iterator` uses the `descending_mergeset` iterator, we know for certain that all remaining blocks in the mergeset will have even lower blue work, hence we should be able to break (note: this is true even for the case we inherited the window from the selected parent and are pushing the top-most mergeset)

tiram88 (changes requested):
Some minor additional issues/enhancements.

The insertion of non DAA items in `push_mergeset` has to be addressed though.

consensus/src/processes/window.rs
```diff
@@ -387,50 +370,91 @@ impl<T: GhostdagStoreReader, U: BlockWindowCacheReader, V: HeaderStoreReader, W:
...
+                    if !heap.deref_mut().try_push(block.hash, block.blue_work) {
```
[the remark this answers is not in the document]
michaelsutton: For some reason the deref_mut call is necessary (at least when I checked)

michaelsutton (approved):

tiram88 (approved):

PR #582, opened 2024-10-18
typo(cli/utils): kaspa wording

IzioDev:


D-Stacks:
hey, can you update and correct the lint warnings / errors as well? these are because of a rust upgrade, in most cases just upgrading to newest rust and following the clippy edit suggestions suffices.

Issue #586, opened 2024-11-03
Feature Request:  signMessageWithoutRand api of kaspa wasm


witter-deland:
**Is your feature request related to a problem? Please describe.**

We have a scenario where we require a stable signature for verification purposes. Currently, the existing functionality in kaspa wasm does not provide a dedicated api for this specific need. Without the signMessageWithoutRand api, we face difficulties in achieving consistent and reliable signature verification, which is crucial for the integrity and security of our operations.

**Describe the solution you'd like**

We would like the addition of the signMessageWithoutRand api in kaspa wasm. This api should allow us to generate signatures in a stable and deterministic manner, without the randomness factor that may cause variability in the signature generation process. It should accept the necessary input parameters, such as the message to be signed and the relevant key information, and return a signature that can be used for reliable verification.

**Describe alternatives you've considered**

We have considered implementing our own custom signature generation mechanism outside of kaspa wasm. However, this would require significant additional development effort and may not integrate as seamlessly with the existing kaspa infrastructure. It also poses risks in terms of compatibility and security compared to having a native api within kaspa wasm. Another alternative was to look for workarounds within the existing api set, but we have not found a satisfactory solution that meets our stability and verification requirements.

**Additional context**

In our application, we have a process that relies on the verification of signatures to ensure the authenticity and integrity of data. The lack of a stable signature generation api has led to inconsistent results in our verification process, which can potentially lead to security vulnerabilities and incorrect data handling. We believe that the addition of the signMessageWithoutRand api would greatly enhance the reliability and security of our overall system.

elichai:
> In our application, we have a process that relies on the verification of signatures to ensure the authenticity and integrity of data. The lack of a stable signature generation api has led to inconsistent results in our verification process, which can potentially lead to security vulnerabilities and incorrect data handling. We believe that the addition of the signMessageWithoutRand api would greatly enhance the reliability and security of our overall system.

Can you explain more on this? how did it affect the results? how are you using those signatures? is it that you hash the data+signature and expect it to reproduce with a new signature? a bit more overview of the usecase would be great and can increase the odds of it merging into `rusty-kaspa`

elichai:
> One use case : use the signature of a specific message as the AES encryption key for some user data

Now that's insecure :) the signature is not uniform and should not be treated as "secret"
Why not derive an AES key from the private key and message? (via (H)MAC)

witter-deland:
[...]

PR #587, opened 2024-11-03
feat: add signMessageWithoutRand method for kaspa wasm

witter-deland:
#586


witter-deland:
https://github.com/kasware-wallet/extension/pull/5

coderofstuff:
Signing without aux rand is dangerous and can leak keys.

apoelstra:
>Signing without aux rand is dangerous and can leak keys.

This is completely untrue.

coderofstuff:
> > Signing without aux rand is dangerous and can leak keys.
> 
> This is completely untrue.

Can you elaborate? I'm not familiar with the underlying implementation here but I'm mainly concerned with this leading to nonce reuse.

elichai:
Andrew is obviously right :) 
The auxiliary randomness exist only to mitigate specific kinds of power analysis side-channel attacks,
Providing it definitely improves security, but omitting it should not be considered dangerous, as most legacy signature schemes don't provide mitigations against such attacks,

To read more about the relevant discussions that arised in adding this randomness please see: https://github.com/sipa/bips/issues/195

coderofstuff:
> Andrew is obviously right :) The auxiliary randomness exists only to mitigate specific kinds of power analysis side-channel attacks, Providing it definitely improves security, but omitting it should not be considered dangerous, as most legacy signature schemes don't provide mitigations against such attacks,
> 
> To read more about the relevant discussions that arised in adding this randomness please see: [sipa/bips#195](https://github.com/sipa/bips/issues/195)

Thanks for the input @elichai.

@witter-deland I've reopened the PR while I read up more onto this.

witter-deland:
@coderofstuff Thanks
@elichai @apoelstra Thank you both for participating in the discussion

> > Andrew is obviously right :) The auxiliary randomness exists only to mitigate specific kinds of power analysis side-channel attacks, Providing it definitely improves security, but omitting it should not be considered dangerous, as most legacy signature schemes don't provide mitigations against such attacks,
> > To read more about the relevant discussions that arised in adding this randomness please see: [sipa/bips#195](https://github.com/sipa/bips/issues/195)
> 
> Thanks for the input @elichai.
> 
> @witter-deland I've reopened the PR while I read up more onto this.

witter-deland:
Excuse me. What's going on? This API is very straightforward. Are there any issues that are blocking the merge? Thank you.

coderofstuff (changes requested):
Apologies for the delay, we're currently packed with other things.

The change itself looks sound. I just have some comments on code duplication.

witter-deland:
fixed :)
> Apologies for the delay, we're currently packed with other things.
> 
> The change itself looks sound. I just have some comments on code duplication.

wallet/core/src/wasm/message.rs
```diff
@@ -14,6 +14,7 @@ const TS_MESSAGE_TYPES: &'static str = r#"
 export interface ISignMessage {
     message: string;
     privateKey: PrivateKey | string;
+    noAuxRand: boolean;
```
coderofstuff, line 17: Please make this optional here. This ensure any existing usage when this is added

wallet/core/src/wasm/message.rs
```diff
@@ -30,10 +31,12 @@ pub fn js_sign_message(value: ISignMessage) -> Result<HexString, Error> {
     if let Some(object) = Object::try_from(&value) {
         let private_key = object.cast_into::<PrivateKey>("privateKey")?;
         let raw_msg = object.get_string("message")?;
+        let no_aux_rand = object.get_bool("noAuxRand")?;
```
coderofstuff, line 34: If not passed, this should default to `false`

coderofstuff (changes requested):

coderofstuff (dismissed):
LGTM. Tested the scenarios and existing demo still works.

As a follow-up, please update the `wasm/examples/nodejs/javascript/general/message-signing.js` to include a demo of the new scenario you added support for (passing `noAuxRand: true`)

witter-deland:
Sure, done :)
> LGTM. Tested the scenarios and existing demo still works.
> 
> As a follow-up, please update the `wasm/examples/nodejs/javascript/general/message-signing.js` to include a demo of the new scenario you added support for (passing `noAuxRand: true`)

coderofstuff (approved):
Thanks for that!

PR #588, opened 2024-11-05
Standartize fork activation logic

someone235:
To avoid confusion where in one place someone will use `>=`  for fork activation logic, and in other place `>`, I introduced a newtype around uint64 to avoid such mistakes and enforce everyone to check with `is_active`

### DIFF
--- a/consensus/src/processes/transaction_validator/transaction_validator_populated.rs
+++ b/consensus/src/processes/transaction_validator/transaction_validator_populated.rs
@@ -44,13 +44,14 @@ impl TransactionValidator {
         let total_in = self.check_transaction_input_amounts(tx)?;
         let total_out = Self::check_transaction_output_values(tx, total_in)?;
         let fee = total_in - total_out;
-        if flags != TxValidationFlags::SkipMassCheck && pov_daa_score > self.storage_mass_activation_daa_score {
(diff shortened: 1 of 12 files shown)


consensus/src/processes/transaction_validator/transaction_validator_populated.rs
```diff
@@ -48,10 +49,11 @@ impl TransactionValidator {
             // Storage mass hardfork was activated
             self.check_mass_commitment(tx)?;
 
-            // TODO: Decide what to do here
-            // if pov_daa_score < self.storage_mass_activation + 10 && self.storage_mass_activation > 0 {
-            //     warn!("--------- Storage mass hardfork was activated successfully!!! --------- (DAA score: {})", pov_daa_score);
-            // }
+            if self.storage_mass_activation.is_within_range_from_activation(pov_daa_score, 10)
+                && self.storage_mass_activation != ForkActivation::always()
```
michaelsutton, line 53: I think this condition can go within the is_within_range_from_activation call (+ a short doc explaining the behavior). Because 
1. activation can be defined as switching on something that was off, so "within range from activation" can logically be interpreted as false in the always case
2. I think all future calls to this method will require the same condition.

michaelsutton (approved):

PR #591, opened 2024-11-10
Enable payloads for non coinbase transactions

someone235:
Simpa results

```
cargo run --release --bin simpa -- --testnet11 -t=5 -n=10000 --long-payload
2024-11-10 17:33:06.044+00:00 [INFO ] Total validation time: 22.023479023s, block processing rate: 454.06 (b/s), transaction processing rate: 2479.35 (t/s)

cargo run --release --bin simpa -- --testnet11 -n=10000
2024-11-10 17:38:16.827+00:00 [INFO ] Total validation time: 14.898869473s, block processing rate: 671.19 (b/s), transaction processing rate: 30463.32 (t/s)
```

TODO:

- [x] Make a test for activation logic
- [x] Make sure txid, txhash and sighash are affected by payload change (maybe we already have tests that verify it, we need to check)
- [x] Search for TODOs related to payloads in the codebase and apply them if needed
- [x] Check perf bottlenecks (Maybe the problem is somehow unique to simpa?)
- [x] Make benchmark for validating a transaction with many inputs and large payloads
- [ ] Launch a network, where txgen creates txns with long payload, and check how the network handles it


michaelsutton:
Note the implementation of `sighash::payload_hash` which currently ignores the payload, as well as the TODO within (reused values)

someone235:
> Note the implementation of `sighash::payload_hash` which currently ignores the payload, as well as the TODO within (reused values)

Added to TODOs

michaelsutton (dismissed):

michaelsutton (approved):

consensus/core/src/hashing/sighash.rs
```diff
@@ -64,6 +64,10 @@ impl<T: SigHashReusedValues> SigHashReusedValues for Arc<T> {
     fn outputs_hash(&self, set: impl Fn() -> Hash) -> Hash {
         self.as_ref().outputs_hash(set)
     }
+
+    fn payload_hash(&self, set: impl Fn() -> Hash) -> Hash {
+        self.as_ref().outputs_hash(set)
```
michaelsutton, line 69: oops this looks like a glitch `outputs_hash` -> `payload_hash`

consensus/core/src/hashing/sighash.rs
```diff
@@ -182,17 +203,17 @@ pub fn sig_op_counts_hash(tx: &Transaction, hash_type: SigHashType, reused_value
...
 
-pub fn payload_hash(tx: &Transaction) -> Hash {
-    if tx.subnetwork_id == SUBNETWORK_ID_NATIVE {
-        return ZERO_HASH;
-    }
+pub fn payload_hash(tx: &Transaction, reused_values: &impl SigHashReusedValues) -> Hash {
+    let hash = || {
+        if tx.subnetwork_id.is_native() && tx.payload.is_empty() {
+            return ZERO_HASH;
+        }
```
michaelsutton, line 210: in all other methods (`sig_op_counts_hash` etc) we do the short-circuit check out of the inner lambda method, perhaps we should keep it consistent?

consensus/core/src/hashing/sighash.rs
michaelsutton: I see this `impl for Arc` is never used anyway, so perhaps we should just delete it

simpa/Cargo.toml
```diff
@@ -40,3 +40,8 @@ tokio = { workspace = true, features = ["rt", "macros", "rt-multi-thread"] }
 [features]
 heap = ["dhat", "kaspa-alloc/heap"]
 semaphore-trace = ["kaspa-utils/semaphore-trace"]
+
+[profile.heap]
+inherits = "release"
+debug = true
+strip = false
```
michaelsutton, line 47: `cargo run simpa` now prints the following:
```
warning: profiles for the non root package will be ignored, specify profiles at the workspace root:
package:   ~/rusty-kaspa/simpa/Cargo.toml
workspace: ~/rusty-kaspa/Cargo.toml
```

and indeed `cargo run --profile heap --bin simpa --` works without it as well

someone235:
Addressed review in #605 @michaelsutton

PR #592, opened 2024-11-11
Implement Retention Period Support

coderofstuff:
### Notes

- New configuration usage `--retention-period-days=N` - set this to some number to keep that many days of data. If this is smaller than the pruning period duration, a pruning period duration amount of data will be kept still.
- Implementation detail - this goes through the selected parent chain to find the first block that is past the N days threshold. There is no extra logic that ensures the anticone of the last SPC block found is also retained as such those will still be pruned.

### DIFF
--- a/consensus/core/src/config/mod.rs
+++ b/consensus/core/src/config/mod.rs
@@ -68,6 +68,9 @@ pub struct Config {
 
     /// A scale factor to apply to memory allocation bounds
     pub ram_scale: f64,
+
+    /// The number of days to keep data for
+    pub retention_period_days: f64,
 }
 
 impl Config {
@@ -95,6 +98,7 @@ impl Config {
             initial_utxo_set: Default::default(),
             disable_upnp: false,
             ram_scale: 1.0,
+            retention_period_days: 3.0,
         }
     }
 
--- a/consensus/core/src/config/mod.rs
+++ b/consensus/core/src/config/mod.rs
@@ -98,7 +98,7 @@ impl Config {
             initial_utxo_set: Default::default(),
             disable_upnp: false,
             ram_scale: 1.0,
-            retention_period_days: 3.0,
+            retention_period_days: 0.0,
         }
     }
 
--- a/consensus/core/src/config/mod.rs
+++ b/consensus/core/src/config/mod.rs
@@ -70,7 +70,7 @@ pub struct Config {
     pub ram_scale: f64,
 
     /// The number of days to keep data for
-    pub retention_period_days: f64,
+    pub retention_period_days: Option<f64>,
 }
 
 impl Config {
@@ -98,7 +98,7 @@ impl Config {
             initial_utxo_set: Default::default(),
             disable_upnp: false,
             ram_scale: 1.0,
-            retention_period_days: 0.0,
+            retention_period_days: None,
         }
     }
 
--- a/consensus/src/consensus/mod.rs
+++ b/consensus/src/consensus/mod.rs
@@ -565,7 +565,7 @@ impl ConsensusApi for Consensus {
         // Verify the block exists and can be assumed to have relations and reachability data
         self.validate_block_exists(hash).ok()?;
 
-        // Verify that the block is in future(source), where Ghostdag data is complete
+        // Verify that the block is in future(retention root), where Ghostdag data is complete
         self.services.reachability_service.is_dag_ancestor_of(self.get_retention_period_root(), hash).then_some(())?;
 
         let sink = self.get_sink();
@@ -626,20 +626,18 @@ impl ConsensusApi for Consensus {
         pruning_point_read.retention_period_root().unwrap_or(pruning_point_read.pruning_point().unwrap())
     }
 
-    /// Estimates number of blocks and headers stored in the node
+    /// Estimates the number of blocks and headers stored in the node database.
     ///
-    /// This is an estimation based on the daa score difference between the node's `source` and `sink`'s daa score,
+    /// This is an estimation based on the DAA score difference between the node's `retention root` and `virtual`'s DAA score,
     /// as such, it does not include non-daa blocks, and does not include headers stored as part of the pruning proof.  
     fn estimate_block_count(&self) -> BlockCount {
-        // PRUNE SAFETY: node is either archival or source is the pruning point which its header is kept permanently
-        let retention_period_root_score =
-            self.headers_store.get_compact_header_data(self.get_retention_period_root()).unwrap().daa_score;
+        // PRUNE SAFETY: retention root is always a current or past pruning point which its header is kept permanently
+        let retention_period_root_score = self.headers_store.get_daa_score(self.get_retention_period_root()).unwrap();
         let virtual_score = self.get_virtual_daa_score();
         let header_count = self
             .headers_store
-            .get_compact_header_data(self.get_headers_selected_tip())
+            .get_daa_score(self.get_headers_selected_tip())
             .unwrap_option()
-            .map(|h| h.daa_score)
             .unwrap_or(virtual_score)
             .max(virtual_score)
             - retention_period_root_score;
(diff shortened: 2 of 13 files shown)


KaffinPX:
Instead of days why dont we keep it by pruning points, would be a more convenient way(for future archival p2p features).

coderofstuff:
Changes to apply:
- Instead of extra days, allow specifying exactly how many days to keep
- Track the history root independent of pruning point

michaelsutton:
> Changes to apply:
> 
> * Instead of extra days, allow specifying exactly how many days to keep
> * Track the history root independent of pruning point
---
1. Also, search for the new history root from the bottom-up. 
2. You might need to add another DB tracking field in addition to history root: `source` (or a better name). 
Flow:
- find new source (by walking up from current source and up the chain)
- set source = new source (right before prune starts)
- prune
- set history root = source

This way the code can safely use `source` in places where data validity is required (e.g., get_source).
In `recover_pruning_workflows_if_needed` the condition for recovery should change to `if history_root != source`

michaelsutton:
> Instead of days why dont we keep it by pruning points, would be a more convenient way(for future archival p2p features)?

This might be a good idea, but we can apply this change later. I still think the cmd arg should be number of days to keep, and we can later translate that to the closest pruning point. 

Btw, perhaps we should log on start-up with warning about approximate max disk usage (for transient data) given the requested days.

kaspad/src/args.rs
```diff
@@ -140,6 +141,7 @@ impl Default for Args {
             disable_dns_seeding: false,
             disable_grpc: false,
             ram_scale: 1.0,
+            retention_period_days: 3.0,
```
michaelsutton, line 144: seems like this will default to 3 days (or am I missing something?), while we want to default to the usual pruning point

consensus/src/pipeline/pruning_processor/processor.rs
```diff
@@ -521,12 +541,47 @@ impl PruningProcessor {
...
 
+    /// Adjusts the retention period root forward until the maximum chain block is reached that covers the retention period.
+    /// This is the last chain block B such that B.timestamp < retention_period_days_ago. This may return the old hash if
+    /// the retention period cannot be covered yet with the node's current history.
+    ///
+    /// This function is expected to be called only when a new pruning point is determined and right before
+    /// doing any pruning.
+    ///
+    /// retention_period_root is guaranteed to be in the past(pruning_point)
+    fn advance_retention_period_root(&self, retention_period_root: Hash, pruning_point: Hash) -> Hash {
```
michaelsutton, line 558: in theory we can now simplify this by iterating through the new pruning samples store from the new pruning point to the prev retention root and searching for the first point with enough days above it. This will fulfill @KaffinPX's request to always have a pruning point as retention root and the code is also very simple to write (+ iteration is efficient and iterates very few steps since we are jumping between samples only).

consensus/src/consensus/mod.rs
```diff
@@ -635,7 +634,7 @@ impl ConsensusApi for Consensus {
     /// as such, it does not include non-daa blocks, and does not include headers stored as part of the pruning proof.  
     fn estimate_block_count(&self) -> BlockCount {
         // PRUNE SAFETY: node is either archival or source is the pruning point which its header is kept permanently
-        let source_score = self.headers_store.get_compact_header_data(self.get_source()).unwrap().daa_score;
+        let source_score = self.headers_store.get_compact_header_data(self.get_retention_root()).unwrap().daa_score;
```
michaelsutton, line 637: are we keeping `source` as a name in other places as well?

michaelsutton (changes requested):

consensus/src/consensus/mod.rs
coderofstuff: No, I just forgot to replace this one.

kaspad/src/args.rs
coderofstuff: It will default to 3 days, which is to keep what we currently have. But setting it to 0 as the default and staying with usual pruning depth is fine too as the default. We can inform users that they can set `--retention-period-days=3` if they want to keep more days of data.

consensus/src/pipeline/pruning_processor/processor.rs
coderofstuff: Say someone sets retention period days to 300 days:

At 10bps, pruning point moves every 12hrs roughly. In that timeframe, that's 432000 blocks in that period. Assuming even only 1/10 of those are chain blocks - a forward chain iteration from retention root to the next pruning point still takes 43,200 steps.

In contrast a pruning point move where the iterations goes through pruning point samples only starting from the new pruning point will take ~600 steps to reach the new retention root (300 days, 2 pruning points per day).

Conclusion: yes, I agree with the above approach.

consensus/src/pipeline/pruning_processor/processor.rs
```diff
@@ -521,12 +536,57 @@ impl PruningProcessor {
...
+
+        trace!(
+            "Adjusting the retention period root to cover the required retention period. Target timestamp: {}",
+            retention_period_root_ts_target,
+        );
+
+        while retention_period_root != new_retention_period_root {
+            let block = new_retention_period_root;
+
+            let timestamp = self.headers_store.get_compact_header_data(block).unwrap().timestamp;
```
michaelsutton, line 568: nit: we have `headers_store.get_timestamp`

consensus/core/src/config/mod.rs
```diff
@@ -68,6 +68,9 @@ pub struct Config {
 
     /// A scale factor to apply to memory allocation bounds
     pub ram_scale: f64,
+
+    /// The number of days to keep data for
+    pub retention_period_days: f64,
```
michaelsutton, line 73: my personal preference is that we make this an `Option` (as well as the cmd arg too) and explicitly handle the `None` case in pruning processor by directly setting the retention root as the pruning point. This makes it easier for me to verify that the changes do not affect the default way of using a node and thus reduces risks to zero.

kaspad/src/args.rs
michaelsutton: see my other comment

michaelsutton:
Opened a PR over this branch with a few more essential leftover fixes I noticed (especially note the DB upgrade for archival cases): https://github.com/coderofstuff/rusty-kaspa/pull/19

michaelsutton (approved):

michaelsutton:
Ran some custom simpa tests with the following diff (and archival switched on and off). All looking good.

command line: `cargo run --release --bin simpa -- -n=6000 -t=1 --test-pruning --loglevel=info`

```diff
git diff
diff --git a/consensus/src/pipeline/pruning_processor/processor.rs b/consensus/src/pipeline/pruning_processor/processor.rs
index 923df58c5..c363c2eaa 100644
--- a/consensus/src/pipeline/pruning_processor/processor.rs
+++ b/consensus/src/pipeline/pruning_processor/processor.rs
@@ -205,6 +205,10 @@ impl PruningProcessor {
             // Inform the user
             info!("Periodic pruning point movement: advancing from {} to {}", current_pruning_info.pruning_point, new_pruning_point);
 
+            let virtual_daa_score = self.lkg_virtual_state.load().daa_score;
+            let retention_daa_score = self.headers_store.get_daa_score(adjusted_retention_period_root).unwrap();
+            warn!("VVVVVVVVVVVVVVVVVVV {}", virtual_daa_score - retention_daa_score);
+
             // Advance the pruning point utxoset to the state of the new pruning point using chain-block UTXO diffs
             if !self.advance_pruning_utxoset(current_pruning_info.pruning_point, new_pruning_point) {
                 info!("Interrupted while advancing the pruning point UTXO set: Process is exiting");
@@ -563,7 +567,11 @@ impl PruningProcessor {
                 let retention_period_ms = (retention_period_days * 86400.0 * 1000.0).ceil() as u64;
 
                 // The target timestamp we would like to find a point below
-                let retention_period_root_ts_target = unix_now().saturating_sub(retention_period_ms);
+                let retention_period_root_ts_target = self
+                    .headers_store
+                    .get_timestamp(self.lkg_virtual_state.load().ghostdag_data.selected_parent)
+                    .unwrap()
+                    .saturating_sub(retention_period_ms);
 
                 // Iterate from the new pruning point to the prev retention root and search for the first point with enough days above it.
                 // Note that prev retention root is always a past pruning point, so we can iterate via pruning samples until we reach it.
diff --git a/simpa/src/main.rs b/simpa/src/main.rs
index 3021ee568..f7f5a3749 100644
--- a/simpa/src/main.rs
+++ b/simpa/src/main.rs
@@ -199,6 +199,8 @@ fn main_impl(mut args: Args) {
         .apply_args(|config| apply_args_to_perf_params(&args, &mut config.perf))
         .adjust_perf_params_to_consensus_params()
         .apply_args(|config| config.ram_scale = args.ram_scale)
+        .apply_args(|config| config.retention_period_days = Some(0.01))
+        // .set_archival()
         .skip_proof_of_work()
         .enable_sanity_checks();
     if !args.test_pruning {
```

PR #594, opened 2024-11-12
Some simplification to script number types

someone235:
Unifying `OpcodeData<i32/i64/Kip10I64>` to use `SizedEncodeInt` for serialization.

Critical lines and explanation:
1. https://github.com/someone235/rusty-kaspa/blob/976313f0f4c492061689d84811b17ee1ea73cc8b/crypto/txscript/src/data_stack.rs#L178
Here we limit any `push_item<i32>` to 4 bytes. This is fine since we nowhere use this method.
2. https://github.com/someone235/rusty-kaspa/blob/976313f0f4c492061689d84811b17ee1ea73cc8b/crypto/txscript/src/data_stack.rs#L166
Here we limit any `push_item<i64>` to 8 bytes. Explanation for safety is in the code.
3. https://github.com/someone235/rusty-kaspa/blob/976313f0f4c492061689d84811b17ee1ea73cc8b/crypto/txscript/src/data_stack.rs#L173
Here we limit any `pop_items<i32>` to 4 bytes, disallowing the use of `i32::MIN`. This is fine since in any place we use `pop_items<i32>` we also return an error if the number is negative.

### DIFF
--- a/crypto/txscript/src/data_stack.rs
+++ b/crypto/txscript/src/data_stack.rs
@@ -3,6 +3,7 @@ use core::fmt::Debug;
 use core::iter;
 use kaspa_txscript_errors::SerializationError;
 use std::cmp::Ordering;
+use std::num::TryFromIntError;
 use std::ops::Deref;
 
 #[derive(PartialEq, Eq, Debug, Default, PartialOrd, Ord)]
@@ -17,6 +18,14 @@ impl<const LEN: usize> From<i64> for SizedEncodeInt<LEN> {
 impl<const LEN: usize> From<i32> for SizedEncodeInt<LEN> {
     fn from(value: i32) -> Self {
         SizedEncodeInt(value as i64)
+    }
+}
+
+impl<const LEN: usize> TryFrom<SizedEncodeInt<LEN>> for i32 {
+    type Error = TryFromIntError;
+
+    fn try_from(value: SizedEncodeInt<LEN>) -> Result<Self, Self::Error> {
+        value.0.try_into()
     }
 }
 
@@ -170,7 +179,7 @@ impl OpcodeData<i64> for Vec<u8> {
 impl OpcodeData<i32> for Vec<u8> {
     #[inline]
     fn deserialize(&self) -> Result<i32, TxScriptError> {
-        OpcodeData::<SizedEncodeInt<4>>::deserialize(self).map(|v| i64::from(v).try_into().expect("number is within i32 range"))
+        OpcodeData::<SizedEncodeInt<4>>::deserialize(self).map(|v| v.try_into().expect("number is within i32 range"))
     }
 
     #[inline]
@@ -195,11 +204,6 @@ impl<const LEN: usize> OpcodeData<SizedEncodeInt<LEN>> for Vec<u8> {
 
     #[inline]
     fn serialize(from: &SizedEncodeInt<LEN>) -> Result<Self, SerializationError> {
-        // This is an optimzation for LEN 8. Without it, the function will fail when checking the serialized data length.
-        if LEN == 8 && from.0 == i64::MIN {
-            return Err(SerializationError::NumberTooLong(from.0));
-        }
-
         let bytes = serialize_i64(&from.0);
         if bytes.len() > LEN {
             return Err(SerializationError::NumberTooLong(from.0));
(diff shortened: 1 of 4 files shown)


biryukovmaxim (approved):
LGTM!

crypto/txscript/src/data_stack.rs
```diff
@@ -218,7 +195,16 @@ impl<const LEN: usize> OpcodeData<SizedEncodeInt<LEN>> for Vec<u8> {
...
     #[inline]
     fn serialize(from: &SizedEncodeInt<LEN>) -> Result<Self, SerializationError> {
-        Ok(serialize_i64(&from.0))
+        // This is an optimzation for LEN 8. Without it, the function will fail when checking the serialized data length.
+        if LEN == 8 && from.0 == i64::MIN {
+            return Err(SerializationError::NumberTooLong(from.0));
+        }
+
+        let bytes = serialize_i64(&from.0);
+        if bytes.len() > LEN {
```
biryukovmaxim, line 204: Is this possible?

crypto/txscript/src/data_stack.rs
```diff
@@ -111,102 +146,44 @@ fn deserialize_i64(v: &[u8]) -> Result<i64, TxScriptError> {
...
 }
 
 impl OpcodeData<i32> for Vec<u8> {
     #[inline]
     fn deserialize(&self) -> Result<i32, TxScriptError> {
-        let res = OpcodeData::<i64>::deserialize(self)?;
-        // TODO: Consider getting rid of clamp, since the call to deserialize should return an error
-        // if the number is not in the i32 range (this should be done with proper testing)?
-        Ok(res.clamp(i32::MIN as i64, i32::MAX as i64) as i32)
+        OpcodeData::<SizedEncodeInt<4>>::deserialize(self).map(|v| i64::from(v).try_into().expect("number is within i32 range"))
```
biryukovmaxim, line 173: Would it be better if tryInto gets implemented on the SizedEncodeInt directly ? Possibly making separate module for the type

crypto/txscript/src/data_stack.rs
```diff
@@ -111,102 +146,44 @@ fn deserialize_i64(v: &[u8]) -> Result<i64, TxScriptError> {
...
     }
 
     #[inline]
     fn serialize(from: &i32) -> Result<Self, SerializationError> {
-        Ok(OpcodeData::<i64>::serialize(&(*from as i64)).expect("should never happen"))
+        OpcodeData::<SizedEncodeInt<4>>::serialize(&(*from).into())
```
michaelsutton, line 178: I think this is a breaking change for `u32::MIN` (which will fail now), so `OpcodeData::<i64>` should be kept here. If i'm correct, then it means that a test is missing for this case (verifying that serialize `u32::MIN` should not err)

crypto/txscript/src/data_stack.rs
```diff
@@ -218,7 +195,16 @@ impl<const LEN: usize> OpcodeData<SizedEncodeInt<LEN>> for Vec<u8> {
 
     #[inline]
     fn serialize(from: &SizedEncodeInt<LEN>) -> Result<Self, SerializationError> {
-        Ok(serialize_i64(&from.0))
+        // This is an optimzation for LEN 8. Without it, the function will fail when checking the serialized data length.
+        if LEN == 8 && from.0 == i64::MIN {
```
michaelsutton, line 199: I argue that marginal optimizations of an error execution path which will never actually happen, are meaningless. I'd suggest explaining in a comment over the condition below `if bytes.len() > LEN` that even for `LEN == 8`, this will fail for `u64::MIN` and that it is by design

michaelsutton (changes requested):

crypto/txscript/src/data_stack.rs
michaelsutton: ok I see the comment above now

michaelsutton:
> 3. https://github.com/someone235/rusty-kaspa/blob/976313f0f4c492061689d84811b17ee1ea73cc8b/crypto/txscript/src/data_stack.rs#L173
>    Here we limit any `pop_items<i32>` to 4 bytes, disallowing the use of `i32::MIN`. This is fine since in any place we use `pop_items<i32>` we also return an error if the number is negative.

For the sake of accuracy, this was limited to 4 bytes also previously, through `DEFAULT_SCRIPT_NUM_LEN`.

crypto/txscript/src/data_stack.rs
michaelsutton: Resolving this since push_items with T=i32 is never called within the engine and hence serialize for OpData<i32> isn't called as well.

michaelsutton (approved):

PR #595, opened 2024-11-13
Create TN11 KIP10 HF activation and KIP9 beta switch

coderofstuff:
### Change

- Hard Fork Activation for:
  - KIP10 on TN11
  - KIP9 Beta version in Consensus
  - Payloads
- Minor cleanups

### Notes
- After this HF, all networks should now use `Kip9Version::Beta` for both mempool and consensus (if activated there). Perhaps a cleanup of KIP9 logic, removing all references to Alpha, is in order after said HF.


michaelsutton:
There is a remaining issue related to the sync gap between mempool state and virtual/block template state. Currently we set the tx mass (https://github.com/kaspanet/rusty-kaspa/blob/master/consensus/src/pipeline/virtual_processor/utxo_validation.rs#L343) when the transaction enters the mempool (which will be based on virtual daa score at the time), however when the transaction enters the block template, virtual might have already progressed ahead, reaching the hf activation, so the mass will contain the wrong commitment.

One obvious solution is to recalc the mass according to the final pov daa score at `validate_block_template_transaction`.

Another option is to ignore the whole thing and to rely on `build_block_template` to remove all the failing transactions and still make progress (if the mempool is full at the time of the hf this might clear the mempool eventually). But this option requires some sort of test to make sure there are no surprises

michaelsutton (approved):

PR #596, opened 2024-11-13
Bitwise-based RPC call limitations/permissions

KaffinPX:
Allows customization over exposed(allowed) RPC calls.

TODO:
- Expose it to CLI.
- Make sure 128 RPC calls are future-proof enough.
- Tests.
- Replace safe mode with specialized permission level by default.


KaffinPX:
telegram-cloud-photo-size-4-5974418355232294236-y

PR #597, opened 2024-11-14
calculate sig ops on fly

biryukovmaxim:
# Migrate to runtime signature operation counting

Changes signature operation counting from static analysis to runtime counting during script execution for improved accuracy.

## Key changes
- Add sig_op tracking fields to TxScriptEngine
- Remove static sig_op counting functions
- Count sig_ops during script execution 
- Add runtime checks for sig_op limits
- Add new error type for sig_op limit violations
- Update transaction validation to use runtime counting

## Breaking changes
- Changes sig_op validation timing from pre-execution to during execution
- May affect transaction fee estimation

## Todo
- [x] Add tests for sig_op calculation
- [x] Add documentation for new sig_op runtime counting behavior
- [ ] Consider adding sig_op estimation helper for fee calculation

Requires manual testing and careful review of sig_op counting logic.

### DIFF
--- a/crypto/txscript/errors/src/lib.rs
+++ b/crypto/txscript/errors/src/lib.rs
@@ -74,7 +74,7 @@ pub enum TxScriptError {
     #[error(transparent)]
     Serialization(#[from] SerializationError),
     #[error("sig op count exceed passed limit of {1}")]
-    SigOpCountTooLow(u16, u8),
+    ExceededSigOpLimit(u16, u8),
 }
 
 #[derive(Error, PartialEq, Eq, Debug, Clone, Copy)]
--- a/crypto/txscript/errors/src/lib.rs
+++ b/crypto/txscript/errors/src/lib.rs
@@ -73,6 +73,8 @@ pub enum TxScriptError {
     InvalidOutputIndex(i32, usize),
     #[error(transparent)]
     Serialization(#[from] SerializationError),
+    #[error("sig op count exceed passed limit of {1}")]
+    SigOpCountTooLow(u16, u8),
 }
 
 #[derive(Error, PartialEq, Eq, Debug, Clone, Copy)]
--- a/crypto/txscript/errors/src/lib.rs
+++ b/crypto/txscript/errors/src/lib.rs
@@ -73,8 +73,8 @@ pub enum TxScriptError {
     InvalidOutputIndex(i32, usize),
     #[error(transparent)]
     Serialization(#[from] SerializationError),
-    #[error("sig op count {0} exceeds passed limit of {1}")]
-    ExceededSigOpLimit(u16, u8),
+    #[error("sig op count exceeds passed limit of {0}")]
+    ExceededSigOpLimit(u8),
 }
 
 #[derive(Error, PartialEq, Eq, Debug, Clone, Copy)]
--- a/crypto/txscript/errors/src/lib.rs
+++ b/crypto/txscript/errors/src/lib.rs
@@ -73,7 +73,7 @@ pub enum TxScriptError {
     InvalidOutputIndex(i32, usize),
     #[error(transparent)]
     Serialization(#[from] SerializationError),
-    #[error("sig op count exceed passed limit of {1}")]
+    #[error("sig op count {0} exceeds passed limit of {1}")]
     ExceededSigOpLimit(u16, u8),
 }
 
--- a/crypto/txscript/src/lib.rs
+++ b/crypto/txscript/src/lib.rs
@@ -81,12 +81,69 @@ enum ScriptSource<'a, T: VerifiableTransaction> {
 /// Unlike the static counting approach which counts all possible signature operations,
 /// this tracks only the actually executed signature operations, leading to more accurate
 /// mass calculations and potentially lower fees for conditional scripts.
-#[derive(Debug, Clone, Copy)]
+#[derive(Debug, Clone)]
 pub struct RuntimeSigOpCounter {
     /// Maximum number of signature operations allowed for this input
     pub sig_op_limit: u8,
     /// Remaining signature operations that can be executed
     pub sig_op_remaining: u8,
+}
+
+impl RuntimeSigOpCounter {
+    /// Attempts to consume the specified number of signature operations.
+    ///
+    /// This method handles:
+    /// - Converting the number of signatures to u8 (failing if > 255)
+    /// - Checking if we have enough remaining operations
+    /// - Updating the remaining count if successful
+    ///
+    /// # Arguments
+    /// * `num_sigs` - The number of signature operations to consume
+    ///
+    /// # Returns
+    /// * `Ok(())` if the operations were successfully consumed
+    /// * `Err(TxScriptError::NumberTooBig)` if num_sigs > 255
+    /// * `Err(TxScriptError::ExceededSigOpLimit)` if not enough operations remain
+    ///
+    /// # Example
+    /// ```
+    /// let mut counter = RuntimeSigOpCounter {
+    ///     sig_op_limit: 10,
+    ///     sig_op_remaining: 5
+    /// };
+    ///
+    /// // Consume 3 operations
+    /// counter.consume_sig_ops(3)?; // Ok(())
+    /// assert_eq!(counter.sig_op_remaining, 2);
+    ///
+    /// // Try to consume too many
+    /// counter.consume_sig_ops(3)?; // Err(ExceededSigOpLimit)
+    /// ```
+    pub fn consume_sig_ops(&mut self, num_sigs: u8) -> Result<(), TxScriptError> {
+        self.sig_op_remaining =
+            self.sig_op_remaining.checked_sub(num_sigs).ok_or(TxScriptError::ExceededSigOpLimit(self.sig_op_limit))?;
+
+        Ok(())
+    }
+}
+
+pub trait SigOpConsumer {
+    fn consume_sig_ops(&mut self, num_sigs: u8) -> Result<(), TxScriptError>;
+}
+
+impl SigOpConsumer for RuntimeSigOpCounter {
+    fn consume_sig_ops(&mut self, num_sigs: u8) -> Result<(), TxScriptError> {
+        RuntimeSigOpCounter::consume_sig_ops(self, num_sigs)
+    }
+}
+impl SigOpConsumer for Option<RuntimeSigOpCounter> {
+    fn consume_sig_ops(&mut self, num_sigs: u8) -> Result<(), TxScriptError> {
+        if let Some(consumer) = self {
+            consumer.consume_sig_ops(num_sigs)
+        } else {
+            Ok(())
+        }
+    }
 }
 
 pub struct TxScriptEngine<'a, T: VerifiableTransaction, Reused: SigHashReusedValues> {
@@ -103,7 +160,7 @@ pub struct TxScriptEngine<'a, T: VerifiableTransaction, Reused: SigHashReusedVal
 
     num_ops: i32,
     kip10_enabled: bool,
-    runtime_sig_op_counting: Option<RuntimeSigOpCounter>,
+    runtime_sig_op_counter: Option<RuntimeSigOpCounter>,
 }
 
 fn parse_script<T: VerifiableTransaction, Reused: SigHashReusedValues>(
@@ -213,7 +270,7 @@ impl<'a, T: VerifiableTransaction, Reused: SigHashReusedValues> TxScriptEngine<'
             cond_stack: vec![],
             num_ops: 0,
             kip10_enabled,
-            runtime_sig_op_counting: None,
+            runtime_sig_op_counter: None,
         }
     }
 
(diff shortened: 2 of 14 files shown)


crypto/txscript/src/lib.rs
```diff
@@ -165,9 +213,18 @@ impl<'a, T: VerifiableTransaction, Reused: SigHashReusedValues> TxScriptEngine<'
...
             kip10_enabled,
+            runtime_sig_op_counting: None,
         }
     }
 
+    /// Returns the number of signature operations that were actually required during script execution.
+    /// Must only be called after script execution completes successfully.
+    ///
+    /// Returns the difference between the input's sig_op_limit and remaining sig ops.
+    pub fn sig_op_required(&self) -> Option<u8> {
```
someone235, line 224: sig_op->sig_ops
Also, maybe `used_sig_ops` is a better name?

crypto/txscript/src/lib.rs
```diff
@@ -225,6 +285,7 @@ impl<'a, T: VerifiableTransaction, Reused: SigHashReusedValues> TxScriptEngine<'
             cond_stack: Default::default(),
             num_ops: 0,
             kip10_enabled,
+            runtime_sig_op_counting: None, // todo?
```
someone235, line 288: I don't think it's needed, but explain why in a comment

crypto/txscript/src/lib.rs
```diff
@@ -973,6 +1047,231 @@ mod tests {
...
+
+    #[derive(Clone)]
+    struct SignatureData {
+        signature: Vec<u8>,
+        public_key: Vec<u8>,
+    }
+
+    enum SignatureScriptBuilder {
+        Single(SignatureData),
+        Multi(Vec<SignatureData>),
```
someone235, line 1059: I think Multisig gives a clearer context. And also comment on each variant would be nice

crypto/txscript/src/lib.rs
```diff
@@ -405,8 +466,13 @@ impl<'a, T: VerifiableTransaction, Reused: SigHashReusedValues> TxScriptEngine<'
         } else if num_sigs > num_keys {
             return Err(TxScriptError::InvalidSignatureCount(format!("more signatures than pubkeys {num_sigs} > {num_keys}")));
         }
+        if let Some(RuntimeSigOpCounter { sig_op_remaining, sig_op_limit }) = &mut self.runtime_sig_op_counting {
+            let num_sigs =
+                num_sigs.try_into().map_err(|_| TxScriptError::NumberTooBig("Signatures count mustn't exceed 255".to_string()))?;
+            *sig_op_remaining =
+                sig_op_remaining.checked_sub(num_sigs).ok_or(TxScriptError::ExceededSigOpLimit(num_sigs as u16, *sig_op_limit))?;
+        }
```
someone235, line 474: I think it would improve readability if `RuntimeSigOpCounter` would have a method that does all of this

crypto/txscript/src/lib.rs
```diff
@@ -973,6 +1047,231 @@ mod tests {
...
+                sig_builder: Box::new(move |tx, reused| {
+                    let sig = create_schnorr_signature(tx, reused);
+                    SignatureScriptBuilder::Multi(vec![sig.clone(), sig])
+                }),
+                expected_sig_ops: 2,
+                sig_op_limit: 2,
+                should_pass: true,
+            },
+            TestCase {
+                name: "Mixed Schnorr and ECDSA",
```
someone235, line 1183: I think it's better not to repeat test names. In this case, adding " - Above limit" or " - Exceeds limit" would be helpful (same for similar tests)

crypto/txscript/src/opcodes/mod.rs
```diff
@@ -720,6 +720,10 @@ opcode_list! {
         match sig.pop() {
             Some(typ) => {
                 let hash_type = SigHashType::from_u8(typ).map_err(|e| TxScriptError::InvalidSigHashType(typ))?;
+                if let Some(RuntimeSigOpCounter{ sig_op_limit, sig_op_remaining} ) = &mut vm.runtime_sig_op_counting {
+                   *sig_op_remaining = sig_op_remaining.checked_sub(1).ok_or(TxScriptError::ExceededSigOpLimit(*sig_op_limit as u16 + 1, *sig_op_limit))?;
+                }
+
```
someone235, line 726: Same comment about a method

crypto/txscript/src/opcodes/mod.rs
```diff
@@ -743,6 +747,9 @@ opcode_list! {
         match sig.pop() {
             Some(typ) => {
                 let hash_type = SigHashType::from_u8(typ).map_err(|e| TxScriptError::InvalidSigHashType(typ))?;
+                if let Some(RuntimeSigOpCounter{ sig_op_limit, sig_op_remaining} ) = &mut vm.runtime_sig_op_counting {
+                   *sig_op_remaining = sig_op_remaining.checked_sub(1).ok_or(TxScriptError::ExceededSigOpLimit(*sig_op_limit as u16 + 1, *sig_op_limit))?;
+                }
```
someone235, line 752: Same comment about a method

someone235 (changes requested):

crypto/txscript/src/lib.rs
biryukovmaxim: 2175dbe5481c4e9e4ff375b2c1d8c7c435e05bfc

crypto/txscript/errors/src/lib.rs
```diff
@@ -73,6 +73,8 @@ pub enum TxScriptError {
     InvalidOutputIndex(i32, usize),
     #[error(transparent)]
     Serialization(#[from] SerializationError),
+    #[error("sig op count exceed passed limit of {1}")]
```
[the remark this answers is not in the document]
biryukovmaxim: 183e859ebf7ff31c0e45d0a7505d6b68a73df418

crypto/txscript/errors/src/lib.rs
```diff
@@ -73,6 +73,8 @@ pub enum TxScriptError {
...
+    ExceededSigOpLimit(u16, u8),
```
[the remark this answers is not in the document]
biryukovmaxim: 183e859ebf7ff31c0e45d0a7505d6b68a73df418

crypto/txscript/src/lib.rs
```diff
@@ -94,6 +112,36 @@ fn parse_script<T: VerifiableTransaction, Reused: SigHashReusedValues>(
     script.iter().batching(|it| deserialize_next_opcode(it))
 }
 
+pub fn get_sig_op_count_via_simulation<T: VerifiableTransaction, Reused: SigHashReusedValues>(
```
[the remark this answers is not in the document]
biryukovmaxim: i was aiming to use it in test, but made them without the function.
the idea is adding some helper function that can return sig op count that is needed for execution by executing it with TxEngine help

crypto/txscript/src/lib.rs
biryukovmaxim: 6ba349b096f149002f4eb4ba2afaeb479b68273c

crypto/txscript/src/lib.rs
biryukovmaxim: dfc20f7043ceb6eb79633905b41a2a436b0031e5

crypto/txscript/src/opcodes/mod.rs
biryukovmaxim: 78dfebd72d7e6a3c0bc0ac3bf410caf9218f7970

crypto/txscript/src/opcodes/mod.rs
biryukovmaxim: 78dfebd72d7e6a3c0bc0ac3bf410caf9218f7970

crypto/txscript/src/lib.rs
biryukovmaxim: 78dfebd72d7e6a3c0bc0ac3bf410caf9218f7970

crypto/txscript/src/lib.rs
biryukovmaxim: f032de3a7ae364278470a15aec4ebdd64c47e9b4

crypto/txscript/src/lib.rs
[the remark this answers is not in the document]
someone235: If the function is only for tests, delete it since it's not used.

If you think it's somehow good for users, add documentation that explains what it does

crypto/txscript/src/lib.rs
```diff
@@ -81,14 +81,69 @@ enum ScriptSource<'a, T: VerifiableTransaction> {
 /// Unlike the static counting approach which counts all possible signature operations,
 /// this tracks only the actually executed signature operations, leading to more accurate
 /// mass calculations and potentially lower fees for conditional scripts.
-#[derive(Debug, Clone, Copy)]
+#[derive(Debug, Clone)]
 pub struct RuntimeSigOpCounter {
     /// Maximum number of signature operations allowed for this input
     pub sig_op_limit: u8,
     /// Remaining signature operations that can be executed
     pub sig_op_remaining: u8,
```
someone235, line 89: Maybe it shouldn't be `pub` anymore?

crypto/txscript/src/lib.rs
```diff
@@ -466,12 +524,9 @@ impl<'a, T: VerifiableTransaction, Reused: SigHashReusedValues> TxScriptEngine<'
...
             return Err(TxScriptError::InvalidSignatureCount(format!("more signatures than pubkeys {num_sigs} > {num_keys}")));
         }
-        if let Some(RuntimeSigOpCounter { sig_op_remaining, sig_op_limit }) = &mut self.runtime_sig_op_counting {
-            let num_sigs =
-                num_sigs.try_into().map_err(|_| TxScriptError::NumberTooBig("Signatures count mustn't exceed 255".to_string()))?;
-            *sig_op_remaining =
-                sig_op_remaining.checked_sub(num_sigs).ok_or(TxScriptError::ExceededSigOpLimit(num_sigs as u16, *sig_op_limit))?;
-        }
+        let num_sigs =
+            num_sigs.try_into().map_err(|_| TxScriptError::NumberTooBig("Signatures count mustn't exceed 255".to_string()))?;
```
someone235, line 528: I had a suspicion that this can create a netsplit, since `num_sigs` arrived from the stack as `i32` so it can theoretically be bigger than 255. But after some reading of the context it seems that this error is actually unreachable, since `num_sigs` is bounded by `num_keys` which is bounded by `MAX_PUB_KEYS_PER_MUTLTISIG=20`.

But just to be sure, can you make a multisig test where the hardfork is inactive and `num_sigs=256`? I expect to see it fails with `Err(TxScriptError::InvalidPubKeyCount(format!("too many pubkeys {num_keys} > {MAX_PUB_KEYS_PER_MUTLTISIG}")))` or `Err(TxScriptError::InvalidSignatureCount(format!("more signatures than pubkeys {num_sigs} > {num_keys}")))`

someone235 (changes requested):

crypto/txscript/src/lib.rs
biryukovmaxim: actually it will face first  MAX_SCRIPT_ELEMENT_SIZE then MAX_STACK_SIZE and only after that, it will fail with error related to MAX_PUB_KEYS_PER_MUTLTISIG.

crypto/txscript/src/lib.rs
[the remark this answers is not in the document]
biryukovmaxim: 4f14ee7e72c87f0523d8d20adcbb4590047f5766

crypto/txscript/src/lib.rs
biryukovmaxim: d23e817a777f25762696b75a7ecdff1762862369

crypto/txscript/src/lib.rs
biryukovmaxim: 7fedccaedf601971b284e7d0659576961835876f

someone235 (approved):

crypto/txscript/src/lib.rs
```diff
@@ -207,6 +361,8 @@ impl<'a, T: VerifiableTransaction, Reused: SigHashReusedValues> TxScriptEngine<'
             cond_stack: Default::default(),
             num_ops: 0,
             kip10_enabled,
+            runtime_sig_op_counter: runtime_sig_op_counting
+                .then_some(RuntimeSigOpCounter { sig_op_limit: input.sig_op_count, sig_op_remaining: input.sig_op_count }),
```
michaelsutton, line 365: we should use `RuntimeSigOpCounter::new` to avoid the manual double mention of `input.sig_op_count` 
(nit: I'd say perhaps this is a reason to move `RuntimeSigOpCounter ` to an inner `mod` to avoid access to the private fields)

michaelsutton:
This is brilliant work btw.

someone235 (approved):

crypto/txscript/src/lib.rs
```diff
@@ -94,8 +100,72 @@ fn parse_script<T: VerifiableTransaction, Reused: SigHashReusedValues>(
...
+/// # Arguments
+/// * `tx` - The transaction containing the input to analyze
+/// * `input_idx` - Index of the input to analyze
+/// * `kip10_enabled` - Whether KIP-10 features are enabled
+///
+/// # Returns
+/// * `Ok(u8)` - The exact number of signature operations executed
+/// * `Err(TxScriptError)` - If script execution fails or input index is invalid
+pub fn get_sig_op_count<T: VerifiableTransaction>(tx: &T, input_idx: usize, kip10_enabled: bool) -> Result<u8, TxScriptError> {
+    let sig_cache = Cache::new(10_000);
```
michaelsutton, line 135: although `get_sig_op_count` is not used in the current codebase, it seem redundant to actually allocate a cache here. I think `size` should be `0` here (there are no expected cache hits).

michaelsutton (approved):

PR #598, opened 2024-11-15
IBD sync: recover sampled window

michaelsutton:
Fixes a rare bug in TN11 where the syncer does not send sufficient DAA window blocks required for recovering the sampled DAA window. 
The sampling mechanism relies on DAA indexes which can only be calculated correctly if the full mergesets covering all sampled blocks are sent.  

Note: fixes a (future) bug from the recent PR where cache entries were not filtered by origin. See [commit](https://github.com/kaspanet/rusty-kaspa/pull/598/commits/f76c71ca3c4bc4c62e5c5c2adf93c00d96c8614a).

Edit: added also a refactor to tx time lock validation and some long due renames for various tx validation modules


consensus/src/processes/transaction_validator/tx_validation_in_header_context.rs
```diff
@@ -0,0 +1,81 @@
...
+
+use super::{
+    errors::{TxResult, TxRuleError},
+    TransactionValidator,
+};
+
+pub(crate) enum LockTimeType {
+    Finalized,
+    DaaScore,
+    Time,
```
D-Stacks, line 13: can we rename this to MedianTime?

coderofstuff (approved):
I'm approving this based on my read that each commit here accomplishes the intended change of:
- ensuring cache reader can only get a window if the window's origin matches the reader's
- the locktime logic doesn't leak outside of the locktime checks
- documentation and refactors/rename

consensus/src/processes/transaction_validator/tx_validation_in_header_context.rs
michaelsutton: I'm actually not sure. Median time is merely the way to get a consensus-deterministic estimation of block time (without depending on block timestamp), but the conceptual lock time type is simply "time"

coderofstuff (approved):

someone235 (approved):

PR #599, opened 2024-11-15
Track the average transaction mass throughout the mempool's lifespan

michaelsutton:
Tracks the average transaction mass throughout the mempool's lifespan using a decaying formulae. 

The intention is to avoid sensitivity to cases where the mempool has only a few transactions with unusually high mass, which affects the fee estimation mechanism (block slots in the estimator are defined by the average mass). 

By using a decaying formula over all history we get:

1. The average reflects samples obtained over a long duration
2. It still gives more weight to more recent samples (eventually "forgetting" old enough samples)
3. The formula for this calculation is very simple and requires no special bookkeeping


mining/src/mempool/model/frontier.rs
```diff
@@ -394,13 +410,13 @@ mod tests {
...
             let estimator = frontier.build_feerate_estimator(args);
-            let estimations = estimator.calc_estimations(1.0);
+            let estimations = estimator.calc_estimations(MIN_FEERATE);
 
             let buckets = estimations.ordered_buckets();
             // Test for the absence of NaN, infinite or zero values in buckets
             for b in buckets.iter() {
                 assert!(
-                    b.feerate.is_normal() && b.feerate >= 1.0,
+                    b.feerate.is_normal() && b.feerate >= MIN_FEERATE - EPS,
```
coderofstuff, line 419: Why is the epsilon here necessary?

mining/src/mempool/model/frontier.rs
```diff
@@ -62,6 +72,11 @@ impl Frontier {
         let mass = key.mass;
         if self.search_tree.insert(key) {
             self.total_mass += mass;
+            // A decaying average formula. Denote ɛ = 1 - AVG_MASS_DECAY_FACTOR. A transaction inserted N slots ago has
+            // ɛ * (1 - ɛ)^N weight within the updated average. This gives some weight to the full mempool history while
+            // giving growing importance to more recent samples.
```
coderofstuff, line 77: When I first read the "giving growing importance" part, I had an impression that the more recent examples continue to increase in importance as time passes (as in something continues to give them importance, making it grow) which is actually the opposite of what happens.

coderofstuff (commented):
I definitely like this approach over the current one. Still need to spend time actually running it, but the change looks sound to me.

mining/src/mempool/model/frontier.rs
michaelsutton: I've seen it fail in CI tests (very rarely, but it happened) due to platform specific floating op precision

coderofstuff (approved):

PR #601, opened 2024-11-21
Ignore cache if policy is empty

D-Stacks:
just a small PR, avoid utilizing a cache, if the cache policy is empty..


D-Stacks:
saw that ops are filtered by max tracked items.

PR #609, opened 2024-12-04
KIP6 logic

freshair18:
Contains most consensus layer infrastracture of Kip6 - with the goal of allowing users to generate and store everlasting proofs of tx inclusion or tx acceptance.
Kip6 requires adding an extra field to the header, this will be covered in a future PR,
for now the required new field is  locally simulated with a test only storage entry called pchmr_store.
New functionality is generally not called to save for testing, but a single placeholder call to calc_pchmr_root is made when building block template (with its value discarded) as a placeholder and in order to see if it has adverse effect on anything.


freshair18:
cryptographic receipts are given by https://github.com/kaspanet/rusty-kaspa/pull/725, obsoleting this PR.

Code logic may still be relevant in the future if logarithmic shortcuts are deemed necessary in a future HF, but for now this day seems far off.

Closing

PR #628, opened 2025-02-07
Mining Rule Engine - Sync Monitor component

coderofstuff:
This is the Sync Monitor component of the mining rule engine discussed in https://research.kas.pa/t/sync-monitor-mining-rule-engine/294

Concepts:
- Sync Rate - observed_number_of_blocks_received_per_interval / expected_number_of_blocks_to_receive_per_interval

Changes:
1. Create a new Mining Rule Engine that currently tracks the sync rate of the node
2. Introduce the `Sync Rate Rule` - allow mining if the sync rate of the node is lower than the sync rate threshold and the finality point is recent, even if the node is not "nearly synced"
3. Move all is_nearly_synced logic to the rule engine area


coderofstuff:
Superceded by #634

PR #634, opened 2025-02-12
Mining Rule Engine

coderofstuff:
# Mining Rule Engine

This is the full implementation the of the mining rule engine discussed in https://research.kas.pa/t/sync-monitor-mining-rule-engine/294

## Concepts
- Sync Rate - observed_number_of_blocks_received_per_interval / expected_number_of_blocks_to_receive_per_interval
- Mining Rule - some rule that affect how and when a node should mine based on the conditions of the node

### Outline

1. Create a new Mining Rule Engine that currently tracks the sync rate of the node
2. Move all is_nearly_synced logic to the rule engine area
3. Introduce "Mining Rules" which are checked every 10s and depending on the conditions of the node triggers rules to be applied.

#### 1. Implements a BlueOnlyParents strategy
  - Trigger: virtual processing time average is above 500ms
  - Recovery:
    - virtual processing time is back below 100ms AND
    - a merge bound amount of blocks has passed since the trigger
#### 2. Implements a NoTransactions (empty blocks) strategy
  - Trigger: BadMerkleRoot count > valid block count in the interval
  - Recovery: 2 intervals have passed since the rule was triggered

#### 3. Implements a SyncRateRule strategy

This is the implementation of the Sync Monitor discussed in the thread and doc above.

1. Create a new Mining Rule Engine that currently tracks the sync rate of the node
2. Introduce the `Sync Rate Rule` - allow mining if the sync rate of the node is lower than the sync rate threshold and the finality point is recent, even if the node is not "nearly synced"
4. Move all is_nearly_synced logic to the rule engine area


michaelsutton:
virtual processing time does not cover things such as `calc_block_parents` which are only called as part of `build_block_template_from_virtual_state`. Perhaps that method should be added to the same timing counter.  

In general I also think that virtual processing time needs to be normalized by the amount of UTXO-validated blocks during the resolve. This can be solved by changing `self.counters.virtual_resolve_counts.fetch_add(1, Ordering::Relaxed)` to fetch add such a count (and renaming the field). See `chain_block_counts` for a reference of such counting. 

Other option (which I prefer): ignore UTXO validation and time only the parts of resolve virtual which prep for block template building: from before `pick_virtual_parents` and until after `calculate_and_commit_virtual_state`. This segment includes all sensitive header composing areas and, added with the time of `build_block_template_from_virtual_state`, should suffice for our cause.

One more comment re blue parents only rule: once triggered, this rule remains for a long period. So it seems wise to activate it only after a few consecutive slow rounds and not only after 10 seconds (or extend the monitor interval to 30 secs?)

michaelsutton:
The general structuring of the rules and the rule engine in terms of code design and style is top-notch btw

consensus/src/pipeline/body_processor/processor.rs
```diff
@@ -193,6 +193,11 @@ impl BlockBodyProcessor {
         let mass = match self.validate_body(block, is_trusted) {
             Ok(mass) => mass,
             Err(e) => {
+                // Count the number of bad merkle root errors as this may trigger NoTransactions rule
+                if matches!(e, RuleError::BadMerkleRoot(_, _)) {
+                    self.counters.bad_merkle_root_count.fetch_add(1, Ordering::Relaxed);
```
michaelsutton, line 198: `BadMerkleRoot` errors are not persisted (as can be seen right below), so this is too easy to artificially increase...

protocol/mining/src/rules/no_transactions_rule.rs
```diff
@@ -0,0 +1,68 @@
...
+                } else {
+                    trace!(
+                        "NoTransactionsRule: recovering | Bad Merkle Root Count: {} | Valid Body Count: {} | Cooldown: {}",
+                        delta.bad_merkle_root_count,
+                        delta.body_counts,
+                        cooldown_count
+                    );
+                }
+            }
+        } else if delta.bad_merkle_root_count > delta.body_counts {
```
michaelsutton, line 49: This condition seems to indicate that majority of blocks have a bad Merkle root (assuming the unique counting problem from the other comment is addressed). Is that the intention?

michaelsutton (changes requested):

protocol/mining/src/rule_engine.rs
```diff
@@ -0,0 +1,184 @@
...
+    pub fn is_nearly_synced(&self, sink_daa_score_timestamp: DaaScoreTimestamp) -> bool {
+        let sink_timestamp = sink_daa_score_timestamp.timestamp;
+
+        if self.config.net.is_mainnet() {
+            // We consider the node close to being synced if the sink (virtual selected parent) block
+            // timestamp is within DAA window duration far in the past. Blocks mined over such DAG state would
+            // enter the DAA window of fully-synced nodes and thus contribute to overall network difficulty
+            //
+            // [Crescendo]: both durations are nearly equal so this decision is negligible
+            unix_now()
```
michaelsutton, line 139: This duration is too long for 10BPS (~44 minutes = 26,400 block depth). To me, part of the motivation for identifying slow sync rate was to allow the definition of nearly synced to be tightened (as in testnets below). It's safer now bcs the network will still mine on a slowdown due to the new rule even if nearly synced = false.

protocol/mining/src/rules/no_transactions_rule.rs
coderofstuff: That is the intent. If there were more blocks that resulted in BadMerkleRoot errors in the last duration than successfully validated blocks, then trigger this rule.

protocol/mining/src/rule_engine.rs
coderofstuff: Should we set it to as tight a bound as the one we have for testnet below? So, change this `fn` to behave the same way in all networks?

If it used to be 44 minutes at 1bps, something like 4.4mins at 10bps is an overwhelming amount of time for blocks not to pass. Perhaps even tighter?

If we consider 1min as the nearly synced bound, what are the consequences of that?

protocol/mining/src/rules/blue_parents_only_rule.rs
```diff
@@ -0,0 +1,66 @@
...
+use super::{mining_rule::MiningRule, ExtraData};
+
+const VIRTUAL_PROCESSING_TRIGGER_THRESHOLD: f64 = 500.0; // 500 milliseconds
+const VIRTUAL_PROCESSING_RECOVERY_THRESHOLD: f64 = 100.0; // 100 milliseconds
+
+/// BlueParentsOnlyRule
+/// Attempt to recover from high virtual processing times possibly caused merging red blocks.
+/// by only pointing to blue parents.
+///
+/// Trigger: virtual processing average time is above threshold
```
someone235, line 18: I think slowness in `build_block_template` should trigger it. @hashdag and I also suggested slowness in header validation to trigger it

protocol/mining/src/rules/sync_rate_rule.rs
```diff
@@ -0,0 +1,67 @@
...
+    atomic::{AtomicBool, Ordering},
+    Arc,
+};
+
+use kaspa_consensus_core::api::counters::ProcessingCountersSnapshot;
+use kaspa_core::{time::unix_now, trace, warn};
+
+use super::{mining_rule::MiningRule, ExtraData};
+
+const SYNC_RATE_THRESHOLD: f64 = 0.10;
```
someone235, line 11: I think this should be 0.95, since processing 10 blocks per second on IBD is a very pessimistic estimation in the first place

rpc/service/src/service.rs
```diff
@@ -472,13 +473,14 @@ NOTE: This error usually indicates an RPC conversion error between the node and
...
-        let is_nearly_synced = self.consensus_manager.consensus().unguarded_session().async_is_nearly_synced().await;
+        let sink_daa_score_timestamp =
+            self.consensus_manager.consensus().unguarded_session().async_get_sink_daa_score_timestamp().await;
         Ok(GetInfoResponse {
             p2p_id: self.flow_context.node_id.to_string(),
             mempool_size: self.mining_manager.transaction_count_sample(TransactionQuery::TransactionsOnly),
             server_version: version().to_string(),
             is_utxo_indexed: self.config.utxoindex,
-            is_synced: self.has_sufficient_peer_connectivity() && is_nearly_synced,
+            is_synced: self.mining_rule_engine.should_mine(sink_daa_score_timestamp),
```
someone235, line 483: Not sure if this should be changed. In non mining contexts I would probably consider the node as not synced. It needs more discussion

rpc/service/src/service.rs
```diff
@@ -1142,7 +1144,8 @@ NOTE: This error usually indicates an RPC conversion error between the node and
         _request: GetServerInfoRequest,
     ) -> RpcResult<GetServerInfoResponse> {
         let session = self.consensus_manager.consensus().unguarded_session();
-        let is_synced: bool = self.has_sufficient_peer_connectivity() && session.async_is_nearly_synced().await;
+        let sink_daa_score_timestamp = session.async_get_sink_daa_score_timestamp().await;
+        let is_synced: bool = self.mining_rule_engine.should_mine(sink_daa_score_timestamp);
```
someone235, line 1148: Same here

rpc/service/src/service.rs
```diff
@@ -1161,8 +1164,9 @@ NOTE: This error usually indicates an RPC conversion error between the node and
         _connection: Option<&DynRpcConnection>,
         _request: GetSyncStatusRequest,
     ) -> RpcResult<GetSyncStatusResponse> {
-        let session = self.consensus_manager.consensus().unguarded_session();
-        let is_synced: bool = self.has_sufficient_peer_connectivity() && session.async_is_nearly_synced().await;
+        let sink_daa_score_timestamp =
+            self.consensus_manager.consensus().unguarded_session().async_get_sink_daa_score_timestamp().await;
+        let is_synced: bool = self.mining_rule_engine.should_mine(sink_daa_score_timestamp);
```
someone235, line 1169: And here

protocol/mining/src/rules/sync_rate_rule.rs
coderofstuff: New agreed value is 0.9. Change the sampling to use a 5 minute window, with samples added every 10s.

consensus/src/pipeline/body_processor/processor.rs
coderofstuff: Moved the counting to RPC submit block context

protocol/mining/src/rules/blue_parents_only_rule.rs
```diff
@@ -0,0 +1,66 @@
...
+/// Recovery: virtual processing average time is below threshold and a merge depth bound has passed
```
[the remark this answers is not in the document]
coderofstuff: Any specific suggested recovery condition? On the trigger, @michaelsutton also made a similar comment that trigger should be dependent on BBT only (not the entirety of virtual processing). Not sure if header processing should be considered here though.

rpc/service/src/service.rs
michaelsutton: idk, doesn't slow sync rate truly indicate that you are synced even if sink timestamp is lagging?

protocol/mining/src/rules/no_transactions_rule.rs
coderofstuff: The rule has been updated to trigger if:
- submit_block_bad_merkle_root_count > 0 && submit_block_successful_count == 0 (there were submit blocks in the period that resulted in BadMerkleRoot and there were no successful submissions).

coderofstuff:
Closing this as it has been split in #654 #655 #656

rpc/service/src/service.rs
coderofstuff: Definitely needs more discussion. I don't think it has to be changed, and prefer to keep it in line with using `should_mine`. We can tackle a full-on rename as a follow-up PR too.

I want to make sure we change behavior here depending on what the caller would want of this info anyway.

protocol/mining/src/rules/blue_parents_only_rule.rs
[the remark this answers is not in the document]
coderofstuff: This is still an open question, but let's discuss it in #656

PR #635, opened 2025-02-12
Ibd/streams.rs HeaderChunkStream next() function refactor

miningexperiments:
## Refactoring
Tried to make it less nested and more readable

## Testing
- Tested it to sync headers on testnet11 without problems


freshair18 (approved):
functionality remains the same as it was.

This refactor appears more comprehensible and more idiomatic to rust.

PR #637, opened 2025-02-13
Bug fixes: ungraceful test crash fixes

ghost:
## Bug fix
- When running cargo tests the testing is crashed ungracefully leaving temp databases and such in tens of gigabytes, if they are run a few times, and still crashing.
- This mainly due errors not propagating to the test functions, but panicking early.
- tokio was underflowing UtxoChangeSubscription,

## What I've done
- Propagated errors from temp dir creation, create_temp_db macro, unwrapper errors in test functions.
- New test functions with consensus.init() were not shutdown, and they were Arc'd so could possibly stay haunting after test is over?
- connection_handler.rs serve function was forcefully panicked, I changed this to error
- added math check to UtxoChangeSubscription

## Additional thoughts
- [protocol/p2p/src/core/connection_handler.rs](https://github.com/kaspanet/rusty-kaspa/compare/master...miningexperiments:rusty-kaspa:testing-fixes?expand=1#diff-5bf29ee7444f7470df1524354eb1db288971e3cc13462c87bcb566aeeb0e81b5) This kept going on for 15 tries, and then shut. Perhaps another check should be added so it can close earlier in the test.
  - It is common for servers and e.g. VPNs to not use ipv6, so graceful error propagation is needed in testing (not panic)
- cargo test without --release is still complaining in current master branch (before any pr changes). This is not tested in ci.yaml.
```sh
error: 2 targets failed:
    `-p kaspa-testing-integration --lib`
    `-p kaspa-wallet-core --lib`
    ```

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -114,7 +114,7 @@ jobs:
           key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
 
       - name: Run cargo tests non--release
-        run: cargo nextest --no-fail-fast
+        run: cargo nextest run --no-fail-fast
 
       - name: Run cargo build with devnet-prealloc feature
         run: cargo build --release --features devnet-prealloc --workspace --all --tests --benches

### CODE REVIEW

**Comment by miningexperiments** [comment]
this is fixed in PR: `test daemon_integration_tests::daemon_utxos_propagation_test ... ok` 

The other test failing is the
```rs
 thread 'tx::generator::test::test_generator_inputs_250k_outputs_2_sweep' has overflowed its stack
fatal runtime error: stack overflow
```
I think it might be just that the vector with 250k elements is in stack, so it overflows. You could allocate it in heap with  fixed size, but then you are fixing the test, and not sure how it reflects to testing the making of utxo_entries in wallet/core/src/tx/generator/generator.rs @michaelsutton @coderofstuff  @someone235

notify/src/subscription/single.rs
```diff
@@ -383,11 +383,17 @@ impl Display for UtxosChangedSubscription {
 
 impl Drop for UtxosChangedSubscription {
     fn drop(&mut self) {
-        trace!(
-            "UtxosChangedSubscription: {} in total (drop {})",
-            UTXOS_CHANGED_SUBSCRIPTIONS.fetch_sub(1, Ordering::SeqCst) - 1,
-            self
-        );
+        // Prevent underflow; [ERROR] thread 'tokio-runtime-worker' panicked at notify/src/subscription/single.rs:388:13: attempt to subtract with overflow
+        let _ = UTXOS_CHANGED_SUBSCRIPTIONS.fetch_update(Ordering::SeqCst, Ordering::SeqCst, |count| {
```
coderofstuff, line 387: What if we just did:

```
UTXOS_CHANGED_SUBSCRIPTIONS.fetch_sub(1, Ordering::SeqCst).saturating_sub(1)
```

Also, I wonder what the root cause of `UTXOS_CHANGED_SUBSCRIPTIONS` being `0` is. The above fixes the underflow, but the incorrect usage that allowed the situation still needs to be determined. the expectation here is that `UTXOS_CHANGED_SUBSCRIPTIONS` should always be above 0 when drop occurs OR document the situations when it can be 0.

.github/workflows/ci.yaml
```diff
@@ -113,6 +113,9 @@ jobs:
             target/
           key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
 
+      - name: Run cargo tests non--release
```
coderofstuff, line 116: We moved away from `test` and switched to using `nextest`. But I do wonder why nextest doesn't capture the errors that test is displaying here.

protocol/p2p/src/core/connection_handler.rs
```diff
@@ -85,9 +85,11 @@ impl ConnectionHandler {
                 .serve_with_shutdown(serve_address.into(), termination_receiver.map(drop))
                 .await;
 
+            // log:error! was panic! before, but crashes testing ungracefully (error! not implemented)
+            // Now error is propagated
             match serve_result {
                 Ok(_) => info!("P2P Server stopped: {}", serve_address),
-                Err(err) => panic!("P2P, Server {serve_address} stopped with error: {err:?}"),
+                Err(err) => log::error!("P2P, Server {serve_address} stopped with error: {err:?}"),
```
coderofstuff, line 92: Add error in the `use` at line 8 and just use that here.

testing/integration/src/consensus_integration_tests.rs
```diff
@@ -1820,7 +1821,7 @@ async fn run_kip10_activation_test() {
     let mut genesis_multiset = MuHash::new();
     consensus.append_imported_pruning_point_utxos(&initial_utxo_collection, &mut genesis_multiset);
     consensus.import_pruning_point_utxo_set(config.genesis.hash, genesis_multiset).unwrap();
-    consensus.init();
+    let wait_handles = consensus.init();
```
coderofstuff, line 1824: Add comments on why this is necessary

testing/integration/src/consensus_integration_tests.rs
```diff
@@ -1872,6 +1873,8 @@ async fn run_kip10_activation_test() {
     let status = consensus.add_utxo_valid_block_with_parents((index + 1).into(), vec![index.into()], vec![spending_tx.clone()]).await;
     assert!(matches!(status, Ok(BlockStatus::StatusUTXOValid)));
     assert!(consensus.lkg_virtual_state.load().accepted_tx_ids.contains(&tx_id));
+    consensus.shutdown(wait_handles);
```
coderofstuff, line 1876: Same here and the rest of similar changes: Add comments on why this is necessary.

coderofstuff (changes requested):
Initial review. I'll probably on be able to get back to this after the next few weeks but please take a look.

testing/integration/src/consensus_integration_tests.rs
miningexperiments: this was done in previous tests as well. 
Here
```rs
  pub fn init(&self) -> Vec<JoinHandle<()>> {
        self.consensus.run_processors()
    }
```
and
```rs
    pub fn shutdown(&mut self) {
        self.core.shutdown();
        self.join();
    }
 ```

where core is wrapped in Arc<core>, so its reference counted. Without it some background threads might stay running, thought it might be related to getting System File Descriptor problems and not getting /tmp/rusty-kaspa cleanups early.

notify/src/subscription/single.rs
miningexperiments: afaik saturating_sub only applies to old value from fetch_sub not the atomic itself, so it doesnt prevent actual atomic variable from becoming negative. Suggested fetch_sub solution only applies if the value is greater than zero

miningexperiments: image
btw this happens frequently in testing

PR #641, opened 2025-02-23
Generalized KIP9--storage mass plurality (Crescendo)

michaelsutton:
Generalizes KIP9 to handle the case where the UTXO script public key is larger than standard keys (max 33 bytes for ECDSA) and thus captures more (persistent) storage.

The idea is based on the following simple reduction:

- Define the base UTXO storage unit (`UTXO_UNIT=100` bytes seems a natural choice)
- For a UTXO entry, let `P := ceiling(entry.size/UTXO_UNIT)` denote the UTXO's "plurality"
- Treat the entry as `P` entries each holding `entry.amount/P` KAS value 

This directly maps the usage of such a big entry to the usage of `P` single-unit entries in a model where all entries have the same size.

The choice of `UTXO_UNIT=100` is driven by the const parts of the UTXO = 63 bytes + max standard public key which is 33 bytes.

Implementing the above reduction in the storage mass calculation is straightforward. The harmonic part of the formula is $\frac{|O|}{H(O)} = \sum_{o \in O} \frac{1}{o}$. For each entry we now generalize to $\frac{P}{o/P} = \frac{P^2}{o}$. The arithmetic part of the formula is $\frac{|I|}{A(I)} = \frac{|I|^2}{sum(I)}$, and is generalized to $\frac{sum(plurality(I))^2}{sum(I)}$ where $sum(plurality(I)) = \sum_{i \in I} plurality(i)$. Note that $sum(I)$ can remain unchanged.

### DIFF
--- a/consensus/core/src/mass/mod.rs
+++ b/consensus/core/src/mass/mod.rs
@@ -103,14 +103,28 @@ impl UtxoPlurality for TransactionOutput {
     }
 }
 
-/// Defines an abstract UTXO storage cell.
+/// An abstract UTXO storage cell.
 ///
-/// The cell includes plurality and amount information required for calculating the storage mass.
+/// # Plurality
+///
+/// Each `UtxoCell` now has a `plurality` field reflecting how many 100-byte "storage units"
+/// this UTXO effectively occupies. This generalizes KIP-0009 to support UTXOs with
+/// script public keys larger than the standard 33-byte limit. For a UTXO of byte-size
+/// `entry.size`, we define:
+///
+/// ```text
+/// P := ceil(entry.size / UTXO_UNIT)
+/// ```
+///
+/// Conceptually, we treat a large UTXO as `P` sub-entries each holding `entry.amount / P`,
+/// preserving the total locked amount but increasing the "count" proportionally to script size.
+///
+/// Refer to the KIP-0009 specification and related documentation for more details.
 #[derive(Clone, Copy)]
 pub struct UtxoCell {
-    /// The plurality of this cell (how many "storage units" it occupies)
+    /// The plurality (number of "storage units") for this UTXO
     pub plurality: u64,
-    /// The amount of KAS value locked in the cell (sompis)
+    /// The amount of KAS (in sompis) locked in this UTXO
     pub amount: u64,
 }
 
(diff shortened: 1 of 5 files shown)


consensus/core/src/tx/script_public_key.rs
```diff
@@ -56,6 +56,14 @@ pub struct ScriptPublicKey {
     pub(super) script: ScriptVec, // Kept private to preserve read-only semantics
 }
 
+impl ScriptPublicKey {
+    pub fn plurality(&self) -> u64 {
+        const UTXO_STORAGE: usize = 32 + 4 + 8 + 8 + 1 + 2 + 8;
```
coderofstuff, line 61: The PR comment hints what these magic numbers are. The full explanation should make it as a comment here. I see you have a TODO too so I'm adding this for you to have something to mark as resolved.

coderofstuff (commented):
The change seems sounds. It's not immediately obvious for those not too familiar with the code but typical UTXOs with standard transaction SPKs will have plurality = 1 with this code.

consensus/core/src/mass/mod.rs
```diff
@@ -235,8 +235,9 @@ pub fn calc_storage_mass(
     //
     // Note: in theory this can be tighten by subtracting input mass in the process (possibly avoiding the overflow),
     // however the overflow case is so unpractical with current mass limits so we avoid the hassle
-    let harmonic_outs =
-        output_values.map(|out| storage_mass_parameter / out).try_fold(0u64, |total, current| total.checked_add(current))?; // C·|O|/H(O)
+    let harmonic_outs = output_values
+        .map(|(plurality, out)| storage_mass_parameter * plurality * plurality / out)
```
someone235, line 239: I think it can overflow here, if SPK size is 430KB and then `10^9*(430000/100)^2 > 2^64`.

someone235 (changes requested):

someone235:
I also recommend to update KIP 9 accordingly

consensus/core/src/mass/mod.rs
michaelsutton: Compute mass gives a weight of 11 gram to each spk byte, so max SPK size is ~45KB

michaelsutton:
- Fixed and significantly enhanced all comments
- Added tests 
- Optimized iterator consumption as much as possible

PR #642, opened 2025-02-24
Change expected pruning point to chain qualification check (Crescendo)

michaelsutton:


coderofstuff:
For this one, the effective change is that if a block has an unexpected pruning point, it will be marked as `StatusDisqualifiedFromChain`, right?

But also, since the check is now in the chain qualification part which is run against chain block candidates only, the actual expected pruning point check can be made simpler?

coderofstuff (approved):

coderofstuff (approved):

Issue #631, opened 2025-02-09
IBD headers optimizations


freshair18:
During IBD, headers are initially supplied to the node at the headers only stage, in order to determine the most advanced selected tip quickly.
The blocks download stage later asks for the full blocks. However currently the headers too are redownloaded despite already being available.

The reasoning for this is mostly due to simplicity and legacy logic from Bitcoin, where headers are incredibly small. In Kaspa the headers consist of multiple parents' hashes instead of 1, and moreso contain higher level parents' hashes as well. Headers are still small compared to body, but no longer negligibly small, resulting in a small but unnecessary elongation of IBD time and extra bandwidth consumption.

**Optimization proposal**

Implement a new p2p message to provide block body only, with structure roughly like:
`struct block_body{
hash: Hash,
txs: Vec<Transaction>,
}`,

 and have nodes receiving this message construct the block body on their own and match and verify it with the already available header.

miningexperiments:
during ibd sync_headers processing the cpu and bandwith are in quite low. Some testing I did for one peer:
1. If I remove all processing the download speed increases from 0.2 to 1 Mbps for that specific peer; e.g. simply downloading chunks with next(). I assume this was the max upload speed for the peer.
2.  Then I tried collecting these into 500 element chunks first as a buffer to see the cpu utilization. It increased, but, maybe from 20% to 30% with multicores.
  - With some late night testing it seemed that each chunks is 0.2Mb, with about 1022-1025 calls to validate_insert_block each.
  - Each call also creates a tokio onechannel, so after downloading 1200chunks it was about 1M channels.
  - Each calls also ends up adding to thread_spawn pool,
  
The IBD seems to work the same way for block processing and there cpu is at 100% use and bandwith massively increased for downloads (multiple peers i suppose?).

I am wondering if there is too much overhead for headers processing to go through?
https://github.com/kaspanet/rusty-kaspa/blob/ac677a0f31a349b8cf77cb839c9d660fb2112508/protocol/flows/src/v5/ibd/flow.rs#L381
https://github.com/kaspanet/rusty-kaspa/blob/cdd43795ac1c82e97058143d5cb607b07df324c9/consensus/src/consensus/mod.rs#L328
https://github.com/kaspanet/rusty-kaspa/blob/cdd43795ac1c82e97058143d5cb607b07df324c9/consensus/src/pipeline/header_processor/processor.rs#L212

freshair18:
[...]

PR #652, opened 2025-03-18
Ibd body optimization

freshair18:
* optimizes redundant downloading and verifications of headers already verified during header only stage of IBD
* creates a new p2p flow version, and a few new p2p messages
*  addresses issue  #631


protocol/flows/src/flow_context.rs
```diff
@@ -762,7 +765,9 @@ impl ConnectionInitializer for FlowContext {
 
         // Register all flows according to version
         let (flows, applied_protocol_version) = match peer_version.protocol_version {
-            v if v >= PROTOCOL_VERSION => (v6::register(self.clone(), router.clone()), PROTOCOL_VERSION),
+            v if v >= PROTOCOL_VERSION => (v7::register(self.clone(), router.clone()), PROTOCOL_VERSION),
+            6 => (v6::register(self.clone(), router.clone()), 6),
+
             5 => (v5::register(self.clone(), router.clone()), 5),
```
miningexperiments, line 771: if only 7 is supported according to line 64,
```rs
/// The P2P protocol version. Currently the only one supported.
const PROTOCOL_VERSION: u32 = 7;
```
should 6 and 5 be ignored?

utils/src/sync/rwlock.rs
```diff
@@ -119,7 +119,7 @@ mod tests {
     use tokio::{sync::oneshot, time::sleep, time::timeout};
 
     const ACQUIRE_TIMEOUT: Duration = Duration::from_secs(5);
-
+    #[ignore = "Performance Dependent Test"]
```
miningexperiments, line 122: if the test is not deterministic in higher performance systems, should the test be changed or ignored?

protocol/flows/src/flow_context.rs
freshair18: probably this comment is legacy from when 5 was the only active version and should be removed.

freshair18:
For all intents and purposes reviewing can start.

Test failures are caused by the fresh rust version. 
Will remerge once new version fixes are pushed to master.

protocol/flows/src/v8/request_ibd_blocks_body.rs
```diff
@@ -0,0 +1,48 @@
...
+    async fn start_impl(&mut self) -> Result<(), ProtocolError> {
+        loop {
+            let (msg, request_id) = dequeue_with_request_id!(self.incoming_route, Payload::RequestIbdBlocksBodies)?;
+            let hashes: Vec<_> = msg.try_into()?;
+
+            debug!("got request for {} IBD blocks bodies", hashes.len());
+            let session = self.ctx.consensus().unguarded_session();
+
+            for hash in hashes {
+                let block_body = session.async_get_block(hash).await?.transactions;
```
michaelsutton, line 42: Initial comment: you should expose a get block body method from consensus to avoid reading the header from the database

utils/src/sync/rwlock.rs
freshair18: addressed

consensus/src/pipeline/header_processor/processor.rs
```diff
@@ -292,8 +292,10 @@ impl HeaderProcessor {
                 let ctx = self.validate_trusted_header(header)?;
                 self.commit_trusted_header(ctx, header);
             }
+            BlockTask::BodyOnly { .. } => {
```
michaelsutton: Assuming the header was indeed already processed, then line 281 above would already return (`Some(status) => return Ok(status)`), and if it wasn't processed, you definitely do not want to do nothing here. In general, consensus code should not rely on correctness by external callers, and this is a good example of violating that principle. 

I suggest removing the `BlockTask::BodyOnly` variant altogether and simply using `Ordinary`. Line 281 will identify this case automatically

components/consensusmanager/src/session.rs
```diff
@@ -174,6 +174,9 @@ impl ConsensusSessionOwned {
     pub fn validate_and_insert_block(&self, block: Block) -> BlockValidationFutures {
         self.consensus.validate_and_insert_block(block)
     }
+    pub fn validate_and_insert_body(&self, block: Block) -> BlockValidationFutures {
```
michaelsutton: in cont. of the previous comment, there are two options here:

1. remove this method and call the usual `validate_and_insert_block` 

2. pass only the body to this call (`Arc<Vec<Txs>>`, possibly with type alias), and fetch the header as part of the impl of `validate_and_insert_body` which will then call the usual `validate_and_insert_block`

I lean towards 2

protocol/flows/src/ibd/flow.rs
```diff
@@ -606,7 +624,41 @@ staging selected tip ({}) is too small or negative. Aborting IBD...",
...
+            ))
+            .await?;
+        for &expected_hash in chunk {
+            let msg = dequeue_with_timeout!(self.incoming_route, Payload::IbdBlockBody)?;
+            //possibly worthwhile to make the header queries in a batch
+            let blk_header = consensus.async_get_header(expected_hash).await?;
+            let mut blk_body: Vec<Transaction> = vec![];
+            for el in msg.transactions {
+                // a bit roundabout, but into doesn't work nicely
+                blk_body.push(el.try_into()?);
```
michaelsutton: 1. The correct syntax for this is `let blk_body: Vec<Transaction> = msg.transactions.into_iter().map(|i| i.try_into()).collect::<Result<Vec<Transaction>, ConversionError>>()?`

2. This logic should actually go into https://github.com/kaspanet/rusty-kaspa/blob/master/protocol/p2p/src/convert/block.rs (converting from `proto::BlockBodyMessage` to `Vec<Transaction>`)
and the code here should be simply 
`let blk_body: Vec<Transaction> = msg.try_into()?;`

protocol/flows/src/ibd/flow.rs
```diff
@@ -606,7 +624,41 @@ staging selected tip ({}) is too small or negative. Aborting IBD...",
...
+        let mut current_timestamp = 0;
+        self.router
+            .enqueue(make_message!(
+                Payload::RequestIbdBlocksBodies,
+                RequestIbdBlocksBodiesMessage { hashes: chunk.iter().map(|h| h.into()).collect() }
+            ))
+            .await?;
+        for &expected_hash in chunk {
+            let msg = dequeue_with_timeout!(self.incoming_route, Payload::IbdBlockBody)?;
+            //possibly worthwhile to make the header queries in a batch
```
michaelsutton: depending on how you fix prev comments, you should probably remove this comment (or make it an actual todo)

protocol/flows/src/v8/request_ibd_blocks_body.rs
```diff
@@ -0,0 +1,48 @@
...
+
+    async fn start_impl(&mut self) -> Result<(), ProtocolError> {
+        loop {
+            let (msg, request_id) = dequeue_with_request_id!(self.incoming_route, Payload::RequestIbdBlocksBodies)?;
+            let hashes: Vec<_> = msg.try_into()?;
+            debug!("got request for {} IBD blocks bodies", hashes.len());
+            let session = self.ctx.consensus().unguarded_session();
+            for chunk in hashes.chunks(IBD_BATCH_SIZE) {
+                let bodies_batch = session.async_get_block_body_batch(chunk.to_vec()).await;
+                for el in bodies_batch {
```
michaelsutton: what does `el` stand for?

protocol/flows/src/v8/request_ibd_blocks_body.rs
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    async fn start_impl(&mut self) -> Result<(), ProtocolError> {
+        loop {
+            let (msg, request_id) = dequeue_with_request_id!(self.incoming_route, Payload::RequestIbdBlocksBodies)?;
+            let hashes: Vec<_> = msg.try_into()?;
+            debug!("got request for {} IBD blocks bodies", hashes.len());
+            let session = self.ctx.consensus().unguarded_session();
+            for chunk in hashes.chunks(IBD_BATCH_SIZE) {
+                let bodies_batch = session.async_get_block_body_batch(chunk.to_vec()).await;
```
michaelsutton: any way to avoid the `to_vec` alloc?

protocol/flows/src/v8/request_ibd_blocks_body.rs
freshair18: see now

protocol/flows/src/ibd/flow.rs
freshair18: Made into a todo.

protocol/flows/src/ibd/flow.rs
freshair18: done

components/consensusmanager/src/session.rs
freshair18: I ended up choosing one - fetching the header internally hinders IBD flow from accessing timestamps and daa scores used for completion tracking

consensus/src/pipeline/header_processor/processor.rs
freshair18: done

protocol/flows/src/v8/request_ibd_blocks_body.rs
freshair18: This appears to be the most natural way -  I cannot move references (array slices) across threads.

PR #653, opened 2025-03-19
refactor: kaspad/src/daemon.rs

miningexperiments:
# Problem
fn create_core_with_runtime is a massive monolithic function

# Fixes
- Attempted to modularize the code by dividing the code into more functions
- Added some optimizations to reduce function calls such as is_empty(), use of references
- Few minor runtime optimizations such as using inverted gigabyte to reduce runtime calculations.

# Tests
- Tested to work with cargo test --release, cargo nextest r --release
- cargo test non-release still has usual suspects 250k vec
- Ran with --reset-db flag, which reset db
- Ran in testnet10 without problems

# Additional notes
- There some todo notes left in the function, not sure of the status of these

### DIFF
--- a/kaspad/src/daemon.rs
+++ b/kaspad/src/daemon.rs
@@ -1,5 +1,3 @@
-use std::{fs, path::PathBuf, process::exit, sync::Arc, time::Duration};
-
 use async_channel::unbounded;
 use kaspa_consensus_core::{
     config::ConfigBuilder,
@@ -21,6 +19,8 @@ use kaspa_utils::git;
 use kaspa_utils::networking::ContextualNetAddress;
 use kaspa_utils::sysinfo::SystemInfo;
 use kaspa_utils_tower::counters::TowerConnectionCounters;
+use std::io::Write;
+use std::{fs, path::PathBuf, process::exit, sync::Arc, time::Duration};
 
 use kaspa_addressmanager::AddressManager;
 use kaspa_consensus::{consensus::factory::Factory as ConsensusFactory, pipeline::ProcessingCounters};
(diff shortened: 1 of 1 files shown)


kaspad/src/daemon.rs
```diff
@@ -205,6 +200,54 @@ pub fn create_core(args: Args, fd_total_budget: i32) -> (Arc<Core>, Arc<RpcCoreS
...
+        let target_time_per_block = config.target_time_per_block().after(); // in ms
+
+        if MINIMUM_RETENTION_PERIOD_DAYS <= retention_period_days {
+            // milliseconds in a day = 24.0 * 60.0 * 60.0 * 1000.0 = 86 400 000
+            let retention_period_milliseconds = (retention_period_days * 86_400_000.0).ceil() as u64;
+            let total_blocks = retention_period_milliseconds / target_time_per_block;
+            // This worst case usage only considers block space. It does not account for usage of
+            // other stores (reachability, block status, mempool, etc.)
+            let worst_case_usage =
+                ((total_blocks + finality_depth) * (config.max_block_mass / TRANSIENT_BYTE_TO_MASS_FACTOR)) as f64 * INV_ONE_GIGABYTE;
```
freshair18, line 229: I think in this case using the inverse is unnecessary and does more obscuring then simplification.

kaspad/src/daemon.rs
```diff
@@ -205,6 +200,54 @@ pub fn create_core(args: Args, fd_total_budget: i32) -> (Arc<Core>, Arc<RpcCoreS
...
 }
 
+/// Helper functions for create_core_with_runtime
+fn build_db_paths(db_dir: &Path) -> (PathBuf, PathBuf, PathBuf) {
+    const CONSENSUS_DB: &str = "consensus";
+    const UTXOINDEX_DB: &str = "utxoindex";
+    const META_DB: &str = "meta";
+    (db_dir.join(CONSENSUS_DB), db_dir.join(UTXOINDEX_DB), db_dir.join(META_DB))
+}
+fn check_retention_days(args: &Args, config: &Config) {
```
freshair18, line 210: Check is not a good descriptor where really this function is about informing the user. In general I'd add a few comment line detailing its point.

kaspad/src/daemon.rs
```diff
@@ -114,23 +113,19 @@ fn get_user_approval_or_exit(message: &str, approve: bool) {
     if approve {
         return;
     }
-    println!("{}", message);
+    println!("{}: ", message);
+    std::io::stdout().flush().expect("Failed to flush stdout");
```
freshair18, line 117: not sure if the explicit flush is really necessary, can you expand on why add it?

freshair18 (changes requested):

PR #654, opened 2025-03-19
Mining Rule Engine Scaffolding and Sync Rate Rule implementation

coderofstuff:
Splitting the PR at #634 into multiple parts, this is part 1 with

- Mining Rule Engine scaffolding
- Sync Rate Rule implementation


rpc/service/src/service.rs
```diff
@@ -472,13 +473,14 @@ NOTE: This error usually indicates an RPC conversion error between the node and
...
-        let is_nearly_synced = self.consensus_manager.consensus().unguarded_session().async_is_nearly_synced().await;
+        let sink_daa_score_timestamp =
+            self.consensus_manager.consensus().unguarded_session().async_get_sink_daa_score_timestamp().await;
         Ok(GetInfoResponse {
             p2p_id: self.flow_context.node_id.to_string(),
             mempool_size: self.mining_manager.transaction_count_sample(TransactionQuery::TransactionsOnly),
             server_version: version().to_string(),
             is_utxo_indexed: self.config.utxoindex,
-            is_synced: self.has_sufficient_peer_connectivity() && is_nearly_synced,
+            is_synced: self.mining_rule_engine.should_mine(sink_daa_score_timestamp),
```
someone235, line 483: [This](https://github.com/kaspanet/rusty-kaspa/pull/634#discussion_r1997603072) is still relevant.

Re @michaelsutton 

> idk, doesn't slow sync rate truly indicate that you are synced even if sink timestamp is lagging?

I'm not sure about that. I think it just means we should mine, to be on the safe side. Let's say KNG wants to know if its node is synced, I'm not sure I would say it's synced if its tip is not recent. Maybe the sync rate is slow because of bandwidth problems etc, and I wouldn't want to show the wallet balance as the "final" balance at that stage

rpc/service/src/service.rs
```diff
@@ -1142,7 +1144,8 @@ NOTE: This error usually indicates an RPC conversion error between the node and
         _request: GetServerInfoRequest,
     ) -> RpcResult<GetServerInfoResponse> {
         let session = self.consensus_manager.consensus().unguarded_session();
-        let is_synced: bool = self.has_sufficient_peer_connectivity() && session.async_is_nearly_synced().await;
+        let sink_daa_score_timestamp = session.async_get_sink_daa_score_timestamp().await;
+        let is_synced: bool = self.mining_rule_engine.should_mine(sink_daa_score_timestamp);
```
someone235, line 1148: Same here

rpc/service/src/service.rs
```diff
@@ -1161,8 +1164,9 @@ NOTE: This error usually indicates an RPC conversion error between the node and
         _connection: Option<&DynRpcConnection>,
         _request: GetSyncStatusRequest,
     ) -> RpcResult<GetSyncStatusResponse> {
-        let session = self.consensus_manager.consensus().unguarded_session();
-        let is_synced: bool = self.has_sufficient_peer_connectivity() && session.async_is_nearly_synced().await;
+        let sink_daa_score_timestamp =
+            self.consensus_manager.consensus().unguarded_session().async_get_sink_daa_score_timestamp().await;
+        let is_synced: bool = self.mining_rule_engine.should_mine(sink_daa_score_timestamp);
```
someone235, line 1169: Same here

someone235 (changes requested):

rpc/service/src/service.rs
coderofstuff: Addressed in https://github.com/kaspanet/rusty-kaspa/pull/654/commits/f2311b606a93c2a6e2eef9905400d1b0edc4bd69

rpc/service/src/service.rs
coderofstuff: Addressed in https://github.com/kaspanet/rusty-kaspa/pull/654/commits/f2311b606a93c2a6e2eef9905400d1b0edc4bd69

rpc/service/src/service.rs
coderofstuff: Addressed in https://github.com/kaspanet/rusty-kaspa/pull/654/commits/f2311b606a93c2a6e2eef9905400d1b0edc4bd69

someone235 (dismissed):

someone235 (approved):

protocol/flows/src/flow_context.rs
```diff
@@ -586,7 +591,7 @@ impl FlowContext {
         }
 
         // Transaction relay is disabled if the node is out of sync and thus not mining
```
michaelsutton, line 593: ~~and thus not mining~~

protocol/flows/src/v5/txrelay/flow.rs
```diff
@@ -116,7 +116,7 @@ impl RelayTransactionsFlow {
             let session = self.ctx.consensus().unguarded_session();
 
             // Transaction relay is disabled if the node is out of sync and thus not mining
```
michaelsutton, line 118: ~~and thus not mining~~

protocol/flows/src/v5/blockrelay/flow.rs
```diff
@@ -116,7 +116,7 @@ impl HandleRelayInvsFlow {
                 }
             }
 
-            if self.ctx.is_ibd_running() && !session.async_is_nearly_synced().await {
+            if self.ctx.is_ibd_running() && !self.ctx.is_nearly_synced(&session).await {
```
michaelsutton, line 119: Not sure about this one. If the node is on IBD but *yet* sync rate is below threshold and thus mining, I think relay blocks should be processed as well.

michaelsutton (changes requested):

michaelsutton (approved):

michaelsutton:
LGTM

PR #655, opened 2025-03-20
Mining Rule Engine: No Transactions Rule

coderofstuff:
Splitting off #634 into parts. This is Part 2/3 and depends on #654


protocol/mining/src/rules/no_transactions_rule.rs
```diff
@@ -0,0 +1,64 @@
...
+    }
+}
+
+impl MiningRule for NoTransactionsRule {
+    fn check_rule(&self, delta: &ProcessingCountersSnapshot, _extra_data: &ExtraData) {
+        let cooldown_count = self.cooldown.load(Ordering::SeqCst);
+
+        if cooldown_count > 0 {
+            // Recovering
+            if delta.submit_block_success_count > 0 || self.cooldown.fetch_sub(1, Ordering::SeqCst) == 1 {
```
miningexperiments, line 38: would it be better to use fetch_update, then you can also avoid the previous load, and if fetch_sub has a minor potential for datarace conditions

protocol/mining/src/rule_engine.rs
```diff
@@ -75,6 +75,8 @@ impl MiningRuleEngine {
                     has_sufficient_peer_connectivity: self.has_sufficient_peer_connectivity(),
                     finality_duration: self.config.finality_duration_in_milliseconds().get(sink_daa_timestamp.daa_score),
                     elapsed_time,
+                    sink_daa_score_timestamp: session.async_get_sink_daa_score_timestamp().await,
```
michaelsutton, line 78: you already have `sink_daa_timestamp` from above (no need to repeat the consensus call)

Issue #666, opened 2025-04-02
Support specifying the number of confirmations required for blocks returned in GetVirtualChainFromBlock


coderofstuff:
**Is your feature request related to a problem? Please describe.**
`GetVirtualChainFromBlock` returns all blocks from `lowHash` until the latest chain block (aka. the `sink`) or up to the threshold of the request. When you poll GVCFB regularly and are already up to date, you may regularly get `removed` entries that you have to reprocess. These `removed` entries are due to re-orgs around the tips and are normal occurrences.

It can be preferable for a user such as an indexer or an exchange to avoid this reorg handling as much as possible (in the interest of reducing I/O, for example) by fetching only up to blocks that already have some amount of confirmation. Take for instance this chain:

```
A <- B <- C <- D <- E <- F <- G <- H <- I
```

In order to avoid processing re-orgs around G, H and I, a user might specify that they only want to fetch blocks that already have 3 confirmations at least. In this case, return only blocks A, B, C, D, E, F and wait until the next iterations before returning G, H and I when they already have sufficient confirmation count.

**Describe the solution you'd like**
Allow passing a `minConfirmationCount` to `GetVirtualChainFromBlock` so that only chain blocks that have that many confirmations (blue depth) are returned.

Something like (but maybe not example):

```
#[derive(Clone, Debug, Serialize, Deserialize)]
#[serde(rename_all = "camelCase")]
pub struct GetVirtualChainFromBlockRequest {
    pub start_hash: RpcHash,
    pub include_accepted_transaction_ids: bool,
    pub min_confirmation_count: u64,
}
```

**Notes**
1. How to count confirmations for chain blocks?
```
sink.blue_score - current_chain_block.blue_score
```

IzioDev:
I'm currently attempting to implement this feature

PR #667, opened 2025-04-02
feat(vspc): min confirmation count

IzioDev:
This draft is a naive implementation of `min_confirmation_count` parameter during VC retreivals, described in #666.

 Currently investigation if we could partially move this logic lower-level (potentially at Consensus API level).

Feel free to make suggestions / comments even if it's not ready yet :)


consensus/src/consensus/mod.rs
```diff
@@ -588,6 +588,12 @@ impl ConsensusApi for Consensus {
         self.headers_store.get_timestamp(self.get_sink()).unwrap()
     }
 
+    fn get_sink_blue_score(&self) -> u64 {
+        let sink = self.get_sink();
+        let compact = self.headers_store.get_compact_header_data(sink).unwrap();
+        compact.blue_score
```
coderofstuff, line 594:
```suggestion
        self.headers_store.get_blue_score(self.get_sink()).unwrap()
```

rpc/core/src/model/message.rs
```diff
@@ -872,6 +873,7 @@ impl Serializer for GetVirtualChainFromBlockRequest {
         store!(u16, &1, writer)?;
         store!(RpcHash, &self.start_hash, writer)?;
         store!(bool, &self.include_accepted_transaction_ids, writer)?;
+        store!(u64, &self.min_confirmation_count, writer)?;
```
coderofstuff, line 876: @aspect for a change like this, does the version (`&1`) above have to be incremented?

And then some handling on `deserialize` performed based on version number?

rpc/service/src/service.rs
```diff
@@ -684,6 +684,25 @@ NOTE: This error usually indicates an RPC conversion error between the node and
...
+
+            match distance.cmp(&request.min_confirmation_count) {
+                std::cmp::Ordering::Equal => (),
+                std::cmp::Ordering::Greater => (),
+                std::cmp::Ordering::Less => {
+                    let to_remove = request.min_confirmation_count.saturating_sub(distance);
+                    let to_keep = u64::try_from(virtual_chain_batch.added.len())?.saturating_sub(to_remove);
+                    virtual_chain_batch.added.truncate(usize::try_from(to_keep)?);
+                }
+            }
```
coderofstuff, line 703: You can keep checking the `last()` element and see if it's within the required threshold, and remove it if not like this (might not be the best code, but this is the idea):

```
            while !virtual_chain_batch.added.is_empty() {
                let vc_last_accepted_block_hash = virtual_chain_batch.added.last().unwrap();
                let vc_last_accepted_block = session.async_get_block(*vc_last_accepted_block_hash).await?;
    
                let distance = sink_blue_score.saturating_sub(vc_last_accepted_block.header.blue_score);

                if distance < min_confirmation_count {
                    virtual_chain_batch.added.pop();
                } else {
                    break;
                }
            }
```

From your implementation, it assumes that the distance between two chain blocks is always 1 based on my understanding of `to_remove`.

Take a look at this snapshot from KGI for example:

image

The encircled blocks are:

|Block|Blue Score|
|---|---|
| [a99748...82634a](https://explorer.kaspa.org/blocks/a9974801fd0dd145a5376cda597e5214c13eb11202cd5c8d778c0061fb82634a)| `105838553`|
| [75eb67...28bab1](https://explorer.kaspa.org/blocks/75eb6747e2aca2504709fc9bf27e63f0587fef943b062da8a0b1573edf28bab1)| `105838555`|

So even though they are consecutive chain blocks, their blue score diff is 2 because `75eb67...28bab1` merges two blue blocks (`a99748...82634a` and `1a3cc3...d24f2e` [which is not a chain block]).

coderofstuff (changes requested):
Initial feedback. :)

consensus/src/consensus/mod.rs
IzioDev: applied in [a4a91042203c23bb688c7a4d3596e010ebb0e5cb](https://github.com/kaspanet/rusty-kaspa/pull/667/commits/a4a91042203c23bb688c7a4d3596e010ebb0e5cb)

rpc/core/src/model/message.rs
```diff
@@ -886,7 +886,7 @@ impl Deserializer for GetVirtualChainFromBlockRequest {
         let include_accepted_transaction_ids = load!(bool, reader)?;
 
         // starting from version 2, min_confirmation_count is expected to be included in the request
```
IzioDev: if accepted, this makes this comment irrelevant

wasm/examples/nodejs/javascript/general/get-virtual-chain.js
```diff
@@ -0,0 +1,36 @@
...
+    // url : "127.0.0.1",
+    // encoding,
+    resolver: new Resolver(),
+    networkId: "mainnet",
+  });
+  console.log(`Resolving RPC endpoint...`);
+  await rpc.connect();
+  console.log(`Connecting to ${rpc.url}`);
+
+  const virtualChainFromBlockResponse = await rpc.getVirtualChainFromBlock({
```
coderofstuff: This example will fail when that `startHash` is pruned. You can instead query for some guaranteed known block, and use that info.

Example:

```
const info = awat rpc.getBlockDagInfo();
const startHash = info.sink;
```

wasm/examples/nodejs/javascript/general/get-virtual-chain.js
```diff
@@ -0,0 +1,36 @@
...
+    startHash:
+      "106145ef74693458e1013819f04cafb65d99fd17d28a8c0a6e3941199c0adf82",
+    includeAcceptedTransactionIds: true,
+    // minConfirmationCount: 1,
```
coderofstuff: It's worth running this twice: one where `minConfirmationCount` is not passed (as it is now) and one where some number is passed. Both must work, which is why it's worth running both.

testing/integration/src/daemon_integration_tests.rs
```diff
@@ -110,6 +110,7 @@ async fn daemon_mining_test() {
         .get_virtual_chain_from_block(
             kaspa_consensus::params::SIMNET_GENESIS.hash, //
             true,
+            Some(0),
```
coderofstuff: If this is left as Option, the replace `Some(0)` with `None`. If this ends up being applied, please apply to all cases where `Some(0)` occurs for this change.

rpc/service/src/service.rs
```diff
@@ -684,6 +684,27 @@ NOTE: This error usually indicates an RPC conversion error between the node and
         // Note: batch_size does not bound removed chain blocks, only added chain blocks.
         let batch_size = (self.config.mergeset_size_limit().upper_bound() * 10) as usize;
         let mut virtual_chain_batch = session.async_get_virtual_chain_from_block(request.start_hash, Some(batch_size)).await?;
+
+        match request.min_confirmation_count {
+            Some(min_confirmation_count) if min_confirmation_count > 0 => {
```
coderofstuff: This can be simplified with:

```
        if let Some(min_confirmation_count) = request.min_confirmation_count {
            if min_confirmation_count > 0 {
               ...
            }
        }
```

rpc/core/src/model/message.rs
coderofstuff: Yeah, you can probably remove this.

rpc/core/src/model/message.rs
```diff
@@ -859,31 +859,36 @@ impl Deserializer for GetSubnetworkResponse {
...
     }
 }
 
 impl Serializer for GetVirtualChainFromBlockRequest {
     fn serialize<W: std::io::Write>(&self, writer: &mut W) -> std::io::Result<()> {
-        store!(u16, &1, writer)?;
+        store!(u16, &2, writer)?;
         store!(RpcHash, &self.start_hash, writer)?;
         store!(bool, &self.include_accepted_transaction_ids, writer)?;
+        store!(Option<u64>, &self.min_confirmation_count, writer)?;
```
coderofstuff, line 876: So, to discuss this a bit: You were asking whether it was more correct to pass this `Option` down from here to the downstream calls or to just deser to/from u64 and default to 0.

I think it's fine as it is where the Option is passed all the way down. wdyt @biryukovmaxim @aspect?

wasm/examples/nodejs/javascript/general/get-virtual-chain.js
IzioDev: wouldn't it degrade the user example comprehension?

if the goal is to provide backwards compatibility "proof", i think we should have dedicated js unit (or even better integrated) test

if the goal is to show capabilities, we probably should add typedoc (text is placeholder in screenshot)
image
image


what do you think?

wasm/examples/nodejs/javascript/general/get-virtual-chain.js
IzioDev: done

testing/integration/src/daemon_integration_tests.rs
IzioDev: done

rpc/service/src/service.rs
IzioDev: done

wasm/examples/nodejs/javascript/general/get-virtual-chain.js
coderofstuff: It’s purpose is more of a demo of all the typical configurations you might use. Typedoc documentation in addition to still doing this is the best, imo.

see message-signing.js example for a quick run of the same demo using different supported configs.

rpc/service/src/service.rs
coderofstuff: We can always iterate on this section of the code if it turns out that we need to do a bisect. But for now, what's here is good enough.

rpc/core/src/wasm/message.rs
```diff
@@ -1230,6 +1230,12 @@ declare! {
     export interface IGetVirtualChainFromBlockRequest {
         startHash : HexString;
         includeAcceptedTransactionIds: boolean;
+        /**
+         * optional parameter, if not provided, it will be interpreted as 0
+         * 
+         * number of confirmation to wait for before a block is considered as part of the virtual chain
```
coderofstuff: > number of confirmation to wait for before a block is considered as part of the virtual chain

Replace with: "If passed, this request will only return blocks that have at least `minConfirmationCount` number of confirmations. Confirmation is counted through the distance from virtual."

coderofstuff (dismissed):
Remaining changes are suggestions

rpc/core/src/wasm/message.rs
IzioDev: fixed in [6816d0f976cc4c7bcf29513304d109c84e370962](https://github.com/kaspanet/rusty-kaspa/pull/667/commits/6816d0f976cc4c7bcf29513304d109c84e370962)

wasm/examples/nodejs/javascript/general/get-virtual-chain.js
IzioDev: fixed in [6816d0f976cc4c7bcf29513304d109c84e370962](https://github.com/kaspanet/rusty-kaspa/pull/667/commits/6816d0f976cc4c7bcf29513304d109c84e370962)

IzioDev:
Updated with upstream

PR #688, opened 2025-05-27
Fix: Overflow in UTXOS_CHANGED_SUBSCRIPTIONS

hmoog:
### Summary

This PR fixes a subtle but non-trivial bug caused by panics during test execution that interrupt thread logic and leading to corruption of a global shared counter used for tracking subscription instances.

### Problem

A panic occurring in a test can cause the thread to unwind *after* a `subscription` object had been partially initialized and placed on the stack, but *before* the counter increment (`fetch_add`) is executed.

**Code before fix:**
```rust
let subscription = Self { data: RwLock::new(self.data().clone()), listener_id: self.listener_id };

// ⚠️ Thread could unwind here caused by an external panic

trace!(
    "UtxosChangedSubscription: {} in total (clone {})",
    UTXOS_CHANGED_SUBSCRIPTIONS.fetch_add(1, Ordering::SeqCst) + 1,
    subscription
);
```

When the thread panics, the Rust runtime begins unwinding the stack, which triggers the `Drop` implementation.

This leads to:
- The counter being decremented even though it was never incremented
- A resulting **underflow**
- Future calls to `fetch_add` causing **overflow**
- Other tests failing unpredictably due to the corrupted global counter

### Solution

The fix is to increment the counter **before** constructing the `subscription` object. This guarantees that any `Drop` call during unwinding will be paired with a prior increment.

### Remaining Edge Case

If a panic happens *after* incrementing the counter but *before* constructing the object, the counter will be artificially high. However:
- This no longer corrupts the logic by causing an underflow
- The counter being too high is now a **purely cosmetic** issue
- Rust does not allow interception of externally-caused unwinds, so no further mitigation is possible


notify/src/subscription/single.rs
```diff
@@ -365,12 +362,9 @@ impl UtxosChangedSubscription {
 
 impl Clone for UtxosChangedSubscription {
     fn clone(&self) -> Self {
+        let subscription_count = UTXOS_CHANGED_SUBSCRIPTIONS.fetch_add(1, Ordering::SeqCst) + 1;
```
michaelsutton, line 365: in the `Drop` impl below, you must extract the call `UTXOS_CHANGED_SUBSCRIPTIONS.fetch_sub(1, Ordering::SeqCst) - 1` to be out of the `trace!` macro call. In some cases (when loglevel is above trace for all modules) the macro is not evaluated at all. In which case the counter will only increase (since we now have the increase side out of the `trace` call)

michaelsutton: for ref
```
// Recursive expansion of trace! macro
// ====================================

{
    let lvl = (log::Level::Trace);
    if lvl <= log::STATIC_MAX_LEVEL && lvl <= log::max_level() {
        log::__private_api::log(
            log::__private_api::format_args!(
                "UtxosChangedSubscription: {} in total (drop {})",
                UTXOS_CHANGED_SUBSCRIPTIONS.fetch_sub(1, Ordering::SeqCst) - 1,
                self
            ),
            lvl,
            &((log::__private_api::module_path!()), log::__private_api::module_path!(), log::__private_api::loc()),
            (),
        );
    }
};
```

PR #697, opened 2025-06-25
Check that trusted blocks are not in the pruning point future

someone235:


coderofstuff (changes requested):
Syncing with this branch fails with:

```
2025-07-20 21:43:59.790-06:00 [WARN ] IbdFlow flow error: got trusted block 141f840e928c0f08dea5b05ad6d7ce46abdd3cb28e8b97c16f68957fde5986c8 in the future of the pruning point, disconnecting from peer [::ffff:198.52.142.144]:16111.
```

Where `141f840e928c0f08dea5b05ad6d7ce46abdd3cb28e8b97c16f68957fde5986c8` is the current pruning point as of this writing.

Issue #700, opened 2025-06-26
Add tx.payload support for PSKT constructions


danwt:
It's misssing

keruch:
+1, it's impossible to create PSKT with custom payload.

It looks like adding a payload should be pretty straightforward. It can be kept in `Global` as `Option`:
https://github.com/kaspanet/rusty-kaspa/blob/eb71df4d284593fccd1342094c37edc8c000da85/wallet/pskt/src/global.rs#L18
And populated on the unchecked tx creation:
https://github.com/kaspanet/rusty-kaspa/blob/eb71df4d284593fccd1342094c37edc8c000da85/wallet/pskt/src/pskt.rs#L153

PR #703, opened 2025-07-04
Adds payload support in PSKT

danwt:
Closes https://github.com/kaspanet/rusty-kaspa/issues/700

### DIFF
--- a/wallet/pskt/src/global.rs
+++ b/wallet/pskt/src/global.rs
@@ -39,6 +39,7 @@ pub struct Global {
     /// Unknown key-value pairs for this output.
     #[serde(flatten)]
     pub unknowns: BTreeMap<String, serde_value::Value>,
+    pub payload: Option<Vec<u8>>,
 }
 
 impl Add for Global {
@@ -122,6 +123,7 @@ impl Default for Global {
             id: None,
             proprietaries: Default::default(),
             unknowns: Default::default(),
+            payload: None,
         }
     }
 }
--- a/wallet/pskt/src/pskt.rs
+++ b/wallet/pskt/src/pskt.rs
@@ -150,7 +150,7 @@ impl<R> PSKT<R> {
             self.determine_lock_time(),
             SUBNETWORK_ID_NATIVE,
             0,
-            vec![],
+            self.global.payload.clone().unwrap_or_default(),
         );
         let entries = self.inputs.iter().filter_map(|Input { utxo_entry, .. }| utxo_entry.clone()).collect();
         SignableTransaction::with_entries(tx, entries)
--- a/wallet/pskt/src/pskt.rs
+++ b/wallet/pskt/src/pskt.rs
@@ -236,6 +236,11 @@ impl PSKT<Constructor> {
         self
     }
 
+    pub fn payload(mut self, payload: Vec<u8>) -> Self {
+        self.inner_pskt.global.payload = Some(payload);
+        self
+    }
+
     /// Returns a PSKT [`Updater`] once construction is completed.
     pub fn updater(self) -> PSKT<Updater> {
         let pskt = self.no_more_inputs().no_more_outputs();


wallet/pskt/src/pskt.rs
```diff
@@ -237,6 +237,11 @@ impl PSKT<Constructor> {
         self
     }
 
+    pub fn payload(mut self, payload: Vec<u8>) -> Self {
+        self.inner_pskt.global.payload = Some(payload);
+        self
+    }
```
IzioDev: Could be wise to keep the underlying signature (`Option<Vec<u8>>`), some might want to reset payload by setting a None value to it.

danwt: I will update

danwt: done

wallet/pskt/src/global.rs
```diff
@@ -39,6 +39,7 @@ pub struct Global {
     /// Unknown key-value pairs for this output.
     #[serde(flatten)]
     pub unknowns: BTreeMap<String, serde_value::Value>,
+    pub payload: Option<Vec<u8>>,
```
biryukovmaxim, line 43: Add attribute `#[serde(with = "kaspa_utils::serde_bytes_optional")]` to serialize it as hex in case of human readable serializer, as well as it has consistency with redeem script serialization

biryukovmaxim:
Hello, @danwt great work.
There are some things that should also be done:
1) version bump. New field makes incompatible at signing and finalization steps. Old clients will make transaction without payload, but new clients will include it.
So we need to ensure that every combining has the same versions, as well as payload can only be added if version is 2. The same rule for unsigned tx method
2) [Add method should be reflected](https://github.com/dymensionxyz/rusty-kaspa/blob/10d15eba8b67c184ce9d7886d7b3f734242eb56b/wallet/pskt/src/global.rs#L45) 
If one of left/right global struct has some payload it should have payload in the output, both none -> None. If they have non equal payloads -> return error.

If overwriting is required - it can be solved via another pr/issue or it should require separate field with vector of payloads so the finalized can combine them properly, or some other deterministic approach. The invariant is left+right==right+left

biryukovmaxim (changes requested):

wallet/pskt/src/global.rs
danwt: Thanks very much for review I'll address soon

danwt: Hi @biryukovmaxim I have updated the pr

wallet/pskt/src/global.rs
```diff
@@ -104,6 +106,19 @@ impl Add for Global {
         self.proprietaries =
             combine_if_no_conflicts(self.proprietaries, rhs.proprietaries).map_err(CombineError::NotCompatibleProprietary)?;
         self.unknowns = combine_if_no_conflicts(self.unknowns, rhs.unknowns).map_err(CombineError::NotCompatibleUnknownField)?;
+
+        // Combine payloads according to the rules:
+        // - Both None -> None
+        // - One has payload -> use that payload
+        // - Both have same payload -> use that payload
+        // - Different payloads -> error
+        self.payload = match (self.payload.clone(), rhs.payload.clone()) {
```
biryukovmaxim: You can use `.take` instead of `clone` here

biryukovmaxim: It also should return error in case of version<1
Could you also add a set_version method to pskt<creator>?

danwt: will do

coderofstuff: PR looks legit. What else is pending here @biryukovmaxim @danwt?

biryukovmaxim: @coderofstuff https://github.com/kaspanet/rusty-kaspa/pull/703/files#r2295402182

danwt: on it

biryukovmaxim:
LGTM,
cc. @coderofstuff

coderofstuff:
Looks good. @danwt Please merge or rebase with latest master

PR #712, opened 2025-07-23
WIP: New VirtualChainFromBlockV2 endpoint. 

D-Stacks:
work in progress...

### DIFF
--- a/wallet/core/src/utxo/scan.rs
+++ b/wallet/core/src/utxo/scan.rs
@@ -152,8 +152,7 @@ impl Scan {
 
         utxo_context.register_addresses(&address_vec).await?;
         let resp = utxo_context.processor().rpc_api().get_utxos_by_addresses(address_vec).await?;
-        let refs: Vec<UtxoEntryReference> =
-            resp.into_iter().map(UtxoEntryReference::try_from).collect::<Result<Vec<_>, _>>()?;
+        let refs: Vec<UtxoEntryReference> = resp.into_iter().map(UtxoEntryReference::try_from).collect::<Result<Vec<_>, _>>()?;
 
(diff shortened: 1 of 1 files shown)


consensus/src/consensus/mod.rs
```diff
@@ -767,11 +770,82 @@ impl ConsensusApi for Consensus {
...
+                                self.get_block_transactions(
+                                    mbad.block_hash,
+                                    Some(mbad.accepted_transactions.iter().map(|atx| atx.index_within_block).collect()),
+                                )
+                            })
+                            .flatten()
+                            .collect::<Vec<_>>(),
+                    )))
+                }
+            }
```
IzioDev, line 844: factorable, but has it been made for readability reasons?

if so, should it deserve two functions instead?

alternatively: evaluate conditional `filter` apply

consensus/src/consensus/mod.rs
```diff
@@ -972,6 +1046,33 @@ impl ConsensusApi for Consensus {
         })
     }
 
+    fn get_block_transactions(&self, hash: Hash, indices: Option<Vec<TransactionIndexType>>) -> ConsensusResult<Vec<Transaction>> {
```
IzioDev, line 1049: Can we think of use-cases for this function signature other than retrieving the transactions after getting the accepted_block?

If not, should we instead tie these two inside a new function?

```rust
fn get_transactions_from_accepting_block(
  accepting_block_hash: Hash,
  tx_ids: Option<Vec<TransactionId>>
)  -> ConsensusResult<Vec<Transaction>>
```

IzioDev:
Superseded with #740

IzioDev:
Closed, continued in #740 that has been merged

PR #715, opened 2025-08-05
Rewrite estimate_network_hashes_per_second_impl

someone235:


consensus/src/consensus/mod.rs
```diff
@@ -469,12 +469,41 @@ impl Consensus {
...
+            return Err(ConsensusError::UnderMinWindowSizeAllowed(window_size, MIN_WINDOW_SIZE));
+        }
+
+        let mut count = 0;
+        let mut red_work: BlueWorkType = 0.into();
+        let mut bottom = ghostdag_data.selected_parent;
+        for chain_block in self.services.reachability_service.default_backward_chain_iterator(ghostdag_data.selected_parent) {
+            let gd = self.get_ghostdag_data(chain_block).unwrap();
+            for red in &gd.mergeset_reds {
+                let red_header = self.headers_store.get_header(*red).unwrap();
```
IzioDev, line 483: Is it safe to assume the header will be in the store for the whole rpc thread execution?

Is there a lock on the store?

edit: should be reasonable to assume the iterator won't go up until pruning point since the window is meant to be reasonable

PR #721, opened 2025-08-19
Support multiple --externalip args

supertypo:
Allows to specify multiple externalips. 
Common use case: specify both ipv4 and ipv6 address.


p4bpj:
I'm working on this feature https://github.com/p4bpj/rusty-kaspa/pull/2 and https://github.com/p4bpj/rusty-kaspa/pull/3
Maybe there should be a discussion how to improve external-ips/addressmanager and the associated systems. I'm also working on support for a dyndns-domain, same logic as upnp but it resolves the dyndns-domain periodically.

components/addressmanager/src/lib.rs
```diff
@@ -113,18 +113,20 @@ impl AddressManager {
...
-                info!("External address is publicly routable {}", local_net_address);
-                return Left(iter::once(local_net_address));
+        if let Some(externalips) = &self.config.externalip {
+            let mut is_publicly_routable = Vec::new();
+            for addr in externalips {
+                if addr.ip.is_publicly_routable() {
+                    info!("External address is publicly routable {}", addr);
+                    is_publicly_routable.push(*addr);
+                } else {
+                    info!("External address is not publicly routable {}", addr);
```
coderofstuff, line 123:
```
            let is_publicly_routable = externalips
                .iter()
                .filter(|addr| {
                    let is_routable = addr.ip.is_publicly_routable();
                    if is_routable {
                        info!("External address is publicly routable {}", addr);
                    } else {
                        info!("External address is not publicly routable {}", addr);
                    }

                    is_routable
                })
                .copied()
                .collect_vec();
```

PR #723, opened 2025-08-22
UPnP: Periodic External IP Refresh & Outbound Peer Reconnect on Change

p4bpj:
## Discussion Invitation
This PR (and upcoming related PRs) is meant to spark discussion on how we can further improve the reliability and usability of Kaspa nodes at home. It is provided as a template and idea starter and does not need to be merged verbatim in its current form.

## Summary
Adds automatic periodic refreshing of the external IP obtained via UPnP and propagates changes to the peer layer. The node no longer requires a restart after WAN IP changes (e.g. ISP lease renewals). When the external IP changes, outbound peers are reconnected gradually so the updated address spreads quickly.

## Motivation
Previously the external (WAN) IP was fetched exactly once at startup (via UPnP) and never validated again. If the router/ISP changed the IP later, the node continued advertising a stale address until manual restart. This degraded inbound connectivity and peer discovery. The goal: make external IP tracking resilient to dynamic WAN environments with minimal overhead and no startup delays.

## Key Changes
- **UPnP Extender**
  - Renews port mapping as before.
  - Also fetches `get_external_ip` each cycle.
  - Detects IP changes and updates AddressManager.
- **AddressManager**
  - Adds `set_best_local_address`.
  - Introduces `ExternalIpChangeSink` trait.
- **ConnectionManager**
  - Implements `ExternalIpChangeSink` to trigger a staggered outbound reconnect (avoids a connection storm).
- **Config / CLI**
  - Adds `--disable-ipv6-interface-discovery` to skip automatic IPv6 interface scanning (explicit IPv6 still honored).
- **Daemon**
  - Registers ConnectionManager as an external IP change sink early in startup.
- **Logging**
  - Informational / debug logs for mapping, extension, fetch results, and IP change events.

## Behavior
1. Startup: UPnP mapping established (existing flow).
2. Each extender interval (half the lease):
   - Renew port mapping.
   - Fetch current external IP.
   - If changed: update best local address + notify sinks.
3. ConnectionManager schedules outbound reconnects so new peers learn the updated address.

- Sink notifications are fire-and-forget tasks 

## Configuration
New flag:
```
--disable-ipv6-interface-discovery
```
Use to suppress automatic IPv6 interface enumeration while still allowing explicit IPv6 configuration.

## Testing / Verification
Manual scenarios:
- Stable IP: Only renewal logs; no change events.
- Forced WAN IP change (router reconnect): “External IP changed …” appears; outbound reconnects occur.
- With `--disable-ipv6-interface-discovery`: No automatically added IPv6 addresses.

## Authorship Note
This PR description was initially drafted with the assistance of an AI tool. I (p4bpj) have carefully reviewed, adjusted where necessary, and confirm that it accurately reflects the implemented changes.

### DIFF
--- a/components/addressmanager/src/lib.rs
+++ b/components/addressmanager/src/lib.rs
@@ -74,7 +74,7 @@ impl AddressManager {
         };
 
         let am = Arc::new(Mutex::new(instance));
-        let extender = Self::init_local_addresses_with_arc(&am, tick_service);
+        let extender = Self::init_local_addresses(&am, tick_service);
 
         (am, extender)
     }
@@ -87,7 +87,7 @@ impl AddressManager {
         self.external_ip_change_sinks.clone()
     }
 
-    fn init_local_addresses_with_arc(this: &Arc<Mutex<Self>>, tick_service: Arc<TickService>) -> Option<Extender> {
+    fn init_local_addresses(this: &Arc<Mutex<Self>>, tick_service: Arc<TickService>) -> Option<Extender> {
         let mut me = this.lock();
         me.local_net_addresses = me.local_addresses().collect();
 
@@ -185,7 +185,7 @@ impl AddressManager {
 
     fn upnp(&self) -> Result<Option<(NetAddress, ExtendHelper)>, UpnpError> {
         info!("[UPnP] Attempting to register upnp... (to disable run the node with --disable-upnp)");
-        let gateway = igd::search_gateway (Default::default())?;
+        let gateway = igd::search_gateway(Default::default())?;
         let ip = IpAddress::new(gateway.get_external_ip()?);
         if !ip.is_publicly_routable() {
             info!("[UPnP] Non-publicly routable external ip from gateway using upnp {} not added to store", ip);
--- a/components/addressmanager/src/lib.rs
+++ b/components/addressmanager/src/lib.rs
@@ -24,6 +24,10 @@ use thiserror::Error;
 
 pub use stores::NetAddress;
 
+pub trait ExternalIpChangeSink: Send + Sync {
+	fn on_external_ip_changed(&self, new_ip: std::net::IpAddr, old_ip: Option<std::net::IpAddr>);
+}
+
 const MAX_ADDRESSES: usize = 4096;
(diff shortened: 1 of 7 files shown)


kaspad/src/daemon.rs
```diff
@@ -611,7 +611,7 @@ do you confirm? (answer y/n or pass --yes to the Kaspad command line to confirm
         notify_service.notifier(),
         index_service.as_ref().map(|x| x.notifier()),
         mining_manager,
-        flow_context,
+        flow_context.clone(),
```
coderofstuff, line 550: Why is this clone necessary?

components/connectionmanager/src/lib.rs
```diff
@@ -93,6 +93,46 @@ impl ConnectionManager {
         });
     }
 
+    /// Synchronously trigger a staggered outbound reconnect (terminates peers one by one with 30s delays)
+    pub fn trigger_outbound_reconnect(&self) {
```
coderofstuff, line 97: Why is this necessary? When your external IP changes, wouldn't a disconnection already have happened (you disconnect from all your peers) and when you try to reconnect to your peers they would then know your new IP?

components/addressmanager/src/lib.rs
```diff
@@ -24,6 +24,10 @@ use thiserror::Error;
 
 pub use stores::NetAddress;
 
+pub trait ExternalIpChangeSink: Send + Sync {
```
coderofstuff, line 27: Why is this called `Sink`?

components/addressmanager/src/lib.rs
```diff
@@ -251,6 +281,14 @@ impl AddressManager {
         }
     }
 
+    pub fn set_best_local_address(&mut self, address: NetAddress) {
```
coderofstuff, line 284: I wonder if it's better to handle the `TODO` in `pub fn best_local_address(&mut self) -> Option<NetAddress> {` instead

coderofstuff (commented):
initial comments :)

coderofstuff:
Also please rebase and fix the lint and test issues

Issue #709, opened 2025-07-17
Everlasting cryptographic receipts using sequencing commitment


freshair18:
Kaspa's pruning means without external data full nodes cannot be used to prove a specific historical transaction was accepted to the DAG.

Instead an alternate paradigm needs to be adopted where users are responsible to create and store cryptographic receipts of transactions they potentially care about. If the need arises, these receipts could be verified by full nodes at any point in time.

Previously [KIP6](https://github.com/kaspanet/kips/blob/master/kip-0006.md) was meant to allow the creation of light weight receipts (via a logarithmic shortcut), however the required changes were left out of the crescendo HF and will not be incorporated in the near future if at all.

[KIP15](https://github.com/kaspanet/kips/blob/master/kip-0015.md) (already implemented) however does allow for similar functionality, which would be useful to support explicitly: 
Say a user wishes to create a receipt for tx0, which was accepted at chain block B, and let C=Post_posterity(B) be the first pruning block at the future of B (which is stored  to eternity). the fact that tx0 was accepted on B can be forever testified to by providing:
(a) all accepted transactions Merkle roots en route from B to C, i.e. the hashes needed to recreate the sequencing commitment of C from that of B.
(b) the Merkle witnness of tx0 in the sequencing commitment of B.
(c) tx0 data itself

**Comparison with KIP6 cryptographic receipts**

KIP6 was not only for transaction receipts per-se but also allowed for a more general primitive called proof of chain membership, which could have had other usages. In particular creating proofs of publication (for unaccepted transactions) as mentioned in KIP6 is not possible via the KIP15 sequencing commitment.

Size wise the logarithmic shortcut was very parsimonious, and KIP6 receipts were expected to be no larger than a few kilobytes in size. KIP15 receipts will not be competitive with that but the net result is still good due to the need to only store hashes and not headers:
If we assume a very conservative lower bound of the average DAG width being 2, there are 144000 chain blocks per pruning period, so the size of the receipts would be a maximum of 
32bytes*144000=4.60800MB + miniscule terms regarding the tx data itself and the Merkle witness for its acceptance in the eventual block. In practice it will be significantly lower.

In the future it may be worthwhile to increase the frequency of "posterity blocks" to further decrease this size, as well as improve the wait time required to be able to create a receipt.

**Previous Implementation**
In implementing this feature, one may base their work or take inspiration from the now obsolete (partial) PR on KIP6:
https://github.com/kaspanet/rusty-kaspa/pull/609

[...]

PR #725, opened 2025-08-25
Consensus level support for cryptographic receipts 

freshair18:
Provides consensus layer support for cryptographic receipts based on KIP15.

Additionally:
* Adds a bfs iterator for the blockdag, currently incorporated only in new functions included in this PR, but potentially should be incorporated to the existing functions using bfs traversal.
* Provides consensus infra for  "primitive" Pochm, proof of publication, and "legacy" transaction receipts, by storing selected chain and its peripharies, as briefly described in https://github.com/kaspanet/kips/blob/master/kip-0006.md.

Supersedes https://github.com/kaspanet/rusty-kaspa/pull/609, addresses https://github.com/kaspanet/rusty-kaspa/issues/709.


consensus/core/src/receipts.rs
```diff
@@ -0,0 +1,68 @@
+use kaspa_merkle::MerkleWitness;
+use std::{collections::HashMap, sync::Arc};
+
+use crate::header::Header;
+use kaspa_hashes::Hash;
+#[derive(Clone)]
+pub struct Pochm {
```
coderofstuff: Please rename this struct to `ProofOfChainMembership`. I know what `pochm` means because I'm familiar with the KIP, but an arbitrary reader of this struct will not have that context immediately.

Also, please add comments to the struct or parts of the struct explaining what this it's intended. Take a different struct for example, it has this:

```
pub struct PruningPointReply {
    /// The most recent pruning sample from POV of the queried block (with distance up to ~F)
    pub pruning_sample: Hash,

    /// The pruning point of the queried block. I.e., the most recent pruning sample with
    /// depth P (except for shortly after the fork where the new P' is gradually reached)
    pub pruning_point: Hash,
}
```

consensus/core/src/api/mod.rs
```diff
@@ -371,6 +371,37 @@ pub trait ConsensusApi: Send + Sync {
...
+    }
+    fn generate_proof_of_pub(
+        &self,
+        tx_id: Hash,
+        publishing_block: Option<Hash>,
+        tx_timestamp: Option<u64>,
+    ) -> ConsensusResult<ProofOfPublication> {
+        unimplemented!()
+    }
+    fn generate_pochm(&self, block: Hash) -> ConsensusResult<Pochm> {
```
coderofstuff: Please use `generate_proof_of_chain_membership`

consensus/core/src/api/mod.rs
```diff
@@ -371,6 +371,37 @@ pub trait ConsensusApi: Send + Sync {
...
+        unimplemented!()
+    }
+    fn verify_tx_receipt(&self, receipt: &TxReceipt) -> bool {
+        unimplemented!()
+    }
+    fn verify_proof_of_pub(&self, proof_of_pub: &ProofOfPublication) -> bool {
```
coderofstuff: Please use `verify_proof_of_publication`

consensus/core/src/api/mod.rs
```diff
@@ -371,6 +371,37 @@ pub trait ConsensusApi: Send + Sync {
...
+        unimplemented!()
+    }
+    fn verify_pochm(&self, chain_purporter: Hash, pochm: &Pochm) -> bool {
```
coderofstuff: Please use `verify_proof_of_chain_membership`

consensus/src/consensus/mod.rs
```diff
@@ -486,6 +487,178 @@ impl Consensus {
             .map(|hash| (hash, self.headers_store.get_compact_header_data(hash).unwrap()))
             .collect_vec()
     }
+    /* this logic may not be the most efficient and reliable as of now */
```
coderofstuff: Inside this function are several large comment blocks. Please update this into a function comment explaining what this function does - maybe move parts of those large comments below to this function comment.

Take a look at this function elsewhere for an example:

```
pub fn select_transactions(&mut self) -> Vec<Transaction>
```

coderofstuff (commented):
Still need to wrap my head around the KIP and PR. Initial cosmetic comments first :)

consensus/core/src/receipts.rs
```diff
@@ -0,0 +1,68 @@
...
+    pub hdr_map: HashMap<Hash, Arc<Header>>,
```
IzioDev: At a first lecture, I thought `hdr` was reference to some specific object I wasn't aware of, would it be possible to rename it `header` instead?

consensus/core/src/receipts.rs
freshair18: see now

consensus/core/src/api/mod.rs
freshair18: addressed

consensus/core/src/api/mod.rs
freshair18: addressed

consensus/src/consensus/mod.rs
freshair18: see now

consensus/core/src/receipts.rs
freshair18: addressed

freshair18:
I removed proof of publication and proof of chain memberships as it became apparent to me these features' implementation is unsatisfactory, and the correct implementation not immediate. Coupled with the fact that it greatly complicated tests, and that  the usecase was always dubious I decided to cut it off in the hopes that it will ease reviewing.

An implementation of proof of publication  as it previously was can be found at  https://github.com/freshair18/rusty-kaspa/tree/crypt_receipts_with_pop.

PR #727, opened 2025-08-28
feat(wrpc/wasm): get network id and fix current url update

IzioDev:
Allows `rpc.networkId` from wasm usage to get the current `NetworkId`.

Fixes an edge-case where `current_url` wasn't updated, this happened specifically when `options.url` was set, because the underlying `ws` library isn't calling the `RpcResolver::resolve_url` trait function in this case, which makes sense.


IzioDev:
cc @aspect

rpc/wrpc/client/src/client.rs
```diff
@@ -372,6 +372,10 @@ impl KaspaRpcClient {
         Ok(())
     }
 
+    pub fn network_id(&self) -> Result<Option<NetworkId>> {
+        Ok(*self.inner.network_id.lock().unwrap())
```
biryukovmaxim: The function can't return an error, result in function signature is redundant

IzioDev: Nice catch, thanks.

IzioDev: updated in [review: remove surrounding result](https://github.com/kaspanet/rusty-kaspa/pull/727/commits/df6a38a39dd9e9ce63cdb55a8d6dae4d8cd94fe6)

PR #736, opened 2025-10-04
Crescendo cleanup

michaelsutton:


consensus/src/model/stores/pruning.rs
```diff
@@ -5,38 +5,34 @@ use kaspa_database::prelude::DB;
...
 use kaspa_database::registry::DatabaseStorePrefixes;
 use kaspa_hashes::Hash;
+use kaspa_hashes::ZERO_HASH;
 use rocksdb::WriteBatch;
 use serde::{Deserialize, Serialize};
 
 #[derive(Clone, Copy, Serialize, Deserialize)]
-pub struct PruningPointInfo {
+struct PruningPointInfo {
     pub pruning_point: Hash,
```
someone235: The fields don't need to be `pub` anymore

consensus/src/model/stores/pruning.rs
```diff
@@ -5,38 +5,34 @@ use kaspa_database::prelude::DB;
...
-    pub candidate: Hash,
+    pub _candidate: Hash, // Obsolete filed. Kept only for avoiding the DB upgrade logic. TODO: remove all together
```
someone235: filed -> field

consensus/src/pipeline/header_processor/processor.rs
```diff
@@ -98,18 +96,14 @@ impl HeaderProcessingContext {
 
     /// Returns the pruning point at the time this header began processing
     pub fn pruning_point(&self) -> Hash {
```
someone235: If the field is `pub`, the method fills redundant

consensus/core/src/hashing/tx.rs
```diff
@@ -2,33 +2,40 @@ use super::HasherExtensions;
...
-/// A bitmask defining which transaction fields we
-/// want to encode and which to ignore.
-type TxEncodingFlags = u8;
-
-pub const TX_ENCODING_FULL: TxEncodingFlags = 0;
-pub const TX_ENCODING_EXCLUDE_SIGNATURE_SCRIPT: TxEncodingFlags = 1;
+bitflags::bitflags! {
+    /// A bitmask defining which transaction fields we want to encode and which to ignore.
+    #[derive(Debug, Clone, Copy, PartialEq, Eq)]
+    pub struct TxEncodingFlags: u8 {
```
someone235, line 8: nit: Maybe it's better just to have a struct of booleans?

consensus/core/src/merkle.rs
```diff
@@ -2,8 +2,8 @@ use crate::{hashing, tx::Transaction};
 use kaspa_hashes::Hash;
 use kaspa_merkle::calc_merkle_root;
 
-pub fn calc_hash_merkle_root<'a>(txs: impl ExactSizeIterator<Item = &'a Transaction>, include_mass_field: bool) -> Hash {
-    calc_merkle_root(txs.map(|tx| hashing::tx::hash(tx, include_mass_field)))
+pub fn calc_hash_merkle_root<'a>(txs: impl ExactSizeIterator<Item = &'a Transaction>) -> Hash {
```
someone235, line 5: I wonder if we should keep a version of this function with a crescendo_activated flag that can be used for archival nodes, proof of publication, etc (where in body_validation you'll always pass true)

consensus/src/pipeline/body_processor/body_validation_in_context.rs
```diff
@@ -129,7 +129,7 @@ mod tests {
     use kaspa_hashes::Hash;
 
     fn calc_hash_merkle_root<'a>(txs: impl ExactSizeIterator<Item = &'a Transaction>) -> Hash {
```
someone235: This wrapper looks redundant

rpc/service/src/converter/consensus.rs
```diff
@@ -124,7 +124,7 @@ impl ConsensusConverter {
         if include_verbose_data {
             let verbose_data = Some(RpcTransactionVerboseData {
                 transaction_id: transaction.id(),
-                hash: hash(transaction, false),
+                hash: hash(transaction),
```
someone235, line 127: This looks like a breaking change and should be documented in the version release notes

someone235 (changes requested):

consensus/core/src/merkle.rs
michaelsutton: How about adding a `calc_hash_merkle_root_pre_crescendo` ?

PR #740, opened 2025-10-14
feat(rpc/vccv2): enhanced vcc data aggregator

IzioDev:
This PR is a follow-up to @D-Stacks initial effort [here](https://github.com/D-Stacks/rusty-kaspa/tree/vspcc-v2-gutted).
Massive thanks to @coderofstuff who helped **a lot** on taking it to the state it currently is.

### What?
VSPCv2 allows gathering more data than the current VSPC endpoint in a non-breaking change way, in particular "acceptance data". It is a way for network observers to solely follow VSPCv2 without having to keep parallel listener(s) on added block (and potential sender resolver on top of it). Example, end-users can now rely on VSPCv2 and get: accepted DAA score, accepting block, sender address for each transaction that got accepted by each mergeset of chain block (accepting block).

Since the endpoint can be chatty, @D-Stacks introduced a granular verbosity selector (at field level). We found it could be difficult to manipulate as an end-users so we introduced a simplified top-level verbosity level instead (full, high, low, none).

Example usage:
```rs
let response = client
    .get_virtual_chain_from_block_v2(
        // start hash
        RpcHash::from_str("ef26630bd11660faaf5bdd2d49dae3b8f39519c8ea9a7b0ccbd8d18a0716d15a").unwrap(),
        // optional verbosity level (default to ::None)
        Some(RpcDataVerbosityLevel::Low),
        // optional min confirmation counts
        None,
    )
    .await?;
```

To avoid user confusion about populated fields, comment documentations have been added. It is yet to be determined how to translate this for `wasm` users.

Example, while I'm in my IDE, I can hover used field to get insight on required verbosity level it needs:
image

### Verbosity Level
They are incremental, meaning a verbosity level of 2 (High) would include every fields defined at a verbosity level of 1 (Low).

To get a detailed view, I suggest two entry-point:
* [per field definition](https://github.com/kaspanet/rusty-kaspa/pull/740/files#diff-4bf2433efc159e4c00ebfb53eb33d5ce3c0a375116139d72ca999c056fc4cfef)
* [struct fields documentation](https://github.com/kaspanet/rusty-kaspa/pull/740/files#diff-13cc641de5a00d3ba2c493214fc798d0d3cb8bf20a5d7898a1c131db3300b66aR11) (valid in other files, this is one of those)

### Use-case
As a network observer needing data tracking at an accepted transaction level, I can now solely rely on this endpoint given the fact i have a block hash checkpoint (last scanned/ingested block).
In such case, I'd call the new endpoint with a verbosity level fitting my requirements, in this example i'll provide "None" and we'll use the wasm sdk:
```js
// Start from node sink / pruning point
let lowHash = info.sink;
console.info("Starting lowHash (sink):", lowHash);

// Main loop - runs forever every 10 seconds
while (true) {
  try {
    const date = new Date();
    const vspc = await rpc.getVirtualChainFromBlockV2({
      startHash: lowHash,
      minConfirmationCount: 10,
      dataVerbosityLevel: "None",
    });
    console.info("VSPC Info:", vspc);

    for (const hash of vspc.removedChainBlockHashes) {
      console.info("Removed block hash:", hash);
    }

    for (const hash of vspc.addedChainBlockHashes) {
      console.info("Added block hash:", hash);
      lowHash = hash;
    }

    for (const cbat of vspc.chainBlockAcceptedTransactions) {
      // Do something with the chain block header
      console.info(cbat.chainBlockHeader);
      // Do something with the accepted transactions
      console.info(cbat.acceptedTransactions);
    }

    console.info("Time span:", Date.now() - date.getTime(), "ms");
  } catch (innerErr) {
    console.error("Error in loop iteration:", innerErr);
    // keep running despite errors
  }

  // wait 10 seconds before next iteration
  await delay(10000);
}
```

An example of usage is provided [here](https://github.com/kaspanet/rusty-kaspa/pull/740/files#diff-ed6d4ce77133c24e7d91772c9986f7af928409c41b684586ccf509c045949fecR56).

### Future possibilities
* While having an on-demand endpoint can be sufficient, reactive (as in react to events) applications might prefer using a notification system (pub/sub model). This endpoint could serve as a base for a notification system: `NotifyVirtualChainChangedV2`

### Latencies in high load environment
in `/testing/integration/src/rpc_perf_benchmarks.rs`, you'll find an attempt at measuring rpc calls. It simulates a high load network with several transactions being published to the mempool regularly, as-well as a group of miner publishing blocks.

Test case 1: 20 parallel clients each doing 20 requests sequentially:
```
 Total requests: 400
 Total duration: 21.9941844s
 Requests per second: 18.19
 --------------------
Latency metrics:
   Min: 1.0905ms
   Max: 1.3626893s
   Avg: 372.957331ms
   p95: 881.6559ms
   p99: 1.0612529s
```

Test case 2: 1 client doing 20 requests sequentially
```
Total requests: 20
Total duration: 1.3679693s
Requests per second: 14.62
--------------------
 Latency metrics:
   Min: 23.2835ms
   Max: 126.0707ms
   Avg: 53.278565ms
   p95: 119.5862ms
   p99: 126.0707ms
```

### Extra
Since I needed to try the `gRPC` protocol, I've also introduced a simple gRPC client example in `/rpc/grpc/examples/simple_client` (same semantic as wRPC client example)

### DIFF
--- a/rpc/service/src/converter/consensus.rs
+++ b/rpc/service/src/converter/consensus.rs
@@ -335,7 +335,7 @@ impl ConsensusConverter {
                         .map_err(|_| AddressError::InvalidAddress)?,
                 )
             } else {
-                None
+                Default::default()
             },
         })
     }
@@ -570,7 +570,7 @@ impl ConsensusConverter {
         })
     }
 
-    async fn get_mergeset_transactions_with_verbosity(
+    async fn get_mergeset_accepted_transactions_with_verbosity(
         &self,
         consensus: &ConsensusProxy,
         accepting_block: Hash,
@@ -638,7 +638,7 @@ impl ConsensusConverter {
 
             if let Some(mergeset_block_acceptance_data_verbosity) = verbosity.mergeset_block_acceptance_data_verbosity.as_ref() {
                 let mergeset_transactions_with_verbosity = self
-                    .get_mergeset_transactions_with_verbosity(
+                    .get_mergeset_accepted_transactions_with_verbosity(
                         consensus,
                         *accepting_chain_hash,
                         chain_block_mergeset_acceptance_data,


IzioDev:
> Comments from Michael:
> 
>     1. remove merged blocks headers from high verbosity (we can always add it with an even higher verbosity)
> 
>     2. support min confirmation count feature on v2 as well
> 
>     3. rename v_2 to v2
> 
>     4. evaluate performance costs of this api under TPS stretch and if required, decide how to keep it under control
> 
> 
> Additional: Ensure that the new RPC, it's arguments and everything that goes in there is added into WASM. There must be an `rpc.getVirtualChainFromBlockV2` that can be used through wasm

All of the point but performance evaluation have been addressed.

Still some cleaning tasks to be done and some change in low-level utxo inquirer (panic-safe related). Other than that, it's in a testable state.

cli/src/modules/rpc.rs
```diff
@@ -290,6 +290,30 @@ impl Rpc {
 
                 self.println(&ctx, result);
             }
+            RpcApiOps::GetVirtualChainFromBlockV2 => {
+                if argv.is_empty() {
+                    return Err(Error::custom("Missing startHash argument"));
+                };
+                let start_hash = RpcHash::from_hex(argv.remove(0).as_str())?;
+
+                argv.reverse(); // reverse so we can pop from the end
```
coderofstuff: Is this really needed? How many arguments does this RPC call have?

rpc/core/src/error.rs
```diff
@@ -141,6 +141,12 @@ pub enum RpcError {
     #[error("utxo return address could not be found -> {0}")]
     UtxoReturnAddressNotFound(UtxoInquirerError),
 
+    #[error("header hashes to {0}, provided is {1}")]
+    BadBlockHash(RpcHash, RpcHash),
```
coderofstuff: This is not used anywhere

consensus/core/src/utxo/utxo_inquirer.rs
```diff
@@ -11,6 +13,8 @@ pub enum UtxoInquirerError {
     NoTxAtScore,
     #[error("Transaction was found but not standard")]
     NonStandard,
+    #[error("no transaction specified")]
```
coderofstuff: Capitalize the first letter to match everything else in here

consensus/core/src/utxo/utxo_inquirer.rs
```diff
@@ -33,6 +35,26 @@ pub enum UtxoInquirerError {
     MissingHashAtIndex(u64),
     #[error("Did not find acceptance data for chain block {0}")]
     MissingAcceptanceDataForChainBlock(Hash),
+    #[error("Did not find accepting block for {0}")]
+    MissingAcceptingBlock(Hash),
```
coderofstuff: This is unused

consensus/core/src/utxo/utxo_inquirer.rs
```diff
@@ -33,6 +35,26 @@ pub enum UtxoInquirerError {
...
+    #[error("Did not find utxo entry for outpoint {0}")]
+    MissingUtxoEntryForOutpoint(TransactionOutpoint),
+    #[error("Did not find queried transactions in acceptance data: {0:?}")]
+    MissingQueriedTransactions(Vec<TransactionId>),
+    #[error("unexpected store error")]
```
coderofstuff: Capitalize the first letter to match the rest of the error text here

testing/integration/src/rpc_perf_benchmarks.rs
```diff
@@ -0,0 +1,174 @@
...
+
+// Constants for transaction generation and mempool pressure
+const MEMPOOL_TARGET: u64 = 10_000;
+const TX_COUNT: usize = 200_000;
+const TX_LEVEL_WIDTH: usize = 5_000;
+const TPS_PRESSURE: u64 = 40;
+const PREALLOC_AMOUNT_SOMPI: u64 = 1;
+const SUBMIT_TX_CLIENTS: usize = 2;
+
+/// `cargo test --release --package kaspa-testing-integration --lib --features devnet-prealloc -- rpc_perf_benchmarks::bench_rpc_high_load --exact --nocapture --ignored`
```
coderofstuff, line 35: Is this a test that's supposed to run indefinitely?

rpc/wrpc/examples/vcc_v2/src/main.rs
```diff
@@ -0,0 +1,111 @@
...
+        strategy: ConnectStrategy::Fallback,
+        ..Default::default()
+    };
+
+    // Connect to selected Kaspa node
+    client.connect(Some(options)).await?;
+
+    let response = client
+        .get_virtual_chain_from_block_v2(
+            RpcHash::from_str("ef26630bd11660faaf5bdd2d49dae3b8f39519c8ea9a7b0ccbd8d18a0716d15a").unwrap(),
```
coderofstuff: This example becomes useless very quickly when this hash becomes pruned. Instead of a hardcoded value, use something like the sink or the pruning point or something in between to demonstrate the example.

consensus/src/pipeline/virtual_processor/utxo_inquirer.rs
```diff
@@ -138,50 +332,159 @@ impl VirtualStateProcessor {
...
+                                .enumerate()
+                                .filter_map(|(i, tx)| {
+                                    if i == next_target_index {
+                                        next_target_index = index_iter.next().unwrap_or(usize::MAX);
+                                        Some(tx.clone())
+                                    } else {
+                                        None
+                                    }
+                                })
+                                .take(mbad.accepted_transactions.len()),
```
coderofstuff: Add comments here to explain the logic of this code.

consensus/src/pipeline/virtual_processor/utxo_inquirer.rs
```diff
@@ -138,50 +332,159 @@ impl VirtualStateProcessor {
...
+                    .map_err(|_| UtxoInquirerError::MissingBlockFromBlockTxStore(containing_block))
+                    .and_then(|block_txs| {
+                        block_txs.get(index).cloned().ok_or(UtxoInquirerError::MissingTransactionIndexOfBlock(index, containing_block))
+                    })?;
+
+                Ok(vec![tx])
+            }
+            Some(_more) => {
+                Err(UtxoInquirerFindTxsFromAcceptanceDataError::TxIdsFilterNeedsLessOrEqualThanOneElementError.into())
+                // artifact implementation that has been commented, keeping it for track record as long as it's unimplemented
```
coderofstuff, line 440: nit. make sure to mark this as TODO

coderofstuff (changes requested):
Great work here! The command itself looks to be running really well and I've been testing it.

testing/integration/src/rpc_perf_benchmarks.rs
michaelsutton: No. It's a benchmark you run when there are significant changes in the benchmarked area. cf mempool bmk

cli/src/modules/rpc.rs
IzioDev: No it's not, we currently have 2 arguments, in this order: <start_hash> [verbosity_level] (note that the last is optional)

I removed the reverse operation as it indeed is useless now in this context 👍
[review(cli): remove useless reverse operation on argv](https://github.com/kaspanet/rusty-kaspa/pull/740/commits/0170f8cf87beebde7c33986a5a930aaa573d4a9c)

rpc/core/src/error.rs
IzioDev: Nice catch
[review(rpc/error): remove unused member](https://github.com/kaspanet/rusty-kaspa/pull/740/commits/bf33dd396501ecd3ec75dc56241081c8031449a3)

consensus/core/src/utxo/utxo_inquirer.rs
IzioDev: [review(utxo-inquierer/errors): removed unused and message format unif…](https://github.com/kaspanet/rusty-kaspa/pull/740/commits/e7d8c3ce152d2bee44527c037152c52578335aaf)

consensus/core/src/utxo/utxo_inquirer.rs
IzioDev: Yes, now handled "natively" by `ConsensusError`.
I've also identified more unused members, now cleaned
[review(utxo-inquierer/errors): removed unused and message format unif…](https://github.com/kaspanet/rusty-kaspa/pull/740/commits/e7d8c3ce152d2bee44527c037152c52578335aaf)

consensus/core/src/utxo/utxo_inquirer.rs
IzioDev: [review(utxo-inquierer/errors): removed unused and message format unif…](https://github.com/kaspanet/rusty-kaspa/pull/740/commits/e7d8c3ce152d2bee44527c037152c52578335aaf)

testing/integration/src/rpc_perf_benchmarks.rs
IzioDev: The benchmark indeed did not stop once the results are produced, the program hanged (kept pushing transaction and submitting blocks) in the background after having printed the metrics.
I've modified the benchmark to stop tasks now, it's not perfect as I have error logs upon stopping:
```
[ERROR] Tx sender channel returned error sending into a closed channel
```
But yet better than hanging, if necessary, I can push further

[review(benchmark/rpc): stop program execution after getting the results](https://github.com/kaspanet/rusty-kaspa/pull/740/commits/7bfde5e40c26c8740db08110d3d8b38a2c4abbe6)

Edit: another way would be to constantly set a pressure, constantly push metrics and keep program opens.

rpc/wrpc/examples/vcc_v2/src/main.rs
IzioDev: Revamped the example to present realistic use-cases
[review(wrpc/example): revamped](https://github.com/kaspanet/rusty-kaspa/pull/740/commits/d368e189eb84c390c631f151f93544e87e188919)

consensus/src/pipeline/virtual_processor/utxo_inquirer.rs
IzioDev: While I was writing the comment (and hence deep re-reading this logic), I found it was too complicated (not very readable) for the "simple" mapping need.

I proposed a more naive approach here, let me know what you think. I believe this implementation is self-explanatory, but if you think it also needs comment tell me
[refacto(utxo-inquirer): find txs from acceptance data](https://github.com/kaspanet/rusty-kaspa/pull/740/commits/b1583d25801add101d427143f6f1185554c9857c)

coderofstuff: It's much better. Thanks

consensus/src/pipeline/virtual_processor/utxo_inquirer.rs
IzioDev: [review(utxo-inquirer): annotate unimplemented part with todo](https://github.com/kaspanet/rusty-kaspa/pull/740/commits/9714f5ae6d31150de5f4ad3ce1e841664afdf352)

IzioDev:
@coderofstuff as per discussed on R&D, updated with upstream/master and set verbosity level to full by default

[review(rpc/vspcv2): set verbosity level to full by default](https://github.com/kaspanet/rusty-kaspa/pull/740/commits/3b7423c49ff6b052f79ab95c54e10b8053036e43)

rpc/grpc/core/src/convert/header.rs
```diff
@@ -65,25 +106,30 @@ try_from!(item: &protowire::RpcBlockHeader, kaspa_rpc_core::RpcHeader, {
         RpcHash::from_str(&item.pruning_point)?,
     );
 
-    header.into()
-});
+    kaspa_rpc_core::RpcHeader::from(header)
```
coderofstuff: Why is this change necessary?

rpc/grpc/core/src/convert/header.rs
```diff
@@ -48,6 +72,23 @@ from!(item: &[RpcHash], protowire::RpcBlockLevelParents, { Self { parent_hashes:
 // protowire to rpc_core
 // ----------------------------------------------------------------------------
 
+try_from!(item: &protowire::RpcBlockHeader, kaspa_rpc_core::RpcRawHeader, {
```
coderofstuff: Any reason this was moved up?

IzioDev: No reason, fixed here [refacto(rpc/grpc): convert header file re-arranged](https://github.com/kaspanet/rusty-kaspa/pull/740/commits/4e556938587056c6dac9286dd52eadcc4e62f4b4)

rpc/grpc/core/src/convert/header.rs
IzioDev: It's not indeed, fixed [refacto(rpc/grpc): convert header file re-arranged](https://github.com/kaspanet/rusty-kaspa/pull/740/commits/4e556938587056c6dac9286dd52eadcc4e62f4b4)

rpc/service/src/service.rs
```diff
@@ -1231,6 +1243,54 @@ NOTE: This error usually indicates an RPC conversion error between the node and
...
+                }
+            }
+        }
+
+        let chain_blocks_accepted_transactions = self
+            .consensus_converter
+            .get_chain_blocks_accepted_transactions(&session, &verbosity, &chain_path, Some(batch_size))
+            .await?;
+
+        chain_path.added.truncate(chain_blocks_accepted_transactions.len());
```
coderofstuff: What's this line for?

rpc/service/src/converter/consensus.rs
```diff
@@ -178,6 +188,477 @@ impl ConsensusConverter {
...
+                        rpc_tx
+                    });
+                }
+
+                converted
+            }
+        })
+    }
+
+    async fn get_mergeset_transactions_with_verbosity(
```
coderofstuff: Rename `get_mergeset_transactions_with_verbosity` to `get_mergeset_accepted_transactions_with_verbosity`

`get_mergeset_transactions_with_verbosity` implies that it includes all transactions in the mergeset, where really this just returns the accepted ones.

rpc/service/src/converter/consensus.rs
```diff
@@ -178,6 +188,477 @@ impl ConsensusConverter {
...
+            } else {
+                Default::default()
+            },
+            script_public_key_address: if verbosity.include_script_public_key_address.unwrap_or(false) {
+                Some(
+                    extract_script_pub_key_address(&output.script_public_key, self.config.prefix())
+                        .map_err(|_| AddressError::InvalidAddress)?,
+                )
+            } else {
+                None
```
coderofstuff: nit. Inconsistency. This is the only place among the converters added by this PR where `None` is used instead of `Default::default()`

rpc/service/src/converter/consensus.rs
IzioDev: Fixed here: [review: small rename and inconsistency fix](https://github.com/kaspanet/rusty-kaspa/pull/740/commits/0f7c2dd54729f7df0e6ab84c73d48f496e55822f)

rpc/service/src/converter/consensus.rs
IzioDev: Fixed here: [review: small rename and inconsistency fix](https://github.com/kaspanet/rusty-kaspa/pull/740/commits/0f7c2dd54729f7df0e6ab84c73d48f496e55822f)

coderofstuff (approved):
Great work here @IzioDev and @D-Stacks :)

PR #741, opened 2025-10-15
Allow Wallet Core to generate transactions with 1 UTXO and a payload

wigwagwent:
The kasia broadcast use this form where you send yourself a transaction with a payload. The only kaspa is spent on the transaction fees so it should only need 1 utxo to cover the cost.

### DIFF
--- a/wallet/core/src/account/pskb.rs
+++ b/wallet/core/src/account/pskb.rs
@@ -453,6 +453,7 @@ pub async fn commit_reveal_batch_bundle(
         CommitRevealBatchKind::Manual { hop_payment, destination_payment } => {
             let addr_commit = match hop_payment.clone() {
                 PaymentDestination::Change => Err(Error::CommitRevealInvalidPaymentDestination),
+                PaymentDestination::PayloadOnly => Err(Error::CommitRevealInvalidPaymentDestination),
                 PaymentDestination::PaymentOutputs(payment_outputs) => {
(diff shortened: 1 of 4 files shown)


IzioDev:
Hello, what are/is the fundamental difference(s) in behaviors between the freshly introduced `PayloadOnly` payment type and the already existing `Change` payment type?

wigwagwent:
I needed to use an empty `vec` of `output` instead of outputs with 0 amount to self when I was trying to make a transaction whose purpose was to send a payload. This makes it easier to understand what is going on but does not add any features. The other option would be make outputs an `option<Outputs>` this would be a breaking change for every wallet using an output.

If you think that this change is not worth it and people should continue with how it is done now, this can be closed.

IzioDev:
> I needed to use an empty `vec` of `output` instead of outputs with 0 amount to self when I was trying to make a transaction whose purpose was to send a payload. This makes it easier to understand what is going on but does not add any features. The other option would be make outputs an `option<Outputs>` this would be a breaking change for every wallet using an output.
> 
> If you think that this change is not worth it and people should continue with how it is done now, this can be closed.

I wonder what's the "best" way of achieving such a sugar syntax.
Ultimately a `PayloadOnly` is a `Change` payment type. In both case, I believe one could to still want to have the ability to control what is the used change (or receive) target address, which is defined by `PaymentOutput` currently in case of a `Change` payment type.

What are your thoughts about it?

PR #742, opened 2025-10-16
Add docker files for all binaries

someone235:
This adds docker support, and an automated script for multi-arch build.

I chose to make a dedicated Dockerfile for each binary, to make the build as fast as possible for people who only need one binary and don't want to compile dependencies for other binaries.

The content of this PR is heavily influenced by the content of @supertypo's repository https://github.com/supertypo/docker-rusty-kaspa, where the main advantage of placing it inside rusty-kaspa repo is to support building docker images for different branches or uncommited work.

**P.S.**
I updated rocksdb to 0.24.0 to avoid [this](https://github.com/rust-rocksdb/rust-rocksdb/issues/953) bug.

### DIFF
--- a/musl-toolchain/build.sh
+++ b/musl-toolchain/build.sh
@@ -85,12 +85,3 @@ rustup target add x86_64-unknown-linux-musl
 
 # Install missing dependencies
 cargo fetch --target x86_64-unknown-linux-musl
-
-# Patch missing include in librocksdb-sys-0.16.0+8.10.0. Credit: @supertypo
-FILE_PATH=$(find $HOME/.cargo/registry/src/ -path "*/librocksdb-sys-0.16.0+8.10.0/*/offpeak_time_info.h")
-
-if [ -n "$FILE_PATH" ]; then
-  sed -i '1i #include <cstdint>' "$FILE_PATH"
-else
-  echo "No such file for sed modification."
-fi


IzioDev:
Since it contains multiple RocksDB major version upgrade, i have:
On Windows, setup a local node with `master` version, let it synced and run for a few minutes.
Then, I built this version (`someone235:docker`) and bootstrap the node with the same datadir.

Node is running since a few hours, no issues.

Cargo.lock
```diff
@@ -3770,14 +3759,13 @@ dependencies = [
 
 [[package]]
 name = "librocksdb-sys"
-version = "0.16.0+8.10.0"
+version = "0.17.3+10.4.2"
```
michaelsutton, line 3762: This can possibly mean we can remove this line (and below it) https://github.com/someone235/rusty-kaspa/blob/6aaaf8e3f873752f71bae036066b1a6031fb0c21/musl-toolchain/build.sh#L90 from the CI build process. I think it's worth testing. 
cc: @saefstroem @supertypo

rubend056:
Thanks for requesting rocksdb bump, I couldn't compile with recent toolchain here on Archlinux and bumping rocksdb locally to  0.24.0  solved it for me.

PR #743, opened 2025-10-16
feat(txscript): introduce human readable viewer

IzioDev:
Usage:
```rust
    let mut s = ScriptBuilder::new();
    s.script_mut().extend_from_slice(&script_vec);

    // print the human readable form
    println!(
        "{}",
        s.string_view::<ValidatedTransaction, SigHashReusedValuesSync>(ScriptViewerOptions {
            contains_redeem_script: true,
            address_prefix: Prefix::Mainnet
        })
    );
```

Input:
```
4130ef124590e4e6627078a658e2eb0b89fe4733f40d8cbfe0d077ae16bb90afb0a5f10e5693352e4b9d19d77a98fe75e395ce60988a0750ab8603a252c9c7290401412294d292317d03d1a5f49a8204c35486da84bbbea604209637e2bbfb5bbfabb36bcb37fc90aeb9836ed950a42b87382880fbd926b362cdbca16e9db9891918850141c2d76d4c64c9b8a8a64fa34a69f7cea953c4f0e564463226d931481ee1fbccafd7c20500a699fc8a10d01d03219d25944081750cdbba89e6a5a64b3224f58a5a014c875320b0a2f302b97271d6d1f20f2168e8b86b037d42a52aaf7ca959bea8a8bbf859a220e040996f44024491881ad4d2f59d4397a5a1f2e169c55624cb9509693fbb7a14204e518f0ecb51eef7db45042e441bb4d99f2c68277359bea369fcb7c80bee5b0120924013135715c9a8076141a33d6528a13fa2e816d3f006897b6d6c8b1da90fd754ae
```

Output:
```
OpData65 hex: 30ef124590e4e6627078a658e2eb0b89fe4733f40d8cbfe0d077ae16bb90afb0a5f10e5693352e4b9d19d77a98fe75e395ce60988a0750ab8603a252c9c7290401
OpData65 hex: 2294d292317d03d1a5f49a8204c35486da84bbbea604209637e2bbfb5bbfabb36bcb37fc90aeb9836ed950a42b87382880fbd926b362cdbca16e9db98919188501
OpData65 hex: c2d76d4c64c9b8a8a64fa34a69f7cea953c4f0e564463226d931481ee1fbccafd7c20500a699fc8a10d01d03219d25944081750cdbba89e6a5a64b3224f58a5a01
OpPushData1 len=135 hex: 5320b0a2f302b97271d6d1f20f2168e8b86b037d42a52aaf7ca959bea8a8bbf859a220e040996f44024491881ad4d2f59d4397a5a1f2e169c55624cb9509693fbb7a14204e518f0ecb51eef7db45042e441bb4d99f2c68277359bea369fcb7c80bee5b0120924013135715c9a8076141a33d6528a13fa2e816d3f006897b6d6c8b1da90fd754ae
  (redeem script:
    Op3
    OpData32 hex: b0a2f302b97271d6d1f20f2168e8b86b037d42a52aaf7ca959bea8a8bbf859a2
    OpData32 hex: e040996f44024491881ad4d2f59d4397a5a1f2e169c55624cb9509693fbb7a14
    OpData32 hex: 4e518f0ecb51eef7db45042e441bb4d99f2c68277359bea369fcb7c80bee5b01
    OpData32 hex: 924013135715c9a8076141a33d6528a13fa2e816d3f006897b6d6c8b1da90fd7
    Op4
    OpCheckMultiSig
    // 3 of 4
    // kaspa:qzc29uczh9e8r4k37g8jz68ghp4sxl2z55427l9ftxl2329mlpv6yjuyyrqd0
    // kaspa:qrsypxt0gspyfyvgrt2d9avagwt6tg0ju95u243yew2sj6flhdapg0aca5fql
    // kaspa:qp89rrcwedg7aa7mg5zzu3qmknve7trgyae4n04rd87t0jqtaedsz0kwjemhq
    // kaspa:qzfyqycn2u2un2q8v9q6x0t99zsnlghgzmflqp5f0dkkezca4y8awuvxkn228
  )
```

An early review has been made by @biryukovmaxim [here](https://github.com/IzioDev/rusty-kaspa/pull/2) for the record. Thanks Maxim 🙏

### DIFF
--- a/crypto/txscript/Cargo.toml
+++ b/crypto/txscript/Cargo.toml
@@ -22,7 +22,6 @@ wasm32-sdk = []
 [dependencies]
 blake2b_simd.workspace = true
 borsh.workspace = true
-bs58.workspace = true
 cfg-if.workspace = true
 faster-hex.workspace = true
 hex.workspace = true
@@ -47,6 +46,7 @@ smallvec.workspace = true
 thiserror.workspace = true
 wasm-bindgen.workspace = true
 workflow-wasm.workspace = true
+js-sys.workspace = true
 
 [dev-dependencies]
 criterion.workspace = true
--- a/crypto/txscript/src/viewer.rs
+++ b/crypto/txscript/src/viewer.rs
@@ -1,14 +1,23 @@
 use crate::{TxScriptError, multi_sig::get_schnorr_multisig_params, opcodes::codes, parse_script};
+use kaspa_addresses::{Address, Prefix, Version};
 use kaspa_consensus_core::{hashing::sighash::SigHashReusedValues, tx::VerifiableTransaction};
 use std::{
     fmt::{Display, Formatter},
     marker::PhantomData,
 };
 
-#[derive(Debug, Clone, Default)]
+#[derive(Debug, Clone)]
 pub struct ScriptViewerOptions {
     /// if true, viewer tries to dissassemble a sub-script
     pub contains_redeem_script: bool,
+    /// prefix used when formatting Kaspa addresses
+    pub address_prefix: Prefix,
+}
+
+impl Default for ScriptViewerOptions {
+    fn default() -> Self {
+        Self { contains_redeem_script: false, address_prefix: Prefix::Mainnet }
+    }
 }
 
 pub struct ScriptViewer<'a, T, Reused> {
@@ -58,7 +67,10 @@ where
 
                 if self.options.contains_redeem_script {
                     // Try to disassemble the pushed data as a redeem script and keep it visually tied to the hex blob.
-                    let sub_viewer = ScriptViewer::<T, Reused>::new(data, ScriptViewerOptions { contains_redeem_script: false });
+                    let sub_viewer = ScriptViewer::<T, Reused>::new(
+                        data,
+                        ScriptViewerOptions { contains_redeem_script: false, address_prefix: self.options.address_prefix },
+                    );
                     if let Ok(sub_disassembly) = sub_viewer.try_to_string()
                         && sub_disassembly.contains("Op")
                     {
@@ -75,7 +87,7 @@ where
                         s.push(')');
                     }
                 }
-            } else if value == codes::OpCheckMultiSig || value == codes::OpCheckMultiSigVerify {
+            } else if value == codes::OpCheckMultiSig {
                 s.push_str(&opcode_display);
                 let multisig_parameters = get_schnorr_multisig_params(&opcodes, i)?;
                 s.push_str(&format!(
@@ -84,7 +96,8 @@ where
                 ));
 
                 for pubkey in multisig_parameters.signers_pubkey.iter() {
-                    s.push_str(&format!("\n// {}", bs58::encode(pubkey.serialize()).into_string()));
+                    let address = Address::new(self.options.address_prefix, Version::PubKey, pubkey.serialize().as_slice());
+                    s.push_str(&format!("\n// {address}"));
                 }
             } else {
                 s.push_str(&opcode_display);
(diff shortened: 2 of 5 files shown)


crypto/txscript/src/opcodes/macros.rs
```diff
@@ -83,11 +83,17 @@ macro_rules! opcode_impl {
...
+            fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+                write!(f, $str_rep)
+            }
+        }
     }
 }
 
 macro_rules! opcode_list {
-    ( $( opcode $(|$alias:ident|)? $name:ident<$num:literal, $length:tt>($self:ident, $vm:ident) $code: expr ) *)  => {
+    ( $( opcode $(|$alias:ident|)? $name:ident<$str_rep:literal, $num:literal, $length:tt>($self:ident, $vm:ident) $code: expr ) *)  => {
```
biryukovmaxim: i think $name:ident can be used instead on str_rep directly and it will be consistent with [enum](https://github.com/kaspanet/rusty-kaspa/blob/7d93df0d6c87f81ed3af1fae0211c4995b2c4033/crypto/txscript/src/wasm/opcodes.rs#L7)

crypto/txscript/src/multi_sig.rs
```diff
@@ -0,0 +1,63 @@
...
+use kaspa_txscript_errors::TxScriptError;
+
+use crate::opcodes::OpCodeImplementation;
+
+#[derive(Debug)]
+pub struct MultiSigScriptParameters {
+    pub required_signatures_count: u8,
+    pub signers_count: u8,
+    pub signers_pubkey: Vec<secp256k1::XOnlyPublicKey>,
+}
```
biryukovmaxim, line 12: that is only correct for schnorr based multisigs. so either add enum or suffix to make it cleat that it can only work with schnorr based mulisigs

crypto/txscript/src/viewer.rs
```diff
@@ -0,0 +1,92 @@
...
+                s.push(' ');
+                s.push_str(&hex::encode(data));
+            } else if value == codes::OpPushData1 || value == codes::OpPushData2 || value == codes::OpPushData4 {
+                let data = opcode.get_data();
+                s.push(' ');
+                s.push_str(&data.len().to_string());
+                s.push(' ');
+                s.push_str(&hex::encode(data));
+
+                // try to disassemble the data as a script
```
biryukovmaxim: maybe it needs explicit param to try disassemble. also output is not clearly specifies that redeem script and last push data are the same. maybe it makes sense to wrap it into `: (redeem script)` or something like that

crypto/txscript/Cargo.toml
```diff
@@ -12,14 +12,19 @@ repository.workspace = true
...
 [features]
 wasm32-core = []
 wasm32-sdk = []
 
 [dependencies]
 blake2b_simd.workspace = true
 borsh.workspace = true
+bs58.workspace = true
 cfg-if.workspace = true
+hex.workspace = true
```
biryukovmaxim, line 27: use faster hex. i think we should use the library everywhere instead of mixing them

biryukovmaxim (changes requested):

crypto/txscript/src/viewer.rs
IzioDev: > maybe it makes sense to wrap it into `: (redeem script)`

Could you please describe a bit more what you mean by that?
I'm not sure to get it properly. Thanks :)

crypto/txscript/src/opcodes/macros.rs
IzioDev: Ok, let's try this 👍

IzioDev: I'm not familiar with macros, but compiler complains on introduced `$name:ident` with error `duplicate matcher binding`

crypto/txscript/src/multi_sig.rs
IzioDev: in [review: make it clear it only support schnorr multiscript](https://github.com/kaspanet/rusty-kaspa/pull/743/commits/fa1603be4b25115416f0f12dfdd097f51bd521fb), precised naming to restraint to schnorr.

It seems to be standard, as the opcode `OpCheckMultiSig` is for Schnorr, while this one: `OpCheckMultiSigECDSA` is specifically for ECDSA

crypto/txscript/Cargo.toml
IzioDev: addressed in [review: replace hex with faster hex for fresh code](https://github.com/kaspanet/rusty-kaspa/pull/743/commits/4a0a6c26faf7be6bd9a5162b5e7fcc4af0cb52f0)

crypto/txscript/src/opcodes/macros.rs
biryukovmaxim: name:ident already there. You can reuse it

IzioDev:
As per requested by @someone235 and @biryukovmaxim, added a parameter to enable sub script disassembling. If false, it will not attempt disassembling redeem script: [review: add parameters to enable subscript disassembling](https://github.com/kaspanet/rusty-kaspa/pull/743/commits/e23d8acfdd83ea92991381ce405d4b957037df92)

crypto/txscript/src/opcodes/macros.rs
IzioDev: Yes, I tried re-using this "variable" (not sure the proper term for macros)

Image

crypto/txscript/src/viewer.rs
biryukovmaxim: I meant formatting.currently it looks like
```
OpCode, OpCodeData data
Redeem script 
```

There's no direct link between data and redeem script. I suggest adding some extra delimiters to make it look like `OpData hex: (redeem script)`

IzioDev: Makes sense, on it

crypto/txscript/src/opcodes/macros.rs
biryukovmaxim: No
You are trying to introduce new one with the same name. Reusing it means remove the newly added variable and use the first "name"

IzioDev:
Converted to draft in the meantime of a cleaner implementation. (cf. review by Maxim)

IzioDev:
retrospectively, we probably should be displaying address directly instead of some other encoding on multisig

PR #744, opened 2025-10-24
feat(header): VRLE encoding for parents_by_level (WIP)

AxiomePro:
**Related commits**

* `feat(header): VRLE encoding for parents_by_level`
* `chore(header): remove VRLE debug logs`

## Summary

This PR introduces **Vector Run-Length Encoding (VRLE)** for the `parents_by_level` field in `Header`.
VRLE collapses consecutive identical vectors into `(cumulative_count, vector)` pairs, reducing serialized size and network bandwidth while preserving JSON readability (human-readable serializers remain unchanged).

> ⚠️ **Status:** Work in progress.
> Before merging into main, we still need a **DB upgrade strategy** to avoid deserialization errors on existing node databases (pre-VRLE).

---

## Context & Objective

The new VRLE format changes how `parents_by_level` is serialized/stored.
Legacy databases (written before VRLE) will fail to deserialize if read with the new scheme. We need a **safe, backward-compatible** upgrade path that:

* Reads legacy data without errors
* Upgrades to VRLE over time (or in one step)
* Avoids duplicated/fragile read paths
* Enables a **progressive, non-blocking** migration

---

## Chosen Upgrade Strategy (WIP)

**Option (ii): dual DB prefixes** — *recommended*

* New writes go to a **new store prefix**: `headers_vrle:`
* Reads try **VRLE first**, then **fallback** to the legacy prefix `headers:`
* For legacy reads, we’ll **deserialize using a local `LegacyHeader` struct** (no VRLE), then convert to the in-memory `Header`
* (Optional) Opportunistic rewrite: after successfully reading legacy, re-encode and write under `headers_vrle:` to migrate naturally

> This approach avoids startup pauses, is explicit and observable (no “try-and-catch” masking of corruption), and allows a later sweep to fully remove legacy.

---

## What’s in this PR

* `Header` now uses custom Serde for binary formats:

  * **Serialize:** scan once, group consecutive identical vectors, emit `(cumulative_count, vec)` pairs
  * **Deserialize:** validate monotonic cumulative counts and expand pairs back to the original sequence
  * **Human-readable serializers (e.g., JSON):** unchanged (emit/accept plain nested arrays)
* Removed temporary VRLE debug logs

> Note: The **dual-prefix store logic** and **`LegacyHeader`** type will be added next (separate commit/PR if preferred), per the agreed plan.

---

## Why VRLE

* Reduces header size on the wire (Michael expects **much larger savings on mainnet**, up to ~90% in some cases)
* Cuts bandwidth and “gas-equivalent” costs tied to payload size
* Keeps JSON debugging readable (no VRLE applied when `is_human_readable()`)

---

## Remaining Work (before merge)

* Implement **dual-prefix store** (`headers_vrle:` + `headers:`)

  * Add `LegacyHeader` (old serde, no VRLE) local to the store module
  * Read path: try VRLE → fallback legacy (convert to current `Header`)
  * (Optional) Opportunistic rewrite of legacy entries into `headers_vrle:`
* Add integration tests:

  * Legacy bytes → load via fallback → ok
  * New VRLE bytes → load → ok
  * Both keys present → prefer VRLE
  * (Optional) Opportunistic rewrite creates VRLE key after legacy read

### DIFF
--- a/consensus/core/src/header.rs
+++ b/consensus/core/src/header.rs
@@ -152,6 +152,9 @@ pub mod parents_by_level_format {
             }
         }
 
+        cumulative = cumulative.checked_add(current_len).ok_or_else(|| S::Error::custom("cumulative length overflow"))?;
+        runs.push(Run { cumulative, vec: current_vec });
+
         let mut seq = serializer.serialize_seq(Some(runs.len()))?;
         for run in &runs {
             let elem = (run.cumulative, run.vec);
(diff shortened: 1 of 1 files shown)


michaelsutton:
Integrated via #765

PR #749, opened 2025-11-01
lints and style

freshair18:
* Fixes recent rust version lints - namely useless vec! and loop variable "improper" indexing.

* improves general style at the proximity of one of the warnings

### DIFF
--- a/rpc/grpc/core/Cargo.toml
+++ b/rpc/grpc/core/Cargo.toml
@@ -34,10 +34,12 @@ tokio-stream.workspace = true
 tonic = { workspace = true, features = ["tls", "gzip"] }
 triggered.workspace = true
 workflow-core.workspace = true
-itertools.workspace = true
 
 [build-dependencies]
 tonic-build = { workspace = true, features = ["prost"] }
+
+[dev-dependencies]
+itertools.workspace = true
 
 [lints]
 workspace = true


rpc/grpc/core/Cargo.toml
```diff
@@ -34,6 +34,7 @@ tokio-stream.workspace = true
 tonic = { workspace = true, features = ["tls", "gzip"] }
 triggered.workspace = true
 workflow-core.workspace = true
+itertools.workspace = true
```
someone235: Should be under `dev-dependencies`

rpc/grpc/core/src/convert/header.rs
```diff
@@ -102,49 +103,49 @@ mod tests {
...
-        }
+    fn test_parents_by_level_rxr(rpc_parents_1: &[Vec<RpcHash>], rpc_parents_2: &[Vec<RpcHash>]) {
+        assert_eq!(rpc_parents_1, rpc_parents_2);
     }
-    fn test_parents_by_level_rxp(r: &[Vec<RpcHash>], p: &[protowire::RpcBlockLevelParents]) {
-        for i in 0..r.len() {
-            for j in 0..r[i].len() {
-                assert_eq!(r[i][j].to_string(), p[i].parent_hashes[j]);
+    fn test_parents_by_level_rxp(rpc_parents: &[Vec<RpcHash>], proto_parents: &[protowire::RpcBlockLevelParents]) {
+        assert_eq!(rpc_parents.len(), proto_parents.len(), "Number of parent levels differ");
```
someone235: Redundant if `zip_eq` is used

PR #751, opened 2025-11-06
Add support for consensus parameters override

someone235:
The goal of this PR is to support changing consensus constants for test purposes, without editing the code and recompiling.

### DIFF
--- a/README.md
+++ b/README.md
@@ -487,5 +487,10 @@ Logging in `kaspad` and `simpa` can be [filtered](https://docs.rs/env_logger/0.1
 
 </details>
 
-## Override consensus parameters
-You can experiment with non-standard consensus parameters by supplying a JSON file with `--override-params-file <path>`. See [docs/override-params.md](docs/override-params.md) for a more detailed explanation.
+<details>
+
+<summary>Override consensus parameters</summary>
+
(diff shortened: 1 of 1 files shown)


simpa/src/main.rs
```diff
@@ -218,6 +221,13 @@ fn main_impl(mut args: Args) {
     if let Some(rocksdb_mem_budget) = args.rocksdb_mem_budget {
         conn_builder = conn_builder.with_mem_budget(rocksdb_mem_budget);
     }
+
+    if let Some(output_path) = args.override_params_output {
```
coderofstuff, line 225: It's worth documenting this usage in `override-params.md` - that you can construct a starter JSON through simpa

README.md
```diff
@@ -485,4 +485,7 @@ Logging in `kaspad` and `simpa` can be [filtered](https://docs.rs/env_logger/0.1
     ```
     In this command we set the `loglevel` to `INFO`.
 
-</details>
\ No newline at end of file
+</details>
+
+## Override consensus parameters
```
coderofstuff: Can you add this documentation as a section below `Logging`?

Screenshot 2025-11-10 at 9 10 09 AM

Also add update the text slightly to: "You can experiment with non-standard consensus parameters in **non-mainnet environments** by supplying..." (I know that's mentioned in the override-params.md, but I think it's also worth mentioning here.)

simpa/src/main.rs
someone235: It's already there, in the last section

coderofstuff (approved):

PR #753, opened 2025-11-12
Pruning: add batched commits with metrics instrumentation

elldeeone:
## Summary

  - Inspired by Michael Sutton’s pruning discussion (June-2025 workshop), this implements reach-only pruning batching in consensus/src/pipeline/pruning_processor/processor.rs. Batch flush on 256 blocks,
    50,000 ops, 4 MB, or 50 ms; lock yields target 5 ms.
  - Block bodies/UTXO deletes are per-block; batching applies only to reachability/headers/relations.
  - Every prune pass emits [PRUNING METRICS] with batch-write stats (count, avg/max ops/bytes/latency).
  - Benchmarking on local NVMe SSD + VM HDD shows this config best overall across fast + slow disks. Full run logs + summaries below.

  ## Context & Motivation

  - Per-block commits keep the pruning lock hot and thrash RocksDB. Batching improves throughput; reach-only batching avoids large/variable body commits.

  ## Field Data

  - I started with 30+ simpa runs at 10BPS, spanning lock targets + batch thresholds (and body batching on/off) to map the performance curve.
  - From that set I narrowed to the top three reach-only configs: reach-maxblocks15, reach-maxbytes1m, reachlock5, plus unbatched as control.
  - I then ran mainnet + testnet on both local NVMe SSD and a VM HDD. reachlock5 was best overall across fast + slow disks; reach-maxblocks15 was fastest on local NVMe only.

  ## Performance (reachlock5 vs unbatched)

  - local mainnet: 6.80 min vs 13.54 min (1.99x)
  - local testnet: 9.82 min vs 53.82 min (5.48x)
  - VM HDD mainnet: 13.44 min vs 16.50 min (1.23x)
  - VM HDD testnet: 52.60 min vs 207.38 min (3.94x)

  ## Run Folders

  - https://github.com/elldeeone/rusty-kaspa/tree/prune-batching-tests/pruning-optimisation/baseline/experiments
  - https://docs.google.com/spreadsheets/d/1qEm37H-CXnHNT839mmOAx6avnbsjXW5--3lVAHkPveo/edit?usp=sharing

  ## Update (2026-01-21)

  - Replaced earlier 25 ms lock + body batching with reach-only batching + 5 ms lock target
  - Removed per-commit/lock metrics; keep batch-write stats only
  - New local SSD + VM HDD runs attached above

### DIFF
--- a/consensus/src/pipeline/pruning_processor/processor.rs
+++ b/consensus/src/pipeline/pruning_processor/processor.rs
@@ -1,6 +1,6 @@
 //! TODO: module comment about locking safety and consistency of various pruning stores
 
-use super::batching::{PruneBatch, PruningPhaseMetrics, PRUNE_LOCK_TARGET_MAX_DURATION_MS};
+use super::batching::{PRUNE_LOCK_TARGET_MAX_DURATION_MS, PruneBatch, PruningPhaseMetrics};
 use crate::{
     consensus::{
         services::{ConsensusServices, DbParentsManager, DbPruningPointManager},
@@ -38,14 +38,14 @@ use kaspa_consensus_core::{
 };
 use kaspa_consensusmanager::SessionLock;
 use kaspa_core::{debug, info, trace, warn};
-use kaspa_database::prelude::{BatchDbWriter, MemoryWriter, StoreResultExt, DB};
+use kaspa_database::prelude::{BatchDbWriter, DB, MemoryWriter, StoreResultExt};
 use kaspa_hashes::Hash;
 use kaspa_muhash::MuHash;
 use kaspa_utils::iter::IterExtensions;
 use parking_lot::RwLockUpgradableReadGuard;
 use rocksdb::WriteBatch;
 use std::{
-    collections::{hash_map::Entry::Vacant, VecDeque},
+    collections::{VecDeque, hash_map::Entry::Vacant},
     ops::Deref,
     sync::{
         Arc,
--- a/simpa/src/main.rs
+++ b/simpa/src/main.rs
@@ -318,7 +318,6 @@ fn main_impl(mut args: Args) {
                     });
                 });
             }
-
         }
 
         drop(consensus);


consensus/src/pipeline/pruning_processor/processor.rs
```diff
@@ -58,6 +59,160 @@ pub enum PruningProcessingMessage {
...
+        self.batch.len() == 0
+    }
+
+    fn take(&mut self) -> WriteBatch {
+        self.block_count = 0;
+        self.started = None;
+        mem::take(&mut self.batch)
+    }
+}
+
```
freshair18: I do support giving off these statistics, but these classes should be in a different file

consensus/src/pipeline/pruning_processor/processor.rs
```diff
@@ -409,7 +574,13 @@ impl PruningProcessor {
...
                 .iter()
                 .copied()
-                .filter(|&h| !reachability_read.is_dag_ancestor_of_result(new_pruning_point, h).unwrap())
+                .filter(|&h| {
+                    !reachability_read
+                        .as_ref()
+                        .expect("reachability guard should be available")
+                        .is_dag_ancestor_of_result(new_pruning_point, h)
+                        .unwrap()
+                })
```
freshair18: could you explain what is the point of this change?

consensus/src/pipeline/pruning_processor/processor.rs
```diff
@@ -442,126 +617,180 @@ impl PruningProcessor {
 
         // Now we traverse the anti-future of the new pruning point starting from origin and going up.
         // The most efficient way to traverse the entire DAG from the bottom-up is via the reachability tree
-        let mut queue = VecDeque::<Hash>::from_iter(reachability_read.get_children(ORIGIN).unwrap().iter().copied());
+        let mut queue = VecDeque::<Hash>::from_iter(
+            reachability_read.as_ref().expect("reachability guard should be available").get_children(ORIGIN).unwrap().iter().copied(),
+        );
```
freshair18: Same comment is above on why this is an option

rpc/grpc/core/src/convert/message.rs
```diff
@@ -31,7 +31,7 @@ use kaspa_utils::hex::*;
 use std::{str::FromStr, sync::Arc};
 
 macro_rules! from {
-    // Response capture
```
freshair18, line 34: I'm not sure how I feel about all these comment changes here and below and above, they require a big context switch from me to review and I don't understand how you got to them anyways.

To what extent are you sure about their veracity?

Anyway, I think you can submit them in a separate cosmetic PR.

consensus/src/pipeline/pruning_processor/processor.rs
```diff
@@ -442,126 +617,180 @@ impl PruningProcessor {
...
-                        // other parts of the code assume the existence of GD data etc.)
-                        statuses_write.set_batch(&mut batch, current, StatusHeaderOnly).unwrap();
+        'staging: loop {
+            // Create staging stores once per batch to maintain consistency across multiple block deletions
+            let mut relations_write = self.relations_store.write();
+            let mut reachability_relations_write = self.reachability_relations_store.write();
+            let mut staging_relations = StagingRelationsStore::new(&mut relations_write);
+            let mut staging_reachability_relations = StagingRelationsStore::new(&mut reachability_relations_write);
+            let mut staging_reachability =
+                StagingReachabilityStore::new(reachability_read.take().expect("reachability guard should be available"));
```
freshair18: same comment

consensus/src/pipeline/pruning_processor/processor.rs
```diff
@@ -574,18 +803,48 @@ impl PruningProcessor {
...
             drop(pruning_point_write);
         }
+
+        self.flush_prune_batch(&mut prune_batch, &mut metrics);
+
+        metrics.set_traversed(traversed, counter);
+        metrics.log_summary();
+    }
+
+    fn should_flush_prune_batch(&self, batch: &PruneBatch, lock_elapsed: Duration) -> bool {
```
freshair18: Similar to comment above, I think these functions to should exist in a separate file

consensus/src/pipeline/pruning_processor/processor.rs
```diff
@@ -442,126 +617,180 @@ impl PruningProcessor {
...
-                        self.pruning_samples_store.delete_batch(&mut batch, current).unwrap();
+                    prune_guard.blocking_yield();
+                    lock_acquire_time = Instant::now();
+                    queue.push_front(current);
+                    reachability_read = Some(self.reachability_store.upgradable_read());
+                    metrics.record_lock_reacquire();
+                    continue 'staging;
+                }
+
+                let skip_due_to_retention = match staging_reachability.is_dag_ancestor_of_result(retention_period_root, current) {
```
freshair18: variable name is peculiar.

 The reason you "skip" this is simply because you retain anything in the future of the the retention block (the pruning point or a few pruning points back if it was set so). So I think a more suitable name would be "is_block_in_retention_root_future" or the contra positive "is_block_in_retention_root_antifuture"

consensus/src/pipeline/pruning_processor/processor.rs
```diff
@@ -442,126 +617,180 @@ impl PruningProcessor {
...
+                    Ok(result) => result,
+                    Err(err) if err.is_key_not_found() => {
+                        // A keyed block might already be staged for deletion in the current batch.
+                        // The underlying store still contains it until the batch is flushed, so consult
+                        // a fresh read guard to answer the reachability query.
+                        let reachability_read_only = self.reachability_store.read();
+                        reachability_read_only.is_dag_ancestor_of_result(retention_period_root, current).unwrap()
```
freshair18: I don't get this code, why wont we check on reachability read directly? If we still do things in a manner that makes sense it shouldn't be harmed

consensus/src/pipeline/pruning_processor/processor.rs
```diff
@@ -442,126 +617,180 @@ impl PruningProcessor {
...
-                self.block_transactions_store.delete_batch(&mut batch, current).unwrap();
-
-                if let Some(&affiliated_proof_level) = keep_relations.get(&current) {
-                    if statuses_write.get(current).optional().unwrap().is_some_and(|s| s.is_valid()) {
-                        // We set the status to header-only only if it was previously set to a valid
-                        // status. This is important since some proof headers might not have their status set
-                        // and we would like to preserve this semantic (having a valid status implies that
-                        // other parts of the code assume the existence of GD data etc.)
-                        statuses_write.set_batch(&mut batch, current, StatusHeaderOnly).unwrap();
+        'staging: loop {
```
freshair18: I don't like this name. Staging is not the defining characteristic of this loop.

consensus/src/pipeline/pruning_processor/processor.rs
```diff
@@ -442,126 +617,180 @@ impl PruningProcessor {
...
+            drop(reachability_write);
+            drop(statuses_write);
+            drop(reachability_relations_write);
+            drop(relations_write);
+            break;
         }
 
-        drop(reachability_read);
+        metrics.finalize_lock_hold(lock_acquire_time.elapsed());
+        if let Some(guard) = reachability_read.take() {
```
freshair18: what is the logic behind changing this?

consensus/src/pipeline/pruning_processor/processor.rs
```diff
@@ -574,18 +803,48 @@ impl PruningProcessor {
         );
 
         if self.config.enable_sanity_checks {
+            // Ensure the pruning changes are persisted before rebuilding proofs from storage.
+            self.flush_prune_batch(&mut prune_batch, &mut metrics);
```
freshair18: this is not related to sanity, you simply should have flushed the batch one last time after you exit the loop.

The bigger problem here, is you already dropped the prune_guard, meaning the deletions are not "safe". this is cause for much trouble.

consensus/src/pipeline/pruning_processor/processor.rs
```diff
@@ -574,18 +803,48 @@ impl PruningProcessor {
...
             // Set the retention checkpoint to the new retention root only after we successfully pruned its past
             let mut pruning_point_write = self.pruning_point_store.write();
-            let mut batch = WriteBatch::default();
-            pruning_point_write.set_retention_checkpoint(&mut batch, retention_period_root).unwrap();
-            self.db.write(batch).unwrap();
+            pruning_point_write.set_retention_checkpoint(&mut prune_batch.batch, retention_period_root).unwrap();
             drop(pruning_point_write);
         }
+
+        self.flush_prune_batch(&mut prune_batch, &mut metrics);
```
freshair18: Here you finally flush the prune batch post the loop, but as mentioned, you are not protected by the prune_guard.

consensus/src/pipeline/pruning_processor/processor.rs
```diff
@@ -58,6 +59,160 @@ pub enum PruningProcessingMessage {
...
+        self.lock_yield_count += 1;
+    }
+
+    fn finalize_lock_hold(&mut self, held: Duration) {
+        self.total_lock_hold += held;
+    }
+
+    fn record_lock_reacquire(&mut self) {
+        self.lock_reacquire_count += 1;
+    }
```
freshair18: I think the lock datas is TMI though

consensus/src/pipeline/pruning_processor/processor.rs
```diff
@@ -352,11 +510,16 @@ impl PruningProcessor {
                     self.ghostdag_store.update_batch(&mut batch, kept, &Arc::new(mutable_ghostdag.into())).unwrap();
                 }
             }
+            let ops = batch.len();
+            let bytes = batch.size_in_bytes();
+            let commit_start = Instant::now();
             self.db.write(batch).unwrap();
+            metrics.record_commit("ghostdag_adjust", ops, bytes, commit_start.elapsed());
```
freshair18: I think the only thing we should collect statistics on is the batch writes.

consensus/src/pipeline/pruning_processor/processor.rs
```diff
@@ -392,8 +555,10 @@ impl PruningProcessor {
         }
 
         prune_guard = self.pruning_lock.blocking_write();
-        let mut lock_acquire_time = Instant::now();
-        let mut reachability_read = self.reachability_store.upgradable_read();
+        lock_acquire_time = Instant::now();
+        metrics.record_lock_reacquire();
+        let mut reachability_read = Some(self.reachability_store.upgradable_read());
```
freshair18: why did we make it an option?

consensus/src/pipeline/pruning_processor/processor.rs
```diff
@@ -433,7 +604,11 @@ impl PruningProcessor {
             }
 
             // Flush the batch to the DB
+            let ops = batch.len();
+            let bytes = batch.size_in_bytes();
+            let commit_start = Instant::now();
             self.db.write(batch).unwrap();
+            metrics.record_commit("tips_and_selected_chain", ops, bytes, commit_start.elapsed());
```
freshair18: same comment, we don't actually care for this statistic.

consensus/src/pipeline/pruning_processor/processor.rs
elldeeone: Agree. These helper structs do not belong in this file. I will move CommitStats, PruningPhaseMetrics, PruneBatch, and the consts into a dedicated module to keep processor.rs focused.

consensus/src/pipeline/pruning_processor/processor.rs
elldeeone: The change is fallout of making reachability_read movable into staging. I will refactor to avoid the Option indirection and keep direct reads here. Refactored to keep reachability guard direct; now take a fresh upgradable read per batch iteration and drop the read-only guard before staging commit to avoid the lock hold. Behavior unchanged; this fixes a deadlock we saw in pruning tests.

consensus/src/pipeline/pruning_processor/processor.rs
elldeeone: Same as above; I will refactor so reachability_read stays non-optional. Refactored to keep reachability guard direct; now take a fresh upgradable read per batch iteration and drop the read-only guard before staging commit to avoid the lock hold. Behavior unchanged; this fixes a deadlock we saw in pruning tests.

rpc/grpc/core/src/convert/message.rs
elldeeone: Agree; cosmetic comment change should be separate. I will revert this comment change in this PR.

consensus/src/pipeline/pruning_processor/processor.rs
elldeeone: Same as above; I will remove the Option/take pattern and pass a scoped guard. Refactored to keep reachability guard direct; now take a fresh upgradable read per batch iteration and drop the read-only guard before staging commit to avoid the lock hold. Behavior unchanged; this fixes a deadlock we saw in pruning tests.

consensus/src/pipeline/pruning_processor/processor.rs
elldeeone: Agree; batching helpers should live in a separate module. I will move should_flush_prune_batch and flush_prune_batch alongside the batch structs.

consensus/src/pipeline/pruning_processor/processor.rs
elldeeone: Agree; name is misleading. I will rename to is_block_in_retention_root_future (or ..._antifuture if you prefer).

consensus/src/pipeline/pruning_processor/processor.rs
elldeeone: Agree; once the Option is removed, I will check on the reachability guard directly. I will keep the fallback only if it is still needed for the staged-deletion edge case. Refactored to keep reachability guard direct; now take a fresh upgradable read per batch iteration and drop the read-only guard before staging commit to avoid the lock hold. Behavior unchanged; this fixes a deadlock we saw in pruning tests.

consensus/src/pipeline/pruning_processor/processor.rs
elldeeone: Agree; staging is not the defining characteristic. I will rename the label (e.g. prune_batch or prune_loop).

consensus/src/pipeline/pruning_processor/processor.rs
elldeeone: This is only here because of the Option. I will remove it with the refactor. Refactored to keep reachability guard direct; now take a fresh upgradable read per batch iteration and drop the read-only guard before staging commit to avoid the lock hold. Behavior unchanged; this fixes a deadlock we saw in pruning tests.

consensus/src/pipeline/pruning_processor/processor.rs
elldeeone: Agree; flush should not be tied to sanity checks and must happen before releasing prune_guard. I will move the final flush under the guard.

consensus/src/pipeline/pruning_processor/processor.rs
elldeeone: Agree; I will ensure the flush happens while prune_guard is held (along with the retention checkpoint).

consensus/src/pipeline/pruning_processor/processor.rs
elldeeone: Agree; lock stats are too noisy. I will drop lock metrics and keep only batch-write stats.

consensus/src/pipeline/pruning_processor/processor.rs
elldeeone: Agree; I will remove per-commit metrics and keep only batch-write stats.

consensus/src/pipeline/pruning_processor/processor.rs
elldeeone: Agree; I will refactor to avoid the Option for reachability read. Refactored to keep reachability guard direct; now take a fresh upgradable read per batch iteration and drop the read-only guard before staging commit to avoid the lock hold. Behavior unchanged; this fixes a deadlock we saw in pruning tests.

consensus/src/pipeline/pruning_processor/processor.rs
elldeeone: Agree; I will drop this per-commit metric and keep batch-write stats only.

elldeeone:
@freshair18

Changes:
- Move prune batching structs/helpers into pruning_processor/batching.rs; processor.rs slimmed.
- Reachability guard refactor: no Option/take; per-iteration upgradable read, drop read-only before staging commit → fixes pruning-test deadlock; pruning behavior unchanged.
- Final flush/retention checkpoint under prune_guard; lock metrics + per-commit stats dropped; comment revert in grpc convert.

Tests:
- cargo fmt --all
- cargo clippy --workspace --tests --benches --examples -- -D warnings
- cargo test --release -p kaspa-testing-integration --lib -- goref_custom_pruning_depth_test --nocapture
- cargo test --release -p kaspa-testing-integration --lib -- integration_tests::pruning_test --nocapture
- ./test (full suite incl wasm) with LLVM toolchain: PATH=/opt/homebrew/opt/llvm/bin CC=…/clang AR=…/llvm-ar RANLIB=…/llvm-ranlib

consensus/src/pipeline/pruning_processor/processor.rs
```diff
@@ -443,125 +446,167 @@ impl PruningProcessor {
...
+                    // An exit signal was received. Exit from this long running process.
+                    if self.is_consensus_exiting.load(Ordering::Relaxed) {
+                        drop(prune_guard);
+                        info!("Header and Block pruning interrupted: Process is exiting");
+                        return;
+                    }
+                    prune_guard.blocking_yield();
+                    lock_acquire_time = Instant::now();
+                    queue.push_front(current);
+                    continue 'prune_batch;
```
freshair18: This structure of repushing to queue and continuing another iteration is not unheard of, but is a bit harder to follow.  Come to think, I'm not sure we really need an outer loop at all.  We can reacquire the write locks here.

consensus/src/pipeline/pruning_processor/processor.rs
```diff
@@ -443,125 +446,167 @@ impl PruningProcessor {
...
-                lock_acquire_time = Instant::now();
-                reachability_read = self.reachability_store.upgradable_read();
-            }
 
-            if traversed % 1000 == 0 {
-                info!("Header and Block pruning: traversed: {}, pruned {}...", traversed, counter);
-            }
+                let is_block_in_retention_root_future = {
+                    let reachability_read_only = self.reachability_store.read();
+                    reachability_read_only.is_dag_ancestor_of_result(retention_period_root, current).unwrap()
```
freshair18: remove only

elldeeone:
@freshair18 addressed both items. Traversal loop now avoids requeue/outer loop: we peek `queue.front()`, commit/flush + yield when the lock budget elapses, then re-acquire locks in place and continue. Also removed the extra read-only reachability guard; retention-future check now uses `staging_reachability.is_dag_ancestor_of_result(...)`.

consensus/src/pipeline/pruning_processor/processor.rs
```diff
@@ -443,19 +446,34 @@ impl PruningProcessor {
...
+                // Commit staging stores and flush the batch so we can yield
+                let reachability_write = staging_reachability.commit(&mut prune_batch.batch).unwrap();
+                staging_reachability_relations.commit(&mut prune_batch.batch).unwrap();
+                staging_relations.commit(&mut prune_batch.batch).unwrap();
+                drop(reachability_write);
+                drop(statuses_write);
+                drop(reachability_relations_write);
+                drop(relations_write);
+
+                prune_batch.flush(&self.db, &mut metrics);
```
D-Stacks, line 472: I noticed you measure the time on how long a batch write executes, do you have some info on this? unsure if i am missing something but since you flush after measuring time elapsed, and initiate a new timer afterwards, after blocking yield, i believe the time it takes to flush is added to the prune lock time? which might be negligible.. but should maybe be considered as this adds to the expected `PRUNE_LOCK_MAX_DURATION_MS` .

consensus/src/pipeline/pruning_processor/processor.rs
```diff
@@ -522,46 +550,74 @@ impl PruningProcessor {
...
+                let reachability_write = staging_reachability.commit(&mut prune_batch.batch).unwrap();
+                staging_reachability_relations.commit(&mut prune_batch.batch).unwrap();
+                staging_relations.commit(&mut prune_batch.batch).unwrap();
                 drop(reachability_write);
                 drop(statuses_write);
                 drop(reachability_relations_write);
                 drop(relations_write);
 
-                reachability_read = self.reachability_store.upgradable_read();
+                prune_batch.flush(&self.db, &mut metrics);
```
D-Stacks: same exceeding of the `PRUNE_LOCK_MAX_DURATION_MS`,  as in the first closure applies here btw.. but maybe in a more obscured "worst case" scenario, as there are other triggers besides `PRUNE_LOCK_MAX_DURATION_MS` that can influence the trigger to enter into this closure. But the fact remains that exceeding `PRUNE_LOCK_MAX_DURATION_MS`  is possible here.

consensus/src/pipeline/pruning_processor/processor.rs
```diff
@@ -450,154 +450,165 @@ impl PruningProcessor {
...
-                queue.extend(staging_reachability.get_children(current).unwrap().iter());
+        // Create staging stores once per batch to maintain consistency across multiple block deletions
+        let mut relations_write = self.relations_store.write();
+        let mut reachability_relations_write = self.reachability_relations_store.write();
+        let mut staging_relations = StagingRelationsStore::new(&mut relations_write);
+        let mut staging_reachability_relations = StagingRelationsStore::new(&mut reachability_relations_write);
+        let mut staging_reachability = StagingReachabilityStore::new(self.reachability_store.upgradable_read());
+        let mut statuses_write = self.statuses_store.write();
+
+        loop {
```
D-Stacks: I think this can be re-written more succinctly: `while !queue.is_empty`

consensus/src/pipeline/pruning_processor/processor.rs
```diff
@@ -443,19 +446,34 @@ impl PruningProcessor {
...
+                break;
             }
-            traversed += 1;
-            // Obtain the tree children of `current` and push them to the queue before possibly being deleted below
-            queue.extend(reachability_read.get_children(current).unwrap().iter());
 
-            // If we have the lock for more than a few milliseconds, release and recapture to allow consensus progress during pruning
-            if lock_acquire_time.elapsed() > Duration::from_millis(5) {
-                drop(reachability_read);
+            if lock_acquire_time.elapsed() > Duration::from_millis(PRUNE_LOCK_MAX_DURATION_MS) {
```
D-Stacks: I actually don't really see the point of this closure, at the end of the loop you call the exact same things as in this closure, only with the should_flush condition, should_flush also implicitly does this same check, so i think you can just use should_flush right here? and remove the closure at the end. So, either a) should_flush triggers at the end, you reset the timer, and you will not enter this closure b)  if should_flush didn't trigger,  we don't expect this to trigger (unless in some extreme edge cases, where condition is met in the few nano secs timeframe upon loop restart).

cli/src/modules/rpc.rs
```diff
@@ -82,7 +82,7 @@ impl Rpc {
             //     self.println(&ctx, result);
             // }
             RpcApiOps::GetMempoolEntries => {
-                // TODO
+                // TODO: expose filter flags in CLI args instead of hard-coded defaults.
```
D-Stacks: not sure how this pertains to the current pr, maybe remove this, or pr it separately?

cli/src/modules/rpc.rs
```diff
@@ -326,7 +326,7 @@ impl Rpc {
     }
 
     async fn display_help(self: Arc<Self>, ctx: Arc<KaspaCli>, _argv: Vec<String>) -> Result<()> {
-        // RpcApiOps that do not contain docs are not displayed
+        // Hide ops without docs so help output does not include empty entries.
```
D-Stacks: not sure how this pertains to the current pr, or why it was changed, maybe remove this?

consensus/src/pipeline/pruning_processor/processor.rs
elldeeone: Good catch, I’d overlooked that. Flush latency is currently inside the lock window, so the limit is best‑effort. We could either document it, or subtract a rolling avg commit time from the check to keep holds closer to the bound (at the cost of potentially smaller batches/more churn though). What do you prefer?

consensus/src/pipeline/pruning_processor/processor.rs
elldeeone: Yep same issue as your first comment. Other triggers can push us over; the budget is best‑effort. We can handle it the same way: document it, or bias the check by subtracting avg commit time. Whats your preference?

consensus/src/pipeline/pruning_processor/processor.rs
elldeeone: Np, happy to switch to the simpler loop.

consensus/src/pipeline/pruning_processor/processor.rs
elldeeone: Good point. I was aiming for a pre‑work budget check so we can yield even when the batch is empty (e.g. keep‑only traversal). should_flush returns false on empty, so moving everything there would drop that yield. If you want one path, I can tweak should_flush or add a should_yield helper to cover both while keeping behaviour identical?

cli/src/modules/rpc.rs
elldeeone: Agree off‑scope. I’ll drop it.

cli/src/modules/rpc.rs
elldeeone: Agree off‑scope. I’ll drop it.

consensus/src/pipeline/pruning_processor/processor.rs
D-Stacks: I think you should just go for an or operator something like:

```rs
if prune_batch.should_flush() || lock_acquire_time.elapsed() > Duration::from_millis(PRUNE_LOCK_MAX_DURATION_MS) {
//Do work
}
```

I think then you can also avoid sending the lock_acquire_time into the should_flush condition and we keep the lock_acquire condition separate and explicit.

consensus/src/pipeline/pruning_processor/processor.rs
D-Stacks: I think for now I would just add a comment and document this potential behavior over this line. Would prefer if other reviewers could also assess relevance of this behavior. 

Taking just the average times (that have been recorded) of a few milliseconds, or a fraction thereof on newer hardware, it seems like it shouldn't influence lock time much. But then again it seems like worst case goes to a few hundred milliseconds, which seems relevant. But I am assuming that removing block bodies from the batch might add some consistency to these times. 

Might also be worth renaming `PRUNE_LOCK_MAX_DURATION_MS` to `PRUNE_LOCK_TARGET_MAX_DURATION_MS`, I think current naming suggests a hard upper bound, which kind of is no longer a given.

elldeeone: Added a comment noting the lock budget is best‑effort and renamed to PRUNE_LOCK_TARGET_MAX_DURATION_MS.

consensus/src/pipeline/pruning_processor/processor.rs
elldeeone: Done. Lock budget check is now explicit (should_flush() + lock_elapsed check), and should_flush no longer takes lock_elapsed.

consensus/src/pipeline/pruning_processor/processor.rs
```diff
@@ -443,19 +446,31 @@ impl PruningProcessor {
...
+        let mut relations_write = self.relations_store.write();
+        let mut reachability_relations_write = self.reachability_relations_store.write();
+        let mut staging_relations = StagingRelationsStore::new(&mut relations_write);
+        let mut staging_reachability_relations = StagingRelationsStore::new(&mut reachability_relations_write);
+        let mut staging_reachability = StagingReachabilityStore::new(self.reachability_store.upgradable_read());
+        let mut statuses_write = self.statuses_store.write();
+
+        while !queue.is_empty() {
+            // Lock budget is best-effort because batch flush happens under the prune lock.
+            if lock_acquire_time.elapsed() > Duration::from_millis(PRUNE_LOCK_TARGET_MAX_DURATION_MS) {
```
D-Stacks: So what I meant is now you can move the closure, with its condition, at line 574, here, and remove the whole closure at line 574.

elldeeone: Sorry i misunderstood earlier, thanks for clarifying. I’ve now moved the flush/yield block to the top of the loop and removed the bottom one, so there’s a single flush path with the same condition.

PR #754, opened 2025-11-12
Crescendo Cleanup - Part 2

someone235:
I tried to clean all fork activation logic that have little effect on existing tests. In the next PR I'll clean activation logic that breaks tests.

### DIFF
--- a/components/addressmanager/src/lib.rs
+++ b/components/addressmanager/src/lib.rs
@@ -536,7 +536,8 @@ mod address_store_with_cache {
         }
 
         // This test is indeterminate, so it is ignored by default.
-        // TODO: Maybe run multiple times and make it pass if the majority of runs pass?
+        // Every developer that changes the logic of the address manager should run this test locally before sending a PR.
+        // TODO: Maybe change statistical parameters to reduce the failure rate?
         #[test]
         #[ignore]
         fn test_network_distribution_weighting() {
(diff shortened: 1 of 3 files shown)


components/addressmanager/src/lib.rs
```diff
@@ -535,7 +535,10 @@ mod address_store_with_cache {
             assert_eq!(iter.count(), 0);
         }
 
+        // This test is indeterminate, so it is ignored by default.
+        // TODO: Maybe run multiple times and make it pass if the majority of runs pass?
```
michaelsutton: Not sure the todo makes sense. Since the test is statistical, there's always some chance of failure. The comment should emphasize that any changes in this area require running the test locally

rpc/service/src/service.rs
```diff
@@ -378,14 +378,7 @@ impl RpcApi for RpcCoreService {
...
 
-            // A simple heuristic check which signals that the mined block is out of date
-            // and should not be accepted unless user explicitly requests
-            //
-            // [Crescendo]: switch to the larger duration only after a full window with the new duration is reached post activation
-            let difficulty_window_duration = self
-                .config
-                .difficulty_window_duration_in_block_units()
-                .get(virtual_daa_score.saturating_sub(self.config.difficulty_window_duration_in_block_units().after()));
+            let difficulty_window_duration = self.config.difficulty_window_duration_in_block_units().after();
```
michaelsutton, line 326: maybe keep the first two lines of the comment

protocol/flows/src/flow_context.rs
```diff
@@ -233,7 +233,7 @@ pub struct FlowContextInner {
     block_event_logger: Option<BlockEventLogger>,
 
     // Bps upper bound
```
michaelsutton: update comment

PR #755, opened 2025-11-13
Tor integration

elldeeone:
## Tor + Kaspa
- Tor routes traffic through layered relays so peers never learn each other’s clearnet IPs.
- Applications integrate by sending every socket through Tor’s SOCKS proxy and, when inbound reachability is needed, publishing a hidden service via the control port.
- This branch wires rusty-kaspa into that flow: when I point the daemon at a Tor control listener and SOCKS endpoint, it authenticates, waits for bootstrap, can publish its own `.onion` address, and forces outbound P2P through Tor.
- I mirrored Bitcoin Core’s Tor methodology so operators see familiar semantics, but all of the implementation is Rust/Kaspa-specific.

## Inspiration & Approach
- I intentionally shadowed Bitcoin Core’s integration strategy: system Tor, control-port orchestration, persistent onion keys, per-connection SOCKS auth isolation (IsolateSOCKSAuth), and ADDRv2/BIP155 signaling for onion gossip.
- Terminology like “BIP155” appears because Tor operators already know it; Kaspa previously lacked that service bit, so I added it to advertise onion capability.
- Everything else—FlowContext plumbing, ConnectionManager hooks, gRPC transport, logging—is native to Kaspa; no Bitcoin code was imported.

## Intent & Expectations
- I’m not expecting this PR to merge as-is; my goal is to surface the work and spark architectural discussion.
- Please treat it as exploratory: prioritize feedback on security, ergonomics, and long-term feasibility over polishing.
- If the direction resonates, I’m happy to chase the invasive follow-ups (DNS plumbing, listener defaults, bundled Tor, automated tests) before proposing anything merge-worthy.

## Proof-of-Concept Status
- End-to-end demo works in my lab: `kaspad --tor-control=… --tor-cookie=… --proxy-net=onion=127.0.0.1:9050 --listen-onion --tor-only` routes outbound traffic exclusively through SOCKS and advertises a persistent hidden service.
- It’s still “P(roof)-ish”: everything lives in the daemon, but I haven’t bundled Tor, failure recovery is minimal, and validation has been manual. Treat it like an advanced feature flag for now.
- Tor bandwidth (roughly 2–5 Mbps) makes full sync over onions impractical—especially with pruning. I expect mixed-mode usage (clearnet IBD, Tor steady-state) until we build smarter paths.

## What Landed
1. **Tor runtime & hidden service plumbing** (`kaspad/src/tor_manager.rs`, `kaspad/src/daemon.rs`)
   - I wrapped `tor-interface`, negotiate password/cookie auth, stream bootstrap events, publish/remove hidden services, and persist Ed25519 onion keys under `~/.rusty-kaspa/<net>/tor/p2p_onion.key`.
2. **CLI & config surface** (`kaspad/src/args.rs`, `kaspad/src/daemon.rs`)
   - I added flags for control/cookie paths, tor-only mode, per-network SOCKS overrides, onion listener publication, bootstrap timeouts, and key locations. `ProxySettings::resolve` now emits `SocksProxyParams` with auth metadata and auto-promotes Tor proxies to randomized credentials.
3. **P2P SOCKS integration** (`protocol/p2p/src/core/connection_handler.rs`, `protocol/flows/src/service.rs`)
   - I updated outbound dials to consult `SocksProxyConfig`, pick default/IPv4/IPv6/onion endpoints, and generate per-stream username/password pairs whenever Tor isolation is requested.
4. **AddrV2/onion awareness** (`protocol/flows/src/flow_context.rs`, `protocol/flows/src/v5/address.rs`, `protocol/p2p/proto/p2p.proto`, `utils/src/networking.rs`, `components/addressmanager/*`)
   - I taught version handshakes to advertise BIP155-style ADDRv2 support, ensured onion gossip only happens when both peers run Tor, updated protobuf/storage layers to encode onion hosts, and tightened the address manager’s `--onlynet`/`--tor-only` filtering.
5. **Runtime orchestration** (`kaspad/src/daemon.rs`, `protocol/flows/src/service.rs`)
   - I gate P2P startup on Tor bootstrap via watch channels, handle onion publication/teardown through control-port commands, and automatically drop onion peers when Tor isn’t active.

## Known Gaps / Risks
- **DNS & hostname leakage:** proxy targets still resolve via the OS, and DNS seeders perform clearnet lookups (`kaspad/src/args.rs`, `components/connectionmanager/src/lib.rs`). Full parity requires host-preserving proxy targets plus a name-proxy pipeline.
- **Listener semantics:** clearnet listeners stay up unless the operator opts out; Bitcoin disables them automatically when Tor is active. Matching that behavior would mean changing default `--listen`/`--onlynet` handling.
- **Tor lifecycle robustness:** I currently assume an externally managed Tor daemon—there’s no bundled binary, no retry logic when the control port drops, and no automatic re-publication after Tor restarts.
- **Testing coverage:** outside of some onion-filtering unit tests, validation was manual (logs, `lsof`, Tor control output). There’s no integration harness that spins up a Tor stub yet.
- **Performance realities:** Tor’s bandwidth ceiling makes Tor-only syncing largely impractical today; expect clearnet IBD followed by Tor steady-state for the foreseeable future.

## Call for Discussion
- Is the security/maintenance cost of native Tor support acceptable for upstream Kaspa?
- Should I finish the DNS/name-proxy refactor and align listener defaults before even asking for mergeability?
- How should we document limitations (bandwidth, Tor restarts, lack of bundled daemon) so operators don’t misinterpret the feature?

### DIFF
--- a/components/addressmanager/src/lib.rs
+++ b/components/addressmanager/src/lib.rs
@@ -138,7 +138,7 @@ impl AddressManager {
     fn local_addresses(&self) -> impl Iterator<Item = NetAddress> + '_ {
         match self.config.externalip {
             // An external IP was passed, we will try to bind that if it's valid
-            Some(local_net_address) if local_net_address.as_ip().map_or(false, |ip| ip.is_publicly_routable()) => {
+            Some(local_net_address) if local_net_address.as_ip().is_some_and(|ip| ip.is_publicly_routable()) => {
                 info!("External address is publicly routable {}", local_net_address);
                 return Left(iter::once(local_net_address));
             }
@@ -155,10 +155,10 @@ impl AddressManager {
         // check whatever was passed as listen address (if routable)
         // otherwise(listen_address === 0.0.0.0) check all interfaces
         let listen_address = self.config.p2p_listen_address.normalize(self.config.default_p2p_port());
-        if listen_address.as_ip().map_or(false, |ip| ip.is_publicly_routable()) {
+        if listen_address.as_ip().is_some_and(|ip| ip.is_publicly_routable()) {
             info!("Publicly routable local address found: {}", listen_address);
             Left(Left(iter::once(listen_address)))
-        } else if listen_address.as_ip().map_or(false, |ip| ip.is_unspecified()) {
+        } else if listen_address.as_ip().is_some_and(|ip| ip.is_unspecified()) {
             let network_interfaces = list_afinet_netifas();
             let Ok(network_interfaces) = network_interfaces else {
                 warn!("Error getting network interfaces: {:?}", network_interfaces);
@@ -187,7 +187,7 @@ impl AddressManager {
         info!("[UPnP] Got external ip from gateway using upnp: {ip}");
 
         let normalized_p2p_listen_address = self.config.p2p_listen_address.normalize(self.config.default_p2p_port());
-        let local_addr = if normalized_p2p_listen_address.as_ip().map_or(false, |ip| ip.is_unspecified()) {
+        let local_addr = if normalized_p2p_listen_address.as_ip().is_some_and(|ip| ip.is_unspecified()) {
             SocketAddr::new(local_ip_address::local_ip().unwrap(), normalized_p2p_listen_address.port)
         } else {
             normalized_p2p_listen_address.to_socket_addr().expect("expected listen address to be IP-based")
@@ -279,11 +279,11 @@ impl AddressManager {
             debug!("[Address manager] skipping onion address {} (onion disabled)", address);
             return;
         }
-        if !self.allow_ipv4 && address.as_ip().map_or(false, |ip| ip.is_ipv4()) {
+        if !self.allow_ipv4 && address.as_ip().is_some_and(|ip| ip.is_ipv4()) {
             debug!("[Address manager] skipping IPv4 address {} (ipv4 disabled)", address);
             return;
         }
-        if !self.allow_ipv6 && address.as_ip().map_or(false, |ip| ip.is_ipv6()) {
+        if !self.allow_ipv6 && address.as_ip().is_some_and(|ip| ip.is_ipv6()) {
             debug!("[Address manager] skipping IPv6 address {} (ipv6 disabled)", address);
             return;
         }
@@ -331,7 +331,7 @@ impl AddressManager {
 
     fn prune_ipv4_addresses(&mut self) {
         let to_remove: Vec<_> =
-            self.address_store.iterate_addresses().filter(|addr| addr.as_ip().map_or(false, |ip| ip.is_ipv4())).collect();
+            self.address_store.iterate_addresses().filter(|addr| addr.as_ip().is_some_and(|ip| ip.is_ipv4())).collect();
         for addr in to_remove {
             self.address_store.remove(addr);
         }
@@ -339,7 +339,7 @@ impl AddressManager {
 
     fn prune_ipv6_addresses(&mut self) {
         let to_remove: Vec<_> =
-            self.address_store.iterate_addresses().filter(|addr| addr.as_ip().map_or(false, |ip| ip.is_ipv6())).collect();
+            self.address_store.iterate_addresses().filter(|addr| addr.as_ip().is_some_and(|ip| ip.is_ipv6())).collect();
         for addr in to_remove {
             self.address_store.remove(addr);
         }
--- a/components/connectionmanager/src/lib.rs
+++ b/components/connectionmanager/src/lib.rs
@@ -389,6 +389,6 @@ impl ConnectionManager {
             .lock()
             .await
             .iter()
-            .any(|(address, request)| request.is_permanent && address.as_ip().map_or(false, |addr_ip| IpAddr::from(addr_ip) == ip))
+            .any(|(address, request)| request.is_permanent && address.as_ip().is_some_and(|addr_ip| IpAddr::from(addr_ip) == ip))
     }
 }
--- a/kaspad/src/args.rs
+++ b/kaspad/src/args.rs
@@ -188,19 +188,18 @@ impl Default for Args {
 impl Args {
     pub fn proxy_settings(&self) -> ProxySettings {
         let mut settings = ProxySettings::default();
-        if let Some(default_proxy) = self.proxy.clone() {
+        if let Some(default_proxy) = self.proxy {
             settings.default = Some(ProxyConfigEntry { address: default_proxy, auth: self.proxy_auth(false) });
         }
         for rule in &self.proxy_net {
-            let entry =
-                ProxyConfigEntry { address: rule.address.clone(), auth: self.proxy_auth(matches!(rule.network, ProxyNetwork::Onion)) };
+            let entry = ProxyConfigEntry { address: rule.address, auth: self.proxy_auth(matches!(rule.network, ProxyNetwork::Onion)) };
             match rule.network {
                 ProxyNetwork::Ipv4 => settings.ipv4 = Some(entry),
                 ProxyNetwork::Ipv6 => settings.ipv6 = Some(entry),
                 ProxyNetwork::Onion => settings.onion = Some(entry),
             }
         }
-        if let Some(tor_specific) = self.tor_proxy.clone() {
+        if let Some(tor_specific) = self.tor_proxy {
             settings.onion = Some(ProxyConfigEntry { address: tor_specific, auth: self.proxy_auth(true) });
         }
         settings
@@ -655,7 +654,7 @@ impl Args {
             connect_peers: arg_match_many_unwrap_or::<ContextualNetAddress>(&m, "connect-peers", defaults.connect_peers),
             add_peers: arg_match_many_unwrap_or::<ContextualNetAddress>(&m, "add-peers", defaults.add_peers),
             listen: m.get_one::<ContextualNetAddress>("listen").cloned().or(defaults.listen),
-            proxy: m.get_one::<ContextualNetAddress>("proxy").cloned().or(defaults.proxy.clone()),
+            proxy: m.get_one::<ContextualNetAddress>("proxy").cloned().or(defaults.proxy),
             proxy_user: m.get_one::<String>("proxy-user").cloned().or(defaults.proxy_user.clone()),
             proxy_pass: m.get_one::<String>("proxy-pass").cloned().or(defaults.proxy_pass.clone()),
             proxy_net: m
--- a/kaspad/src/daemon.rs
+++ b/kaspad/src/daemon.rs
@@ -64,7 +64,10 @@ use kaspa_mining::{
     monitor::MiningMonitor,
     MiningCounters,
 };
-use kaspa_p2p_flows::{flow_context::FlowContext, service::P2pService};
+use kaspa_p2p_flows::{
+    flow_context::{FlowContext, ProxyEndpoints, TorConfig},
+    service::P2pService,
+};
 
 use itertools::Itertools;
 use kaspa_perf_monitor::{builder::Builder as PerfMonitorBuilder, counters::CountersSnapshot};
@@ -204,8 +207,8 @@ fn compute_tor_system_config(args: &Args) -> Option<TorSystemConfig> {
     let socks_source = proxy_settings
         .onion
         .as_ref()
-        .or_else(|| proxy_settings.default.as_ref())
-        .map(|entry| entry.address.clone())
+        .or(proxy_settings.default.as_ref())
+        .map(|entry| entry.address)
         .unwrap_or_else(|| ContextualNetAddress::loopback().with_port(9050));
     let socks_addr = contextual_to_socket(socks_source, 9050);
 
@@ -592,7 +595,7 @@ pub fn create_core_with_runtime(runtime: &Runtime, args: &Args, fd_total_budget:
     let effective_tor_proxy = tor_proxy_from_manager
         .clone()
         .or_else(|| tor_proxy_override_addr.clone())
-        .or_else(|| default_proxy_addr.as_ref().map(|entry| promote_tor_proxy(entry)));
+        .or_else(|| default_proxy_addr.as_ref().map(promote_tor_proxy));
     if let Some(proxy) = effective_tor_proxy.as_ref() {
         info!("Effective Tor proxy: {}", proxy.address);
         if default_proxy_addr.is_none() {
@@ -1032,6 +1035,14 @@ do you confirm? (answer y/n or pass --yes to the Kaspad command line to confirm 
         (None, None)
     };
 
+    let proxy_endpoints = ProxyEndpoints::new(default_proxy_addr.clone(), proxy_ipv4_addr.clone(), proxy_ipv6_addr.clone());
+    let tor_config = TorConfig {
+        proxy: effective_tor_proxy.clone(),
+        tor_only: args.tor_only,
+        onion_service: onion_service_info.as_ref().map(|info| (info.id.clone(), info.virt_port)),
+        bootstrap_rx: tor_bootstrap_rx,
+    };
+
     let flow_context = Arc::new(FlowContext::new(
         consensus_manager.clone(),
         address_manager,
@@ -1041,13 +1052,8 @@ do you confirm? (answer y/n or pass --yes to the Kaspad command line to confirm 
         notification_root,
         hub.clone(),
         mining_rule_engine.clone(),
-        default_proxy_addr,
-        proxy_ipv4_addr,
-        proxy_ipv6_addr,
-        effective_tor_proxy,
-        args.tor_only,
-        onion_service_info.as_ref().map(|info| (info.id.clone(), info.virt_port)),
-        tor_bootstrap_rx,
+        proxy_endpoints,
+        tor_config,
     ));
     let tor_async_service = tor_manager.as_ref().map(|manager| {
         Arc::new(TorRuntimeService::new(
--- a/protocol/flows/src/flow_context.rs
+++ b/protocol/flows/src/flow_context.rs
@@ -215,6 +215,27 @@ impl BlockEventLogger {
     }
 }
 
+#[derive(Clone, Default)]
+pub struct ProxyEndpoints {
+    pub default: Option<SocksProxyParams>,
+    pub ipv4: Option<SocksProxyParams>,
+    pub ipv6: Option<SocksProxyParams>,
+}
+
+impl ProxyEndpoints {
+    pub fn new(default: Option<SocksProxyParams>, ipv4: Option<SocksProxyParams>, ipv6: Option<SocksProxyParams>) -> Self {
+        Self { default, ipv4, ipv6 }
+    }
+}
+
+#[derive(Clone)]
+pub struct TorConfig {
+    pub proxy: Option<SocksProxyParams>,
+    pub tor_only: bool,
+    pub onion_service: Option<(V3OnionServiceId, u16)>,
+    pub bootstrap_rx: Option<watch::Receiver<bool>>,
+}
+
 pub struct FlowContextInner {
     pub node_id: PeerId,
     pub consensus_manager: Arc<ConsensusManager>,
@@ -244,9 +265,7 @@ pub struct FlowContextInner {
 
     // Mining rule engine
     mining_rule_engine: Arc<MiningRuleEngine>,
-    proxy: Option<SocksProxyParams>,
-    proxy_ipv4: Option<SocksProxyParams>,
-    proxy_ipv6: Option<SocksProxyParams>,
+    proxy_endpoints: ProxyEndpoints,
     tor_proxy: Option<SocksProxyParams>,
     tor_only: bool,
     onion_service: Option<(V3OnionServiceId, NetAddress)>,
@@ -325,16 +344,13 @@ impl FlowContext {
         notification_root: Arc<ConsensusNotificationRoot>,
         hub: Hub,
         mining_rule_engine: Arc<MiningRuleEngine>,
-        proxy_default: Option<SocksProxyParams>,
-        proxy_ipv4: Option<SocksProxyParams>,
-        proxy_ipv6: Option<SocksProxyParams>,
-        tor_proxy: Option<SocksProxyParams>,
-        tor_only: bool,
-        onion_service: Option<(V3OnionServiceId, u16)>,
-        tor_bootstrap_rx: Option<watch::Receiver<bool>>,
+        proxy_endpoints: ProxyEndpoints,
+        tor_config: TorConfig,
     ) -> Self {
         let bps_upper_bound = config.bps().upper_bound() as usize;
         let orphan_resolution_range = BASELINE_ORPHAN_RESOLUTION_RANGE + (bps_upper_bound as f64).log2().ceil() as u32;
+
+        let TorConfig { proxy: tor_proxy, tor_only, onion_service, bootstrap_rx } = tor_config;
 
         let onion_service = onion_service.and_then(|(id, port)| {
             let onion_host = format!("{}.onion", id);
@@ -377,13 +393,11 @@ impl FlowContext {
                 max_orphans,
                 config,
                 mining_rule_engine,
-                proxy: proxy_default,
-                proxy_ipv4,
-                proxy_ipv6,
+                proxy_endpoints,
                 tor_proxy,
                 tor_only,
                 onion_service,
-                tor_bootstrap_rx: Mutex::new(tor_bootstrap_rx),
+                tor_bootstrap_rx: Mutex::new(bootstrap_rx),
             }),
         }
     }
@@ -401,15 +415,15 @@ impl FlowContext {
     }
 
     pub fn proxy(&self) -> Option<SocksProxyParams> {
-        self.proxy.clone()
+        self.proxy_endpoints.default.clone()
     }
 
     pub fn proxy_ipv4(&self) -> Option<SocksProxyParams> {
-        self.proxy_ipv4.clone()
+        self.proxy_endpoints.ipv4.clone()
     }
 
     pub fn proxy_ipv6(&self) -> Option<SocksProxyParams> {
-        self.proxy_ipv6.clone()
+        self.proxy_endpoints.ipv6.clone()
     }
 
     pub fn tor_only(&self) -> bool {
--- a/protocol/p2p/src/core/connection_handler.rs
+++ b/protocol/p2p/src/core/connection_handler.rs
@@ -320,5 +320,5 @@ async fn connect_via_socks(params: SocksProxyParams, target: String) -> io::Resu
             Socks5Stream::connect_with_password(address, target, username.as_str(), password.as_str()).await
         }
     };
-    result.map_err(|err| io::Error::new(io::ErrorKind::Other, err))
+    result.map_err(io::Error::other)
 }
--- a/rpc/grpc/core/src/convert/peer.rs
+++ b/rpc/grpc/core/src/convert/peer.rs
@@ -45,8 +45,8 @@ try_from!(item: &protowire::GetConnectedPeerInfoMessage, kaspa_rpc_core::RpcPeer
 });
 
 try_from!(item: &protowire::GetPeerAddressesKnownAddressMessage, kaspa_rpc_core::RpcPeerAddress, {
-    Self::from_str(&item.addr).map_err(NetAddressError::from)?
+    Self::from_str(&item.addr).map_err(RpcError::from)?
 });
 try_from!(item: &protowire::GetPeerAddressesKnownAddressMessage, kaspa_rpc_core::RpcIpAddress, {
-    Self::from_str(&item.addr).map_err(NetAddressError::from)?
+    Self::from_str(&item.addr).map_err(|err| RpcError::from(NetAddressError::from(err)))?
 });
--- a/utils/src/networking.rs
+++ b/utils/src/networking.rs
@@ -438,7 +438,7 @@ impl BorshDeserialize for AddressKind {
         match variant_idx {
             0 => Ok(AddressKind::Ip(BorshDeserialize::deserialize_reader(reader)?)),
             1 => Ok(AddressKind::Onion(BorshDeserialize::deserialize_reader(reader)?)),
-            _ => Err(std::io::Error::new(std::io::ErrorKind::InvalidInput, "invalid address kind variant").into()),
+            _ => Err(std::io::Error::new(std::io::ErrorKind::InvalidInput, "invalid address kind variant")),
         }
     }
 }
--- a/kaspad/src/tor_manager.rs
+++ b/kaspad/src/tor_manager.rs
@@ -176,8 +176,8 @@ impl TorManager {
                 return Err(TorManagerError::Control("unexpected EOF while waiting for ADD_ONION response".into()));
             }
             let trimmed = line.trim();
-            if trimmed.starts_with("250-ServiceID=") {
-                service_id = Some(trimmed["250-ServiceID=".len()..].to_string());
+            if let Some(stripped) = trimmed.strip_prefix("250-ServiceID=") {
+                service_id = Some(stripped.to_string());
             } else if trimmed.starts_with("250 ") {
                 break;
             } else if trimmed.starts_with('5') {


saefstroem:
Running P2P over Tor will impact block discovery for other peers due to slow performance. This could have negative effects on the anticone count of the DAG and in general be penalizing in terms of network connectivity. I am wondering if this has been considered and if-so what are the mitigations in place?

D-Stacks:
> Running P2P over Tor will impact block discovery for other peers due to slow performance. This could have negative effects on the anticone count of the DAG and in general be penalizing in terms of network connectivity. I am wondering if this has been considered and if-so what are the mitigations in place?
Just my 2 cents, but, 

1) in general the p2p network is really only as slow as the fastest link(s), if we ensure all clear-net nodes only have partial connectivity to the TOR network (something like a set inbound / outbound limit at maybe something like 10% - 20% of total network) I think the impact will be minimal. As I have been measuring peer connectivity with my perigee implementation, I can ascertain that block delivery is really only dependent on the fastest link(s) in practice. Or to visualize it as an analogy, Imagine a highway, TOR nodes will be on the creeper lane, all other nodes will be on the fast lane.. as long as we ensure TOR nodes do not congest the whole highway, in practice their impact should be minimal. 

2) This doesn't hold true if miners actually mine blocks directly into the TOR network, thereby emitting latency to the whole network but I do think this would be considered bad practice on a wide array of metrics, including economically, for the miner, as kaspa mining is about a race to mine the tips.

In short: I think it is just about proper segregation of the TOR network from the clear-net, and limiting it's exposure over clear-net, so it cannot dominate latency in the clear-net, which will remain the part where blocks will be actually mined and communicated to other miners.

D-Stacks:
Btw when I have time I'll try to review this, TOR has also been on my p2p bucket list. Just some overall input from my side on how I  see it:  

1) I think TOR Addresses and handling IMO should be segregated on a much finer granularity imo, probably with a separate address store, hub hash-map etc.. 

2) If we think about integrating TOR we should think about designing a frame-work to include different overlay-networks as a whole, and place TOR within it, although the only one besides TOR that I can really think of is I2P.. but in case we want to include more overlay networks we will have a clear and easy pathway to do this. 

_Is the security/maintenance cost of native Tor support acceptable for upstream Kaspa?_

It depends what you mean with native, I think the maintenance to support connecting to, and proxying over TOR should be minimal, I guess it is mostly just about expanding the address handling to support TOR addresses, keeping some other types of sockets up, and to ensure segregation of TOR and clear-net addresses. I wouldn't directly support the use of in-process spawned TOR, i.e. something like arti. This puts the maintenance burden back on us, if things end up breaking with them, or there is some versioning misalignment etc.. etc.. we get the brunt of it. If it is their client we are proxying over, then they get to at least share some of it. Also I think the people who wish to use TOR are probably the same that don't mind reading up on some set-up guide, or probably don't even need to, once they know how to point kaspa into the right direction, and don't need some all-encompassing `--run-over-tor` flag.  Even if We do wish to support it natively, I think getting out a "minimal viable product" out first is always the way to go.  

_Should I finish the DNS/name-proxy refactor and align listener defaults before even asking for mergeability?_

I'll be honest, I didn't understand the issue with dns seeders, is the problem simply that we also need to PR to the dns seeders to make them hand out tor addresses, or do we need a .onion dns seeder?

_How should we document limitations (bandwidth, Tor restarts, lack of bundled daemon) so operators don’t misinterpret the feature?_

Personally I would just include a set-up guide in our README.md, probably over or under the `Start a devnet node`  section, with instructions on how to proxy over TOR with a bunch of disclaimers. 

That being said, I am not sure what the overall enthusiasm for TOR is here, I have less qualms about them dragging down the network as a whole, but more so about them being so slow, that we cannot support them into the future.. Imo it needs to be bench-marked under high load testnet conditions, It is a very real possibility that TOR might simply be too slow to tango with kaspa's throughput. Maintaining and putting effort into something that cannot catch-up anyway probably isn't worth the effort.

cli/src/modules/rpc.rs
```diff
@@ -200,15 +201,15 @@ impl Rpc {
                 if argv.is_empty() {
                     return Err(Error::custom("Please specify peer IP address"));
                 }
-                let ip: RpcIpAddress = argv.remove(0).parse()?;
+                let ip: RpcIpAddress = argv.remove(0).parse().map_err(NetAddressError::from)?;
```
D-Stacks, line 204: can NetAddressError be emitted from the parse() call?

components/addressmanager/src/stores/address_store.rs
```diff
@@ -118,3 +168,9 @@ impl AddressesStore for DbAddressesStore {
         self.set(key, Entry { connection_failed_count, address: entry.address })
     }
 }
+
+impl DbAddressesStore {
+    pub fn clear(&self) -> StoreResult<()> {
+        self.access.delete_all(DirectDbWriter::new(&self.db))
+    }
+}
```
D-Stacks: I'll just leave one comment for this file, so I think we should stat with:

```rs
#[derive(Clone)]
pub struct DbAddressesStore {
    db: Arc<DB>,
    clear_net_access: CachedDbAccess<DbAddressKey, Entry>,
    tor_access: CachedDbAccess<DbTorAddressKey, TorEntry>
}
```

leave as much current code untouched, give the tor_access its own Serialization and key for it, own methods etc..

components/addressmanager/src/lib.rs
```diff
@@ -286,6 +322,29 @@ impl AddressManager {
         self.address_store.set(address, 0);
     }
 
+    fn prune_onion_addresses(&mut self) {
```
D-Stacks, line 325: see comment at line 85

utils/src/networking.rs
```diff
@@ -218,21 +222,284 @@ impl BorshDeserialize for IpAddress {
...
+            AddressKind::Onion(onion) => {
+                let mut bytes = [0u8; 8];
+                bytes.copy_from_slice(&onion.raw()[..8]);
+                PrefixBucket(u64::from_be_bytes(bytes))
+            }
+        }
+    }
+}
+
 /// A network address, equivalent of a [SocketAddr].
```
D-Stacks, line 459: the comment here is no longer accurate.

utils/src/networking.rs
```diff
@@ -218,21 +222,284 @@ impl BorshDeserialize for IpAddress {
...
 #[derive(PartialEq, Eq, Hash, Copy, Clone, Serialize, Deserialize, Debug, BorshSerialize, BorshDeserialize)]
 pub struct NetAddress {
-    pub ip: IpAddress,
+    pub ip: AddressKind,
     pub port: u16,
 }
 
 impl NetAddress {
     pub fn new(ip: IpAddress, port: u16) -> Self {
```
D-Stacks, line 467: instead of `new`, `new_onion` and `new_kind`, I'd rather have this just accept an `AddressKind` explictly in new, i.e. remove `new`, `new_onion`, and rename `from_kind` to `new`. refractor accordingly.

utils/src/networking.rs
```diff
@@ -218,21 +222,284 @@ impl BorshDeserialize for IpAddress {
...
+
+    pub fn new_onion(address: OnionAddress, port: u16) -> Self {
+        Self { ip: AddressKind::Onion(address), port }
+    }
+
+    pub fn from_kind(kind: AddressKind, port: u16) -> Self {
+        Self { ip: kind, port }
+    }
+
+    pub fn is_ip(&self) -> bool {
```
D-Stacks, line 479: I think if we have a use-case for it somewhere in the code-base, which we probably do, we should add `is_onion` as well. in cases where we explicitly want to check if it is onion, we can then use that call, rather then having to infer that the inverse of is_ip is an onion.

utils/src/networking.rs
```diff
@@ -218,21 +222,284 @@ impl BorshDeserialize for IpAddress {
...
+
+    pub fn as_onion(&self) -> Option<OnionAddress> {
+        self.ip.as_onion()
+    }
+
+    pub fn kind(&self) -> AddressKind {
+        self.ip
     }
 
     pub fn prefix_bucket(&self) -> PrefixBucket {
```
D-Stacks, line 495: the prefix_bucket is used for a rough geographic dispersal estimate in clear net ips, I believe in .onion addresses this measures nothing really, so it should return `Option<PrefixBucket>` and `None` for onion addresses.

utils/src/networking.rs
```diff
@@ -218,21 +222,284 @@ impl BorshDeserialize for IpAddress {
...
+
+    pub fn from_kind(kind: AddressKind, port: u16) -> Self {
+        Self { ip: kind, port }
+    }
+
+    pub fn is_ip(&self) -> bool {
+        self.ip.is_ip()
+    }
+
+    pub fn as_ip(&self) -> Option<IpAddress> {
```
D-Stacks, line 483: I think if we follow and segregate ips throughout the whole codebase at each location we should always know with what kind of address we are dealing with, and there should be no need to optionalize anything. I think this as_ip should either panic, and be renamed to unwrap_ip(), or should be renamed to try_as_ip(), returning a result, which we can unwrap where needed.

utils/src/networking.rs
```diff
@@ -218,21 +222,284 @@ impl BorshDeserialize for IpAddress {
...
+        self.ip.as_ip()
+    }
+
+    pub fn as_onion(&self) -> Option<OnionAddress> {
```
D-Stacks, line 487: see comment regarding `as_ip`

utils/src/networking.rs
```diff
@@ -218,21 +222,284 @@ impl BorshDeserialize for IpAddress {
...
+
+    pub fn kind(&self) -> AddressKind {
+        self.ip
     }
 
     pub fn prefix_bucket(&self) -> PrefixBucket {
         PrefixBucket::from(self)
     }
+
+    pub fn to_socket_addr(&self) -> Result<SocketAddr, NetAddressError> {
```
D-Stacks, line 499: should be renamed to `try_to_socket_addr` as it emits a result.

utils/src/networking.rs
```diff
@@ -218,21 +222,284 @@ impl BorshDeserialize for IpAddress {
...
+                bytes.copy_from_slice(&onion.raw()[..8]);
+                PrefixBucket(u64::from_be_bytes(bytes))
+            }
+        }
+    }
+}
+
 /// A network address, equivalent of a [SocketAddr].
 #[derive(PartialEq, Eq, Hash, Copy, Clone, Serialize, Deserialize, Debug, BorshSerialize, BorshDeserialize)]
 pub struct NetAddress {
```
D-Stacks, line 461: So, even though I left a bunch of comments here, on how I would implement it (in case other reviewers decide it is the preferred method). I generally think this whole merging of Socket and Onion addresses into one struct shouldn't be required in the codebase. it should suffice to simply keep the NetAddress, as it was, maybe rename it to `StdNetAddress` or something to indicate that it pertains to clearnet, and simply define some sort of enum, like:

```rs
enum NetAddress {
    StdAddress(StdNetAddress)
    OnionAddress(OnionAddress)
} 
```

D-Stacks: this way we can simply match and use the addresses underlying methods directly in the codebase.

components/addressmanager/src/stores/mod.rs
```diff
@@ -1,34 +1,39 @@
...
 
-pub use kaspa_utils::networking::NetAddress;
+pub use kaspa_utils::networking::{AddressKind, NetAddress};
 
 pub(super) mod address_store;
 pub(super) mod banned_address_store;
 
 #[derive(PartialEq, Eq, Hash, Clone, Copy)]
-pub struct AddressKey(Ipv6Addr, u16);
+pub struct AddressKey {
```
D-Stacks, line 9: Also Here use StdAddressKey and OnionAddressKey, keep separation of concerns going.

components/addressmanager/src/stores/mod.rs
```diff
@@ -1,34 +1,39 @@
...
+    pub fn new(kind: AddressKind, port: u16) -> Self {
+        Self { kind, port }
     }
 
     pub fn is_ip(&self, ip: IpAddr) -> bool {
-        match ip {
-            IpAddr::V4(ip) => ip.to_ipv6_mapped() == self.0,
-            IpAddr::V6(ip) => ip == self.0,
+        match self.kind {
+            AddressKind::Ip(stored) => IpAddr::from(stored) == ip,
```
D-Stacks, line 21: old code had `IpAddr::V4 => ip.to_ipv6_mapped`, unless I am missing something, this seems like an unrequired change?

components/addressmanager/src/stores/mod.rs
```diff
@@ -1,34 +1,39 @@
...
 impl From<NetAddress> for AddressKey {
     fn from(value: NetAddress) -> Self {
-        AddressKey::new(
-            match value.ip.0 {
-                IpAddr::V4(ip) => ip.to_ipv6_mapped(),
-                IpAddr::V6(ip) => ip,
-            },
-            value.port,
-        )
+        AddressKey::new(value.ip, value.port)
```
D-Stacks, line 37: same here, keep `IpAddr::V4 => ip.to_ipv6_mapped`.

components/addressmanager/src/lib.rs
```diff
@@ -56,19 +56,42 @@ pub struct AddressManager {
     address_store: address_store_with_cache::Store,
     config: Arc<Config>,
     local_net_addresses: Vec<NetAddress>,
+    allow_ipv4: bool,
+    allow_ipv6: bool,
```
D-Stacks, line 60: why do we have allow_ipv4 and allow_ipv6? would it not suffice to simply have something like allow_std_ip and allow_onion?

components/addressmanager/src/lib.rs
```diff
@@ -56,19 +56,42 @@ pub struct AddressManager {
...
             local_net_addresses: Vec::new(),
             config,
+            allow_ipv4,
+            allow_ipv6,
+            allow_onion,
         };
 
         let extender = instance.init_local_addresses(tick_service);
 
+        if !instance.allow_onion {
```
D-Stacks, line 85: I'm not sure we need these prunings, if we segregate on the db level, this way when the node is restarted with / or without tor, we can simply re-use the addresses already stored in the appropriate db access and don't need to re-trigger dns seeding.

components/addressmanager/src/lib.rs
```diff
@@ -286,6 +322,29 @@ impl AddressManager {
...
     }
 
+    fn prune_onion_addresses(&mut self) {
+        let to_remove: Vec<_> = self.address_store.iterate_addresses().filter(|addr| addr.as_onion().is_some()).collect();
+        for addr in to_remove {
+            self.address_store.remove(addr);
+        }
+    }
+
+    fn prune_ipv4_addresses(&mut self) {
```
D-Stacks, line 332: see comment at line 85

components/addressmanager/src/lib.rs
```diff
@@ -286,6 +322,29 @@ impl AddressManager {
...
+        let to_remove: Vec<_> =
+            self.address_store.iterate_addresses().filter(|addr| addr.as_ip().is_some_and(|ip| ip.is_ipv4())).collect();
+        for addr in to_remove {
+            self.address_store.remove(addr);
+        }
+    }
+
+    fn prune_ipv6_addresses(&mut self) {
```
D-Stacks, line 340: see comment at line 85

components/addressmanager/src/lib.rs
```diff
@@ -286,6 +322,29 @@ impl AddressManager {
...
+        let to_remove: Vec<_> =
+            self.address_store.iterate_addresses().filter(|addr| addr.as_ip().is_some_and(|ip| ip.is_ipv6())).collect();
+        for addr in to_remove {
+            self.address_store.remove(addr);
+        }
+    }
+
```
D-Stacks, line 347: related to changes in address_store.rs: We probably should have a `iterate_onion_addresses`, `iterate_std_addresses`, iterate_all_addresses` (which chains both), depending on requirements.

components/addressmanager/src/stores/address_store.rs
D-Stacks: this also means renaming methods to set_std / set_onion, get_std / get_onion etc.. you get the drift.

components/addressmanager/src/lib.rs
```diff
@@ -362,8 +422,30 @@ mod address_store_with_cache {
             // We manage the cache ourselves on this level, so we disable the inner builtin cache
```
D-Stacks, line 422: again I'll just leave a comment for this whole store, keep two hashmaps, one for clearnet / std ips i.e. StdAddressKey, and one for onion addresses i.e. OnionAddressKey, fill them accordingly with the appropriate db iterators (that should be established). for operations get / set etc.. we can match the netaddress (refer to comments in the utils/networking) to query and supply into the right hashmap.

components/addressmanager/src/lib.rs
```diff
@@ -362,8 +422,30 @@ mod address_store_with_cache {
...
             let db_store = DbAddressesStore::new(db, CachePolicy::Empty);
             let mut addresses = HashMap::new();
-            for (key, entry) in db_store.iterator().map(|res| res.unwrap()) {
-                addresses.insert(key, entry);
+            let mut error_count = 0usize;
```
D-Stacks, line 425: all these error things shouldn't be required if done as specified. potentially if we know the node will not allow cearnet, or onion, we can skip filling the hashmap and leave it empty. or optionalize the hashmap to be more explicit that we never require it to be populated.

components/addressmanager/src/lib.rs
D-Stacks: It is also important to note: iterating by prioritised addresses which weights the ip bucket should not be applied to onion addresses. none-the-less we probably still want to weight by exceptions. so some similar but not the same iteration needs to be done to onion addresses.

components/addressmanager/src/lib.rs
```diff
@@ -286,6 +322,29 @@ impl AddressManager {
...
+        let to_remove: Vec<_> =
+            self.address_store.iterate_addresses().filter(|addr| addr.as_ip().is_some_and(|ip| ip.is_ipv6())).collect();
+        for addr in to_remove {
+            self.address_store.remove(addr);
+        }
+    }
+
     pub fn iterate_addresses(&self) -> impl Iterator<Item = NetAddress> + '_ {
         self.address_store.iterate_addresses()
     }
```
D-Stacks, line 350: My vs code github review extension thing, won't let me comment on unchanged code, so just placing this as close to the relevant location as possible.. but banning should also handle .onion addresses.

components/connectionmanager/src/lib.rs
```diff
@@ -24,16 +24,50 @@ use tokio::{
     time::{interval, MissedTickBehavior},
 };
 
+#[derive(Clone, Copy, Debug)]
+pub struct AllowedNetworks {
+    allow_ipv4: bool,
```
D-Stacks, line 29: I think i mentioned this before, but why do we differentiate ip_v4 and ip_v6?

components/connectionmanager/src/lib.rs
```diff
@@ -24,16 +24,50 @@ use tokio::{
     time::{interval, MissedTickBehavior},
 };
 
+#[derive(Clone, Copy, Debug)]
+pub struct AllowedNetworks {
```
D-Stacks, line 28: I prefer allowance based on an arg specified number of tor outbound / inbound connections.. not as a boolean. I also think this struct complicates things, lets just keep a `tor_outbound_target`, and `tor_inbound_target` in the connection manager, along with a method `should_allow(..)` that handles the logic based on currently active peers and the targets.

D-Stacks: or maybe just put it right into the connection manager, where outbound target also lives.. i think this whole struct is just overly verbose for what it does, and not required for handling or tracking.. this can just be done in the connection manger directly.

components/connectionmanager/src/lib.rs
```diff
@@ -159,9 +203,9 @@ impl ConnectionManager {
         *requests = new_requests;
     }
 
-    async fn handle_outbound_connections(self: &Arc<Self>, peer_by_address: &HashMap<SocketAddr, Peer>) {
+    async fn handle_outbound_connections(self: &Arc<Self>, peer_by_address: &HashMap<NetAddress, Peer>) {
```
D-Stacks, line 206: so in this whole function we should use two types of missing 1) missing_std_connections 2) missing_tor_connections. 

missing_std_connections = active_outbound_std_connections.saturating_sub( connections outbound_target - tor_outbound_target)
missing_tor_connections = active_outbound_tor_connections.saturating_sub(tor_outbound_target)

handle both independently.. the std_connections should use the `iterate_prioritized_random_addresses` while for the tor addresses, as aforementioned, should use something similar but without weighting the ip prefixes. 

then have two while loops, one that first populates the jobs from one, then the other. 

it is worth noting that this function is a bit buggy, i know this because i also re-worked this for another PR, but if you don't have missing connections you should still check that the proportions of tor to std addresses are correct, if not terminate any excess in either of the one that overflows, and handle it as a missing connection for the other. If you want you can check out the perigee PR, as it handles similar logic to what you require here.

components/connectionmanager/src/lib.rs
```diff
@@ -238,7 +285,7 @@ impl ConnectionManager {
         }
     }
 
-    async fn handle_inbound_connections(self: &Arc<Self>, peer_by_address: &HashMap<SocketAddr, Peer>) {
+    async fn handle_inbound_connections(self: &Arc<Self>, peer_by_address: &HashMap<NetAddress, Peer>) {
```
D-Stacks, line 288: here is similar to the handle outbound, evict based on tor_inbound_target & tor_inbound_target.

components/connectionmanager/src/lib.rs
```diff
@@ -330,11 +379,16 @@ impl ConnectionManager {
 
     /// Returns whether the given address is a permanent request.
     pub async fn is_permanent(&self, address: &SocketAddr) -> bool {
-        self.connection_requests.lock().await.contains_key(address)
+        let net_address: NetAddress = (*address).into();
+        self.connection_requests.lock().await.contains_key(&net_address)
     }
 
     /// Returns whether the given IP has some permanent request.
     pub async fn ip_has_permanent_connection(&self, ip: IpAddr) -> bool {
```
D-Stacks, line 387: don't see why we shouldn't be able to supply a permanent onion address?

components/connectionmanager/src/lib.rs
```diff
@@ -316,8 +363,10 @@ impl ConnectionManager {
             return;
         }
         for peer in self.p2p_adaptor.active_peers() {
-            if peer.net_address().ip() == ip {
-                self.p2p_adaptor.terminate(peer.key()).await;
+            if let Some(peer_ip) = peer.net_address().as_ip() {
+                if IpAddr::from(peer_ip) == ip {
```
D-Stacks, line 367: should be able to ban an onion address as well.

kaspad/src/args.rs
```diff
@@ -50,6 +51,27 @@ pub struct Args {
     pub add_peers: Vec<ContextualNetAddress>,
     #[serde_as(as = "Option<DisplayFromStr>")]
     pub listen: Option<ContextualNetAddress>,
+    #[serde_as(as = "Option<DisplayFromStr>")]
```
D-Stacks, line 54: I'll be honest, i don't understand most of these settings, i am not a tor user myself btw, but i would like to specify a node where we can say we allow 'x' amount of tor inbound and 'y' amount of tor outbound connections. this should allow for a cross-over between clear-net kaspa and tor-net kaspa based on user-supplied params.

kaspad/src/daemon.rs
```diff
@@ -1,31 +1,53 @@
-use std::{fs, path::PathBuf, process::exit, sync::Arc, time::Duration};
+use std::{
```
D-Stacks: following my comment in args, I'll refrain from commenting this as it pertains to a lot of settings etc.. that i do not understand. but I would like to mention that a lot of this seems very verbose, and really takes away from a lot of the readability of the daemon code. can you see, if maybe, you can get as much of this set-up and init code as possible out of here, and maybe into the tor_manager?

protocol/flows/src/v5/address.rs
```diff
@@ -18,6 +16,22 @@ const MAX_ADDRESSES_SEND: usize = 1000;
 /// If a peer exceeds this value we consider it a protocol error.
 const MAX_ADDRESSES_RECEIVE: usize = 2500;
```
D-Stacks, line 18: I think this flow should be re-worked, request addresses should take an argument of what kind of addresses you are searching for tor / std, then if you are tor only you query for only tor addresses, if you are none-tor, you only query for std addresses. if you are combined you query based on the ratio of outbound target - tor target vs tor target. Note: this might require a new protocol version if actually implemented properly, but on protobuf, you can add a new field on the message, and old protobuf-users will not see it, but new ones will. this might be a decent work-around for now and for testing purposes.

protocol/flows/src/flow_context.rs
```diff
@@ -751,7 +836,10 @@ impl ConnectionInitializer for FlowContext {
 
         // Build the local version message
         // Subnets are not currently supported
+        // NOTE: We currently avoid advertising onion addresses in the handshake since legacy peers
```
D-Stacks, line 839: Turn this into a todo comment for the next protocol version update.

protocol/flows/src/flow_context.rs
```diff
@@ -212,6 +215,27 @@ impl BlockEventLogger {
     }
 }
 
+#[derive(Clone, Default)]
```
D-Stacks, line 218: I will refrain from reviewing this file, due to lack of knowledge of tor things that were added, might come back to it later.

protocol/flows/src/v5/address.rs
D-Stacks: although current work-around with one-of might also suffice.

D-Stacks: and it should be a protocol error if wrong addresses are returned.

protocol/p2p/src/convert/net_address.rs
```diff
@@ -1,73 +1,60 @@
...
-            },
-            port: port as u32,
-        }
-    }
-}
-
 impl From<NetAddress> for protowire::NetAddress {
     fn from(item: NetAddress) -> Self {
-        (item.ip, item.port).into()
+        let address = match item.kind() {
```
D-Stacks, line 14: see comments i left at utils/networking they might be relevant here.

protocol/p2p/src/core/connection_handler.rs
```diff
@@ -51,15 +62,60 @@ pub struct ConnectionHandler {
...
+
+#[derive(Clone, Debug)]
+pub enum SocksAuth {
+    None,
+    Static { username: Arc<str>, password: Arc<str> },
+    Randomized,
+}
+
+#[derive(Clone, Debug)]
+pub struct SocksProxyParams {
```
D-Stacks, line 76: idk, I see a lot of these configs / params etc.. scattered around in the codebase, some seem similar or near identical. maybe just keep and hide all this in utils/tor.rs somewhere? and see if some can be re-used across the code-base?

protocol/p2p/src/core/connection_handler.rs
```diff
@@ -221,3 +291,34 @@ impl ProtoP2p for ConnectionHandler {
         Ok(Response::new(Box::pin(ReceiverStream::new(outgoing_receiver).map(Ok)) as Self::MessageStreamStream))
     }
 }
+
+fn random_isolation_prefix() -> String {
+    let mut bytes = [0u8; 8];
+    OsRng.fill_bytes(&mut bytes);
```
D-Stacks, line 297: Why OsRng? just seems like a weird choice, if you need security, you should probably use StdRng, else what is wrong with ThreadRng?

protocol/p2p/src/core/peer.rs
```diff
@@ -41,7 +42,7 @@ impl Peer {
     }
```
D-Stacks, line 43: lets add some sort of is_tor, you can use it with Itertools partition to quickly extract two lists of peers based on if they are tor, or not, simplifying things in handle outbound, and handle inbound.

protocol/p2p/src/core/router.rs
```diff
@@ -255,8 +254,8 @@ impl Router {
         *self.identity.lock_write() = identity;
     }
```
D-Stacks, line 256: lets add some sort of is_onion, you can use it with Itertools partition to quickly extract two lists of peers based on if they are tor, or not, simplifying things in handle outbound, and handle inbound.

protocol/p2p/src/core/peer.rs
```diff
@@ -70,23 +71,23 @@ impl Peer {
 #[derive(Debug, Copy, Clone, Hash, PartialEq, Eq)]
 pub struct PeerKey {
     identity: PeerId,
-    ip: IpAddress,
+    address: AddressKind,
 }
 
 impl PeerKey {
```
D-Stacks, line 77: PeerKey is pretty sensitive, it's hash defines a unique peer, it can safeguard against eclipse attacks. since I am unsure how enums derive their hash etc.. i think it is better here to define a custom hash explicitly, that gets the IpAddress / OnionAddress and hashes that into its state instead of just hashing the AddressKind.. also I am not sure if maybe the host bytes part of an onion address has any relevance here in hashing.. just something that should be checked.

rpc/grpc/core/src/convert/message.rs
```diff
@@ -848,10 +848,14 @@ try_from!(item: &protowire::GetSinkBlueScoreResponseMessage, RpcResult<kaspa_rpc
     Self { blue_score: item.blue_score }
 });
 
-try_from!(item: &protowire::BanRequestMessage, kaspa_rpc_core::BanRequest, { Self { ip: RpcIpAddress::from_str(&item.ip)? } });
+try_from!(item: &protowire::BanRequestMessage, kaspa_rpc_core::BanRequest, {
+    Self { ip: RpcIpAddress::from_str(&item.ip).map_err(NetAddressError::from)? }
```
D-Stacks, line 852: rework rpc.proto, define a new address type that can be both, ip based and onion based, allow this to pass through.

rpc/grpc/core/src/convert/message.rs
```diff
@@ -848,10 +848,14 @@ try_from!(item: &protowire::GetSinkBlueScoreResponseMessage, RpcResult<kaspa_rpc
...
+});
 try_from!(&protowire::BanResponseMessage, RpcResult<kaspa_rpc_core::BanResponse>);
 
-try_from!(item: &protowire::UnbanRequestMessage, kaspa_rpc_core::UnbanRequest, { Self { ip: RpcIpAddress::from_str(&item.ip)? } });
+try_from!(item: &protowire::UnbanRequestMessage, kaspa_rpc_core::UnbanRequest, {
```
D-Stacks, line 856: same here

rpc/grpc/core/src/convert/peer.rs
```diff
@@ -43,5 +44,9 @@ try_from!(item: &protowire::GetConnectedPeerInfoMessage, kaspa_rpc_core::RpcPeer
     }
 });
 
-try_from!(item: &protowire::GetPeerAddressesKnownAddressMessage, kaspa_rpc_core::RpcPeerAddress, { Self::from_str(&item.addr)? });
-try_from!(item: &protowire::GetPeerAddressesKnownAddressMessage, kaspa_rpc_core::RpcIpAddress, { Self::from_str(&item.addr)? });
+try_from!(item: &protowire::GetPeerAddressesKnownAddressMessage, kaspa_rpc_core::RpcPeerAddress, {
```
D-Stacks, line 47: same as in message.rs, this shouldn't error, but be defined in rpc.proto.

rpc/grpc/core/src/convert/peer.rs
```diff
@@ -43,5 +44,9 @@ try_from!(item: &protowire::GetConnectedPeerInfoMessage, kaspa_rpc_core::RpcPeer
...
+    Self::from_str(&item.addr).map_err(RpcError::from)?
+});
+try_from!(item: &protowire::GetPeerAddressesKnownAddressMessage, kaspa_rpc_core::RpcIpAddress, {
```
D-Stacks, line 50: same here

rpc/wrpc/server/src/address.rs
```diff
@@ -1,8 +1,8 @@
 use crate::service::WrpcEncoding;
 use kaspa_consensus_core::network::NetworkType;
-use kaspa_utils::networking::ContextualNetAddress;
+use kaspa_utils::networking::{ContextualNetAddress, NetAddressError};
 use serde::Deserialize;
-use std::{net::AddrParseError, str::FromStr};
+use std::str::FromStr;
 
 #[derive(Clone, Debug, Deserialize)]
 #[serde(rename = "lowercase")]
```
D-Stacks, line 8: not too familiar with Wrpc, maybe something for coderofstuff or IzioDev to go over. but that should probably be after potential changes to Netaddress in core.

.gitignore
```diff
@@ -14,3 +14,4 @@ testing/integration/perflogs*
 Servers.toml
 release
 package-sizes.js
+CLAUDE.md
```
D-Stacks, line 17: I understand this is something to do with AI assistance, and it probably helps regarding not accidentally uploading this, consider PRing this separately, or put a big `TODO: remove this, and CLAUDE.md, before merging` over it. As I think it is independent from actual PR.

components/connectionmanager/src/lib.rs
D-Stacks: although i realized this struct / logic is re-used elsewhere, so maybe keep this then.

D-Stacks (commented):
So, I think I want larger sturctural changes, so not really requesting individual changes per-se, more of a general archtectual overhaul of how I think the addresses should be managed and integrated into kaspad. I think the finer grained changes will come as they are implemented, and i probably cannot account for all changes required with a simple review. I refrained from reviewing tor set-up code, and code involved with proxying over tor, as i do not think i am familiar enough with that part. Review is more of a guidance on the route I think this pr should go.

components/addressmanager/src/stores/address_store.rs
D-Stacks, line 1: I'll just leave one comment for this file, so I think we should start with:

```rs
#[derive(Clone)]
pub struct DbAddressesStore {
    db: Arc<DB>,
    // a cached db access for standard "clear-net" addresses
    std_access: CachedDbAccess<DbStdAddressKey, StdEntry>,
    // a cached db access for tor overlay addresses 
    onion_access: CachedDbAccess<DbOnionAddressKey, OnionEntry>
}
```

this implies separate serialization of the keys and entries and setting a new Database prefix for the new store etc.. etc.. for keys I would go for SocketAddr, or the current NetAddress (see comments regarding it in utils/networking), for StdEntry. For OnionEntry use the appropriate OnionAddress. 

this also means renaming methods to set_std / set_onion, get_std / get_onion etc.. you get the drift.

Issue #733, opened 2025-09-29
Why `SCRIPT_VECTOR_SIZE` is 36?


mahnunchik:
The standard size of `script_public_key` is 34 bytes, as shown in the comment here:

https://github.com/kaspanet/rusty-kaspa/blob/v1.0.1/consensus/core/src/tx/script_public_key.rs#L22-L25

One might think this size already includes 2 bytes for the version. But when calculating the size of a standard output, those 2 bytes are added again:

https://github.com/kaspanet/rusty-kaspa/blob/v1.0.1/wallet/core/src/tx/mass.rs#L208-L211

Where is the mistake, or why is it not a mistake?

someone235:
`rusty-kaspa` uses `smallvec` without the `const_generics` feature. This only allows you to choose from a closed list. See [here](https://github.com/servo/rust-smallvec/blob/v1.11.1/src/lib.rs#L2419). We might change this in the future to allow some tighter sizes.

mahnunchik:
@someone235 Yes, the internal implementation can use any vector size. The question is more about how `transaction_standard_output_serialized_byte_size` is calculated.

In the current implementation, it is calculated as: `8 + 2 + 8 + 36 = 54`

But it should be: `8 + 2 + 8 + 34 = 52`

Maybe we should create a separate constant for calculating this size.

someone235:
You are correct, it looks like a mistake in the way `transaction_standard_output_serialized_byte_size` is implemented

aspect:
Good catch, it looks like this was inherited from consensus https://github.com/kaspanet/rusty-kaspa/blob/v1.0.1/consensus/core/src/mass/mod.rs#L57 but incorrectly optimized.

freshair18:
@someone235 @aspect can we close this issue?

PR #757, opened 2025-11-13
Enable smallvec const_generics and optimize script vector size 

### LINKED ISSUE #748: Enable `const_generics` feature on `smallvec` to allow array sizes of any size
We currently use `smallvec` without the `const_generics` feature, which allows to choose vector capacity only from a [closed list](https://github.com/servo/rust-smallvec/blob/v1.11.1/src/lib.rs#L2419). I suggest enabling it and tighten vector capacities (e.g. Reduce `SCRIPT_VECTOR_SIZE` from 36 to 35 to accommodate a p2sh script).

avishai12321:
Fixes #748

### DIFF
--- a/kaspad/Cargo.toml
+++ b/kaspad/Cargo.toml
@@ -44,7 +44,7 @@ kaspa-wrpc-server.workspace = true
 
 async-channel.workspace = true
 cfg-if.workspace = true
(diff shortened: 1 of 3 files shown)


gcpreston:
I figure this also fixes #733 ?

PR #694, opened 2025-06-12
windows asm support for Keccak

x100111010:
It follows the same approach used in https://github.com/elichai/kaspa-miner/commit/f759ce3c6e680a8b2dbd88a38e7edb601b91c638 by @elichai adapted for the hashes crate.

Benchmarks with:
`cargo bench -p kaspa-hashes --features no-asm` and `cargo bench -p kaspa-hashes`

```
PoWHash including timestamp
                        time:   [225.16 ns 225.50 ns 225.86 ns]
                        change: [-20.500% -20.193% -19.890%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 7 outliers among 100 measurements (7.00%)
  3 (3.00%) high mild
  4 (4.00%) high severe

PoWHash without timestamp
                        time:   [220.29 ns 220.71 ns 221.19 ns]
                        change: [-21.103% -20.809% -20.508%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 5 outliers among 100 measurements (5.00%)
  2 (2.00%) high mild
  3 (3.00%) high severe
KHeavyHash              time:   [215.44 ns 215.73 ns 216.02 ns]
                        change: [-20.636% -20.326% -19.935%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 8 outliers among 100 measurements (8.00%)
  6 (6.00%) high mild
  2 (2.00%) high severe
```

PR #760, opened 2025-11-14
Android asm support for Keccak

wigwagwent:
In pull request https://github.com/kaspanet/rusty-kaspa/pull/694 android support was removed from the build script. This adds it back in. From what I understand linux and android should be similar in the assembly file. I am not sure the correct way to test this. I can build the `x86_64-linux-android` target after this change.

### DIFF
--- a/crypto/hashes/build.rs
+++ b/crypto/hashes/build.rs
@@ -12,7 +12,7 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
         builder.flag("-c");
         match target_os.as_str() {
             "macos" => builder.file("src/asm/keccakf1600_x86-64-osx.s"),
-            "linux" => builder.file("src/asm/keccakf1600_x86-64-elf.s"),
+            "linux" | "android" => builder.file("src/asm/keccakf1600_x86-64-elf.s"),
             "windows" if target_env == "gnu" => builder.file("src/asm/keccakf1600_x86-64-mingw64.s"),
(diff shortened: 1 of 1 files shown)


michaelsutton:
> In pull request #694 android support was removed from the build script. This adds it back in. From what I understand linux and android should be similar in the assembly file. I am not sure the correct way to test this. I can build the `x86_64-linux-android` target after this change.

so previous code was
```
    if target_arch == "x86_64" && target_os != "windows" && target_os != "macos" {
        cc::Build::new().flag("-c").file("src/keccakf1600_x86-64.s").compile("libkeccak.a");
    }
```

any other linux-related os that might have been dropped other than android? 

cc @elichai

elichai:
Interesting, yeah I probably didn't think of other platforms when I wrote it(in Kaspa-miner) 😅

All Linux based OSs should probably work,
Maybe some openbsd might work? Or it might be more similar to Macos than Linux,
The best way to check would be to run the tests with the asm while emulating those OSs, but I don't know off of hand

jcroger:
I tested this patch on xx86_64-linux-android and it works correctly.
One small suggestion: we could also add "ios" to the macOS arm, to cover the iOS simulator on Intel Macs (x86_64-apple-ios):
```
rust"macos" | "ios" => builder.file("src/asm/keccakf1600_x86-64-osx.s"),
"linux" | "android" => builder.file("src/asm/keccakf1600_x86-64-elf.s"),
```
The case is rare but it makes the code more complete and explicit.

PR #761, opened 2025-11-15
Add PQ signature scheme(Falcon512Padded) support to txscript

biryukovmaxim:
This PR introduces full support for Falcon-512 (padded) signatures in the UTXO and script verification layers.

~~build fails in wasm. maybe we can replact the impl for wasm:
https://github.com/partylikeits1983/falcon512-stm32/tree/main/falcon-wasm~~

wasm build will be fixed when i add support of wasm-unknown-unknown into pqcrypto repo. local fork works fine if i statically include [wasm32-unknown-unknown-openbsd-libc](https://github.com/trevyn/wasm32-unknown-unknown-openbsd-libc/tree/main)


michaelsutton:
I understand this is a long-shot PR, but why not price a Falcon sig as N sigops? (say, 5 usual sigops). It seems redundant to add a new field

biryukovmaxim:
> I understand this is a long-shot PR, but why not price a Falcon sig as N sigops? (say, 5 usual sigops). It seems redundant to add a new field

Makes sense. I initially was thinking that it would lead to u8 overflow at tx level

biryukovmaxim:
corresponding pr https://github.com/rustpq/pqcrypto/pull/94

PR #765, opened 2025-11-19
VRLE parents by level refactor

Manyfestation:
# RLE compression for parents_by_level in block header

This is a continuation of PR #744.

### NOTE- 

This change introduces a new database format for headers.

*  Upgrade is automatic and instant- A fallback logic ensures older header formats can still be read without a migration process.
* Once the DB runs with this version, it is **not compatible** with older versions. Going back to master would require deleting the DB.

---

### Motivation

It is common for many headers to share identical parent sets across parent levels. This redundancy offers an opportunity for compression.

The PR introduces Run-Length Encoding (RLE) to compress the `parents_by_level` field in block headers. Based on a sample of 5 random headers, the compression reduces the header size by ~80%.

### Changes

* Replaced `Vec<Vec<Hash>>` with a new `CompressedParents` struct for `header.parents_by_level` to store data in RLE format.
* **DB Fallback:** Updated `DbHeadersStore` to handle existing databases. It uses a `read_with_fallback` approach: it tries to read from the new `CompressedHeaders` prefix first. If not found, it looks in the old `Headers` prefix.
* Added temporary `Header2` and `HeaderWithBlockLevel2` structs to deserialize the legacy format, which are then converted to the new compressed format in memory.
* New headers are always written to the new compressed prefix.

### DIFF
--- a/kaspad/src/daemon.rs
+++ b/kaspad/src/daemon.rs
@@ -377,6 +377,10 @@ do you confirm? (answer y/n or pass --yes to the Kaspad command line to confirm 
         // If I'm at version 3 and latest version is 7, I need to be able to upgrade to that version following the intermediate
         // steps without having to delete the DB
         if version == 4 {
+            let msg = "NOTE: Node database is from an older version. Proceeding with the upgrade is instant and safe.
+However, downgrading to an older node version later will require deleting the database.
+Do you confirm? (y/n)";
+            get_user_approval_or_exit(msg, args.yes);
             mcms.set_version(kaspa_consensus::consensus::factory::LATEST_DB_VERSION).unwrap();
         } else {
             let msg =


consensus/src/model/stores/headers.rs
```diff
@@ -38,6 +38,54 @@ pub trait HeaderStore: HeaderStoreReader {
     fn delete(&self, hash: Hash) -> Result<(), StoreError>;
 }
 
+/// A temporary struct for backward compatibility. This struct is used to deserialize old header data with
+/// parents_by_level as Vec<Vec<Hash>>.
+#[derive(Clone, Debug, Deserialize)]
+pub struct Header2 {
```
freshair18: Probably better to have a more descriptive name.

Header_V2 would be slightly better.

michaelsutton: It's an internal sturuct with no outside usage. The name is not important as long as doc msg is clear. Also V2 sounds like a later version (whereas this is the prev)

freshair18: my misunderstanding, then - because of the name. Maybe header_legacy?

consensus/core/src/header.rs
```diff
@@ -66,19 +146,26 @@ impl Header {
...
     }
 
+    pub fn parents_by_level_vec(&self) -> Vec<Vec<Hash>> {
+        self.parents_by_level.clone().into()
+    }
+
+    pub fn set_parents_by_level_vec(&mut self, parents: Vec<Vec<Hash>>) {
+        self.parents_by_level = parents.try_into().unwrap();
+    }
+
```
freshair18: is this function only for tests? If so perhaps it should be written explicitly somewhere.
(what caught my eye is the unwrap)

consensus/core/src/header.rs
```diff
@@ -1,8 +1,88 @@
...
+impl CompressedParents {
+    pub fn len(&self) -> usize {
+        self.0.last().map(|(cum, _)| *cum as usize).unwrap_or(0)
+    }
+
+    pub fn is_empty(&self) -> bool {
+        self.0.is_empty()
+    }
+
+    pub fn get(&self, index: usize) -> Option<&Vec<Hash>> {
```
biryukovmaxim: Returning slice instead of reference to Vec would look better

consensus/core/src/header.rs
```diff
@@ -1,8 +1,88 @@
...
+            return None;
+        }
+        // `partition_point` returns the index of the first element for which the predicate is false.
+        // The predicate `cum - 1 < index` checks if a run is before the desired `index`.
+        // The first run for which this is false is the one that contains our index.
+        let i = self.0.partition_point(|(cum, _)| (*cum as usize) - 1 < index);
+        Some(&self.0[i].1)
+    }
+
+    pub fn iter(&self) -> impl Iterator<Item = &'_ Vec<Hash>> {
```
biryukovmaxim: Same comment regarding slice

consensus/src/model/stores/headers.rs
michaelsutton: Maybe. More importantly - it should be a private struct (so need to remove the pub)

consensus/client/src/header.rs
```diff
@@ -284,6 +286,8 @@ impl TryCastFromJs for Header {
                     })
                     .collect::<std::result::Result<Vec<Vec<Hash>>, Error>>()?;
 
+                let parents_by_level = parents_by_level_vec.try_into().unwrap();
```
michaelsutton: This function returns a Result, so you shouldn't unwrap but rather propagate the err via `?` operator and proper error casting

consensus/client/src/header.rs
```diff
@@ -246,6 +246,8 @@ impl Header {
             .unwrap_or_else(|err| {
                 panic!("{}", err);
             });
+
+        self.inner_mut().parents_by_level = parents.try_into().unwrap();
```
michaelsutton, line 250: It seems like this function panics on errors, so I guess the `unwrap` is fine here? 
cc @biryukovmaxim @aspect (ctx: this will only fail if someone submits a consensus-illegal vec with more than 255 levels)

consensus/core/src/header.rs
```diff
@@ -1,8 +1,88 @@
...
-use kaspa_utils::mem_size::MemSizeEstimator;
+use kaspa_utils::{
+    iter::{IterExtensions, IterExtensionsRle},
+    mem_size::MemSizeEstimator,
+};
 use serde::{Deserialize, Serialize};
+use std::mem::size_of;
+
+#[derive(Clone, Debug, Default, PartialEq, Eq, Serialize, Deserialize, BorshSerialize, BorshDeserialize)]
+
```
michaelsutton: blank line

consensus/core/src/header.rs
```diff
@@ -1,8 +1,88 @@
...
 use borsh::{BorshDeserialize, BorshSerialize};
 use kaspa_hashes::Hash;
-use kaspa_utils::mem_size::MemSizeEstimator;
+use kaspa_utils::{
+    iter::{IterExtensions, IterExtensionsRle},
+    mem_size::MemSizeEstimator,
+};
 use serde::{Deserialize, Serialize};
+use std::mem::size_of;
+
```
michaelsutton, line 10: a doc comment over the struct explaining the basic encoding/compression logic would be helpful

consensus/core/src/header.rs
```diff
@@ -1,8 +1,88 @@
...
+}
+
+impl From<CompressedParents> for Vec<Vec<Hash>> {
+    fn from(value: CompressedParents) -> Self {
+        value.iter().cloned().collect()
+    }
+}
+
+impl CompressedParents {
+    pub fn len(&self) -> usize {
```
michaelsutton: I would consider renaming to `expanded_len`

consensus/core/src/header.rs
michaelsutton: I would consider renaming to `expanded_iter`

consensus/core/src/header.rs
```diff
@@ -1,8 +1,88 @@
...
+
+impl TryFrom<Vec<Vec<Hash>>> for CompressedParents {
+    type Error = &'static str;
+
+    fn try_from(parents: Vec<Vec<Hash>>) -> Result<Self, Self::Error> {
+        if parents.len() > u8::MAX as usize {
+            return Err("Parents by level exceeds maximum levels of 255");
+        }
+
+        if parents.is_empty() {
```
michaelsutton: the call `rle_cumulative` will return an empty iter if parents is empty, so this check seems redundant

consensus/src/processes/parents_builder.rs
```diff
@@ -179,19 +183,14 @@ impl<T: HeaderStoreReader, U: ReachabilityStoreReader, V: RelationsStoreReader>
...
-            // If is genesis
-            &[]
-        } else if header.parents_by_level.len() > level as usize {
-            &header.parents_by_level[level as usize][..]
-        } else {
-            std::slice::from_ref(&self.genesis_hash)
-        }
+        header.parents_by_level.get(level as usize).map(Vec::as_slice).unwrap_or_else(|| {
+            if header.parents_by_level.is_empty() {
+                &[]
```
michaelsutton, line 190: keep the comment here 
`// If is genesis`

database/src/access.rs
```diff
@@ -67,6 +67,51 @@ where
...
+            if self.db.get_pinned(&db_key)?.is_some() {
+                Ok(true)
+            } else {
+                let db_key = DbKey::new(fallback_prefix, key.clone());
+                Ok(self.db.get_pinned(&db_key)?.is_some())
+            }
+        }
+    }
+
+    pub fn read_with_fallback<TDeser>(&self, fallback_prefix: &[u8], key: TKey) -> Result<TData, StoreError>
```
michaelsutton: maybe rename `TDeser` -> `TFallbackDeser`

protocol/p2p/src/convert/header.rs
```diff
@@ -44,7 +44,12 @@ impl TryFrom<protowire::BlockHeader> for Header {
     fn try_from(item: protowire::BlockHeader) -> Result<Self, Self::Error> {
         Ok(Self::new_finalized(
             item.version.try_into()?,
-            item.parents.into_iter().map(Vec::<Hash>::try_from).collect::<Result<Vec<Vec<Hash>>, ConversionError>>()?,
+            item.parents
+                .into_iter()
+                .map(Vec::<Hash>::try_from)
+                .collect::<Result<Vec<Vec<Hash>>, ConversionError>>()?
+                .try_into()
+                .unwrap(),
```
michaelsutton: most critical comment: this must be `?`

rpc/core/src/model/header.rs
```diff
@@ -85,7 +85,7 @@ impl From<&Header> for RpcHeader {
         Self {
             hash: header.hash,
             version: header.version,
-            parents_by_level: header.parents_by_level.clone(),
+            parents_by_level: Vec::from(&header.parents_by_level),
```
michaelsutton: for consistency use `.into()` here as well

rpc/core/src/model/header.rs
```diff
@@ -240,7 +240,7 @@ impl From<&Header> for RpcRawHeader {
     fn from(header: &Header) -> Self {
         Self {
             version: header.version,
-            parents_by_level: header.parents_by_level.clone(),
+            parents_by_level: Vec::from(&header.parents_by_level),
```
michaelsutton: for consistency use `.into()` here as well

rpc/grpc/core/src/convert/header.rs
```diff
@@ -52,7 +52,7 @@ try_from!(item: &protowire::RpcBlockHeader, kaspa_rpc_core::RpcHeader, {
     // We re-hash the block to remain as most trustless as possible
     let header = Header::new_finalized(
         item.version.try_into()?,
-        item.parents.iter().map(Vec::<RpcHash>::try_from).collect::<RpcResult<Vec<Vec<RpcHash>>>>()?,
+        item.parents.iter().map(Vec::<RpcHash>::try_from).collect::<RpcResult<Vec<Vec<RpcHash>>>>()?.try_into().unwrap(),
```
michaelsutton: same critical comment: must be handled via `?` and not an unwrap. Any code paths coming from the outside world (p2p, grpc, wrpc) must be dealt with caution

rpc/core/src/model/header.rs
```diff
@@ -105,7 +105,7 @@ impl From<RpcHeader> for Header {
         Self {
             hash: header.hash,
             version: header.version,
-            parents_by_level: header.parents_by_level,
+            parents_by_level: header.parents_by_level.try_into().unwrap(),
```
michaelsutton: replace `unwrap` with `?`. This will require changing `From<RpcHeader>` to `TryFrom<RpcHeader>` and updating usages, but should be an easy change (conversion paths are expected to be fallible on code leading from rpc to consensus. ie the code calling this conversion is expected to be within functions that return a `Result` already) 

one might argue that `RpcHeader` and `RpcRawHeader` should use `CompressedParents` in the first place, but that would be a breaking API change and should be done as a possible followup task

rpc/core/src/model/header.rs
```diff
@@ -125,7 +125,7 @@ impl From<&RpcHeader> for Header {
         Self {
             hash: header.hash,
             version: header.version,
-            parents_by_level: header.parents_by_level.clone(),
+            parents_by_level: header.parents_by_level.clone().try_into().unwrap(),
```
michaelsutton: same

rpc/core/src/model/header.rs
```diff
@@ -202,7 +202,7 @@ impl From<RpcRawHeader> for Header {
     fn from(header: RpcRawHeader) -> Self {
         Self::new_finalized(
             header.version,
-            header.parents_by_level,
+            header.parents_by_level.try_into().unwrap(),
```
michaelsutton: same

rpc/core/src/model/header.rs
```diff
@@ -221,7 +221,7 @@ impl From<&RpcRawHeader> for Header {
     fn from(header: &RpcRawHeader) -> Self {
         Self::new_finalized(
             header.version,
-            header.parents_by_level.clone(),
+            header.parents_by_level.clone().try_into().unwrap(),
```
michaelsutton: same

consensus/client/src/header.rs
Manyfestation: Proper casting means to create a default error casting using `thiserror`? 
This line stays `let parents_by_level = parents_by_level_vec.try_into()?;`?

consensus/core/src/header.rs
Manyfestation: Yes it is. Can you think of a better place for it?

protocol/p2p/src/convert/header.rs
Manyfestation: 🙏

consensus/client/src/header.rs
Manyfestation: Let me know if this needs to be changed

consensus/client/src/header.rs
Manyfestation: By proper error casting your mean to use `thiserror` to create a new error type, true?

consensus/core/src/header.rs
Manyfestation: Good point

consensus/core/src/header.rs
Manyfestation: Yes it is, can you think of a better placement for this? Thanks!

consensus/src/model/stores/headers.rs
Manyfestation: Thanks, will do, along side `HeaderWithBlockLevel2`

protocol/p2p/src/convert/header.rs
Manyfestation: 🙏

rpc/core/src/model/header.rs
Manyfestation: Got you, TY

PR #770, opened 2025-11-30
feat(wasm): pskt & pskb sign, get input addresses and iterate pskts from pskb

IzioDev:
### What
This PR is trying to add `sign`, `addresses` (get input addresses) and `finalize_p2pk` to pskt and pskb on wasm exposed struct.
It also add a way of iterating pskts from a pskb.

Since the current rust (native) implementation is tied to wallet & account struct, I had to do it off the native implementation, but ideally we would have the wasm part using the native implementation instead of duplicating logic.

### Why
This is needed for pskb/pskt wasm usage, which is needed for KIP-12.

Please, carefully review as this is the first time I contribute to this part of the codebase (which has many specificity i'm still unfamiliar with)

### Extra
Fix existing wasm tests and added wasm test suites execution as a Github Action (CI)

### DIFF
--- a/wallet/pskt/src/pskt.rs
+++ b/wallet/pskt/src/pskt.rs
@@ -131,7 +131,7 @@ impl<ROLE> Deref for PSKT<ROLE> {
 }
 
 impl<R> PSKT<R> {
-    pub fn unsigned_tx(&self) -> SignableTransaction {
+    fn unsigned_tx(&self) -> SignableTransaction {
         let tx = Transaction::new(
             self.global.tx_version,
             self.inputs


wallet/pskt/src/pskt.rs
```diff
@@ -131,7 +131,7 @@ impl<ROLE> Deref for PSKT<ROLE> {
 }
 
 impl<R> PSKT<R> {
-    fn unsigned_tx(&self) -> SignableTransaction {
+    pub fn unsigned_tx(&self) -> SignableTransaction {
```
IzioDev: It may be intended to not be public

wallet/pskt/src/wasm/signer.rs
```diff
@@ -0,0 +1,29 @@
+// todo: this is a copy/paste from wallet/core/src/wasm
+// i tried to mutualize it in wallet/core/keys, but it conflicted (circular dep with tx_script)
+// it also feels overkill to create a package only for that
+// need guidance on how to procede with architecturing
```
IzioDev, line 4: as highlighted

wallet/pskt/src/wasm/pskt.rs
```diff
@@ -384,4 +393,196 @@ impl PSKT {
...
+        let prefix: Prefix = network_type.try_into()?;
+
+        let private_keys: Vec<PrivateKey> =
+            private_keys.try_into().map_err(|e| Error::Custom(format!("Invalid private keys: {:?}", e)))?;
+
+        let mut key_map: HashMap<Address, PrivateKey> = HashMap::new();
+        for pk in private_keys {
+            // todo: find a better way to convert to address
+            // here we unwrap because we checked network type -> prefix didn't produce error
+            // but this isn't future proof (if internal implementation change)
```
IzioDev: as highlighted

wallet/pskt/src/wasm/pskt.rs
```diff
@@ -384,4 +393,196 @@ impl PSKT {
...
+                    }
+                }
+            }
+        }
+        let signed_pskt: Native<Signer> = inner_pskt.into();
+        self.replace(State::Signer(signed_pskt))
+    }
+}
+
+#[cfg(test)]
```
IzioDev, line 588: first wasm test i write, please carefully review

wallet/pskt/src/wasm/pskt.rs
```diff
@@ -384,4 +396,296 @@ impl PSKT {
...
+        });
+
+        let finalized_pskt = match result {
+            Ok(finalized_pskt) => finalized_pskt,
+            Err(e) => return Err(Error::from(e.to_string())),
+        };
+
+        let mutable_transaction = finalized_pskt
+            .extractor()?
+            .extract_tx(&network_type.into())
```
IzioDev, line 561: This get `&Params` for network type

wallet/pskt/src/wasm/pskt.rs
```diff
@@ -384,4 +396,296 @@ impl PSKT {
...
+            .map_err(|e| Error::custom(format!("Failed to extract transaction: {e}")))?;
+
+        Ok(mutable_transaction.tx.into())
```
IzioDev, line 564: Is there a better interface to expose here, so end-user would still be able to modify priority_fees and other parameters of their choices?

wallet/pskt/src/wasm/pskt.rs
```diff
@@ -384,4 +396,296 @@ impl PSKT {
...
+                            input
+                                .redeem_script
+                                .as_ref()
+                                .map(|redeem_script| ScriptBuilder::new().add_data(redeem_script.as_slice()).unwrap().drain().to_vec())
+                                .unwrap_or_default(),
+                        )
+                        .collect()
+                })
+                .collect())
+        });
```
IzioDev: This has been re-taken from `account/pskb.rs` (`finalize_pskt_one_or_more_sig_and_redeem_script`), not sure what's the best approach here. I feel like i'm re-implementing most of the logic, and this seems wrong

wallet/pskt/src/wasm/pskt.rs
```diff
@@ -384,4 +396,296 @@ impl PSKT {
...
+        };
+
+        let mutable_transaction = finalized_pskt
+            .extractor()?
+            .extract_tx(&network_type.into())
+            .map_err(|e| Error::custom(format!("Failed to extract transaction: {e}")))?;
+
+        Ok(mutable_transaction.tx.into())
+    }
+}
```
biryukovmaxim: the function (or name) is incorrect according to design and purpose. every kaspa script can accept arbitraty arguments. for example multisig expects count and signtures. htlc would expect pre-image and etc. that is why it finalization must be aligned with purpose. here you created a function that according to name is applicable to any script which is wrong.

IzioDev: Summarizing a deeper discussion Maksim and I had:
* (paraphrasing) Here we deal with potentially different script public key kind, and it's wrong to assume they only are p2pk related 
* Maksim suggested that in an ideal world, we would allow end-user to finalize each inputs of a pskt according to their script kind (ex: with two inputs of different kinds: `pskt.finalize([Enum.P2PK, Enum.OtherKind])`. It has been identified this would represent a difficulty when the entry point is a pskb (list of pskt) as the API could be quite confusing (list of list of spk kind)
* We ended up by agreeing that only allowing P2PK finalization and return `Err` in other cases is sufficient in an initial implementation

naming suggestion: `finalize_p2pk`

IzioDev: fixed here: [review(pskt): reject finalization if inputs script class isn't p2pk](https://github.com/kaspanet/rusty-kaspa/pull/770/commits/d9b5e3da8532cb8d2e747837a33b7a452e8850ef)

wallet/pskt/src/wasm/error.rs
```diff
@@ -31,6 +32,9 @@ pub enum Error {
     #[error("PSKT must be initialized with a payload or CREATE role")]
     NotInitialized,
 
+    #[error("Invalid ScriptClass, expected {0}, got {1}")]
+    InvalidScriptClassError(ScriptClass, ScriptClass),
```
biryukovmaxim: Id say its unexpected rather than invalid

wallet/pskt/src/wasm/pskt.rs
```diff
@@ -664,7 +685,7 @@ mod tests {
     }
 
     #[wasm_bindgen_test]
-    fn _test_finalize_and_extract_transaction() {
+    fn _test_finalize_p2pk() {
```
biryukovmaxim, line 688: what is the reason to start it with underscore?

.zed/debug.json
```diff
@@ -0,0 +1,82 @@
...
+    ],
+
+    "cwd": "D:\\Dev\\kaspa\\rusty-kaspa",
+    "env": {
+      "RUST_BACKTRACE": "full",
+      "RUST_LOG": "info"
+    },
+    "sourceLanguages": ["rust"]
+  }
+]
```
biryukovmaxim: im not sure the file should be commited

wallet/pskt/src/wasm/pskt.rs
```diff
@@ -384,4 +396,317 @@ impl PSKT {
...
+                        })
+                    })
+                })
+                .collect::<Result<Vec<_>>>()?;
+
+            Ok(signatures)
+        })?;
+
+        self.replace(State::Signer(signed_pskt))
+    }
```
biryukovmaxim, line 487: 1) utxo can belong to p2sh. it doesnt mean we want to skip signing. user must decide if it want sign or not
2) its possible to sign with ecdsa too
3) you can sign the same input with multiple keys which is useful for multisig like scripts.

only finalizer is responsible to form signature script. its not an issue if it has more signatures than needed.

4) secp256k1 private key can be  used for both ecdsa or schnorr, so caller must decide what signatures it want to pass. and for what input

wallet/pskt/src/wasm/pskt.rs
```diff
@@ -384,4 +396,317 @@ impl PSKT {
...
+                .inputs
+                .iter()
+                .map(|input| {
+                    if input.partial_sigs.len() != 1 {
+                        return Err(format!(
+                            "finalize_p2pk error: input for outpoint {} must have exactly one signature, but has {}",
+                            input.previous_outpoint,
+                            input.partial_sigs.len()
+                        ));
+                    }
```
biryukovmaxim, line 538: im not sure it must be error. you can extract what sig you need and leave others

wallet/pskt/src/wasm/pskt.rs
```diff
@@ -384,4 +396,317 @@ impl PSKT {
...
+            Ok(finalized_pskt) => finalized_pskt,
+            Err(e) => return Err(Error::from(e.to_string())),
+        };
+
+        let mutable_transaction = finalized_pskt
+            .extractor()?
+            .extract_tx(&network_type.into())
+            .map_err(|e| Error::custom(format!("Failed to extract transaction: {e}")))?;
+
+        Ok(mutable_transaction.tx.into())
```
biryukovmaxim, line 564: finalizer can only build tx id and signature scripts.

extractor may perform simulation(by default) and calculates mass, also metable tx has tx, mass and utxo entries. pskt interface returns all of them to caller. if you assume that caller doesnt need them - u can extract and return tx as you r doing now

wallet/pskt/src/pskt.rs
biryukovmaxim: why do you need to expose it. basicaly its private because method of signing modifies pskt passing signature to the right place

wallet/pskt/src/wasm/pskt.rs
```diff
@@ -384,4 +396,317 @@ impl PSKT {
...
+
+        Ok(addresses)
+    }
+
+    /// Sign the PSKT with the provided private keys.
+    /// The method will find the inputs corresponding to the private keys and sign them.
+    /// This method performs partial signing, so if no private keys are provided, it will
+    /// return an unmodified PSKT.
+    #[wasm_bindgen]
+    pub fn sign(&self, private_keys: PrivateKeyArrayT, network_type: &NetworkTypeT) -> Result<PSKT> {
```
biryukovmaxim, line 435: also for multisig or other purposes Input.bip32_derivations and global.xpubs may be checked to define what keys must be used. but i think its unrelevant to sign fn that must allow caller to pass any signature to any input

wallet/pskt/src/wasm/pskt.rs
IzioDev: I'm unsure why, but if i don't set it, i have clippy warnings about unused. I guess it has to do with targets.
If you know a better way, happy to make changes on it

.zed/debug.json
IzioDev: Yes totally a mistake 👍

wallet/pskt/src/wasm/pskt.rs
IzioDev: Ack on 2, 3 and 4.

Could you please detail 1. A bit more, I'm unsure how to proceed with it. Should the function handle all of the ScriptClass possible?

wallet/pskt/src/wasm/pskt.rs
IzioDev: I didn't consider multi sig for this implementation indeed, is it what you meant? That there are cases where we could have more than 1 signatue?

wallet/pskt/src/wasm/pskt.rs
biryukovmaxim: Nah. The caller is who decides what signature it wants to pass. Like for multisig each participant signs the same utxo. Basically if they pass wrong signature finalized will fail. You can either add helper method that only works for p2pk or just assume caller reads pskt/psbt fields and signs with required key. They can look at redeem script and/or utxo. Basically it's passing signature to the map that may be reused by finaliser

wallet/pskt/src/wasm/error.rs
IzioDev: done here: [review: rename wasm unexpected script class error](https://github.com/kaspanet/rusty-kaspa/pull/770/commits/5c2cb93b0a41c7b5e3054ef7a308a2c5d7fa27a9)

.zed/debug.json
IzioDev: done: [review: remove dot zed files](https://github.com/kaspanet/rusty-kaspa/pull/770/commits/e205e4d6f87e23ffea772cf91db2cf7391268a67)

PR #773, opened 2025-12-03
On the fly higher relations

freshair18:
Significant resources on the node are currently devoted to maintaining "higher order" children<->father relations between blocks, within an array of relation stores. 
Notably though, the only true usage of these stores beyond the base level, regards to building a pruning proof - where  forward traversal on the DAG is desired and supplied by these stores. The pruning proof itself however ultimately only contains circa 2000 blocks per level. This PR hence computes these higher level stores on the fly during pruning proof construction, for only a limited set of relevant blocks, and discards these stores as a whole following this.  This parallels a past PR https://github.com/kaspanet/rusty-kaspa/pull/494 on ghostdag stores.

**WIP:**
(a) Currently old stores are still present in the node, just not updated, deleted, or used anywhere. A clean up remains to be done removing them from the codebase, as does a concentrated deletion of them from the datadir in any node updating to a new version.
(b) the transient relations store is currently constructed very naively, with multiple attempts per level until a sufficiently deep root for that level is found. Unlike the case of ghostdag store, much of the data of stores is shared between iterations. A smarter implementation would build upon the store created in the last iteration. Considering this only effects the (relatively) short and uncommon process of creating a pruning proof, this could wait for a subsequent PR.

**Local Performance Initial Impressions (relatively weak node, windows):**
* Pruning of higher relations was previously isolated as a major local bottleneck in the pruning process
* Accordingly, an x10-x11 improvement was observed in pruning speed following this PR, simply by not having to prune higher relations. Numerically, from pruning 1000 blocks every 70 seconds (i.e. not keeping up), to pruning 1000 blocks every 6-7 seconds
* datadir size is about 13GB, which I believe is a decrease.
*  headers IBD appears to be about x3-x5 faster.
* Pruning proof verification section of IBD feels a tad slower, no exact measurements currently. It remains a very short stage regardless.

### DIFF
--- a/components/consensusmanager/src/session.rs
+++ b/components/consensusmanager/src/session.rs
@@ -495,11 +495,11 @@ impl ConsensusSessionOwned {
     pub async fn async_set_pruning_utxoset_stable(&self) {
         self.clone().spawn_blocking(move |c| c.set_pruning_utxoset_stable_flag(true)).await
     }
-    pub async fn async_verify_is_pruning_sample(&self, candidate_hash: Hash) -> ConsensusResult<()> {
-        self.clone().spawn_blocking(move |c| c.verify_is_pruning_sample(candidate_hash)).await
-    }
     pub async fn async_intrusive_pruning_point_update(&self, new_pruning_point: Hash, syncer_sink: Hash) -> ConsensusResult<()> {
         self.clone().spawn_blocking(move |c| c.intrusive_pruning_point_update(new_pruning_point, syncer_sink)).await
+    }
+    pub async fn async_get_n_last_pruning_points(&self, n: usize) -> Vec<Hash> {
+        self.clone().spawn_blocking(move |c| c.get_n_last_pruning_points(n)).await
     }
 }
 
--- a/consensus/core/src/api/mod.rs
+++ b/consensus/core/src/api/mod.rs
@@ -422,11 +422,12 @@ pub trait ConsensusApi: Send + Sync {
         unimplemented!()
     }
 
-    fn verify_is_pruning_sample(&self, candidate_hash: Hash) -> ConsensusResult<()> {
+    fn intrusive_pruning_point_update(&self, new_pruning_point: Hash, syncer_sink: Hash) -> ConsensusResult<()> {
         unimplemented!()
     }
 
-    fn intrusive_pruning_point_update(&self, new_pruning_point: Hash, syncer_sink: Hash) -> ConsensusResult<()> {
+    /// Returns the n most recent pruning points (including the current pruning point)
+    fn get_n_last_pruning_points(&self, n: usize) -> Vec<Hash> {
         unimplemented!()
     }
 }
--- a/consensus/src/consensus/mod.rs
+++ b/consensus/src/consensus/mod.rs
@@ -511,7 +511,25 @@ impl Consensus {
         // The new pruning point P can be "finalized" into consensus if:
         // 1) P satisfies P.blue_score>Nf and selected_parent(P).blue_score<=NF
         // where N is some integer (i.e. it is a valid pruning point based on score)
-        // *this condition is assumed to have already been checked externally and we do not repeat it here*.
+        let Ok(candidate_ghostdag_data) = self.get_ghostdag_data(new_pruning_point) else {
+            return Err(ConsensusError::General(
+                "Catchup cannot be continued since the syncer pruning point could not be confirmed to be a valid pruning point",
+            ));
+        };
+        let Ok(selected_parent_ghostdag_data) = self.get_ghostdag_data(candidate_ghostdag_data.selected_parent) else {
+            return Err(ConsensusError::General(
+                "Catchup cannot be continued since the syncer pruning point could not be confirmed to be a valid pruning point",
+            ));
+        };
+        self.services
+            .pruning_point_manager
+            .is_pruning_sample(
+                candidate_ghostdag_data.blue_score,
+                selected_parent_ghostdag_data.blue_score,
+                self.config.params.finality_depth(),
+            )
+            .then_some(())
+            .ok_or(ConsensusError::General("the alleged pruning point is not a valid pruning point, aborting catchup attempt"))?;
 
         // 2) There are sufficient headers built on top of it, specifically,
         // a header is validated whose blue_score is greater than P.B+p:
@@ -1374,27 +1392,6 @@ impl ConsensusApi for Consensus {
         pruning_meta_write.utxo_set.clear().unwrap();
     }
 
-    fn verify_is_pruning_sample(&self, pruning_candidate: Hash) -> ConsensusResult<()> {
-        if pruning_candidate == self.config.genesis.hash {
-            return Ok(());
-        }
-        let Ok(candidate_ghostdag_data) = self.get_ghostdag_data(pruning_candidate) else {
-            return Err(ConsensusError::General("pruning candidate missing ghostdag data"));
-        };
-        let Ok(selected_parent_ghostdag_data) = self.get_ghostdag_data(candidate_ghostdag_data.selected_parent) else {
-            return Err(ConsensusError::General("pruning candidate selected parent missing ghostdag data"));
-        };
-        self.services
-            .pruning_point_manager
-            .is_pruning_sample(
-                candidate_ghostdag_data.blue_score,
-                selected_parent_ghostdag_data.blue_score,
-                self.config.params.finality_depth(),
-            )
-            .then_some(())
-            .ok_or(ConsensusError::General("pruning candidate is not a pruning sample"))
-    }
-
     /// The usual flow consists of the pruning point naturally updating during pruning, and hence maintains consistency by default
     /// During pruning catchup, we need to manually update the pruning point and
     /// make sure that consensus looks "as if" it has just moved to a new pruning point.
@@ -1426,5 +1423,10 @@ impl ConsensusApi for Consensus {
     fn is_consensus_in_transitional_ibd_state(&self) -> bool {
         let pruning_meta_read = self.pruning_meta_stores.read();
         pruning_meta_read.is_in_transitional_ibd_state()
+    }
+
+    fn get_n_last_pruning_points(&self, n: usize) -> Vec<Hash> {
+        let (_pruning_point, pruning_index) = self.pruning_point_store.read().pruning_point_and_index().unwrap();
+        (0..=pruning_index).rev().take(n).map(|ind| self.past_pruning_points_store.get(ind).unwrap()).collect_vec()
     }
 }
--- a/database/src/errors.rs
+++ b/database/src/errors.rs
@@ -10,7 +10,7 @@ pub enum StoreError {
     #[error("key {0} already exists in store")]
     KeyAlreadyExists(String),
 
-    /// Specialization of key not found for the common `Hash` case.
+    /// Specialization of key already exists for the common `Hash` case.
     /// Added for avoiding the `String` allocation
     #[error("hash {0} already exists in store")]
     HashAlreadyExists(Hash),
--- a/database/src/utils.rs
+++ b/database/src/utils.rs
@@ -49,7 +49,7 @@ pub fn get_kaspa_tempdir() -> TempDir {
     db_tempdir
 }
 
-/// Creates a DB within a temp directory under `<OS SPECIFIC TEMP DIR>/kaspa-rust`
+/// Creates a DB within a temp directory under `<OS SPECIFIC TEMP DIR>/rusty-kaspa`
 /// Callers must keep the `TempDbLifetime` guard for as long as they wish the DB to exist.
 #[macro_export]
 macro_rules! create_temp_db {
--- a/protocol/flows/src/ibd/flow.rs
+++ b/protocol/flows/src/ibd/flow.rs
@@ -59,11 +59,16 @@ impl Flow for IbdFlow {
         self.start_impl().await
     }
 }
-
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum SyncerSkew {
+    Lagging,
+    Aligned,
+    Leading,
+}
 pub enum IbdType {
-    Sync,
+    Sync { highest_known_syncer_chain_hash: Hash, is_utxo_stable: bool, is_pp_anticone_synced: bool },
     DownloadHeadersProof,
-    PruningCatchUp,
+    PruningCatchUp { highest_known_syncer_chain_hash: Hash },
 }
 
 struct QueueChunkOutput {
@@ -115,7 +120,7 @@ impl IbdFlow {
             )
             .await?;
         match ibd_type {
-            IbdType::Sync => {
+            IbdType::Sync { highest_known_syncer_chain_hash, is_utxo_stable, is_pp_anticone_synced } => {
                 let pruning_point = session.async_pruning_point().await;
 
                 info!("syncing ahead from current pruning point");
@@ -126,10 +131,10 @@ impl IbdFlow {
                 // Notice that this is degenerate following sync_with_headers_proof
                 // but not necessarily so after sync_headers -
                 // as it might sync following a previous pruning_catch_up that crashed before this stage concluded
-                if !session.async_is_pruning_point_anticone_fully_synced().await {
+                if !is_pp_anticone_synced {
                     self.sync_missing_trusted_bodies(&session).await?;
                 }
-                if !session.async_is_pruning_utxoset_stable().await
+                if !is_utxo_stable
                 // Utxo might not be available even if the pruning point block data is.
                 // Utxo must be synced before all so the node could function
                 {
@@ -144,7 +149,7 @@ impl IbdFlow {
                 self.sync_headers(
                     &session,
                     negotiation_output.syncer_virtual_selected_parent,
-                    negotiation_output.highest_known_syncer_chain_hash.unwrap(),
+                    highest_known_syncer_chain_hash,
                     &relay_block,
                 )
                 .await?;
@@ -174,9 +179,9 @@ impl IbdFlow {
                     }
                 }
             }
-            IbdType::PruningCatchUp => {
+            IbdType::PruningCatchUp { highest_known_syncer_chain_hash } => {
                 info!("catching up to new pruning point {} ", negotiation_output.syncer_pruning_point);
-                match self.pruning_point_catchup(&session, &negotiation_output, &relay_block).await {
+                match self.pruning_point_catchup(&session, &negotiation_output, &relay_block, highest_known_syncer_chain_hash).await {
                     Ok(()) => {
                         info!("header stage of pruning catchup from peer {} completed", self.router);
                         self.sync_missing_trusted_bodies(&session).await?;
@@ -233,39 +238,39 @@ impl IbdFlow {
             info!("current sink is:{}", sink);
             info!("current pruning point is:{}", pruning_point);
             if consensus.async_is_chain_ancestor_of(pruning_point, highest_known_syncer_chain_hash).await? {
-                if syncer_pruning_point == pruning_point {
-                    // The node is only missing a segment in the future of its current pruning point, and the chains
-                    // agree as well, so we perform a simple sync IBD and only download the missing data
-                    return Ok(IbdType::Sync);
+                let syncer_skew = if syncer_pruning_point == pruning_point {
+                    SyncerSkew::Aligned
+                } else if consensus.async_is_chain_ancestor_of(pruning_point, syncer_pruning_point).await.unwrap_or(false) {
+                    SyncerSkew::Leading
+                } else if consensus.async_get_n_last_pruning_points(4 /*syncer lag tolerance*/).await.contains(&syncer_pruning_point) {
+                    SyncerSkew::Lagging
                 } else {
-                    consensus.async_verify_is_pruning_sample(syncer_pruning_point).await?;
-                    // The node is missing a segment in the near future of its current pruning point, but the syncer is ahead
-                    // and already pruned the current pruning point.
+                    return Err(ProtocolError::Other(
+                        "The syncer purports to have data in the recent future but their pruning point could not be easily recognized",
+                    ));
+                };
+                let is_utxo_stable = consensus.async_is_pruning_utxoset_stable().await;
+                let is_pp_anticone_synced = consensus.async_is_pruning_point_anticone_fully_synced().await;
 
-                    if consensus.async_get_block_status(syncer_pruning_point).await.is_some_and(|b| b.has_block_body())
-                        && !consensus.async_is_consensus_in_transitional_ibd_state().await
-                    {
-                        // The data pruned by the syncer is already available from within the node (from relay or past ibd attempts)
-                        // and the consensus is not in a transitional state requiring data on the previous pruning point,
-                        // hence we can carry on syncing as normal.
-                        return Ok(IbdType::Sync);
-                    } else {
-                        // Two options:
-                        // 1: syncer_pruning_point is in the future, and there is a need to partially resync from syncer_pruning_point
-                        // 2: syncer_pruning_point is in the past of current pruning point, or is unknown on which case the syncing node is flawed,
-                        // and IBD should be stopped
-
-                        if consensus
-                            .async_is_chain_ancestor_of(pruning_point, syncer_pruning_point)
-                            .await
-                            .map_err(|_| ProtocolError::Other("syncer pruning point is corrupted"))?
-                        {
-                            return Ok(IbdType::PruningCatchUp);
+                return match (syncer_skew, is_utxo_stable && is_pp_anticone_synced) {
+                    (SyncerSkew::Aligned, _) => {
+                        Ok(IbdType::Sync { highest_known_syncer_chain_hash, is_utxo_stable, is_pp_anticone_synced })
+                    }
+                    (SyncerSkew::Lagging, true) => {
+                        Ok(IbdType::Sync { highest_known_syncer_chain_hash, is_utxo_stable, is_pp_anticone_synced })
+                    }
+                    (SyncerSkew::Lagging, false) => Err(ProtocolError::Other("Local node is in a transitional state requiring external data to stablize, but the syncer lags behind and is unable to provide said data")),
+                    (SyncerSkew::Leading, true) => {
+                        if consensus.async_get_block_status(syncer_pruning_point).await.is_some_and(|b| b.has_block_body()) {
+                            // While a leading syncer skew often indicates the need for catchup, in this case
+                            // the node is just missing a segment in the future of its current pruning point, that is available to the syncer
+                            Ok(IbdType::Sync { highest_known_syncer_chain_hash, is_utxo_stable, is_pp_anticone_synced })
                         } else {
-                            return Err(ProtocolError::Other("syncer pruning point is outdated"));
+                            Ok(IbdType::PruningCatchUp { highest_known_syncer_chain_hash })
                         }
                     }
-                }
+                    (SyncerSkew::Leading, false) => Ok(IbdType::PruningCatchUp { highest_known_syncer_chain_hash }),
+                };
             }
 
             // If the pruning point is not in the chain of `highest_known_syncer_chain_hash`, it
@@ -311,12 +316,13 @@ impl IbdFlow {
         consensus: &ConsensusProxy,
         negotiation_output: &ChainNegotiationOutput,
         relay_block: &Block,
+        highest_known_syncer_chain_hash: Hash,
     ) -> Result<(), ProtocolError> {
         // Before attempting to update to the syncers pruning point, sync to the latest headers of the syncer,
         // to ensure that  we will locally have sufficient headers on top of  the syncer's pruning point
         let syncer_pp = negotiation_output.syncer_pruning_point;
         let syncer_sink = negotiation_output.syncer_virtual_selected_parent;
-        self.sync_headers(consensus, syncer_sink, negotiation_output.highest_known_syncer_chain_hash.unwrap(), relay_block).await?;
+        self.sync_headers(consensus, syncer_sink, highest_known_syncer_chain_hash, relay_block).await?;
 
         // This function's main effect is to confirm the syncer's pruning point can be finalized into the consensus, and to update
         // all the relevant stores


consensus/src/processes/pruning_proof/apply.rs
coderofstuff: I think you only need to do this now for level = 0

so no outer for loop on all levels anymore and instead just look at level 0 directly to fill the relations there.

freshair18:
Finished (a).

(b) is probably best differed for a subsequent PR.

Wondering about keep_relations during pruning - I believe the logic as a whole remains necessary- but I'm not sure if the name keep_relations is really descriptive for what it does anymore.

freshair18:
Note: there appears to currently at times be a deadlock of sorts, which results in Block relay ceasing activity. This sporadically happens past an IBD when pruning also needs to occur, but the exact conditions to replicate remain unclear at the moment. Do not merge yet until found and addressed

consensus/src/pipeline/header_processor/processor.rs
```diff
@@ -313,28 +308,24 @@ impl HeaderProcessor {
...
+    fn collect_known_direct_parents(&self, header: &Header) -> BlockHashes {
+        let relations_read = self.relations_store.read();
+        Arc::new(
+            self.parents_manager
+                        .parents_at_level(header,0)
                         .iter()
                         .copied()
-                        .filter(|parent| relations_read[level as usize].has(*parent).unwrap())
+                        // TODO(relaxed): revisit a more transparent filtering condition
+                        .filter(|&parent| relations_read.has(parent).unwrap_option().is_some())
```
coderofstuff: Why is this check an `is_some()`? Wouldn't `.has(parent)` can possibly return `Some(true)` or `Some(false)`

consensus/src/processes/pruning_proof/validate.rs
```diff
@@ -145,23 +144,23 @@ impl PruningProofManager {
...
-                }
-                None => {
-                    // If the current pruning point doesn't have a parent at this level, we consider the proof state to be better.
-                    return Ok(());
-                }
+            let parents_at_level = self.parents_manager.parents_at_level(&current_pp_header, level);
+            if parents_at_level
+                .iter()
+                .copied()
+                .filter_map(|parent| current_consensus_ghostdag_stores[level_idx].get_blue_score(parent).unwrap_option())
```
coderofstuff: The filtering applied here needs to be applied during the initialization at line L147 because the `parents_at_level.is_empty()` check in L157 depends on it.

So make sure that L147 is actually the known parents at level.

consensus/src/processes/pruning_proof/build.rs
```diff
@@ -69,7 +82,16 @@ impl PruningProofManager {
 
         let (_db_lifetime, temp_db) = kaspa_database::create_temp_db!(ConnBuilder::default().with_files_limit(10));
         let pp_header = self.headers_store.get_header_with_block_level(pp).unwrap();
-        let (ghostdag_stores, selected_tip_by_level, roots_by_level) = self.calc_gd_for_all_levels(&pp_header, temp_db);
+        let (transient_ghostdag_stores, transient_relations_stores, selected_tip_by_level, roots_by_level) = self
+            .calc_ghostdag_and_relations_for_all_levels(&pp_header, temp_db)
+            .into_iter()
+            .fold((vec![], vec![], vec![], vec![]), |(mut a, mut b, mut c, mut d), (x0, x1, x2, x3)| {
```
coderofstuff: What is this for?

consensus/src/processes/pruning_proof/build.rs
```diff
@@ -173,42 +196,83 @@ impl PruningProofManager {
             .collect_vec()
     }
 
-    fn calc_gd_for_all_levels(
-        &self,
-        pp_header: &HeaderWithBlockLevel,
-        temp_db: Arc<DB>,
-    ) -> (Vec<Arc<DbGhostdagStore>>, Vec<Hash>, Vec<Hash>) {
+    fn calc_ghostdag_and_relations_for_all_levels(&self, pp_header: &HeaderWithBlockLevel, temp_db: Arc<DB>) -> Vec<LevelProofStores> {
```
coderofstuff: Suggestion: Rename to `calc_all_level_proof_stores`

consensus/src/processes/pruning_proof/build.rs
```diff
@@ -285,27 +348,30 @@ impl PruningProofManager {
                 common_ancestor.hash
             };
 
-            if level == 0 {
-                return Ok((self.ghostdag_store.clone(), selected_tip, root));
-            }
+            // Step 3 - Fill the ghostdag data from root to tip.
+            //
+            // First, derive the relevant relations down to the candidate root.
+            let transient_relation_store = self.populate_relation_store_at_level(temp_db.clone(), selected_tip, root, level, tries);
```
coderofstuff: I suppose the reason this is here is because it needs to reference the `root` and `selected_tip` as an argument. But what if:

- You move this all the way to the top (maybe even outside this function for all levels)
- Replace selected tip with `pp_header.header.hash` regardless of level
  - If level > pp.level - the would-be SP is guaranteed to be an existing parent of pp at that level
  - If level <= pp.level - the SP is pp itself
- Instead of requiring `root`, keep traversing until you encounter ORIGIN - which you will always encounter

This way, you can use this `relations_store` in anything that depended on the high level store before, such as `find_approximate_selected_parent_header_at_level`

consensus/src/pipeline/header_processor/processor.rs
```diff
@@ -313,28 +308,24 @@ impl HeaderProcessor {
...
+    fn collect_known_direct_parents(&self, header: &Header) -> BlockHashes {
+        let relations_read = self.relations_store.read();
+        Arc::new(
+            self.parents_manager
+                        .parents_at_level(header,0)
                         .iter()
                         .copied()
-                        .filter(|parent| relations_read[level as usize].has(*parent).unwrap())
+                        // TODO(relaxed): revisit a more transparent filtering condition
+                        .filter(|&parent| relations_read.has(parent).unwrap_option().unwrap_or(false))
```
coderofstuff: This conversion from `.unwrap()` to `.unwrap_option().unwrap_or(false)` changes the semantics of this such in case there is a store error, this will silently continue with `false` (when it would have been possible for that value to be `true` if there was no such store error).

consensus/src/pipeline/header_processor/processor.rs
```diff
@@ -313,28 +308,24 @@ impl HeaderProcessor {
...
-                        .parents_at_level(header, level)
+    fn collect_known_direct_parents(&self, header: &Header) -> BlockHashes {
+        let relations_read = self.relations_store.read();
+        Arc::new(
+            self.parents_manager
+                        .parents_at_level(header,0)
                         .iter()
                         .copied()
-                        .filter(|parent| relations_read[level as usize].has(*parent).unwrap())
+                        // TODO(relaxed): revisit a more transparent filtering condition
```
coderofstuff: What kind of "more transparent filtering condition" are you referring to here?

(Thinking if this TODO is necessary)

consensus/src/processes/pruning_proof/build.rs
freshair18: its some destructuring from a vector of tuples to a tuple of vectors, I agree it is ugly, do suggest a better syntax

consensus/src/pipeline/header_processor/processor.rs
freshair18: This is Plato's cave - relations data as a filtering condition is a shadow cast by the real condition that logically needs to be considered here. In high likelihood this condition is either of the following:
a) existence of reachability data
b) existence of ghostdag data
c) perhaps even just existence of header in this case, though we've been burned before

The first two are currently somewhat correlated with the existence of relations data, since they all are pruned together or kept in (the badly named) "keep relations" flow.  For safety we should currently keep changes to a minimum, but it is nonsensical to treat the existence or inexistence of relations as the reason we filter out certain parents - relations store is currently  mostly a legacy in the codebase.

consensus/src/processes/pruning_proof/build.rs
freshair18: Do you mean BFS backwards until origin? this will be problematic for archivals if we don't stop somewhere.

consensus/src/pipeline/header_processor/processor.rs
freshair18: addressed

consensus/src/processes/pruning_proof/build.rs
freshair18: addressed

consensus/src/processes/pruning_proof/validate.rs
freshair18: addressed

consensus/src/pipeline/header_processor/processor.rs
freshair18: addressed

consensus/src/processes/pruning_proof/build.rs
coderofstuff: EDIT: need to think deeper, since the below will loop. Leaving it here anyway for reference.

---

Good point. The main reason that `root` is an input here is so that there is a stopping point for the backward BFS. Instead of ORIGIN all the time, there is a way for us to find a reasonable root to end with:

1. Get Current Proof (current_proof) - either the node has an existing proof (likely in cache) or it does not (it's the first time it will build)
2. From the current_proof, we use the oldest element in this proof (`current_proof[level][0]`) as the `root`

Essentially:

```
let current_proof = pruning_manager.get_pruning_point_proof(); // EDIT: Can't actually use this here since it get_pruning_point_proof calls build so infinite loop. Need to think.
// In an empty node, this will be genesis. In a regular node, it'd be the deepest hash known in that level
let root = current_proof[level].get(0).unwrap_or(ORIGIN);

let level_relations_store = self.populate_relation_store_at_level(temp_db.clone(), pp.header.hash, root, level, 0);

// Then use level_relations_store wherever it was used before, like in `find_approximate_selected_parent_header_at_level`
```

freshair18:
> Upgrading from an older DB versions asks to confirm twice. The user should only have to confirm an upgrade once.
> 
> ```
> NOTE: Node database is from an older version. Proceeding with the upgrade is instant and safe.
> However, downgrading to an older node version later will require deleting the database.
> Do you confirm? (y/n)
> y
> Node database currently at or below version 5. Upgrade process to version 6 needs to be applied. Continue? (y/n)
> y
> ```

While I agree it is probably better to ask once, I want to note that different upgrades do different things and have different implications, a hyper informed user might care that the latter upgrade fundementally changes the existing datadir while the former only effects future blocks

consensus/src/processes/pruning_proof/build.rs
coderofstuff: try this patch [fold-diff.patch](https://github.com/user-attachments/files/24332795/fold-diff.patch)

consensus/src/processes/pruning_proof/build.rs
```diff
@@ -173,42 +198,99 @@ impl PruningProofManager {
...
+                continue;
+            }
+            let header = self.headers_store.get_header(hash).unwrap();
+            let parents = Arc::new(
+                self.parents_manager
+                    .parents_at_level(&header, level)
+                    .iter()
+                    .copied()
+                    .filter(|&p| self.headers_store.has(p).unwrap())
+                    .filter(|&p| self.reachability_service.is_dag_ancestor_of_result(root, p).unwrap_or(false))
```
coderofstuff: I think this condition alone is sufficient and the header filter above can be removed

consensus/src/processes/pruning_proof/build.rs
```diff
@@ -173,42 +198,99 @@ impl PruningProofManager {
...
+        let level_relation_store =
+            Arc::new(RwLock::new(DbRelationsStore::new_temp(temp_db.clone(), level, try_number, cache_policy, cache_policy)));
+        let mut relations_write = level_relation_store.write();
+
+        relations_write.insert(ORIGIN, BlockHashes::new(vec![])).unwrap();
+
+        // sanity check
+        assert!(self.reachability_service.is_dag_ancestor_of(root, tip));
+        while let Some(hash) = queue.pop_front() {
+            //TODO(relaxed): consider tightening to is_dag_ancestor_of
```
coderofstuff: is this todo a remnant of earlier impls?

consensus/src/processes/pruning_proof/build.rs
coderofstuff: no longer relevant for now

consensus/src/processes/pruning_proof/build.rs
freshair18: addressed

consensus/src/processes/pruning_proof/build.rs
freshair18: addressed

consensus/src/processes/pruning_proof/build.rs
freshair18: addressed

consensus/src/pipeline/header_processor/processor.rs
```diff
@@ -313,28 +308,24 @@ impl HeaderProcessor {
...
-        (0..=block_level)
-            .map(|level| {
-                Arc::new(
-                    self.parents_manager
-                        .parents_at_level(header, level)
+    fn collect_known_direct_parents(&self, header: &Header) -> BlockHashes {
+        let relations_read = self.relations_store.read();
+        Arc::new(
+            self.parents_manager
+                        .parents_at_level(header,0)
```
coderofstuff:
```suggestion
                        .parents_at_level(header, 0)
```

kaspad/src/daemon.rs
```diff
@@ -366,31 +375,74 @@ do you confirm? (answer y/n or pass --yes to the Kaspad command line to confirm
...
+                        DatabaseStorePrefixes::RelationsChildren.into_iter().chain(end_level_bytes).collect();
+
+                    // Apply delete of range from level 1 to max (+1) for Ghostdag and GhostdagCompact:
+                    writer.delete_range(start_parents_prefix_vec.clone(), end_parents_prefix_vec.clone()).unwrap();
+                    writer.delete_range(start_children_prefix_vec.clone(), end_children_prefix_vec.clone()).unwrap();
+
+                    //  update the version to one higher:
+                    mcms.set_version(6).unwrap();
+                    info!("deprecated stores have been removed from database, storage will be gradually cleared in due time.");
+                    info!("database is now in version 6");
```
coderofstuff:
```suggestion
                    info!("Deprecated stores have been removed from database, storage will be gradually cleared in due time.");
                    info!("Database is now in version 6");
```

kaspad/src/daemon.rs
```diff
@@ -366,31 +375,74 @@ do you confirm? (answer y/n or pass --yes to the Kaspad command line to confirm
...
-        // If I'm at version 3 and latest version is 7, I need to be able to upgrade to that version following the intermediate
-        // steps without having to delete the DB
-        if version == 4 {
-            let msg = "NOTE: Node database is from an older version. Proceeding with the upgrade is instant and safe.
+        if version <= 3 {
+            is_db_reset_needed = request_database_deletion_approval(args.yes);
+            continue 'db_upgrade;
+        }
+
+        let msg = "NOTE: Node database is from an older version. Proceeding with the upgrade is instant and safe.
```
coderofstuff, line 390: In the GD upgrade, we would give the user info on how many records would be deleted. Do you think that's useful here too?

consensus/src/pipeline/header_processor/processor.rs
coderofstuff: Ok, if this comment is going to remain here please make sure to follow up on it and address it in a future PR maybe track it in a GH issue. Or remove it.

consensus/src/processes/pruning_proof/build.rs
```diff
@@ -173,42 +198,100 @@ impl PruningProofManager {
...
+            );
+        }
+
+        let (transient_ghostdag_stores, transient_relations_stores, selected_tip_by_level, roots_by_level) =
+            level_proof_stores_vec.into_iter().map(Option::unwrap).multiunzip();
+
+        MultiLevelProofContext { transient_ghostdag_stores, transient_relations_stores, selected_tip_by_level, roots_by_level }
+    }
+
+    fn populate_relation_store_at_level(
```
coderofstuff, line 238: Add function docs to describe what this function does.

consensus/src/processes/pruning_proof/build.rs
```diff
@@ -427,26 +513,27 @@ impl PruningProofManager {
...
+        // Parents manager parents_at_level may return parents that aren't currently in database and those are filtered out.
+        // This is ok because this function is called in the context of deriving a root deep enough for a proof at this level,
+        // not to find the "best" such proof
         let sp = self
             .parents_manager
             .parents_at_level(header, level)
             .iter()
             .copied()
-            .filter(|p| self.level_relations_services[level as usize].has(*p).unwrap())
+            .filter(|&p| self.reachability_service.has_reachability_data(p))
```
coderofstuff, line 536: Please add a comment to this line to remind us that it's very important. It's how we filter for "known" parents at a level.

consensus/src/processes/pruning_proof/build.rs
```diff
@@ -173,42 +198,100 @@ impl PruningProofManager {
...
+        try_number: u8,
+    ) -> Arc<RwLock<DbRelationsStore>> {
+        //TODO(relaxed): currently we rebuild the relation store in its entirety for each try.
+        // A more efficient implementation could instead extend the store constructed in the previous try
+        //TODO: Remove sanity checks when convinced code is stable and correct
+        let mut queue = VecDeque::new();
+        let mut visited = BlockHashSet::new();
+        queue.push_back(tip);
+        let cache_policy = CachePolicy::Count(2 * self.pruning_proof_m as usize);
+        let level_relation_store =
```
coderofstuff: I'm curious, is the Arc<RwLock<>> necessary here or could this have returned just the `DbRelationsStore` and have the caller be the one to create the Arc<RwLock<>> around it?

kaspad/src/daemon.rs
freshair18: The problem is archival nodes.
According to a tester, the scanning alone is 5+ hours on archivals.

What I'm deliberating on is whether or not we should compact on the spot - the problem is that compacting on the spot takes several hours for archivals and boarders on them losing sync. However the current situation results in sudden hours long compacting at random time, without any notice to the user. The tester said they prefer it that way, but I'm not certain that I agree.

consensus/src/pipeline/header_processor/processor.rs
freshair18: I have no trouble opening an issue but feels a bit of an overkill - i feel todo fits this really.

consensus/src/pipeline/header_processor/pre_pow_validation.rs
```diff
@@ -12,7 +12,7 @@ impl HeaderProcessor {
     }
 
     fn check_pruning_violation(&self, ctx: &HeaderProcessingContext) -> BlockProcessResult<()> {
-        let known_parents = ctx.direct_known_parents();
+        let known_parents = ctx.known_direct_parents.clone();
```
michaelsutton:
```suggestion
        let known_parents = ctx.known_direct_parents.as_slice();
```
`clone` clones the Arc which is not needed here

consensus/src/pipeline/header_processor/processor.rs
```diff
@@ -313,28 +308,24 @@ impl HeaderProcessor {
...
-        (0..=block_level)
-            .map(|level| {
-                Arc::new(
-                    self.parents_manager
-                        .parents_at_level(header, level)
+    fn collect_known_direct_parents(&self, header: &Header) -> BlockHashes {
+        let relations_read = self.relations_store.read();
+        Arc::new(
+            self.parents_manager
+                        .parents_at_level(header, 0)
```
michaelsutton:
```suggestion
            header
                .direct_parents()
```
explanation: `parents_manager` was only required in the case of higher levels where we fallback to genesis if no such level ancestor exists (ie extremely high levels)

consensus/src/pipeline/header_processor/processor.rs
michaelsutton: Terminology: 

Let B be the set of blocks with block body status. 
Let R be the set of blocks with relations entry. 
Let C be the set of blocks with reachability entry. 
Let H be the set of blocks with header entry. 

pp = pruning point
cut(pp) = { pp } ∪ [anticone(pp) ∩ past(virtual)]

Conceptually we know that (up to noise during prune):
B = past(virtual) /setminus past(pp)
R = B ∪ DAA windows of cut(pp) ∪ proof level 0
C = R ∪ all proof levels
H = C ∪ past pruning points

Hence B ⊆ R ⊆ C ⊆ H

In words, R is the consecutive *level 0* DAG area we hold. Mainly consisting of future(pp) but crucially it also holds a consecutive area of recent blocks (~1 hour) below the pruning point.  

All this to say, that the main reason for filtering parents here is because we are going to insert this block into the relations sore (=R) with the filtered parents as relations, and therefor we want to verify that each of these parents is in R as well.    

Hence this is the exact and correct condition to check for here.

consensus/src/pipeline/header_processor/processor.rs
```diff
@@ -399,16 +390,12 @@ impl HeaderProcessor {
...
-        ctx.known_parents.into_iter().enumerate().for_each(|(level, parents_by_level)| {
-            relations_write[level].insert_batch(&mut batch, header.hash, parents_by_level).unwrap();
-        });
+        let mut relations_write = self.relations_store.write();
+        relations_write.insert_batch(&mut batch, ctx.hash, ctx.known_direct_parents.clone()).unwrap();
 
         // Write reachability relations. These relations are only needed during header pruning
         let mut reachability_relations_write = self.reachability_relations_store.write();
-        reachability_relations_write.insert_batch(&mut batch, ctx.hash, reachability_parents).unwrap();
+        reachability_relations_write.insert_batch(&mut batch, ctx.hash, ctx.known_direct_parents.clone()).unwrap();
```
michaelsutton:
```suggestion
        reachability_relations_write.insert_batch(&mut batch, ctx.hash, ctx.known_direct_parents).unwrap();
```
no need for the clone here (you can consume the field from the ctx since this is the last usage)

consensus/src/pipeline/header_processor/processor.rs
```diff
@@ -437,11 +424,8 @@ impl HeaderProcessor {
         // This data might have been already written when applying the pruning proof.
         self.ghostdag_store.insert_batch(&mut batch, ctx.hash, ghostdag_data).unwrap_or_exists();
 
-        let mut relations_write = self.relations_stores.write();
-        ctx.known_parents.into_iter().enumerate().for_each(|(level, parents_by_level)| {
-            // This data might have been already written when applying the pruning proof.
-            relations_write[level].insert_batch(&mut batch, ctx.hash, parents_by_level).unwrap_or_exists();
-        });
+        let mut relations_write = self.relations_store.write();
+        relations_write.insert_batch(&mut batch, ctx.hash, ctx.known_direct_parents.to_vec().into()).unwrap_or_exists();
```
michaelsutton:
```suggestion
        relations_write.insert_batch(&mut batch, ctx.hash, ctx.known_direct_parents).unwrap_or_exists();
```

consensus/src/pipeline/pruning_processor/processor.rs
```diff
@@ -498,16 +498,12 @@ impl PruningProcessor {
...
-                    // This preserves the semantic that for each level, relations represent a contiguous DAG area in that level
-                    for lower_level in 0..affiliated_proof_level as usize {
-                        let mut staging_level_relations = StagingRelationsStore::new(&mut level_relations_write[lower_level]);
-                        relations::delete_level_relations(MemoryWriter, &mut staging_level_relations, current).unwrap_option();
-                        staging_level_relations.commit(&mut batch).unwrap();
-
-                        if lower_level == 0 {
-                            self.ghostdag_store.delete_batch(&mut batch, current).unwrap_option();
-                        }
+                    // delete level 0 relations and ghostdag only unless those are in the 0 level of the pruning proof
```
michaelsutton: the comment can be more clear. maybe drop the word "only"?

consensus/src/processes/pruning_proof/build.rs
```diff
@@ -173,42 +198,100 @@ impl PruningProofManager {
...
+            if hash == ORIGIN {
+                continue;
+            }
+            let header = self.headers_store.get_header(hash).unwrap();
+            let parents = Arc::new(
+                self.parents_manager
+                    .parents_at_level(&header, level)
+                    .iter()
+                    .copied()
+                    .filter(|&p| self.reachability_service.is_dag_ancestor_of_result(root, p).unwrap_or(false))
```
michaelsutton:
```suggestion
                    .filter(|&p| self.reachability_service.is_dag_ancestor_of_result(root, p).unwrap_option().unwrap_or(false))
```
The current `unwrap_or` swallows all DB/IO errors as well, whereas you only want to ignore a missing entry and treat it as false

michaelsutton (changes requested):
still reviewing the pruning proof parts, but this review covers everything else

consensus/src/processes/pruning_proof/mod.rs
```diff
@@ -360,8 +354,9 @@ impl PruningProofManager {
                 let ghostdag = (&*self.ghostdag_store.get_data(current).unwrap()).into();
                 e.insert(TrustedHeader { header, ghostdag });
             }
-            let parents = self.relations_stores.read()[0].get_parents(current).unwrap();
-            for parent in parents.iter().copied() {
+            // TODO(relaxed): revisit the need for relations store as opposed to a more precise (if any) filtering of the direct parents
```
michaelsutton: from my pov there's nothing to revisit here (see comment re R above). If you think a comment explaining why R is used makes sense here, add the comment, but not a todo

consensus/src/processes/pruning_proof/validate.rs
```diff
@@ -145,23 +144,24 @@ impl PruningProofManager {
...
+                .collect();
+            if parents_at_level.iter().copied().any(|parent_bscore| parent_bscore < 2 * self.pruning_proof_m) {
+                return Ok(());
+            }
+
+            if parents_at_level.is_empty() {
+                // If the current pruning point doesn't have a parent at this level, we consider the proof state to be better.
+                return Ok(());
```
michaelsutton:
```suggestion
            if self
                .parents_manager
                .parents_at_level(&current_pp_header, level)
                .iter()
                .copied()
                .filter_map(|parent| current_consensus_ghostdag_stores[level_idx].get_blue_score(parent).unwrap_option())
                .all(|parent_bscore| parent_bscore < 2 * self.pruning_proof_m)
            {
                return Ok(());
```

The original logic should have been `.all(..)` rather than `.any(..)`, because we should prefer current consensus even if only one of pp parents at level has `blue score >= 2M` (think of it as the selected tip at level for current consensus).  

`all` returns `true` for an empty iterator, so the above suggestion is cleaner (and slightly more efficient) 

cc @someone235

consensus/src/pipeline/header_processor/processor.rs
freshair18: As discussed, we should think of some suitable place to include this context

consensus/src/processes/pruning_proof/build.rs
```diff
@@ -38,7 +53,15 @@ struct RelationsStoreInFutureOfRoot<T: RelationsStoreReader, U: ReachabilityServ
...
     fn get_parents(&self, hash: Hash) -> Result<BlockHashes, kaspa_database::prelude::StoreError> {
         self.relations_store.get_parents(hash).map(|hashes| {
-            Arc::new(hashes.iter().copied().filter(|h| self.reachability_service.is_dag_ancestor_of(self.root, *h)).collect_vec())
+            Arc::new(
+                hashes
+                    .iter()
+                    .copied()
+                    // TODO(relaxed): consider removing this filtering altogether - in the context this is currently called
+                    // blocks should not be inserted if they do not fulfill this condition
+                    .filter(|h| self.reachability_service.is_dag_ancestor_of_result(self.root, *h).unwrap_or(false))
```
michaelsutton: add `unwrap_option`

consensus/src/processes/pruning_proof/build.rs
```diff
@@ -335,33 +420,34 @@ impl PruningProofManager {
...
+        transient_ghostdag_store: &Arc<DbGhostdagStore>,
         required_block: Option<Hash>,
         level: BlockLevel,
+        transient_relations_store: &Arc<RwLock<DbRelationsStore>>,
         ghostdag_k: KType,
     ) -> bool {
-        let relations_service = RelationsStoreInFutureOfRoot {
-            relations_store: self.level_relations_services[level as usize].clone(),
+        let transient_relations_service = RelationsStoreInFutureOfRoot {
+            relations_store: transient_relations_store.read().clone(),
```
michaelsutton: this line should be 
```
relations_store: MTRelationsService::new(transient_relations_store.clone())
```

cloning the inner store is not the way to go (effectively you are deep-cloning it in a way that it can be used without the lock)

michaelsutton (commented):
yet another batch of comments

consensus/src/processes/pruning_proof/validate.rs
michaelsutton: Actually I have a better suggestion. In L69 replace `_` with `current_consensus_selected_tip_by_level`

and then edit the above condition to 

```
            if current_consensus_ghostdag_stores[level_idx].get_blue_score(current_consensus_selected_tip_by_level[level_idx]).unwrap()
                < 2 * self.pruning_proof_m
            {
                return Ok(());
            }
```

someone235: lgtm

consensus/src/processes/pruning_proof/build.rs
```diff
@@ -173,42 +198,100 @@ impl PruningProofManager {
...
-        let mut ghostdag_stores: Vec<Option<Arc<DbGhostdagStore>>> = vec![None; self.max_block_level as usize + 1];
-        let mut selected_tip_by_level = vec![None; self.max_block_level as usize + 1];
-        let mut root_by_level = vec![None; self.max_block_level as usize + 1];
+        let mut level_proof_stores_vec: Vec<Option<LevelProofContext>> = vec![None; (self.max_block_level + 1).into()];
         for level in (0..=self.max_block_level).rev() {
             let level_usize = level as usize;
             let required_block = if level != self.max_block_level {
-                let next_level_store = ghostdag_stores[level_usize + 1].as_ref().unwrap().clone();
+                let (next_level_gd_store, _relation_store_at_next_level, selected_tip_at_next_level, _root_at_next_level) =
+                    level_proof_stores_vec[level_usize + 1].as_ref().unwrap().clone();
```
michaelsutton: remove the `.clone()` and change below to `.block_at_depth(next_level_gd_store.as_ref(), *selected_tip_at_next_level`

kaspad/src/daemon.rs
coderofstuff: You could skip the scan for archivals.

But also, it looks like the DB upgrade process for archivals is bigger than this. A future-proof approach for any DB cleanup is to do such cleanup on a different thread (like a cleanup worker).

Something like:
1. Ask the user if they want to upgrade
2. If yes, thread spawns to apply the delete. Presumably this doesn't affect anything else since the records being cleaned up are deprecated.
3. Worker thread deletes the records over time, without blocking anything

freshair18: >You could skip the scan for archivals.

I could, but feels weird given that they are the ones it is probably most relevant to.

Regarding different threads for the deletion -  I can't read between the lines, do you want me to try something like that? Maybe it will be better for tracking purposes, but I doubt it doesn't effect performance - or it wouldn't have slowed down the tester's node (I assume rocksDB built in attempts this in different threads.)

Second, noting if node restarts while this hasn't finished, then it just revert to stealth compacting next time it starts - unless you mean we  become stateful over the state of the cleanup

consensus/src/processes/pruning_proof/build.rs
freshair18: see now

consensus/src/processes/pruning_proof/build.rs
freshair18: see now

consensus/src/pipeline/pruning_processor/processor.rs
freshair18: see now

consensus/src/processes/pruning_proof/build.rs
freshair18: see now

consensus/src/processes/pruning_proof/build.rs
freshair18: addressed

consensus/src/processes/pruning_proof/build.rs
freshair18: addressed

consensus/src/processes/pruning_proof/validate.rs
freshair18: addressed

consensus/src/processes/pruning_proof/build.rs
freshair18: Seems like that's possible yes, I don't think its very important but I suppose it simplifies a bit. You can have a look.

consensus/src/processes/pruning_proof/mod.rs
freshair18: addressed

kaspad/src/daemon.rs
coderofstuff: Nvm, the scanning here is the slow part. The delete itself appears to be fast. You can leave it as it is. The value that the scanning provides is marginal.

coderofstuff:
Before this is merged, please update the title to remove the (WIP)

kaspad/src/daemon.rs
```diff
@@ -366,31 +375,74 @@ do you confirm? (answer y/n or pass --yes to the Kaspad command line to confirm
...
+                        DatabaseStorePrefixes::RelationsParents.into_iter().chain(start_level_bytes).collect();
+                    let end_parents_prefix_vec: Vec<_> =
+                        DatabaseStorePrefixes::RelationsParents.into_iter().chain(end_level_bytes).collect();
+
+                    let start_children_prefix_vec: Vec<_> =
+                        DatabaseStorePrefixes::RelationsChildren.into_iter().chain(start_level_bytes).collect();
+                    let end_children_prefix_vec: Vec<_> =
+                        DatabaseStorePrefixes::RelationsChildren.into_iter().chain(end_level_bytes).collect();
+
+                    // Apply delete of range from level 1 to max (+1) for Ghostdag and GhostdagCompact:
```
coderofstuff:
```suggestion
                    // Apply delete of range from level 1 to max (+1) for RelationsParents and RelationsChildren:
```

michaelsutton (approved):

PR #774, opened 2025-12-06
Tx script engine cleanup

biryukovmaxim:


testing/integration/src/consensus_integration_tests.rs
```diff
@@ -1788,28 +1787,30 @@ async fn staging_consensus_test() {
...
-/// Uses OpInputSpk opcode as an example
+/// Tests the KIP-10 transaction introspection opcodes by verifying that:
+/// 1. Transactions using these opcodes are accepted from genesis (DAA score 0)
+/// 2. The introspection opcodes (like OpInputSpk) function correctly in transaction validation
+///
+/// KIP-10 is now enabled by default from the genesis block, allowing scripts to access
+/// transaction data through introspection opcodes for advanced smart contract capabilities.
 #[tokio::test]
 async fn run_kip10_activation_test() {
     use kaspa_consensus_core::subnets::SUBNETWORK_ID_NATIVE;
```
IzioDev, line 1798: Does it still make sense to keep this test now that KIP-10 is assumed activated in implementation?

If there are specs we'd like to ensure, should we instead create dedicated test case for it? (not related to activation of KIP-10: most likely KIP-10 features)

testing/integration/src/consensus_integration_tests.rs
```diff
@@ -2019,15 +1993,28 @@ async fn payload_for_native_tx_test() {
...
+/// multiple signature opcodes to pass validation if only a subset are actually executed
+/// (e.g., in if-branches), as opposed to static counting which counts all sig ops regardless
+/// of execution path.
+///
+/// Test scenario: A P2SH script with sig_op_count=1 contains 3 CheckSig opcodes, but only 1
+/// is executed due to conditional logic. With runtime counting enabled from genesis, this
+/// transaction is accepted because only 1 sig op is actually executed, even though static
+/// analysis would see 3 sig ops.
 #[tokio::test]
 async fn runtime_sig_op_counting_test() {
```
IzioDev, line 2007: Similar comment to above, is `RUNTIME_SIGOP_ACTIVATION_DAA_SCORE` still relevant?

crypto/txscript/src/opcodes/mod.rs
```diff
@@ -3349,56 +3310,38 @@ mod test {
...
-                            // Test input count
-                            op_input_count.execute(&mut vm).unwrap();
-                            assert_eq!(
-                                vm.dstack,
-                                vec![<Vec<u8> as OpcodeData<i64>>::serialize(&(input_count as i64)).unwrap()],
-                                "Input count mismatch for {} inputs",
-                                input_count
-                            );
-                            vm.dstack.clear();
+                // Test with KIP-10 enabled and disabled
```
someone235: Remove comment

crypto/txscript/src/lib.rs
```diff
@@ -1520,19 +1484,16 @@ mod bitcoind_tests {
...
-                let reader = BufReader::new(file);
-
-                // Read the JSON contents of the file as an instance of `User`.
-                let tests: Vec<JsonTestRow> = serde_json::from_reader(reader).expect("Failed Parsing {:?}");
-                for row in tests {
-                    if let Err(error) = row.test_row(kip10_enabled, runtime_sig_op_counting) {
-                        panic!("Test: {:?} failed for {}: {:?}", row.clone(), file_name, error);
-                    }
-                }
+        let file_name = "script_tests-kip10.json";
```
someone235: Maybe rename to `script_tests.json`?

testing/integration/src/consensus_integration_tests.rs
```diff
@@ -1788,28 +1787,30 @@ async fn staging_consensus_test() {
...
 async fn run_kip10_activation_test() {
     use kaspa_consensus_core::subnets::SUBNETWORK_ID_NATIVE;
     use kaspa_txscript::opcodes::codes::{Op0, OpTxInputSpk};
     use kaspa_txscript::pay_to_script_hash_script;
     use kaspa_txscript::script_builder::ScriptBuilder;
 
-    // KIP-10 activates at DAA score 3 in this test
-    const KIP10_ACTIVATION_DAA_SCORE: u64 = 3;
+    // KIP-10 is enabled from genesis (DAA score 0)
+    const KIP10_ACTIVATION_DAA_SCORE: u64 = 0;
```
someone235: This const looks redundant

testing/integration/src/consensus_integration_tests.rs
```diff
@@ -2019,15 +1993,28 @@ async fn payload_for_native_tx_test() {
...
 async fn runtime_sig_op_counting_test() {
     use kaspa_consensus_core::{
         hashing::sighash::SigHashReusedValuesUnsync, hashing::sighash_type::SIG_HASH_ALL, subnets::SUBNETWORK_ID_NATIVE,
     };
     use kaspa_txscript::{opcodes::codes::*, script_builder::ScriptBuilder};
 
-    // Runtime sig op counting activates at DAA score 3
-    const RUNTIME_SIGOP_ACTIVATION_DAA_SCORE: u64 = 3;
+    // Runtime sig op counting is enabled from genesis (DAA score 0)
+    const RUNTIME_SIGOP_ACTIVATION_DAA_SCORE: u64 = 0;
```
someone235: const looks redundant

testing/integration/src/consensus_integration_tests.rs
someone235: We can just rename it to `kip10_test`

crypto/txscript/src/lib.rs
biryukovmaxim: I kept it because it could raise questions why does it differ from bitcoin tests, meantime the name explains why. On the other hand any new opcode would require adding suffix if we go this way so I'm okay to rename it

testing/integration/src/consensus_integration_tests.rs
biryukovmaxim: My idea was that we can't simply remove test that checks anything. This one covers transactions that uses kip10 opcodes and ensures blocks containing them are valid. So renaming is way

crypto/txscript/src/lib.rs
someone235: Yeah, I think it's best to rename it. We can add a comment in the first lines of the JSON that explains it's copied from bitcoin core tests with some Kaspa related additions and adaptations.

PR #777, opened 2025-12-08
P2p compressed header

Manyfestation:
Enables P2P support for compressed block headers
- Bump protocol version to v9
- Sending p2p headers based on peer version
- Auto detecting received p2p headers based on the presence of the new `cumulative_level` field

### DIFF
--- a/protocol/flows/src/flow_context.rs
+++ b/protocol/flows/src/flow_context.rs
@@ -751,7 +751,8 @@ impl ConnectionInitializer for FlowContext {
 
         // Register all flows according to version
         let (flows, applied_protocol_version) = match peer_version.protocol_version {
-            v if v >= PROTOCOL_VERSION => (v8::register(self.clone(), router.clone()), PROTOCOL_VERSION),
+            v if v >= PROTOCOL_VERSION => (v8::register(self.clone(), router.clone(), PROTOCOL_VERSION), PROTOCOL_VERSION),
+            8 => (v8::register(self.clone(), router.clone(), 8), 8),
             7 => (v7::register(self.clone(), router.clone()), 7),
             v => return Err(ProtocolError::VersionMismatch(PROTOCOL_VERSION, v)),
         };
--- a/protocol/flows/src/ibd/flow.rs
+++ b/protocol/flows/src/ibd/flow.rs
@@ -84,8 +84,8 @@ impl IbdFlow {
         incoming_route: IncomingRoute,
         relay_receiver: JobReceiver<Block>,
         body_only_ibd_permitted: bool,
+        header_format: HeaderFormat,
     ) -> Self {
-        let header_format = HeaderFormat::from(router.properties().protocol_version);
         Self { ctx, router, incoming_route, relay_receiver, body_only_ibd_permitted, header_format }
     }
 
--- a/protocol/flows/src/v7/blockrelay/handle_requests.rs
+++ b/protocol/flows/src/v7/blockrelay/handle_requests.rs
@@ -13,6 +13,7 @@ pub struct HandleRelayBlockRequests {
     ctx: FlowContext,
     router: Arc<Router>,
     incoming_route: IncomingRoute,
+    header_format: HeaderFormat,
 }
 
 #[async_trait::async_trait]
@@ -27,8 +28,8 @@ impl Flow for HandleRelayBlockRequests {
 }
 
 impl HandleRelayBlockRequests {
-    pub fn new(ctx: FlowContext, router: Arc<Router>, incoming_route: IncomingRoute) -> Self {
-        Self { ctx, router, incoming_route }
+    pub fn new(ctx: FlowContext, router: Arc<Router>, incoming_route: IncomingRoute, header_format: HeaderFormat) -> Self {
+        Self { ctx, router, incoming_route, header_format }
     }
 
     async fn start_impl(&mut self) -> Result<(), ProtocolError> {
@@ -36,7 +37,6 @@ impl HandleRelayBlockRequests {
         // state even if no new blocks arrive for some reason.
         // Note: in go-kaspad this was done via a dedicated one-time flow.
         self.send_sink().await?;
-        let header_format = HeaderFormat::from(self.router.properties().protocol_version);
 
         loop {
             let (msg, request_id) = dequeue_with_request_id!(self.incoming_route, Payload::RequestRelayBlocks)?;
@@ -46,7 +46,7 @@ impl HandleRelayBlockRequests {
 
             for hash in hashes {
                 let block = session.async_get_block(hash).await?;
-                self.router.enqueue(make_response!(Payload::Block, (header_format, &block).into(), request_id)).await?;
+                self.router.enqueue(make_response!(Payload::Block, (self.header_format, &block).into(), request_id)).await?;
                 debug!("relayed block with hash {} to peer {}", hash, self.router);
             }
         }
--- a/protocol/flows/src/v7/blockrelay/flow.rs
+++ b/protocol/flows/src/v7/blockrelay/flow.rs
@@ -64,6 +64,8 @@ pub struct HandleRelayInvsFlow {
     msg_route: IncomingRoute,
     /// A channel sender for sending blocks to be handled by the IBD flow (of this peer)
     ibd_sender: JobSender<Block>,
+    /// Header format determined by protocol version
+    header_format: HeaderFormat,
 }
 
 #[async_trait::async_trait]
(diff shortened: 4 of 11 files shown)


consensus/core/src/header.rs
```diff
@@ -91,6 +95,22 @@ impl TryFrom<Vec<Vec<Hash>>> for CompressedParents {
...
+    type Error = CompressedParentsError;
+    fn try_from(parents: Vec<(u8, Vec<Hash>)>) -> Result<Self, Self::Error> {
+        let mut last_cum: u8 = 0;
+        for (cum, _) in &parents {
+            // Make sure any next cumulative is strictly greater than the last
+            if *cum <= last_cum {
+                return Err(CompressedParentsError::LevelsNotStrictlyIncreasing);
+            }
+            last_cum = *cum;
+        }
```
IzioDev: renaming proposal: `*cum` -> `*cumulative`

side note: we often refer as `cumulative` or `cumulative_level`, suggestion: unify variable naming representation

protocol/flows/src/v7/blockrelay/handle_requests.rs
```diff
@@ -43,7 +45,7 @@ impl HandleRelayBlockRequests {
 
             for hash in hashes {
                 let block = session.async_get_block(hash).await?;
-                self.router.enqueue(make_response!(Payload::Block, (&block).into(), request_id)).await?;
+                self.router.enqueue(make_response!(Payload::Block, (header_format, &block).into(), request_id)).await?;
```
IzioDev: What do you think of creating dedicated structure in the objective of trying to make conversion `From/To` clearer? (this pattern happens a few time in other places)

protocol/p2p/src/convert/header.rs
```diff
@@ -5,15 +5,43 @@ use kaspa_hashes::Hash;
...
+}
+
+/// Determines the header format based on the protocol version.
+pub fn determine_header_format(version: u32) -> HeaderFormat {
+    if version >= 9 {
+        HeaderFormat::Compressed
+    } else {
+        HeaderFormat::Legacy
+    }
+}
```
IzioDev, line 23: We could also use `From` trait here instead of the function (can also be seen as confusing in usage since the version is an atomic type, it's just a note)

protocol/flows/src/v7/blockrelay/handle_requests.rs
Manyfestation: I agree it would look cleaner, the down side is that we need 3 such new structures , for `DaaBlockV4` `PruningPointProofHeaderArray` and `BlockHeader`, which to me feels a bit bloated. 
Let's see if there are more opinions on this

consensus/core/src/header.rs
Manyfestation: Done, TY!
I think the only remaining use of `cumulative` is in the `IterExtensions` module, which is preferable as this tries to be a generic implementation of vectors with cumulative runs

protocol/p2p/src/convert/header.rs
Manyfestation: I did not think of that, something like - 

```
impl From<u32> for HeaderFormat {
    fn from(version: u32) -> Self {
        if version >= 9 {
            Self::Compressed
        } else {
            Self::Legacy
        }
    }
}


let header_format = HeaderFormat::from(p2p_version);
```

Fine by me, @michaelsutton wdyt?

protocol/flows/src/flow_context.rs
```diff
@@ -59,7 +59,7 @@ use tokio_stream::{wrappers::UnboundedReceiverStream, StreamExt};
 use uuid::Uuid;
 
 /// The P2P protocol version.
-const PROTOCOL_VERSION: u32 = 8;
+const PROTOCOL_VERSION: u32 = 9;
```
coderofstuff, line 62: I don't think we need to bump the protocol version to `v9` since `v8` isn't even in a release yet. Last release with `v1.0.1` had `v7`

wdyt @someone235 @michaelsutton?

protocol/flows/src/v7/blockrelay/handle_requests.rs
michaelsutton: not sure bcs it will be required in several places. I tend to think that type safety of `HeaderFormat` is enough (even when part of a tuple, it still makes the tuple strongly typed and pretty clear)

protocol/p2p/src/convert/header.rs
michaelsutton: since version is not strongly typed, using `From` might actually result in less clear code. my 2 cents

protocol/flows/src/flow_context.rs
michaelsutton: you will still have various `master` nodes bumping into each other. Good question. 

same discussion can be applied to the DB version bumps

protocol/p2p/src/convert/header.rs
```diff
@@ -42,9 +71,32 @@ impl From<&[Hash]> for protowire::BlockLevelParents {
 impl TryFrom<protowire::BlockHeader> for Header {
```
michaelsutton: I tend to think that the P2P receiving side should also convert based on the expected `HeaderFormat` (driven from the applied protocol version), and not rely on `p.cumulative_level > 0` check (but rather should verify that `is_compressed` matches the expected format and err otherwise). 

It basically boils down to the question of what happens when we deprecate older versions. If we don't do the above, the deprecation will essentially change the expectations (since from that point on it will _demand_ the RLE format)  

cc @someone235 @coderofstuff

protocol/p2p/src/convert/header.rs
michaelsutton: but `let header_format = HeaderFormat::from(p2p_version);` actually looks pretty clean

protocol/flows/src/flow_context.rs
coderofstuff: I think we should stick to a single version upgrade per release even if various `master`-branch based nodes would bump into each other. And I think the same for DB version bumps.

This is representative of the behavior we expect most of the network will actually do. For example, I don't expect pools or exchanges to be running the latest version of `master` as we merge PRs. Instead, they would transition to the latest version upon prompting.

I assume that anyone running `master` must be willing to work with nightly or unstable-style builds, and assume instability that entails.

protocol/p2p/src/convert/header.rs
Manyfestation: I've added a generic `Versioned<T>(pub HeaderFormat, pub T)` wrapper to handle conversions for all related proto types

protocol/p2p/src/convert/header.rs
Manyfestation: Done

consensus/core/src/header.rs
```diff
@@ -95,6 +95,22 @@ impl TryFrom<Vec<Vec<Hash>>> for CompressedParents {
     }
 }
 
+impl TryFrom<Vec<(u8, Vec<Hash>)>> for CompressedParents {
+    type Error = CompressedParentsError;
+    fn try_from(parents: Vec<(u8, Vec<Hash>)>) -> Result<Self, Self::Error> {
+        let mut last_cumulative_level: u8 = 0;
+        for (cumulative_level, _) in &parents {
```
michaelsutton: we also need to verify max compression. ie verify that parents are different in each two consecutive runs. 
you can do both checks (including the check that `cumulative_level` is strictly increasing) using a `tuple_windows` iterator

protocol/flows/src/ibd/flow.rs
```diff
@@ -360,7 +363,8 @@ impl IbdFlow {
 
         // Pruning proof generation and communication might take several minutes, so we allow a long 10 minute timeout
         let msg = dequeue_with_timeout!(self.incoming_route, Payload::PruningPointProof, Duration::from_secs(600))?;
-        let proof: PruningPointProof = msg.try_into()?;
+        let header_format = HeaderFormat::from(self.router.properties().protocol_version);
```
michaelsutton: you should add an `header_format` field to the flow and use it all over

michaelsutton: the same tactic should be applied to blockrelay flow as well or any other flow which calls the line `self.router.properties().protocol_version` more than once (so no need to change the handle_*_request flows where you only call it once before the loop)

Manyfestation: I like the comment, and changed it in here (IbdFlow). This is the only flow which has more than one fn who use the header_format. So this the only place I done it.

Manyfestation:
> Lgtm. Can you please run final tests vs. v7, v8, v9 syncers?

@michaelsutton I synced the new code against all three version (stopped after seeing that IBD headers sync is OK)

Issue #778, opened 2025-12-09
GetBlocks returns the lowHash when called around the sink


coderofstuff:
**Describe the bug**
The RPC method `GetBlocks` is expected to return a block hashes starting from the provided lowHash to up to a limit (or the sink + anticones). It should return a unique set of hashes such that you can do:

```
lowHash = <some start hash>

loop forever:
  blocks = GetBlocks(lowHash)
  
  for block in blocks:
    // do something
    lowHash = block.hash
```

However, when querying blocks close to the sink, it's possible that this RPC will return the `lowHash` as part of the response due to the line: https://github.com/kaspanet/rusty-kaspa/blob/f25a737bea1b5d32902d402f9c1daf1f8efd020a/rpc/service/src/service.rs#L458

Around the tips, `sink_anticone` may contain the start hash if the lowHash is in the anticone of sink. This results in a repeat of a call to `GetBlocks` for the same lowHash.

In this example, sink at the time of call is shown. The highlighted part is what was returned as the `sink_anticone` which is appended onto the chain.

<img width="632" height="639" alt="Image" src="https://github.com/user-attachments/assets/ad2b6daf-ed72-4d6c-b312-886cf0da49f7" />

**To Reproduce**
1. Find a block `X` that is the anticone of `sink`
2. Call `GetBlocks` with `X`

Expectation: All hashes returned are unique
Result: if lowHash is in the anticone of sink, the lowHash appears both at the beginning and at the (or close to) the end of the result

**Additional context**
- Dev note: Ensure that `sink_anticone` is filtered to not contain any of the other blocks in `block_hashes` or `low_hash`

coderofstuff:
Use this nodejs code to replicate the issue with WASM SDK in the same folder. Use `./build-node` from the instructions here to build the WASM SDK from the repo: https://kaspa.aspectron.org/docs/#md:building-from-source

```javascript
// Tests replication of block hashes from a given starting point

globalThis.WebSocket = require('websocket').w3cwebsocket;

const kaspa = require('./kaspa');
const { RpcClient, Encoding } = kaspa;

async function delay(ms) {
    return new Promise((resolve) => setTimeout(resolve, ms));
}

async function run() {
    const networkArg = process.argv[2] || 'mainnet';

    const rpc = new RpcClient({
        url: '127.0.0.1:17110',
        encoding: Encoding.Borsh,
        networkId: networkArg,
    });

    await rpc.connect();
    console.info('Connected to RPC');

[...]

PR #779, opened 2025-12-09
Fix GetBlocks returning duplicate hashes (issue #778)

LiveLaughLove13:
- Filter sink_anticone to remove low_hash and existing block_hashes
- Prevents low_hash from appearing twice when it's in sink anticone
- Fixes GitHub issue #778

@coderofstuff

### DIFF
--- a/rpc/service/src/service.rs
+++ b/rpc/service/src/service.rs
@@ -455,16 +455,15 @@ NOTE: This error usually indicates an RPC conversion error between the node and 
         // If the high hash is equal to sink it means get_hashes_between didn't skip any hashes, and
         // there's space to add the sink anticone, otherwise we cannot add the anticone because
         // there's no guarantee that all of the anticone root ancestors will be present.
-        let sink_anticone = if high_hash == sink_hash { session.async_get_anticone(sink_hash).await? } else { vec![] };
-
-        // Filter out duplicates: remove low_hash and any blocks already in block_hashes from sink_anticone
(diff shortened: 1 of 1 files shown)


rpc/service/src/service.rs
```diff
@@ -456,8 +456,18 @@ NOTE: This error usually indicates an RPC conversion error between the node and
         // there's space to add the sink anticone, otherwise we cannot add the anticone because
         // there's no guarantee that all of the anticone root ancestors will be present.
         let sink_anticone = if high_hash == sink_hash { session.async_get_anticone(sink_hash).await? } else { vec![] };
```
coderofstuff: I think you can add the logic below here in the `if high_hash == sink_hash` condition. As in:

```
let sink_anticone = if high_hash == sink_hash {
  // get the sink anticone and filter out seen hashes
} else {
  vec![];
}
```

---

For the seen_hashes for loop, you can also do:

```
let seen_hashes: : HashSet<_> = once(low_hash).chain(block_hashes.iter().copied()).collect();
```

LiveLaughLove13:
@coderofstuff / @freshair18  Fixed linting issue - removed extra blank lines to pass CI formatting checks.

Ready for review!

LiveLaughLove13:
@coderofstuff  test  finished its checks... I have also done the refactor based on your suggestions. Pushed

LiveLaughLove13:
@coderofstuff All checks have passed :)

PR #781, opened 2025-12-10
perf: batch UTXO set writes during IBD (issue #768)

### LINKED ISSUE #768: IBD: Batch UTXO set writes
In `consensus/src/model/sstores/utxo_set.rs` the `write_many` function will use a `DirectDBWriter` to write to the UTXOs to the database.

It seems like converting using a `WriteBatch` to write the data as an entire batch would make this area less hot (See Extras section below)

Writing to cache can also be skipped

## TODO
- [ ] Investigate if converting to `BatchDbWriter` indeed results in performance improvements (either faster completion or less CPU intensiveness)
- [ ] (if so) Implement a `write_many_without_cache`, and update `UtxoSetStore::write_many` to use `write_many_without_cache` passing in a `BatchDbWriter` instead of a `DirectDbWriter` instance

### Flamegraph when using Direct DB Writer (Current)
Image

### Flamegraph when using Batched Writer (suggested)

Image

LiveLaughLove13:
- Add write_many_batch() method using BatchDbWriter instead of DirectDbWriter
- Skip cache updates during batch writes for better performance
- Integrate batch writing into IBD UTXO import (batch every 50 chunks)
- Reduces database write operations significantly during IBD
- Fixes: https://github.com/kaspanet/rusty-kaspa/issues/768


consensus/src/pipeline/virtual_processor/processor.rs
```diff
@@ -1178,9 +1178,59 @@ impl VirtualStateProcessor {
...
+            let mut chunk_count = 0u64;
+            let mut utxo_count = 0u64;
+            let mut last_log_time = std::time::Instant::now();
+
+            // Batch size: write every 50 chunks to balance performance and memory usage
+            const BATCH_SIZE: u64 = 50;
+            let mut batched_utxos = Vec::new();
+
             for chunk in &pruning_meta_read.utxo_set.iterator().map(|iter_result| iter_result.unwrap()).chunks(1000) {
-                virtual_write.utxo_set.write_from_iterator_without_cache(chunk).unwrap();
```
coderofstuff, line 1587: The entire part here is incorrect. The change to call `write_many_batch` needs to happen in the `write_many` call inside `append_imported_pruning_point_utxos`

See https://github.com/kaspanet/rusty-kaspa/blob/6a94fac31ea9382355c17cb9c2b82b0d0b265177/consensus/src/consensus/mod.rs#L932

coderofstuff (changes requested):
After applying the tests, run this against mainnet and make sure to verify that IBD still completes.

consensus/src/pipeline/virtual_processor/processor.rs
```diff
@@ -1167,9 +1167,22 @@ impl VirtualStateProcessor {
...
+            let mut chunk_count = 0u64;
+            let mut last_log_time = std::time::Instant::now();
+
             for chunk in &pruning_meta_read.utxo_set.iterator().map(|iter_result| iter_result.unwrap()).chunks(1000) {
                 virtual_write.utxo_set.write_from_iterator_without_cache(chunk).unwrap();
+                chunk_count += 1;
+
+                // Log progress every 10,000 chunks or every 5 seconds, whichever comes first
+                let now = std::time::Instant::now();
+                if chunk_count % 10000 == 0 || now.duration_since(last_log_time).as_secs() >= 5 {
```
D-Stacks: personally i feel there is no need to log by chunk amount, I'd rather just stay time-based. It'll give it more consistency, and stop it from hogging the log output on potentially very fast machines.

coderofstuff: Remove all the changes in this file. There's no reason for it to be changed w.r.t. the performance update being made elsewhere.

LiveLaughLove13: Understood, I can stop this PR then correct. Just want to verify

PR #783, opened 2025-12-12
chore(p2p/hub): reduce mismatch version error to debug log

IzioDev:
reduce mismatch version error to debug log

[source discussion](https://t.me/kasparnd/1/7478)

### DIFF
--- a/rpc/wrpc/client/src/client.rs
+++ b/rpc/wrpc/client/src/client.rs
@@ -372,6 +372,10 @@ impl KaspaRpcClient {
         Ok(())
     }
 
+    pub fn network_id(&self) -> Option<NetworkId> {
+        *self.inner.network_id.lock().unwrap()
+    }
+
     pub fn set_network_id(&self, network_id: &NetworkId) -> Result<()> {
         self.inner.network_id.lock().unwrap().replace(*network_id);
         Ok(())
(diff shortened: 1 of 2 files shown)


protocol/p2p/src/core/hub.rs
```diff
@@ -51,10 +51,16 @@ impl Hub {
...
                                     new_router.close().await;
-                                    if matches!(err, ProtocolError::LoopbackConnection(_) | ProtocolError::PeerAlreadyExists(_)) {
-                                        debug!("P2P, handshake failed for inbound peer {}: {}", new_router, err);
-                                    } else {
-                                        warn!("P2P, handshake failed for inbound peer {}: {}", new_router, err);
+
+                                    match err {
+                                        ProtocolError::LoopbackConnection(_)
+                                        | ProtocolError::PeerAlreadyExists(_)
+                                        | ProtocolError::VersionMismatch(_, _) => {
```
michaelsutton: wondering whether we should silence this only for pre crescendo versions 
(ie the question is -- when we deprecate version 7 in the future, should this log be silent or not)

IzioDev: interesting point. so silence <= v6 ?

on the next release, RK will release p2p v8 (right?), so in this context, I believe our consoles will be filled with mismatch log v7 (theirs) / v8 (mine). Is it the expected behavior?

my initial intuition was to "ignore" (ban for a small period of time) because the peer tried to connect regularly (every 10 or 20 sec or so) and this cluttered the logs. but it has been estimated (rightfully) to be too much effort for such a "small" need

michaelsutton: no, v7<->v8 will not be a version mismatch err (but rather the handshake will select v7). 

anyway, yes I think we should only silence < v7 for now

IzioDev: done here: [review: only silencing for deprecated version mismatch](https://github.com/kaspanet/rusty-kaspa/pull/783/commits/e7ef6957f3d01f91e86ca986fa034e23cefa0fb6)

PR #784, opened 2025-12-15
BridgeCrate 

LiveLaughLove13:
Kaspa Cargo and FMT,  Stramtum crate checking if passes.

### DIFF
--- a/components/kaspa-stratum-bridge/src/log_colors.rs
+++ b/components/kaspa-stratum-bridge/src/log_colors.rs
@@ -63,11 +63,6 @@ impl LogColors {
         }
     }
 
-    #[cfg(not(windows))]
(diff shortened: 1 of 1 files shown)


LiveLaughLove13:
"776, If I understand properly should be closed in favor of 784.
Later, if maintainers recommend to also include this service as part of kaspad, it should be integrated using the fresh crate from 784 (example of usage is provided as a binary, as is, shouldn't be difficult to adapt to a new AsyncService that will be used by Kaspad)"--@iziodev

Precisely.. It should be a seamless easy integration.

Should 784 be closed or left as a reference point... which is in draft mode ... as conceptual approaches for integration of enabled flag direct node sync to kaspad

PR #785, opened 2025-12-16
Add multisig PSKT redeem scripts

LiveLaughLove13:
Populate multisig PSKB inputs with derived redeem scripts so cosigners can sign without external data; expose redeem script derivation per address; heap-allocate 250k input fixture to avoid stack overflow in generator test.

### DIFF
--- a/cli/src/modules/pskb.rs
+++ b/cli/src/modules/pskb.rs
@@ -251,11 +251,9 @@ impl Pskb {
                 let mut priority_fee_sompi: i64 = 0;
 
                 // Try to parse first arg as optional priority fee
-                if let Ok(fee_opt) = try_parse_optional_kaspa_as_sompi_i64(Some(&args[0])) {
-                    if let Some(fee_val) = fee_opt {
-                        priority_fee_sompi = fee_val;
-                        args.remove(0);
-                    }
+                if let Ok(Some(fee_val)) = try_parse_optional_kaspa_as_sompi_i64(Some(&args[0])) {
+                    priority_fee_sompi = fee_val;
+                    args.remove(0);
                 }
 
                 if args.len() % 2 != 0 {


LiveLaughLove13:
**What was implemented**

Multisig PSKB signing now auto-populates redeem scripts per input using derived pubkeys (Schnorr/EC-DSA aware), making bundles self-contained for cosigners and preserving consensus expectations. (`wallet/core/src/account/variants/multisig.rs`, `wallet/core/src/account/mod.rs`)

`MultiSig::redeem_script_for_address` derives pubkeys at the correct receive/change index and builds the multisig redeem script using `minimum_signatures` and the configured key format. The PSKB signing path detects multisig and injects these scripts; non-multisig bundles are cloned unchanged.

Stabilized the large UTXO generator test by heap-allocating the 250k input fixture to avoid stack overflow; the test now completes successfully. (`wallet/core/src/tx/generator/test.rs`)

**Why it should work**
Redeem scripts are deterministically reconstructed from the multisig xpub set, cosigner index, and signature threshold, matching the derivation used for receive/change addresses, so each PSKB input carries the exact redeem script needed for validation and sighash.

Prefilling redeem scripts removes reliance on external metadata; cosigners can sign directly from the PSKB with correct sighash context, preserving consensus correctness.

Heap allocation for the 250k-input test prevents stack exhaustion without changing transaction semantics; generator invariants and mass/fee checks remain intact.

**Validation**
 
 cargo test -p kaspa-wallet-core --lib -- --nocapture` (all passing, including `test_generator_inputs_250k_outputs_2_sweep`).
 
 88test tx::generator::test::test_generator_inputs_250k_outputs_2_sweep ... ok

test result: ok. 47 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 9.43s**
 
 **Why multisig matters (Tether, stablecoins, and broader adoption)**
 
 **Operational security for treasuries:** Threshold control (m-of-n) reduces single-key compromise risk for large issuer treasuries (e.g., Tether/USDT) and custodial stablecoin desks; enables hot/cold key separation and time-/role-based policies.

**Regulated and auditable flows:** Clear signer sets per address allow explicit custody policies, rotation, and revocation without changing deposit endpoints; supports compliance/audit trails for mint/burn and treasury rebalancing.

**Resilience and continuity:** Survives individual key loss/rotation while keeping funds live; avoids brittle single signer hot wallets in high-throughput issuance/redemption flows.

**Deterministic script surface:** Redeem scripts are reconstructed from xpubs + cosigner index + min-sigs, so every PSKB/transaction carries the exact script required for consensus validation and sighash no hidden metadata or off-ledger dependencies.


**Interoperability with PSKB/PSBT style flows:** Self contained PSKBs let downstream cosigners sign with correct sighash context; works for distributed treasury ops, external signers, and hardware/MPC bridges.

**Mass/fee correctness preserved:** Multisig sigop counts and mass accounting remain accurate; generator mass/fee invariants still hold (validated across full wallet-core tests, including the 250k-input stress).

**Adoption-friendly integration:** Wallets and services can expose multisig accounts without protocol changes; addresses derive normally (receive/change), and redeem scripts are auto-populated during signing, minimizing integration risk for exchanges, OTC desks, and stablecoin issuers.

 @coderofstuff @IzioDev @michaelsutton @someone235

LiveLaughLove13:
What was added (multisig integration)

https://github.com/kaspanet/rusty-kaspa/pull/785/commits/eb8fe609bc8b6f19a8ad59a5be90cd28b5ef9c91

Redeem-script derivation per address: MultiSig::redeem_script_for_address rebuilds scripts from the multisig xpub set + cosigner index + minimum_signatures, honoring Schnorr/ECDSA modes. (wallet/core/src/account/variants/multisig.rs)
Self-contained PSKBs: During PSKB signing, redeem scripts are injected per input, so cosigners need no external metadata. (wallet/core/src/account/mod.rs)

Wallet API parity for multi-user builds: Added no-op select/account stubs under feature="multi-user" to keep callers compiling; no behavior change for non-multi-user. (wallet/core/src/wallet/mod.rs)

CLI enhancements
pskb build [priority_fee] <address amount> [...]: build PSKB bundles with one or more outputs.
pskb redeem [address]: export redeem script hex for the current multisig (or specified) address.
pskb sign-key <privkey-hex> <pskb>: offline Schnorr signing (SIG_HASH_ALL) with a raw 32-byte hex key; no wallet secrets needed.
Help text updated. (cli/src/modules/pskb.rs)

No consensus changes; no test logic changes.
Redeem scripts are deterministic from xpubs + cosigner index + threshold; mass/sigop accounting is unchanged.
PSKBs are now self-contained for multisig, making cosigner workflows and offline signing straightforward.
CLI gains parity-style flows (build/redeem/offline-sign) without altering core behavior.

LiveLaughLove13:
@coderofstuff  @freshair18  Multisig with CLI for Rusty Kaspa officially ready for review! Woot!

notify/src/subscription/single.rs
```diff
@@ -383,11 +383,11 @@ impl Display for UtxosChangedSubscription {
...
-            UTXOS_CHANGED_SUBSCRIPTIONS.fetch_sub(1, Ordering::SeqCst) - 1,
-            self
-        );
+        let prev = UTXOS_CHANGED_SUBSCRIPTIONS
+            .fetch_update(Ordering::SeqCst, Ordering::SeqCst, |v| v.checked_sub(1))
+            .unwrap_or_else(|current| current);
+        let now = prev.saturating_sub(1);
+        trace!("UtxosChangedSubscription: {} in total (drop {})", now, self);
     }
 }
```
IzioDev, line 392: Seems unrelated to the PR, why is this change introduced?

cli/src/modules/pskb.rs
```diff
@@ -253,6 +367,8 @@ impl Pskb {
             &[
                 ("pskb create <address> <amount> <priority fee>", "Create a PSKB from single send transaction"),
                 ("pskb sign <pskb>", "Sign given PSKB"),
+                ("pskb sign-key <privkey-hex|wif> <pskb>", "Offline sign PSKB with provided private key"),
```
IzioDev, line 370: Why this command is taking a private key as an input / What is the difference with `pskb sign <pskb>` ?

I'm unsure the term "offline" here is adapted, aren't all of the commands defined here (except `send`) offline (they don't require internet)?

cli/src/modules/pskb.rs
```diff
@@ -253,6 +367,8 @@ impl Pskb {
...
+                ("pskb build [priority fee] <address amount> [address amount]...", "Build PSKB bundle with one or more outputs"),
```
IzioDev, line 371: What is the expected difference with this already existing command `pskb create <address> <amount> <priority fee>`?

cli/src/modules/pskb.rs
```diff
@@ -261,6 +377,7 @@ impl Pskb {
                 ("pskb script sign <pskb>", "Sign all PSKB's P2SH locked inputs"),
                 ("pskb script sign <pskb>", "Sign all PSKB's P2SH locked inputs"),
                 ("pskb script address <pskb>", "Prints P2SH address"),
+                ("pskb redeem [address]", "Print redeem script for current multisig account (defaults to current receive)"),
```
IzioDev, line 380: What should a user do once they have a redeem script?

IzioDev (commented):
General high level question before investigating deeper into implementation details

cli/src/modules/pskb.rs
IzioDev: If the goal is to allow multi outputs for a single bundle, maybe we should consider modifying the `create` command instead of creating a new `build` command. As is `create` and `build` living together is confusing from a end-user point of view

PR #786, opened 2025-12-17
tweak(vspcv2): return compressed parents instead of plain parents by level

IzioDev:
#765 introduced a compressed representation of parents by level.

This PR include this representation in vspcv2 RPC responses instead of unzipping prior transmission.

### DIFF
--- a/wasm/Cargo.toml
+++ b/wasm/Cargo.toml
@@ -68,10 +68,10 @@ targets = [
 ]
 folder = "setup"
 
-# Workaround for Rust 1.87.0
-# https://github.com/rust-lang/rust/issues/141048
+
 [package.metadata.wasm-pack.profile.release]
-wasm-opt = ["-O","--enable-bulk-memory"]
+# We require these options:
+wasm-opt = ["-O","--enable-bulk-memory","--enable-nontrapping-float-to-int"]
 
 [lints]
 workspace = true
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1709,9 +1709,11 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "899def5c37c4fd7b2664648c28120ecec138e4d395b459e5ca34f9cce2dd77fd"
 dependencies = [
  "cfg-if 1.0.0",
+ "js-sys",
  "libc",
  "r-efi",
  "wasip2",
+ "wasm-bindgen",
 ]
 
 [[package]]
@@ -2282,9 +2284,9 @@ dependencies = [
 
 [[package]]
 name = "js-sys"
-version = "0.3.77"
+version = "0.3.83"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "1cfaf33c695fc6e08064efbc1f72ec937429614f25eef83af942d0e227c3a28f"
+checksum = "464a3709c7f55f1f721e5389aa6ea4e3bc6aba669353300af094b29ffbdde1d8"
 dependencies = [
  "once_cell",
  "wasm-bindgen",
@@ -2343,7 +2345,7 @@ dependencies = [
  "borsh",
  "bs58",
  "faster-hex",
- "getrandom 0.2.15",
+ "getrandom 0.3.4",
  "hmac",
  "js-sys",
  "kaspa-consensus-core",
@@ -2512,7 +2514,7 @@ dependencies = [
  "criterion",
  "faster-hex",
  "futures-util",
- "getrandom 0.2.15",
+ "getrandom 0.3.4",
  "itertools 0.13.0",
  "js-sys",
  "kaspa-addresses",
@@ -3805,9 +3807,9 @@ dependencies = [
 
 [[package]]
 name = "libm"
-version = "0.2.8"
+version = "0.2.15"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "4ec2a862134d2a7d32d7983ddcdd1c4923530833c9f2ea1a44fc5fa473989058"
+checksum = "f9fbbcab51052fe104eb5e5d351cf728d30a5be1fe14d9be8a3b097481fb97de"
 
 [[package]]
 name = "libmimalloc-sys"
@@ -4278,6 +4280,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "nu-ansi-term"
+version = "0.50.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "7957b9740744892f114936ab4a57b3f487491bbeafaf8083688b16841a4240e5"
+dependencies = [
+ "windows-sys 0.59.0",
+]
+
+[[package]]
 name = "num"
 version = "0.4.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -4412,9 +4423,9 @@ checksum = "1261fe7e33c73b354eab43b1273a57c8f967d0391e80353e51f764ac02cf6775"
 
 [[package]]
 name = "oorandom"
-version = "11.1.4"
+version = "11.1.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "b410bbe7e14ab526a0e86877eb47c6996a2bd7746f027ba551028c925390e4e9"
+checksum = "d6790f58c7ff633d8771f42965289203411a5e5c68388703c06e14f24770b41e"
 
 [[package]]
 name = "opaque-debug"
@@ -6377,9 +6388,9 @@ dependencies = [
 
 [[package]]
 name = "wasm-bindgen"
-version = "0.2.100"
+version = "0.2.106"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "1edc8929d7499fc4e8f0be2262a241556cfc54a0bea223790e71446f2aab1ef5"
+checksum = "0d759f433fa64a2d763d1340820e46e111a7a5ab75f993d1852d70b03dbb80fd"
 dependencies = [
  "cfg-if 1.0.0",
  "once_cell",
@@ -6387,27 +6398,14 @@ dependencies = [
  "serde",
  "serde_json",
  "wasm-bindgen-macro",
-]
-
-[[package]]
-name = "wasm-bindgen-backend"
-version = "0.2.100"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "2f0a0651a5c2bc21487bde11ee802ccaf4c51935d0d3d42a6101f98161700bc6"
-dependencies = [
- "bumpalo",
- "log",
- "proc-macro2",
- "quote",
- "syn 2.0.110",
  "wasm-bindgen-shared",
 ]
 
 [[package]]
 name = "wasm-bindgen-futures"
-version = "0.4.50"
+version = "0.4.56"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "555d470ec0bc3bb57890405e5d4322cc9ea83cebb085523ced7be4144dac1e61"
+checksum = "836d9622d604feee9e5de25ac10e3ea5f2d65b41eac0d9ce72eb5deae707ce7c"
 dependencies = [
  "cfg-if 1.0.0",
  "js-sys",
@@ -6418,9 +6416,9 @@ dependencies = [
 
 [[package]]
 name = "wasm-bindgen-macro"
-version = "0.2.100"
+version = "0.2.106"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7fe63fc6d09ed3792bd0897b314f53de8e16568c2b3f7982f468c0bf9bd0b407"
+checksum = "48cb0d2638f8baedbc542ed444afc0644a29166f1595371af4fecf8ce1e7eeb3"
 dependencies = [
  "quote",
  "wasm-bindgen-macro-support",
@@ -6428,34 +6426,42 @@ dependencies = [
 
 [[package]]
 name = "wasm-bindgen-macro-support"
-version = "0.2.100"
+version = "0.2.106"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "8ae87ea40c9f689fc23f209965b6fb8a99ad69aeeb0231408be24920604395de"
+checksum = "cefb59d5cd5f92d9dcf80e4683949f15ca4b511f4ac0a6e14d4e1ac60c6ecd40"
 dependencies = [
+ "bumpalo",
  "proc-macro2",
  "quote",
  "syn 2.0.110",
- "wasm-bindgen-backend",
  "wasm-bindgen-shared",
 ]
 
 [[package]]
 name = "wasm-bindgen-shared"
-version = "0.2.100"
+version = "0.2.106"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "1a05d73b933a847d6cccdda8f838a22ff101ad9bf93e33684f39c1f5f0eece3d"
+checksum = "cbc538057e648b67f72a982e708d485b2efa771e1ac05fec311f9f63e5800db4"
 dependencies = [
  "unicode-ident",
 ]
 
 [[package]]
 name = "wasm-bindgen-test"
-version = "0.3.50"
+version = "0.3.56"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "66c8d5e33ca3b6d9fa3b4676d774c5778031d27a578c2b007f905acf816152c3"
+checksum = "25e90e66d265d3a1efc0e72a54809ab90b9c0c515915c67cdf658689d2c22c6c"
 dependencies = [
+ "async-trait",
+ "cast",
  "js-sys",
+ "libm",
  "minicov",
+ "nu-ansi-term",
+ "num-traits",
+ "oorandom",
+ "serde",
+ "serde_json",
  "wasm-bindgen",
  "wasm-bindgen-futures",
  "wasm-bindgen-test-macro",
@@ -6463,9 +6469,9 @@ dependencies = [
 
 [[package]]
 name = "wasm-bindgen-test-macro"
-version = "0.3.50"
+version = "0.3.56"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "17d5042cc5fa009658f9a7333ef24291b1291a25b6382dd68862a7f3b969f69b"
+checksum = "7150335716dce6028bead2b848e72f47b45e7b9422f64cccdc23bedca89affc1"
 dependencies = [
  "proc-macro2",
  "quote",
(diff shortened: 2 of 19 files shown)


rpc/grpc/core/src/convert/block.rs
```diff
@@ -7,14 +7,6 @@ use std::str::FromStr;
 // rpc_core to protowire
 // ----------------------------------------------------------------------------
 
-from!(item: &kaspa_rpc_core::RpcOptionalBlock, protowire::RpcBlock, {
-    Self {
-        header: item.header.as_ref().map(protowire::RpcBlockHeader::from),
-        transactions: item.transactions.iter().map(protowire::RpcTransaction::from).collect(),
-        verbose_data: item.verbose_data.as_ref().map(|x| x.into()),
-    }
-});
```
IzioDev, line 16: was unused (and it simplified conversion to remove it), same for the RpcOptionalHeader -> protowire::RpcBlockHeader conversion below

rpc/grpc/core/src/convert/optional/header.rs
```diff
@@ -1,7 +1,30 @@
...
+    runs: &[protowire::RpcCompressedBlockLevelParent],
+) -> RpcResult<kaspa_rpc_core::RpcCompressedParents> {
+    let mut tuples = Vec::with_capacity(runs.len());
+    for run in runs.iter() {
+        let cumulative_level = u8::try_from(run.cumulative_level)?;
+        let parents = run.parent_hashes.iter().map(|h| RpcHash::from_str(h).map_err(RpcError::from)).collect::<RpcResult<Vec<_>>>()?;
+        tuples.push((cumulative_level, parents));
+    }
+    Ok(kaspa_rpc_core::RpcCompressedParents::new(tuples))
+}
```
IzioDev: I wanted to implement proper from! and try_from! traits for `protowire::RpcCompressedBlockLevelParent` <-> `kaspa_rpc_core::RpcCompressedParents`.

It seems because `RpcCompressedParents` is a type alias (to consensus_core), it is treated as foreign from `rpc_core` PoV, hence it blocks the impl of these traits

rpc/core/src/model/optional/header.rs
```diff
@@ -14,7 +14,7 @@ pub struct RpcOptionalHeader {
     /// Level: Low
     pub version: Option<u16>,
     /// Level: High
-    pub parents_by_level: Vec<Vec<Hash>>,
+    pub parents_by_level: Option<RpcCompressedParents>,
```
IzioDev, line 17: Maybe it's worth renaming to avoid end-user confusion.
Proposal: `compressed_parents` to match with upstream type name

rpc/grpc/core/src/convert/optional/header.rs
```diff
@@ -1,7 +1,30 @@
...
+        .collect()
+}
+
+fn compressed_parents_from_protowire(
+    runs: &[protowire::RpcCompressedBlockLevelParent],
+) -> RpcResult<kaspa_rpc_core::RpcCompressedParents> {
+    let mut tuples = Vec::with_capacity(runs.len());
+    for run in runs.iter() {
+        let cumulative_level = u8::try_from(run.cumulative_level)?;
+        let parents = run.parent_hashes.iter().map(|h| RpcHash::from_str(h).map_err(RpcError::from)).collect::<RpcResult<Vec<_>>>()?;
```
IzioDev, line 21: Use: `Vec::<RpcHash>::try_from` (remove manual mapping)

consensus/core/src/header.rs
```diff
@@ -15,9 +16,21 @@ use std::mem::size_of;
...
 
 impl CompressedParents {
+    /// Creates a `CompressedParents` from precomputed runs `(cumulative_level, parents)`.
+    /// The caller must guarantee the runs are valid (monotonic cumulative levels, <= 255, etc.).
+    pub fn new(runs: Vec<(u8, Vec<Hash>)>) -> Self {
+        Self(runs)
+    }
+
+    /// Returns the internal run-length encoded representation.
+    pub fn runs(&self) -> &[(u8, Vec<Hash>)] {
```
Manyfestation: In https://github.com/kaspanet/rusty-kaspa/pull/777 I added 
   ```
 pub fn inner(&self) -> &[(u8, Vec<Hash>)] {
        &self.0
    }
```

Comment (IzioDev) [reply consensus/core/src/header.rs]: Is #777 planned to be merged in the near future?

If yes, we'll wait for it, else I propose we define it here (i'll rename to `inner` to avoid conflicts)

File: rpc/grpc/core/proto/rpc.proto [hunk 6]
```diff
@@ -996,7 +1002,7 @@ enum RpcDataVerbosityLevel {
 message RpcOptionalHeader {
   optional string hash = 2;
   optional uint32 version = 3;
-  repeated RpcBlockLevelParents parentsByLevel = 4;
```
Manyfestation, line 999: In order to send compressed parents In  https://github.com/kaspanet/rusty-kaspa/pull/777 I added a cumulative field to the BlockLevelParents, maybe we should do the same here to be aligned? 
```
message BlockLevelParents {
  repeated Hash parentHashes = 1;
  uint32 cumulativeLevel = 2; // This field was added 
}
```

IzioDev:
```proto
// legacy (kept untouched)
message RpcBlockLevelParents {
  repeated string parentHashes = 1;
}

// new
message RpcCompressedBlockLevelParent {
  // must be <= 255 (u8)
  uint32 cumulativeLevel = 1;
  repeated string parentHashes = 2;
}
```

Since only `RpcOptionalHeader` adds a comprehension of `CompressedHeader` (not `RpcHeader`), I kept the legacy structure untouched in favor of adding a new one dedicated to the compressed representation.

I indeed could have used the same (legacy) struct, but it would have mean that the same structure would have been used for two different data representation (compressed/unzipped).

Let me know if this makes sense to you, and if I should re-consider this alternative

rpc/core/src/wasm/message.rs
```diff
@@ -1328,16 +1330,32 @@ declare! {
...
+        element.set("chainBlockHeader", &OptionalHeader::from(&entry.chain_block_header).into())?;
+        element.set("acceptedTransactions", &to_value(&entry)?)?;
+
+        chain_block_accepted_transactions.push(&element);
+    }
+
+    value.set("chainBlockAcceptedTransactions", chain_block_accepted_transactions.as_ref())?;
+
+    Ok(IGetVirtualChainFromBlockV2Response { obj: value })
 });
```
IzioDev, line 1351: This whole flow only exists because I wanted `object.chainBlockAcceptedTransactions.chainBlockHeader` to be `OptionalHeader`. I'm still unsure why a manual mapping is needed as is, something that I missed for sure.

cc @aspect I'd love your help on this if I can afford your time 🙏

wasm/Cargo.toml
```diff
@@ -71,7 +71,7 @@ folder = "setup"
 # Workaround for Rust 1.87.0
 # https://github.com/rust-lang/rust/issues/141048
 [package.metadata.wasm-pack.profile.release]
-wasm-opt = ["-O","--enable-bulk-memory"]
+wasm-opt = ["-O", "--enable-bulk-memory"]
```
IzioDev: Note: I have to set it to `= false` for my local build to work

consensus/core/src/header.rs
Manyfestation: In #787 (merged to master) @someone235 added it as `raw` https://github.com/kaspanet/rusty-kaspa/commit/3c2115d9fd70935fc6d18777d6a66784d0466a95#diff-77d6c8926daba7cac593b508c6a9b9994e58af21c7862ac1b9537f75741c52b9R78

wasm/Cargo.toml
IzioDev: It works with this one: `wasm-opt = ["-O", "--enable-bulk-memory", "--enable-nontrapping-float-to-int"]`

consensus/core/src/header.rs
IzioDev: Thanks for the notice, done :)

rpc/core/src/model/optional/header.rs
```diff
@@ -198,7 +196,7 @@ impl Deserializer for RpcOptionalHeader {
             1 => {
                 let hash = load!(Option<Hash>, reader)?;
                 let version = load!(Option<u16>, reader)?;
-                let parents_by_level = load!(Vec<Vec<Hash>>, reader)?;
+                let compressed_parents = load!(Option<RpcCompressedParents>, reader)?;
```
coderofstuff: Be mindful of this kind of change. It's probably fine here since vspc2 isn't fully out in a release yet. If it were in a release already, you'd have to use a new version. Just FYI.

rpc/core/src/model/optional/header.rs
```diff
@@ -53,13 +53,6 @@ impl RpcOptionalHeader {
             && self.blue_score.is_none()
             && self.pruning_point.is_none()
     }
-    pub fn direct_parents(&self) -> &[Hash] {
```
coderofstuff, line 56: Why was this removed?

wasm/examples/nodejs/javascript/general/get-virtual-chain-v2.js
coderofstuff, line 1: Please rename this file to `get-virtual-chain-v2.js`?

rpc/core/src/wasm/message.rs
coderofstuff: Did you find out why you needed the manual implementation here?

wasm/Cargo.toml
```diff
@@ -71,7 +71,7 @@ folder = "setup"
 # Workaround for Rust 1.87.0
 # https://github.com/rust-lang/rust/issues/141048
 [package.metadata.wasm-pack.profile.release]
-wasm-opt = ["-O","--enable-bulk-memory"]
+wasm-opt = ["-O", "--enable-bulk-memory", "--enable-nontrapping-float-to-int"]
```
coderofstuff: I tried building the wasm sdk via ./build-node without `--enable-nontrapping-float-to-int`. What is this for?

michaelsutton:
This must wait for #777 to merge since you will need to reuse its `impl TryFrom<Vec<(u8, Vec<Hash>)>> for CompressedParents` (see also https://github.com/kaspanet/rusty-kaspa/pull/777/files/c2b0f870b0516481dc02715e4741c3653eaab0f8#r2657638009)

rpc/grpc/core/proto/rpc.proto
```diff
@@ -42,7 +42,13 @@ message RpcBlockLevelParents {
   repeated string parentHashes = 1;
 }
 
-message RpcBlockVerboseData{
+message RpcCompressedBlockLevelParent {
+  // must be <= 255 (u8)
+  uint32 cumulativeLevel = 1;
```
michaelsutton: I suggest flipping the order (`uint32 cumulativeLevel = 2`) so that we have more flexibility in the future when we deprecate the non-compressed format (ie it will be easier to converge the two structs if the first field is the same and they only differ on whether there exist a second field; cf the p2p proto change in #777 which is backwards compatible).

wasm/Cargo.toml
IzioDev: in short is saturate float to int conversions (do not trap).

I have no idea why on my local machine the WASM build doesn't work without this option. I can keep it uncommitted, as long as the CI environment is fine without it

rpc/grpc/core/src/convert/optional/header.rs
```diff
@@ -1,7 +1,30 @@
...
+        tuples.push((cumulative_level, parents));
+    }
+    Ok(kaspa_rpc_core::RpcCompressedParents::new(tuples))
```
michaelsutton: should be `tuples.try_into()`

consensus/core/src/header.rs
```diff
@@ -18,6 +18,12 @@ use std::mem::size_of;
 pub struct CompressedParents(Vec<(u8, Vec<Hash>)>);
 
 impl CompressedParents {
+    /// Creates a `CompressedParents` from precomputed runs `(cumulative_level, parents)`.
+    /// The caller must guarantee the runs are valid (monotonic cumulative levels, <= 255, etc.).
+    pub fn new(runs: Vec<(u8, Vec<Hash>)>) -> Self {
```
michaelsutton: delete this `new`, and use `.try_into()` instead 

see https://github.com/kaspanet/rusty-kaspa/blob/73c6d5d2c670a0c5789cc7b755fdbd68ff470d8f/consensus/core/src/header.rs#L101 to understand why this conversion is sensitive

rpc/core/src/model/optional/header.rs
michaelsutton: Please address this comment

michaelsutton (changes requested):

rpc/core/src/model/optional/header.rs
IzioDev: It wasn't used internally.
If this is a convenience API (for external developers) we'd like to keep, I can re-introduce

wasm/examples/nodejs/javascript/general/get-virtual-chain-v2.js
IzioDev: [review: rename wasm vspcv2 example file](https://github.com/kaspanet/rusty-kaspa/pull/786/commits/452315d358dd69e639666eae2bae525728154f36)

rpc/grpc/core/proto/rpc.proto
IzioDev: [review: align proto prop order with other message](https://github.com/kaspanet/rusty-kaspa/pull/786/commits/9ad649ad039b885cbb80c224745493da480c65ca)

rpc/grpc/core/src/convert/optional/header.rs
IzioDev: [review: use tuple to compressed parent conversion](https://github.com/kaspanet/rusty-kaspa/pull/786/commits/f3e144a4f1f59e84e2e1e1bbdfc47ef440cc1abc)

consensus/core/src/header.rs
IzioDev: [review: use tuple to compressed parent conversion](https://github.com/kaspanet/rusty-kaspa/pull/786/commits/f3e144a4f1f59e84e2e1e1bbdfc47ef440cc1abc)

wasm/Cargo.toml
IzioDev: wasm-pack 0.13.1
rustc 1.92.0

Maybe a platform difference?

IzioDev: After some testing and thanks to coder's help I now have a better understanding.

If `--enable-nontrapping-float-to-int` isn't present, we need env var: `RUSTFLAGS=-Ctarget-cpu=mvp`.
Not sure what's best, but if we insist in the rustflags solution, I'd recommend adding `.cargo/config.toml` in `/wasm`:
```toml
[build]
rustflags = ["-Ctarget-cpu=mvp"]
```
So it's added automatically on any wasm build

cc @biryukovmaxim, could you please give me/us a bit of context on the initial fix and what is the approach you'd recommend?

biryukovmaxim: .cargo/config.toml in /wasm is not applicable if you build from the root. I as well as CI build everything from the root. I dont mind if it's added and it may help someone to build it simpler. but myself and CI wont be affected anyhow by config

consensus/core/src/header.rs
```diff
@@ -76,6 +76,7 @@ impl CompressedParents {
         *self = parents.try_into().unwrap();
     }
 
+    /// Returns the internal run-length encoded representation.
```
michaelsutton: RLE can be a bit misleading. Maybe write 
`/// Returns the internal cumulative-sum run-length encoded representation.`

IzioDev:
> Not sure about the renames:
> 
> ```
> parents_by_level -> compressed_parents
> include_parents_by_level -> include_compressed_parents
> ```
> 
> in all other places we still call it `parents_by_level` even though the representation changed to a compressed one.

Let me know what is the desired path about renames.
It felt natural to me to rename, from SDK usage it makes more sense: `header.compressedParents.toExpanded()`, as it better reflect the underlying object we're manipulating.
But I don't have a strong opinion, I'm fine with both, type system prevent you from doing mistakes regardless

wasm/Cargo.toml
IzioDev: I'm reverting this change, in favor of #808 
cc @michaelsutton

consensus/core/src/header.rs
IzioDev: [docs(compressed-parents): raw fn wording](https://github.com/kaspanet/rusty-kaspa/pull/786/commits/e1259dcb3475392856fe6a3faec3bb6b471cab84)

consensus/client/src/parents.rs
```diff
@@ -0,0 +1,90 @@
...
+            let array = js_sys::Array::from(value.as_ref());
+            let runs = array
+                .iter()
+                .map(|run_js| {
+                    let run_array = js_sys::Array::from(&run_js);
+                    if run_array.length() != 2 {
+                        return Err(Error::Custom("Invalid parent run, must be a tuple of [level, parents[]]".to_string()));
+                    }
+                    let level =
+                        run_array.get(0).as_f64().ok_or_else(|| Error::Custom("Invalid level in parent runs".to_string()))? as u8;
```
michaelsutton: `as u8` can panic here

consensus/client/src/parents.rs
```diff
@@ -0,0 +1,90 @@
+//!
+//! WASM implementation of a CompressedParents.
+//! This struct is composed of primitives and other WASM types.
+//!
+
+use crate::imports::*;
+use kaspa_consensus_core::header as native;
+use kaspa_hashes::Hash;
+
+/// An efficient run-length encoding for the parent-by-level vector in the block header.
```
michaelsutton: add `cumulative-sum` to this comment as well

rpc/grpc/core/src/convert/optional/header.rs
michaelsutton: it's because you are in **grpc**::core and the _local_ protowire isn't used here directly (wrapped by array/vec)

rpc/grpc/core/proto/rpc.proto
```diff
@@ -42,7 +42,13 @@ message RpcBlockLevelParents {
   repeated string parentHashes = 1;
 }
 
-message RpcBlockVerboseData{
+message RpcCompressedBlockLevelParent {
```
michaelsutton: nit: should be RpcCompressedBlockLevelParent**s** or just RpcBlockLevelRun (i.e., on the per run level, this isn't exactly compressed, the compression comes from the overall vec of runs)

consensus/client/src/parents.rs
IzioDev: nice catch.
proposed fix here: [review(consensus/client): panic-safe u8 cumulative level conversion](https://github.com/kaspanet/rusty-kaspa/pull/786/commits/eed66f8c22cdf59a3aee0e907d547fefab7c2bc7)

consensus/client/src/parents.rs
IzioDev: [docs(consensus/client): wasm compressed parent struct](https://github.com/kaspanet/rusty-kaspa/pull/786/commits/50f962ec828f94a60fa8ad96cc0a353d282bd5df)

rpc/grpc/core/proto/rpc.proto
IzioDev: [refacto(rpc/grpc): rename rpc compressed parents to a better reflecting name](https://github.com/kaspanet/rusty-kaspa/pull/786/commits/67bcadf541315a085ac8e60fabd91c5db9cf095a)

PR #787, opened 2025-12-18
Cleanup - Part 4

someone235:


consensus/core/src/merkle.rs
```diff
@@ -10,6 +10,11 @@ pub fn calc_hash_merkle_root_pre_crescendo<'a>(txs: impl ExactSizeIterator<Item
     calc_merkle_root(txs.map(hashing::tx::hash_pre_crescendo))
 }
 
+pub fn calc_accepted_id_merkle_root_pre_crescendo(mut accepted_tx_ids: Vec<Hash>) -> Hash {
+    accepted_tx_ids.sort();
+    kaspa_merkle::calc_merkle_root(accepted_tx_ids.into_iter())
+}
+
```
freshair18, line 17: Can you expand on why we need this?

consensus/core/src/tx.rs
```diff
@@ -198,6 +197,22 @@ impl Transaction {
...
+        gas: u64,
+        payload: Vec<u8>,
+        mass: u64,
+    ) -> Self {
+        let mut tx = Self::new_non_finalized(version, inputs, outputs, lock_time, subnetwork_id, gas, payload);
+        tx.set_mass(mass);
+        tx.finalize();
+        tx
+    }
+
```
freshair18, line 216: is this test only?

simpa/src/main.rs
```diff
@@ -479,10 +481,10 @@ fn submit_chunk(
     futures
 }
 
-fn topologically_ordered_hashes(src_consensus: &Consensus, genesis_hash: Hash) -> Vec<Hash> {
+pub(crate) fn topologically_ordered_hashes(src_consensus: &Consensus, genesis_hash: Hash, include_genesis: bool) -> Vec<Hash> {
```
freshair18, line 490: I am confused about this function's name.  Why call it topologically ordered if it is ordered by the timestamps?

simpa/src/blocks_json.rs
```diff
@@ -0,0 +1,112 @@
...
+                LockTime: tx.lock_time,
+                SubnetworkID: tx.subnetwork_id.to_string(),
+                Gas: tx.gas,
+                Payload: tx.payload.to_hex(),
+                mass: tx.mass(),
+            })
+            .collect(),
+        VerboseData: RPCBlockVerboseData { Hash: block.hash().to_string() },
+    }
+}
```
freshair18: I assume this  is blocked by some visibility considersations, but don't we have standard conversions from rpc to consensus etc.?

testing/integration/src/common/json.rs
```diff
@@ -0,0 +1,311 @@
...
+                    tx.LockTime,
+                    SubnetworkId::from_str(&tx.SubnetworkID).unwrap(),
+                    tx.Gas,
+                    hex_decode(&tx.Payload),
+                    tx.mass,
+                )
+            })
+            .collect(),
+    )
+}
```
freshair18, line 320: I'm a bit surprised we don't have from implementations for these sort of things, I guess visibility circumstances make this cumbersome?

consensus/src/pipeline/body_processor/body_validation_in_context.rs
```diff
@@ -103,9 +103,9 @@ mod tests {
 
     #[tokio::test]
     async fn validate_body_in_context_test() {
-        let config = ConfigBuilder::new(DEVNET_PARAMS)
+        let config = ConfigBuilder::new(MAINNET_PARAMS)
             .skip_proof_of_work()
-            .edit_consensus_params(|p| p.deflationary_phase_daa_score = 2)
+            .edit_consensus_params(|p| p.deflationary_phase_daa_score = GENESIS.daa_score + 2)
```
michaelsutton:
```suggestion
            .edit_consensus_params(|p| p.deflationary_phase_daa_score = p.genesis.daa_score + 2)
```
and remove `GENESIS` import

consensus/src/processes/transaction_validator/tx_validation_in_utxo_context.rs
```diff
@@ -35,15 +35,14 @@ impl TransactionValidator {
...
         flags: TxValidationFlags,
         mass_and_feerate_threshold: Option<(u64, f64)>,
     ) -> TxResult<u64> {
         self.check_transaction_coinbase_maturity(tx, pov_daa_score)?;
         let total_in = self.check_transaction_input_amounts(tx)?;
         let total_out = Self::check_transaction_output_values(tx, total_in)?;
         let fee = total_in - total_out;
-        if flags != TxValidationFlags::SkipMassCheck && self.crescendo_activation.is_active(block_daa_score) {
+        if flags != TxValidationFlags::SkipMassCheck {
             // Storage mass hardfork was activated
```
michaelsutton: delete/modify comment

consensus/core/src/merkle.rs
michaelsutton: iirc for archival validation

someone235: For reference for clients that wants to implement proof of acceptance on pre-crescendo blocks, and possibly for syncing archival nodes.

consensus/core/src/tx.rs
someone235: For now yes, but might be useful in the future for other uses

testing/integration/src/common/json.rs
someone235: Maybe it should be clarified in the comments, but these structs are the go-RPC structs that were used to create these tests in the first place, so they're not used anywhere else in the code.

simpa/src/main.rs
michaelsutton: bcs in the specific case of simpa, the simulation guarantees that timestamp respects topology

someone235: @michaelsutton wrote this function. I think it relies on the fact that in Simpa chronological order gurantees topological order

testing/integration/src/common/json.rs
michaelsutton: since we are no longer committed to go compatibility (field names, capitalization etc), we can probably migrate to use the usual `Rpc*` structs, but that can happen as a separate task with lower priority

PR #793, opened 2025-12-23
Kaspa Stratum Bridge binary

LiveLaughLove13:
# Kaspa Stratum Bridge (BridgeBinary) is a standalone Stratum bridge for Kaspa built on top of the [rusty-kaspa] stack.

- Accepts Stratum connections from ASIC miners.
- Fetches block templates from a Kaspa node via gRPC.
- Validates and submits found blocks to the node.
- Exposes detailed console stats and Prometheus metrics.

**The bridge supports two node modes:**

**External mode** – connects to an already running [kaspad] node.
**In‑process mode** – starts an embedded [kaspad] instance inside the bridge process, then talks to it via gRPC.
Both modes share the same mining logic and stats; only node lifecycle is different.

## Configuration
The bridge is configured via a YAML file, typically:
```text
components/kaspa-stratum-bridge/config.yaml

Key fields:

Global section
kaspad_address: gRPC address of the Kaspa node, e.g. "127.0.0.1:16110".
Instances section
One or more Stratum instances, each with:
stratum_port: e.g. ":5555", ":5556", …
Difficulty / var‑diff configuration.
Optional Prometheus port and logging options.
The YAML is parsed into:

GlobalConfig – shared settings for all instances.
InstanceConfig – per‑instance Stratum configuration.
BridgeConfig – contains global and instances and is passed into the bridge startup logic.
Command‑line interface
Binary: stratum-bridge (or kaspa-stratum-bridge)

Important CLI options (from src/main.rs):

--config <PATH>
Path to the YAML configuration.
--node-mode <external|inprocess>
external: connect to an already running kaspad.
inprocess: start an embedded kaspad node.

**If omitted:**
If any --node-args / --node-arg are supplied → inprocess is inferred.
Otherwise → external is inferred.
--node-args "<ARGS>"

Quoted string of arguments passed directly to kaspad in in‑process mode.
--node-arg <ARG> (repeatable)

Alternate way to supply individual node args.
**External node mode**
In external mode, the bridge runs as a pure gRPC client. You start kaspad yourself, then run the bridge and point it to the node via kaspad_address in config.yaml.

1. Start kaspad separately
Example (mainnet, with high peer/RPC capacity and perf metrics):

bash
cargo run --release --bin kaspad -- --utxoindex --listen=0.0.0.0:16111 --ram-scale=3 --rpclisten=127.0.0.1:16110 --rpclisten-borsh=127.0.0.1:17110 --rpclisten-json=127.0.0.1:18110 --outpeers=128 --maxinpeers=512 --rpcmaxclients=512 --perf-metrics --perf-metrics-interval-sec=1 --yes

**Important:**

--rpclisten must match kaspad_address in components/kaspa-stratum-bridge/config.yaml.
2. Start the bridge
PowerShell example (one logical command):

powershell
cargo run -p kaspa-stratum-bridge --release --bin stratum-bridge -- `
  --config components/kaspa-stratum-bridge/config.yaml `
  --node-mode external
Or on a single line (any shell):

bash
cargo run -p kaspa-stratum-bridge --release --bin stratum-bridge -- --config components/kaspa-stratum-bridge/config.yaml --node-mode external
If --node-mode is omitted and no --node-args are supplied, the bridge also infers external mode.

In‑process node mode
In in‑process mode, the bridge:

Parses kaspad CLI arguments (kaspad::args::Args).
Starts an embedded node via kaspad_daemon::create_core_with_runtime.
Connects to that node via gRPC at kaspad_address from the config.
The mining and stats behavior is the same as in external mode; only the node lifecycle differs.

**Example command (PowerShell)**
powershell

cargo run -p kaspa-stratum-bridge --release --bin stratum-bridge -- --config bridge/config.yaml --node-mode inprocess --node-args="--utxoindex --rpclisten=127.0.0.1:16110 --rpclisten-borsh=127.0.0.1:17110 --perf-metrics --perf-metrics-interval-sec=1 --outpeers=128 --disable-upnp "

Notes:

In PowerShell, use backtick ` for line continuation, or keep the command on one line.
If --node-mode is omitted but --node-args is present, the bridge infers inprocess mode.
--rpclisten must match kaspad_address in config.yaml (e.g. 127.0.0.1:16110).
Embedded node lifecycle (InProcessNode)
The embedded node is represented by:

rust
struct InProcessNode {
    core: Arc<kaspa_core::core::Core>,
    workers: Vec<std::thread::JoinHandle<()>>,
}
Startup (simplified):

rust
impl InProcessNode {
    fn start_from_args(args: kaspad_args::Args) -> Result<Self, anyhow::Error> {
        // Raise soft FD limit to match kaspad main, avoiding negative fd_budget
        let _ = fd_budget::try_set_fd_limit(kaspad_daemon::DESIRED_DAEMON_SOFT_FD_LIMIT);
        let runtime = kaspad_daemon::Runtime::from_args(&args);
        let fd_total_budget =
            fd_budget::limit()
            - args.rpc_max_clients as i32
            - args.inbound_limit as i32
            - args.outbound_target as i32;
        let (core, _) = kaspad_daemon::create_core_with_runtime(&runtime, &args, fd_total_budget);
        let workers = core.start();
        Ok(Self { core, workers })
    }
    fn shutdown(self) {
        self.core.shutdown();
        self.core.join(self.workers);
    }
}


**Key points:**

Uses the same initialization path as the standalone kaspad binary (create_core_with_runtime).
FD soft limit is raised before computing fd_total_budget, eliminating the fd_budget has to be positive panic with large peer/RPC settings.
Worker threads run inside the bridge process; shutdown is coordinated and blocking via core.shutdown() + core.join(...).
After InProcessNode starts, the bridge constructs a KaspaApi connected to grpc://<kaspad_address>, just like in external mode.

Kaspa API integration (KaspaApi)
File: components/kaspa-stratum-bridge/src/kaspaapi.rs

KaspaApi is a gRPC client wrapper used by the Stratum server to interact with the Kaspa node.

**Responsibilities**
Manage a kaspa_grpc_client::GrpcClient.
Provide high‑level operations:
get_block_template
submit_block
get_balances_by_addresses
Wait for node sync before starting the bridge.
Periodically fetch network stats and DAG info for Prometheus and console output.
Maintain a NODE_STATUS snapshot for display in the [NODE] ... line.
Connection setup
rust
let grpc_address = if address.starts_with("grpc://") {
    address.clone()
} else {
    format!("grpc://{}", address)
};
tracing::debug!("[API] Establishing RPC connection to Kaspa node: {}", grpc_address);
let client = Arc::new(
    GrpcClient::connect_with_args(
        NotificationMode::Direct,
        grpc_address.clone(),
        None,        // subscription_context
        true,        // reconnect: auto reconnection with re‑subscribed notifications
        None,        // no connection event channel
        false,       // override_handle_stop_notify = false
        Some(500_000), // request timeout (ms) - larger than default to reduce spurious timeouts
        Default::default(), // counters
    )
    .await
    .context("Failed to connect to Kaspa node")?,
);
// Start RPC services (notification collector)
client.start(None).await;
// Subscribe to NewBlockTemplate notifications
client
    .start_notify(ListenerId::default(), NewBlockTemplateScope {}.into())
    .await
    .context("Failed to subscribe to block template notifications")?;
Sync and background tasks
After connecting:

wait_for_sync(true):
Periodically calls client.get_sync_status().
Logs and waits until the node reports is_synced = true.
Starts background tasks:
start_stats_thread():
Every 30 seconds:
Calls get_block_dag_info_call and estimate_network_hashes_per_second_call.
Updates Prometheus network stats (hashrate, block count, difficulty).
start_node_status_thread():
Every 10 seconds:
Calls get_server_info_call, get_block_dag_info_call, get_connected_peer_info_call, get_info_call.
Updates NODE_STATUS (is_synced, network_id, server_version, peers, block/header counts, difficulty, tip hash, mempool size).
RPC methods used by Stratum:
get_block_template:
Wraps client.get_block_template_call(...).
Converts RpcRawBlock → Block and validates header serialization.
Retries transient errors up to 3 times, with special handling for malformed hex / “Odd number of digits” issues.
submit_block:
Wraps client.submit_block_call(...).
Logs detailed acceptance/rejection reasons at info/warn/error.
Because both in‑process and external modes use the same KaspaApi, the Stratum logic does not need to care which mode is in use.

**Stratum server and share handling**
Key files:

components/kaspa-stratum-bridge/src/stratum_server.rs
components/kaspa-stratum-bridge/src/share_handler.rs
Stratum server
For each InstanceConfig, the bridge:
Starts a TCP listener on instance.stratum_port, e.g. :5555, :5556, etc.
Creates a ShareHandler associated with an instance_id (e.g. Ins01, Ins02).
Wires in an Arc<dyn KaspaApiTrait + Send + Sync> implemented by KaspaApi.
Share handler
ShareHandler:

Accepts and validates shares from miners:
Parses stratum requests.
Manages per‑worker state (job IDs, difficulty, extranonce, etc.).
Uses KaspaApiTrait to:
Fetch fresh block templates.
Submit valid blocks to the Kaspa node.
Maintains work statistics:
Shares per minute vs target.
Hashrates per worker and per instance.
Accepted / stale / invalid share counts.
Periodically prints a table:
text
+------------------+-------+-------------+--------+-------------+------+--------------+--------+---------+
| Worker           | Inst  |        Hash |   Diff |     SPM/tgt | Trnd |  Acc/Stl/Inv | Blocks |    Time |
+------------------+-------+-------------+--------+-------------+------+--------------+--------+---------+
| ks2lite          | Ins01 |    2.11TH/s |   2048 |   14.4/20.0 | down |      339/0/0 |      0 |   23.6m |
...
| TOTAL            | ALL   |    7.41TH/s |      - |   29.0/20.0 | -    |      686/0/0 |      0 |   23.7m |
+------------------+-------+-------------+--------+-------------+------+--------------+--------+---------+
The share handling code is identical regardless of node mode; it depends only on the KaspaApiTrait abstraction.


Both the bridge and the node share a process.
The bridge still talks to the node via gRPC (loopback), using the same KaspaApi implementation as external mode.
Summary
kaspa-stratum-bridge (BridgeBinary) is the Stratum front‑end for Kaspa in the rusty-kaspa repo.
It supports:
External mode: connects to a separately running kaspad.
In‑process mode: embeds kaspad in the same process and then connects via gRPC.
The technical integration with rusty-kaspa reuses:
kaspad’s Args, Runtime, and create_core_with_runtime for in‑process startup.
kaspa-grpc-client + kaspa-rpc-core for all gRPC interactions.
Shared mining logic in stratum_server.rs and share_handler.rs, abstracted behind KaspaApiTrait.

### DIFF
--- a/README.md
+++ b/README.md
@@ -383,7 +383,7 @@ wRPC
 
 </details>
 
-## Stratum Bridge
+## Stratum Bridge Beta
 
 Check out the [README.md](bridge/docs/README.md) for instructions on how to run the stratum bridge.
 

### CODE REVIEW

File: core/src/log/mod.rs [hunk 1]
```diff
@@ -52,7 +52,7 @@ pub fn init_logger(log_dir: Option<&str>, filters: &str) {
         )
         .unwrap();
 
-    let _handle = log4rs::init_config(config).unwrap();
+    let _ = log4rs::init_config(config);
```
coderofstuff: Why is this change necessary? If it isn't, please revert it.

Cargo.toml
```diff
@@ -51,6 +51,7 @@ members = [
     "components/addressmanager",
     "components/connectionmanager",
     "components/consensusmanager",
+    "components/kaspa-stratum-bridge",
```
coderofstuff: Once the code is moved into `bridge`, make sure to update this reference here.

README.md
```diff
@@ -287,7 +287,80 @@ The framework is compatible with all major desktop and mobile browsers.
   ```bash
 cargo run --release --bin kaspad -- --testnet
   ```
-  
+
+## Stratum Bridge
```
coderofstuff: Move these documentations for the bridge to a dedicated README.md file in `bridge/docs/README.md`

Then, add a reference to that README file here. Something like:

```
## Stratum Bridge

Check out the [README.md](link here) for instructions on how to run the stratum bridge.
```

coderofstuff (changes requested):
Please move the files from within `components/kaspa-stratum-bridge` into a new folder in the root folder called `bridge`

core/src/log/mod.rs
LiveLaughLove13: It is needed to not cause backtrace error on logging for the stats of bridge

README.md
LiveLaughLove13: completed

bridge/Cargo.toml
```diff
@@ -0,0 +1,62 @@
+[package]
+name = "kaspa-stratum-bridge"
+version = "1.0.1"
+edition = "2021"
+authors = ["Kaspa developers"]
+description = "Kaspa Stratum Bridge - Mining pool connectivity for Kaspa ASIC miners"
+license = "ISC"
+repository = "https://github.com/kaspanet/rusty-kaspa"
```
coderofstuff: These should conform with the other Cargo.toml files that aren't in the root directory. Check out `database/Cargo.toml` for example.

bridge/config.yaml
```diff
@@ -0,0 +1,98 @@
...
+#    - Instance 1: localhost:5555 (min_share_diff: 2048) - Low difficulty
+#    - Instance 2: localhost:5556 (min_share_diff: 4096) - Medium difficulty
+#    - Instance 3: localhost:5557 (min_share_diff: 8192) - High difficulty
+#    - Instance 4: localhost:5558 (min_share_diff: 16384) - Very high difficulty
+# 5. Check Prometheus metrics:
+#    - Instance 1: http://localhost:2114/metrics
+#    - Instance 2: http://localhost:2115/metrics
+#    - Instance 3: http://localhost:2116/metrics
+#    - Instance 4: http://localhost:2117/metrics
+# 6. All logs will include instance identifiers like [Instance 1], [Instance 2], etc.
```
coderofstuff: Remove these

bridge/src/kaspaapi.rs
```diff
@@ -0,0 +1,664 @@
...
+    /// Create a new Kaspa API client
+    pub async fn new(address: String, _block_wait_time: Duration) -> Result<Arc<Self>> {
+        info!("Connecting to Kaspa node at {}", address);
+
+        // GrpcClient requires explicit "grpc://" prefix for connection
+        // Always add it if not present (avoids unnecessary connection failure)
+        let grpc_address = if address.starts_with("grpc://") { address.clone() } else { format!("grpc://{}", address) };
+
+        // Log connection attempt (detailed logs moved to debug)
+        tracing::debug!("{} {}", LogColors::api("[API]"), LogColors::label("Establishing RPC connection to Kaspa node:"));
```
coderofstuff: In Line 20 above, you have the other log levels imported, but here you're using `tracing::debug` directly. Why is that?

bridge/src/jsonrpc_event.rs
```diff
@@ -0,0 +1,122 @@
...
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub id: Option<Value>,
+    #[serde(default = "default_version")]
+    pub jsonrpc: String,
+    pub method: String, // We'll parse this as string and convert to StratumMethod when needed
+    pub params: Vec<Value>,
+}
+
+fn default_version() -> String {
+    "2.0".to_string()
```
coderofstuff, line 69: Where is this 2.0 value from?

LiveLaughLove13: Its the Version string for JSON-RPC 2.0

LiveLaughLove13: During serialization (JsonRpcEvent -> JSON):
serde will serialize whatever is currently in self.jsonrpc (it does not automatically “force” "2.0" unless you set it).

bridge/src/kaspaapi.rs
LiveLaughLove13: pulls in three logging macros from the tracing crate so the module can emit structured logs at different severity levels without having to write tracing::info!(...) everywhere.
 
 lets user know when block is found

bridge/config.yaml
LiveLaughLove13: removed

bridge/Cargo.toml
LiveLaughLove13: Updated

bridge/src/main.rs
```diff
@@ -0,0 +1,310 @@
...
+    node_arg: Vec<String>,
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, clap::ValueEnum)]
+enum NodeMode {
+    External,
+    Inprocess,
+}
+
+fn split_shell_words(input: &str) -> Result<Vec<String>, anyhow::Error> {
```
coderofstuff: What is this for? I'm not clear what it's actually doing, but it's strange that it is trying to parse things char by char

bridge/src/main.rs
```diff
@@ -0,0 +1,310 @@
...
+        tracing::info!("\t  stratum:       {}", instance.stratum_port);
+        tracing::info!("\t  min diff:      {}", instance.min_share_diff);
+        if let Some(ref prom_port) = instance.prom_port {
+            tracing::info!("\t  prom:          {}", prom_port);
+        }
+        if let Some(log_to_file) = instance.log_to_file {
+            tracing::info!("\t  log to file:   {}", log_to_file);
+        }
+    }
+    tracing::info!("----------------------------------");
```
coderofstuff: Refactor this initial log into it's own function in this file.

bridge/src/main.rs
```diff
@@ -0,0 +1,310 @@
...
+    let inferred_mode = if !node_args.is_empty() { NodeMode::Inprocess } else { NodeMode::External };
+    let node_mode = cli.node_mode.unwrap_or(inferred_mode);
+
+    // Load config first to check if file logging is enabled
+    let config_path = cli.config.as_path();
+    let fallback_path = std::path::Path::new("bridge").join(config_path);
+    let exe_base = std::env::current_exe().ok().and_then(|p| p.parent().map(|p| p.to_path_buf()));
+    let exe_root = exe_base.as_ref().and_then(|p| p.parent()).and_then(|p| p.parent()).map(|p| p.to_path_buf());
+
+    let mut candidates: Vec<std::path::PathBuf> = vec![config_path.to_path_buf(), fallback_path.clone()];
```
coderofstuff: What are these candidates for?

coderofstuff (changes requested):

bridge/src/main.rs
LiveLaughLove13: candidates vector contains multiple paths where the bridge will look for the config.yaml file. This ensures the bridge can find its configuration in different deployment scenarios:


Development: Running from different directories in the repo
Deployment: Different installation layouts inprocess versus external
Flexibility: Users can place config in various logical locations
Robustness: Bridge finds config regardless of how it's launched
The bridge iterates through these candidates and uses the first one that exists, making it resilient to different directory structures and deployment scenarios.

bridge/src/main.rs
LiveLaughLove13: refactored

bridge/src/main.rs
LiveLaughLove13: It is used to parse the --node-args "<string>" CLI option into a proper argv vector for kaspad_args::Args::parse. It implements minimal shell-style tokenization: splits on whitespace, preserves quoted segments ("..." / '...') as a single argument, and errors on unterminated quotes. The char-by-char scan is just to track “inside quotes” state; .split_whitespace() would break quoted args.

Since we did not add args to the kaspad, section directly as requested not to edit the kaspad area.

bridge/docs/README.md
```diff
@@ -0,0 +1,72 @@
...
+This repository contains a standalone Stratum bridge binary at:
+
+`bridge`
+
+The bridge can run against:
+
+- **External** node (you run `kaspad` yourself)
+- **In-process** node (the bridge starts `kaspad` in the same process)
+
+The bridge no longer supports spawning `kaspad` as a subprocess.
```
coderofstuff: Remove this line. It's not relevant since there was never a release that spawned it as a subprocess.

Make sure to go through this README and check that everything makes sense from a first release POV.

bridge/src/main.rs
coderofstuff: Hmm, is it possible to pass through arguments from command line to the inprocess node without using `--node-args`?

Something like:

```
./bridge --appdir=/dir/blah --disable-upnp
```

And these would work equivalent to as if you directly called:

```
./kaspad --appdir=/dir/blah --disable-upnp
```

Even an approach where these arguments are copied as needed to `struct Cli` could work better than doing `--node-args` this way.

What do you think of this approach?

bridge/src/main.rs
```diff
@@ -0,0 +1,323 @@
...
+    config: PathBuf,
+
+    #[arg(long, value_enum)]
+    node_mode: Option<NodeMode>,
+
+    #[arg(long, default_value = "--utxoindex --rpclisten=127.0.0.1:16110 --rpclisten-borsh=127.0.0.1:17110 --disable-upnp")]
+    node_args: Option<String>,
+
+    #[arg(long, action = clap::ArgAction::Append)]
+    node_arg: Vec<String>,
```
coderofstuff: Why is there a `node_args` and `node_arg`? Although, I think with the suggestion to copy relevant arguments to cli, you can just get rid of both of these.

bridge/src/main.rs
```diff
@@ -0,0 +1,323 @@
...
+            "--datadir".to_string(),
+            "bridge-datadir".to_string(),
+            "--rpclisten".to_string(),
+            "127.0.0.1:16111".to_string(), // Different port to avoid conflicts
+            "--utxoindex".to_string(),
+        ]);
+    }
+
+    let inferred_mode = NodeMode::Inprocess;
+    let node_mode = cli.node_mode.unwrap_or(inferred_mode);
```
coderofstuff:
```suggestion
    let node_mode = cli.node_mode.unwrap_or(NodeMode::Inprocess);
```

bridge/src/main.rs
coderofstuff: Ok, can you refactor these into a `fn initialize_config() -> BridgeConfig` (Currently L152 - L184) so that the remaining code in `main` is something like:

```
  let config = initialize_config();
```

bridge/src/stratum_listener.rs
```diff
@@ -0,0 +1,839 @@
...
+        } else if self.config.port.chars().all(|c| c.is_ascii_digit()) {
+            format!("0.0.0.0:{}", self.config.port)
+        } else {
+            self.config.port.clone()
+        };
+
+        let listener =
+            TcpListener::bind(&addr_str).await.map_err(|e| format!("failed listening to socket {}: {}", self.config.port, e))?;
+
+        tracing::debug!("Stratum listener started on {}", self.config.port);
```
coderofstuff: Why is tracing debug used directly, while others are in `use tracing::{error, info, warn};`? Please make it consistent, and please make sure to apply this to all files where such inconsistency lies. Just add `debug` to the `use tracing::{error, info, warn};` line in this case.

core/src/log/mod.rs
```diff
@@ -52,7 +52,7 @@ pub fn init_logger(log_dir: Option<&str>, filters: &str) {
         )
         .unwrap();
 
-    let _handle = log4rs::init_config(config).unwrap();
+    let _handle = log4rs::init_config(config).ok();
```
coderofstuff: I understand this is needed because when running the bridge in `inprocess` mode, when this is set to `unwrap` it will panic at this call (or is it somewhere else that panics?)

But why does it? What is happening on the bridge side that makes this call panic? When the node is ran on its own, it will unwrap cleanly.

README.md
```diff
@@ -383,6 +383,9 @@ wRPC
 
 </details>
 
+## Stratum Bridge
+
+Check out the [README.md](bridge/docs/README.md) for instructions on how to run the stratum bridge.
```
michaelsutton, line 388: should have a beta warning

bridge/docs/README.md
```diff
@@ -0,0 +1,73 @@
+## Stratum Bridge
+
+This repository contains a standalone Stratum bridge binary at:
```
michaelsutton, line 3: this doc should have a beta warning

LiveLaughLove13: ## Stratum Bridge changed to ## Stratum Bridge Beta

README.md
LiveLaughLove13: ## Stratum Bridge changed to ## Stratum Bridge Beta

coderofstuff (approved):

PR #794, opened 2025-12-23
Crescendo Cleanup - Part 6 (FINAL)

michaelsutton:
Cleans up the final leftovers from Crescendo.  

Notes:
- I moved all bps dependent arguments from `CrescendoParams` into a struct called `BlockrateParams` which remains a field within the main `Params`. Everything else from  `CrescendoParams` was flattened back into `Params` (tx related changes and window sizes). This struct should be useful on any future bps hardfork since you will only need to add another (forked) version of it into `Params`. I implemented `Deref/DerefMut` from `Params` into  `BlockrateParams` so that access to the subfields is transparent 
- Refactored Jtf (json test format) to mostly use rpc core structures

### DIFF
--- a/consensus/src/processes/difficulty.rs
+++ b/consensus/src/processes/difficulty.rs
@@ -95,7 +95,6 @@ impl CrescendoLogger {
         if self.steps.compare_exchange(step, step + 1, AtomicOrdering::SeqCst, AtomicOrdering::SeqCst).is_ok() {
             match step {
                 Self::ACTIVATE => {
-                    // TODO (Crescendo): finalize mainnet ascii art
                     info!(target: CRESCENDO_KEYWORD,
                         r#"
         ____                                  _             
--- a/consensus/src/processes/coinbase.rs
+++ b/consensus/src/processes/coinbase.rs
@@ -118,19 +118,11 @@ impl CoinbaseManager {
         // single output rewarding all to the current block (the "merging" block)
         let mut red_reward = 0u64;
 
-        // bps activation = crescendo activation
-        if self.bps.activation().is_active(daa_score) {
-            for red in ghostdag_data.mergeset_reds.iter() {
-                let reward_data = mergeset_rewards.get(red).unwrap();
-                if mergeset_non_daa.contains(red) {
-                    red_reward += reward_data.total_fees;
-                } else {
-                    red_reward += reward_data.subsidy + reward_data.total_fees;
-                }
-            }
-        } else {
-            for red in ghostdag_data.mergeset_reds.iter().filter(|h| !mergeset_non_daa.contains(h)) {
-                let reward_data = mergeset_rewards.get(red).unwrap();
+        for red in ghostdag_data.mergeset_reds.iter() {
+            let reward_data = mergeset_rewards.get(red).unwrap();
+            if mergeset_non_daa.contains(red) {
+                red_reward += reward_data.total_fees;
+            } else {
                 red_reward += reward_data.subsidy + reward_data.total_fees;
             }
         }


consensus/core/src/config/params.rs
```diff
@@ -489,27 +451,23 @@ impl Params {
...
-        ForkedParam::new(self.prior_max_signature_script_len, self.crescendo.max_signature_script_len, self.crescendo_activation)
+    pub fn max_signature_script_len(&self) -> usize {
+        self.crescendo.max_signature_script_len
     }
 
-    pub fn max_script_public_key_len(&self) -> ForkedParam<usize> {
-        ForkedParam::new(self.prior_max_script_public_key_len, self.crescendo.max_script_public_key_len, self.crescendo_activation)
+    pub fn max_script_public_key_len(&self) -> usize {
+        self.crescendo.max_script_public_key_len
     }
```
someone235: I think nothing here should be wrapped in a method

consensus/src/processes/pruning.rs
```diff
@@ -144,7 +143,7 @@ impl<
             if current_blue_score + pruning_depth <= ghostdag_data.blue_score {
                 break current;
             }
-            // For samples: special clamp for the period right after the fork (where we reach ceiling(P/F) steps before reaching P' depth)
+            // For samples: special clamp for the period right after a fork (where we reach ceiling(P/F) steps before reaching P' depth)
```
someone235: Should define P', or just write "reaching the new pruning depth"

consensus/src/processes/pruning.rs
```diff
@@ -222,11 +221,7 @@ impl<
         }
 
         let tip_bs = self.ghostdag_store.get_blue_score(tip).unwrap();
-        // [Crescendo]: for new nodes syncing right after the fork, it might be difficult to determine whether the
-        // new pruning depth is expected, so we use the DAA score of the pruning point itself as an indicator.
-        // This means that in the first few days following the fork we err on the side of a shorter period which is
-        // a weaker requirement
-        let pruning_depth = self.pruning_depth.after();
+        let pruning_depth = self.pruning_depth;
```
someone235: nit: Redundant `let`

consensus/core/src/config/params.rs
```diff
@@ -334,7 +334,7 @@ impl Params {
     #[inline]
     #[must_use]
     pub fn past_median_time_window_size(&self) -> usize {
-        self.crescendo.past_median_time_sampled_window_size as usize
+        self.past_median_time_window_size as usize
     }
```
someone235: I think this also shouldn't be in a method

consensus/core/src/config/params.rs
```diff
@@ -348,7 +348,7 @@ impl Params {
     #[inline]
     #[must_use]
     pub fn difficulty_window_size(&self) -> usize {
```
someone235: I think this also shouldn't be in a method

someone235 (changes requested):

consensus/src/processes/difficulty.rs
someone235, line 85: Can we delete this?

consensus/src/processes/window.rs
```diff
@@ -416,6 +416,7 @@ impl<T: GhostdagStoreReader, U: BlockWindowCacheReader + BlockWindowCacheWriter,
                     // [Crescendo]: for each block in the original window save its selected parent as well
                     // since it is required for checking whether the block was activated (when building the
                     // window by the syncee or when rebuilding following pruning)
+                    // TODO (relaxed): remove this once most of the network upgrades from crescendo version
                     cover.insert(self.ghostdag_store.get_selected_parent(merged).unwrap());
```
someone235, line 420: Can't we delete this line? Or do we want to keep it for future hardforks? If yes, update comment accordingly

someone235 (changes requested):

consensus/src/processes/difficulty.rs
michaelsutton: I'll do a separate dead code cleanup soon and we can negotiate there

consensus/src/processes/window.rs
michaelsutton: That's exactly the todo (I think it's still early to remove assuming the syncee can be v1.0.0)

PR #795, opened 2025-12-23
Catchup flow tweak - misleading warning fix 

freshair18:
More descriptive warning when syncing with an honest but outdated peer

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,73 @@
+# Contributing to Kaspa
+
+Thanks for your interest in contributing to Kaspa!
+
+We welcome contributions of all sizes and there are many opportunities to contribute at any level — from clarifying documentation and fixing small bugs to implementing full features and reviewing pull requests.
+
+Reach out to `@Node Developers` in Discord in the [#development](https://discord.com/channels/599153230659846165/755890250643144788) channel.
+
+Follow along the R&D Telegram group [@kasparnd](https://t.me/c/kasparnd).
+
+## Quick summary
+
+- Open a GitHub Issue or Pull Request to start any discussion. Use Issues for design or spec discussions and PRs when you have code to share.
+- Look for `good first issue` if you're getting started; these are intentionally approachable.
+- Write detailed pull requests descriptions: explain what you changed, why, design decisions, and any trade-offs.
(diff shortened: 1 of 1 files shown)


freshair18:
Based on @michaelsutton's input, I've instead reworked the code so that is pruning sample is checked internally in the catchup flow, where it is most relavant, and do not attempt to filter this ahead of time.

freshair18:
@someone235 It makes sense for you to review really, since it is a tweak of the pruning catch up PR.

consensus/src/consensus/mod.rs
```diff
@@ -1372,25 +1375,18 @@ impl ConsensusApi for Consensus {
...
     }
 
-    fn verify_is_pruning_sample(&self, pruning_candidate: Hash) -> ConsensusResult<()> {
-        if pruning_candidate == self.config.genesis.hash {
-            return Ok(());
-        }
+    fn verify_is_pruning_sample(&self, pruning_candidate: Hash) -> bool {
         let Ok(candidate_ghostdag_data) = self.get_ghostdag_data(pruning_candidate) else {
-            return Err(ConsensusError::General("pruning candidate missing ghostdag data"));
+            return false;
```
someone235: This should be an error, not false, since we can't determine in this case.

I suggest to do one of the following:

1. Revert this function to the way it was before.
2. Remove it and inline its use in its only caller site, `get_and_verify_path_to_new_pruning_point`

protocol/flows/src/ibd/flow.rs
someone235: Since it might belong to a block with pruned reachability data, add this option to the error message

someone235 (changes requested):

someone235 (approved):

protocol/flows/src/ibd/flow.rs
```diff
@@ -233,37 +240,38 @@ impl IbdFlow {
             info!("current sink is:{}", sink);
             info!("current pruning point is:{}", pruning_point);
             if consensus.async_is_chain_ancestor_of(pruning_point, highest_known_syncer_chain_hash).await? {
+                const SYNCER_LAGGING_TOLERANCE: u64 = 3;
+                let recent_pruning_points = consensus.async_get_n_last_pruning_points(SYNCER_LAGGING_TOLERANCE).await;
```
freshair18: hmm maybe we can move these two lines after the first if.

PR #800, opened 2026-01-01
Refactor pruning proof validate algo

michaelsutton:


protocol/flows/src/ibd/flow.rs
```diff
@@ -238,6 +233,13 @@ impl IbdFlow {
             info!("current sink is:{}", sink);
             info!("current pruning point is:{}", pruning_point);
             if consensus.async_is_chain_ancestor_of(pruning_point, highest_known_syncer_chain_hash).await? {
+                /// Categorizes the syncer's pruning point position relative to local
```
coderofstuff, line 236: why did this need to move in here?

michaelsutton: in my view it's a local enum used for code clarity

michaelsutton: having it local signals this to the reader

coderofstuff: It probably shouldn't be `pub` then

coderofstuff (approved):

protocol/flows/src/ibd/flow.rs
michaelsutton: done

consensus/src/processes/pruning_proof/mod.rs
```diff
@@ -381,7 +381,7 @@ trait GhostdagReaderExt
 where
     Self: GhostdagStoreReader,
 {
-    /// Extension method to get the block at blue depth from `high` via this store reader. Used by build and validate.
+    /// Extension method to get the block at blue `depth` from `high` via this store reader. Used by build and validate.
```
someone235: Should be "blue depth `depth`"

someone235 (changes requested):

consensus/src/processes/pruning_proof/mod.rs
michaelsutton: done

PR #808, opened 2026-01-06
Fix: (and update) wasm-bindgen

D-Stacks:
I have noticed some issues regarding wasm-bindgen requiring a new version of getrandom i.e. v0.3.4, (after rebuilding the cargo.lock from scratch). 

After updating it from v0.2.10 to v0.3.4, specifying the new `wasm_js` feature instead of the seemingly deprecated `js` feature, and updating  wasm-bindgen from v2.100 to v2.106.. some other errors and issues occurred, mostly revolving around an old work-around left from #691, which have here been removed, as well as a necessity  to add `--enable-nontrapping-float-to-int` to the wasm-opts. Which still left some issues around not being able to handle `Symbol.dispose () void` in the type docs, which were mitigated by explictly specifying `"tsconfig": "./tsconfig.json"` in the typdoc.json as well as supplying  `"lib": ["ES2020", "esnext.disposable"]` in the tsconfig.json.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -332,7 +332,7 @@ jobs:
       - name: Install NodeJS
         uses: actions/setup-node@v4
         with:
-          node-version: '24'
+          node-version: '20'
 
       - name: Install NodeJS dependencies
         run: npm install --global typedoc typescript


D-Stacks:
> LTGM. I am approving this PR, but we need to ask JS developers to test this in the wild (binaries can be downloaded from PR actions)

you know of any js devs we can maybe tag / contact, to make them aware of this?

IzioDev:
Question: Is there a way to only define `getrandom` dependency when the target is wasm? (or feature is wasm32-sdk)

biryukovmaxim:
> Question: Is there a way to only define `getrandom` dependency when the target is wasm? (or feature is wasm32-sdk)

Yes you can specify dependency or specific target. But there's no way to enable feature for specific target

```toml
[dependencies]
[target.wasm32-unknown-unknown.dependencies]
getrandom = "*"
```

IzioDev:
> > Question: Is there a way to only define `getrandom` dependency when the target is wasm? (or feature is wasm32-sdk)
> 
> Yes you can specify dependency or specific target. But there's no way to enable feature for specific target
> 
> ```toml
> [dependencies]
> [target.wasm32-unknown-unknown.dependencies]
> getrandom = "*"
> ```

Thanks for the insight. I was reading the [getrandom documentation here](https://docs.rs/getrandom/latest/getrandom/#webassembly-support) and it says:
> We do not include support for this target in the default configuration because our JS backend (supporting web browsers, web workers and Node.js v19 or later) requires [wasm-bindgen](https://github.com/rustwasm/wasm-bindgen), bloating Cargo.lock and potentially breaking builds on non-web WASM platforms.

IzioDev:
LGTM, from a practical standpoint.
Tested node and web, also tested random functionalities such as `Mnemonic.random()`

IzioDev (approved):

PR #813, opened 2026-01-12
Covenant id scheme + Seq commit opcode

someone235:


consensus/core/src/tx.rs
```diff
@@ -121,14 +123,22 @@ impl std::fmt::Debug for TransactionInput {
...
     pub fn new(value: u64, script_public_key: ScriptPublicKey) -> Self {
-        Self { value, script_public_key }
+        Self { value, script_public_key, cov_out_info: None }
     }
 }
 
+#[derive(Debug, Serialize, Deserialize, Clone, PartialEq, Eq, BorshSerialize, BorshDeserialize, Copy)]
+#[serde(rename_all = "camelCase")]
+pub struct CovOutInfo {
+    pub authorizing_input: u16,
```
IzioDev, line 153: Is this intended to be chosen by the transaction?
2 inputs -> 1 output, can the output pick its parent "authority"?

michaelsutton: Yes, but it must be pre-planned by the script. I.e., we expect the covenant script to verify all its authorized outputs (and corresponding opcodes to access these "child" outputs will be available). 
In general, for a covenant 2:1 merge operation, I'd expect that the second input will delegate authorization to a primary input.

consensus/core/src/tx.rs
```diff
@@ -121,14 +123,22 @@ impl std::fmt::Debug for TransactionInput {
 pub struct TransactionOutput {
     pub value: u64,
     pub script_public_key: ScriptPublicKey,
+    pub cov_out_info: Option<CovOutInfo>,
```
michaelsutton: can we call it simply `covenant` ? (and `CovenantInfo` for the type). I think the inner field names speak for themselves.

consensus/core/src/mass/mod.rs
```diff
@@ -77,29 +77,23 @@ pub fn utxo_plurality(spk: &ScriptPublicKey) -> u64 {
     // Hence, all standard SPKs end up with a plurality of 1.
     const UTXO_UNIT_SIZE: usize = 100;
 
-    (UTXO_CONST_STORAGE + spk.script().len()).div_ceil(UTXO_UNIT_SIZE) as u64
+    (UTXO_CONST_STORAGE + spk.script().len() + if has_covenant_id { HASH_SIZE } else { 0 }).div_ceil(UTXO_UNIT_SIZE) as u64
```
biryukovmaxim, line 80: in what form the size is calculated? Option<Hash> is 33 bytes

consensus/src/processes/transaction_validator/tx_validation_in_utxo_context.rs
```diff
@@ -169,6 +170,34 @@ impl TransactionValidator {
...
+                    if kaspa_consensus_core::hashing::covenant_id::covenant_id(authorizing_input.previous_outpoint)
+                        != cov_out_info.covenant_id
+                    {
+                        return Err(TxRuleError::WrongCovenantId(i));
+                    }
+                }
+            }
+        }
+        Ok(())
+    }
```
biryukovmaxim:
```suggestion
    fn check_covenant_info(&self, tx: &impl VerifiableTransaction, block_daa_score: u64) -> TxResult<()> {
        if !self.covenants_activation.is_active(block_daa_score) {
            return Ok(());
        }

        for (i, output) in tx.outputs().iter().enumerate() {
            let Some(cov_out_info) = &output.cov_out_info else { continue };
            let Some(utxo_entry) = tx.utxo(cov_out_info.authorizing_input as usize) else {
                // TODO: Change to another error
                return Err(TxRuleError::MissingTxOutpoints);
            };
            match utxo_entry.covenant_id {
                Some(covenant_id) if covenant_id != cov_out_info.covenant_id => return Err(TxRuleError::WrongCovenantId(i)),
                Some(_) => {}
                None => {
                    let authorizing_input = &tx.inputs()[cov_out_info.authorizing_input as usize]; // It's guaranteed to exist from the earlier check on the UTXO entry.
                    if kaspa_consensus_core::hashing::covenant_id::covenant_id(authorizing_input.previous_outpoint)
                        != cov_out_info.covenant_id
                    {
                        return Err(TxRuleError::WrongCovenantId(i));
                    }
                }
            }
        }
        Ok(())
    }
```

crypto/txscript/src/lib.rs
```diff
@@ -97,6 +98,7 @@ pub struct TxScriptEngine<'a, T: VerifiableTransaction, Reused: SigHashReusedVal
     num_ops: i32,
     runtime_sig_op_counter: RuntimeSigOpCounter,
     flags: EngineFlags,
+    cov_out_indices: Option<Lazy<Vec<Vec<usize>>, Box<dyn FnOnce() -> Vec<Vec<usize>> + 'a>>>,
```
biryukovmaxim: LazyCell is presented in std
but id suggest using requiredField OnceCell, that can be settled when its needed(wont be ever called from standalone scripts) and it would'nt require dyn and boxing

rpc/core/src/model/tx.rs
```diff
@@ -75,8 +84,9 @@ impl Deserializer for RpcUtxoEntry {
         let script_public_key = load!(ScriptPublicKey, reader)?;
         let block_daa_score = load!(u64, reader)?;
         let is_coinbase = load!(bool, reader)?;
+        let covenant_id = load!(Option<RpcHash>, reader)?;
```
biryukovmaxim: this one requires version update. In borsh option tag is 1 byte. none needs that byte otherwise it will return error during borsh-deser

rpc/core/src/model/optional/tx.rs
```diff
@@ -102,8 +106,9 @@ impl Deserializer for RpcOptionalUtxoEntry {
                 let block_daa_score = load!(Option<u64>, reader)?;
                 let is_coinbase = load!(Option<bool>, reader)?;
                 let verbose_data = deserialize!(Option<RpcOptionalUtxoEntryVerboseData>, reader)?;
+                let covenant_id = load!(Option<RpcHash>, reader)?;
```
biryukovmaxim, line 109: this one requires version bump. In borsh option tag is 1 byte. none needs that byte otherwise it will return error during borsh-deser

rpc/core/src/model/tx.rs
biryukovmaxim: it also needs serializer impl

rpc/core/src/model/optional/tx.rs
biryukovmaxim: it also needs serializer impl

consensus/core/src/tx.rs
someone235: Note: we should probably add db upgrade logic for the new fields

mining/src/mempool/model/utxo_set.rs
```diff
@@ -40,7 +40,7 @@ impl MempoolUtxoSet {
 
         for (i, output) in transaction.tx.outputs.iter().enumerate() {
             let outpoint = TransactionOutpoint::new(transaction_id, i as u32);
-            let entry = UtxoEntry::new(output.value, output.script_public_key.clone(), UNACCEPTED_DAA_SCORE, false);
+            let entry = UtxoEntry::new(output.value, output.script_public_key.clone(), UNACCEPTED_DAA_SCORE, false, None);
```
michaelsutton: should map `output.covenant`  instead of None
```suggestion
            let entry = UtxoEntry::new(output.value, output.script_public_key.clone(), UNACCEPTED_DAA_SCORE, false, output.covenant.map(|x| x.covenant_id));
```

protocol/p2p/proto/p2p.proto
```diff
@@ -176,6 +176,7 @@ message UtxoEntry {
   ScriptPublicKey scriptPublicKey = 2;
   uint64 blockDaaScore = 3;
   bool isCoinbase = 4;
+  Hash covenantId = 5;
 }
```
michaelsutton, line 187: Missing P2P changes for transaction output

consensus/core/src/tx.rs
```diff
@@ -121,14 +132,24 @@ impl std::fmt::Debug for TransactionInput {
 pub struct TransactionOutput {
     pub value: u64,
     pub script_public_key: ScriptPublicKey,
+    pub covenant: Option<CovenantBinding>,
 }
 
 impl TransactionOutput {
     pub fn new(value: u64, script_public_key: ScriptPublicKey) -> Self {
```
michaelsutton, line 139: add `with_covenant_binding` and use it (will be needed for eg P2P conversion)

michaelsutton:
Merging this with several pending todos -- mainly around client and rpc code -- so that multiple people can work on this branch to fix the issues and build upon it

PR #821, opened 2026-01-15
Custom identifier

LiveLaughLove13:
Bridge: inprocess and external-mode coinbase tag + missing inprocess CLI args
Inprocess mode base tag: Kaspa Stratum Bridge
External mode base tag: KSB-External

Optional coinbase_tag_suffix is sanitized and appended as /SUFFIX in both modes

Missing inprocess CLI args (forwarded to embedded kaspad)

Added: --rpclisten-json, --listen, --ram-scale, --outpeers, --maxinpeers, --rpcmaxclients, --perf-metrics, --perf-metrics-interval-sec

Config docs
Updated bridge/config.yaml comments to document both base tags + suffix examples

### DIFF
--- a/bridge/config.yaml
+++ b/bridge/config.yaml
@@ -39,13 +39,11 @@ extranonce_size: 2
 
 # Optional: append a custom identifier to the built-in coinbase tag (applies to BOTH node modes).
 # You CANNOT change/override the built-in base tag; you can only append a suffix.
-# Built-in base tag depends on node mode:
-# - Inprocess mode: "Kaspa Stratum Bridge"
-# - External mode:  "KSB-External"
+# Built-in base tag is fixed:
+# - "RK-Stratum"
 # Set this to a short identifier (no leading slash). It will be appended as `/SUFFIX`.
 # Examples (if coinbase_tag_suffix: "supertypo"):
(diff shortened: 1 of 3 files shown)


supertypo:
I've done a quick review of the custom identifier implementation:
Seems to work similar with our without the identifier specified, so the addition of the custom suffix argument is fine.

I mined to it on testnet-12 using SRBMiner-Multi v.2.4.4 (latest which support Kaspa, gpu only) using a rtx 3090 and noticed the following which requires follow up work:

General:
* I would like the config file to be optional, all settings hould be accessible through *command line arguments*. defaults should be sensible for the given networkId
* If possible a single stratum port should be enabled by default with vardiff on (starting sensibly low for the given network id). Ie ks0 ultra type difficulty for mainnet and cpu type for testnets. Multiple ports only if strictly neccessary (are we sure its needed?).
* output should be cleaned up, no emojis - reduce / remove use of colored output (except block win line maybe).

Testnet mining:
* Initial difficulty in config.yaml is too high for testnets (cpu/gpu mining)
* block_wait_time of 1000ms is too long for testnets(?).
* I was unable to produce more than ~1 block / sec on the 3090 on tn12, where it should totally dominate (by 100x) if the miner / bridge combo worked properly.

PR #822, opened 2026-01-15
Validate full levels have tips with 2M blue score

coderofstuff:


consensus/core/src/errors/pruning.rs
```diff
@@ -30,6 +30,9 @@ pub enum PruningImportError {
     #[error("the pruning proof selected tip {0} at level {1} is not a parent of the pruning point on the same level")]
     PruningProofSelectedTipNotParentOfPruningPoint(Hash, BlockLevel),
 
+    #[error("the pruning proof selected tip {0} at level {1} blue score {2} <= 2M")]
```
michaelsutton: nit
```suggestion
    #[error("the pruning proof selected tip {0} at level {1} blue score {2} < 2M and root is not genesis")]
```

PR #823, opened 2026-01-16
P2P block relay flow: Patch - Check and qualify low pow blocks as a protocol violation. 

D-Stacks:
Currently low pow blocks supplied to a node, while being outside the min. required blue work threshold to enter consensus, still get downloaded from the node without any verification.

This provides a p2p ddos attack vector to the node whereby a malicious actor can signal and send an unlimited amount of forged blocks to a target node, thereby wasting, or in worst case, depleting the target node's bandwidth.  

This PR attempts to patch this condition by calculating an expected target difficulty at the supplied daa score, and comparing the supplied block's pow against the expected target with a generous 50% tolerance. This ensures no false positive detections, while ensuring significant pow must have been applied to the supplied block. 

Note: It rounds the expected target difficulty to the nearest daa found on the node's virtual chain, lower bounded to the last pruning point.

### DIFF
--- a/consensus/src/pipeline/virtual_processor/utxo_inquirer.rs
+++ b/consensus/src/pipeline/virtual_processor/utxo_inquirer.rs
@@ -333,7 +333,7 @@ impl VirtualStateProcessor {
                             closest_accepting_hash = Some((hash, cur_diff));
                         }
                     }
-                    high_index = mid.saturating_sub(1);
+                    high_index = mid - 1;
                 }
                 cmp::Ordering::Less => {
                     if round_to_closest {


consensus/src/pipeline/virtual_processor/utxo_inquirer.rs
```diff
@@ -280,7 +281,9 @@ impl VirtualStateProcessor {
         &self,
         target_daa_score: u64,
         retention_period_root_hash: Hash,
+        round_to_closest: bool,
     ) -> UtxoInquirerResult<Hash> {
```
IzioDev, line 285: This is a personal take, and not a general recommendation in any way.

I tend to push creating new function instead of introducing flags that change the internal behavior.
This is for the following reasons:
1. it complexifies the internal logic, that now becomes more difficult when doing a top down analysis.
2. it forces reviewers to check that current behavior hasn't changed
3. function now became more difficult to unit test (side-effect of 1.)

D-Stacks: Idk, I tend to try and keep verbosity in check, rather have a few more Params, instead of having x amount of functions.. that all basically do the same thing. But all personal preferences. But that being said, I will most likely close this solution in favor of another solution anyway (that won't need this query)..

D-Stacks:
closing this in favor of #829

Issue #826, opened 2026-01-17
RK 1.1.0 BUG。I used Release Candidate v1.1.0 (RC 1) to sync the mainnet nodes, but my CLI wallet shows a zero balance. I checked the node data and found that the DAA height is consistent, but the UTXO sets are completely different.


kasfans:
**Describe the bug**

**To Reproduce**

**Expected behavior**

**Screenshots**

**Desktop (please complete the following information):**
 - OS: [e.g. Ubuntu 20.04]
 - Kaspad version: [e.g. 0.12.11]

**Additional context**

kasfans:
I used Release Candidate v1.1.0 (RC 1) to sync the mainnet nodes, but my CLI wallet shows a zero balance. I checked the node data and found that the DAA height is consistent, but the UTXO sets are completely different.

IzioDev:
Hello @kasfans,

Can you confirm the bug is no longer existent in the latest release? (rc-3)

Issue #825, opened 2026-01-17
IBD: indexing UTXOs not complete


vinigo2025:
node 1.0.2 have an issue with indexing UTXOs, the database after IBD contains small chunk of UTXOs

using kaspa-wrpc-simple-client-example:
```
let supp2 = client.get_coin_supply().await?;
println!("circulating_sompi = {}", supp2.circulating_sompi);
```
```
Node version: 1.0.2
Network: mainnet
Node is synced: true
Node is indexing UTXOs: true
Block count: 1418410
Header count: 1418410
Pruning point hash: 615a6e79f67872ab251d2d011d3b812f258c7f7faa26e81509d4b141789e2768
Virtual DAA score: 332412110
Sink: edaf8aa6a42b649fc39055067fe2014077d04e11c23330ccb70dcb1e231ee119

circulating_sompi = 491550134577456
```

 - OS: Ubuntu 24.04
 - Kaspad version: 1.0.2

vinigo2025:
duplicate, to close

vinigo2025:
fixed

PR #827, opened 2026-01-17
Proper utxoindex resyncing post utxo download

freshair18:
Past version has not automatically triggered a resync of the utxo version after downloading a new utxoset. Previously this was triggered when a new consensus is committed, but in recent versions a commit happens before downloading the utxoset.

Addresses recent  reports in https://github.com/kaspanet/rusty-kaspa/issues/825 and https://github.com/kaspanet/rusty-kaspa/issues/826


components/consensusmanager/src/lib.rs
```diff
@@ -171,6 +171,12 @@ impl ConsensusManager {
     pub fn delete_staging_entry(&self) {
         self.factory.delete_staging_entry();
     }
+    pub fn reset_handlers(&self) {
```
michaelsutton: a better name would be `invoke_consensus_reset_handlers` + a short doc for this function

components/consensusmanager/src/lib.rs
```diff
@@ -206,10 +212,7 @@ impl StagingConsensus {
         prev.ctl.stop();
         g.current.ctl.make_active();
         drop(g);
-        let handlers = self.manager.inner.read().consensus_reset_handlers.iter().cloned().collect_vec();
-        for handler in handlers {
-            handler.handle_consensus_reset();
-        }
+        self.manager.reset_handlers();
```
michaelsutton: should we still call this here?

protocol/flows/src/ibd/flow.rs
```diff
@@ -592,6 +592,7 @@ impl IbdFlow {
         consensus.async_clear_pruning_utxo_set().await; // this deletes the old pruning utxoset and also sets the pruning utxo as invalidated
         self.sync_pruning_point_utxoset(consensus, pruning_point).await?;
         consensus.async_set_pruning_utxoset_stable().await; //  only if the function has reached here, will the utxo be considered "final"
+        self.ctx.consensus_manager.reset_handlers();
```
michaelsutton: perhaps add a clear comment above this call explaining the utxoindex case so that future maintainers remain aware

michaelsutton: important: use `spawn_blocking`

components/consensusmanager/src/lib.rs
freshair18: I argue yes?  whatever the utxoindex previously said, it is now rather misleading, and an empty one is more representative (I am working under the assumption that resyncing to empty is very fast... correct me if wrong.).

More generally, we should keep in mind other handlers but the utxoindex handler which might potentially need a reset at this moment - that said, it does seem like this actually is the only handler, so if it is the way it is this structure is  in general a bit off and deserving of a rewrite later on.

michaelsutton: Yeah it should be fast, so keep it. 

I don't think there's any reason to rewrite anything.

michaelsutton (approved):

PR #830, opened 2026-01-19
fix(wasm/build): revert build configuration

IzioDev:
#808 introduced changes regarding the wasm build configuration (getrandom upgrade, wasm bindgen upgrade and optimization flag changes). This produced bugs at runtime on the wasm build (release only), likely indicating wasm opt isn't configured properly with this new configuration.

To not bottleneck the upcoming release(s), it is proposed to revert the changes (+improving DX, when building from `/wasm`). 

How to test:
1. `cargo clean`
2. `cd wasm`
3. `wasm-pack build --weak-refs --target nodejs --out-name kaspa --out-dir nodejs/kaspa --features wasm32-sdk` (build the wasm sdks for nodejs)
5. (optional): install nodejs dependencies from `./examples`: `cd examples && npm i`
6. `cd examples/nodejs/javascript/general`
6. `node resolver.js`

Should not show this error:
```
D:\Dev\kaspa\rusty-kaspa\wasm\nodejs\kaspa\kaspa.js:14004
    const ret = getObject(arg0).data;
                               ^

TypeError: Cannot read properties of undefined (reading 'data')
    at exports.__wbg_data_8bf4ae669a78a688 (D:\Dev\kaspa\rusty-kaspa\wasm\nodejs\kaspa\kaspa.js:14004:32)
    at wasm://wasm/02a5c1da:wasm-function[4840]:0x6ec81f
    at wasm://wasm/02a5c1da:wasm-function[2775]:0x58736e
    at wasm://wasm/02a5c1da:wasm-function[9144]:0x7f1312
    at __wasm_bindgen_func_elem_17796 (D:\Dev\kaspa\rusty-kaspa\wasm\nodejs\kaspa\kaspa.js:319:10)
    at W3CWebSocket.real [as onmessage] (D:\Dev\kaspa\rusty-kaspa\wasm\nodejs\kaspa\kaspa.js:194:20)
    at W3CWebSocket._dispatchEvent [as dispatchEvent] (D:\Dev\kaspa\rusty-kaspa\wasm\examples\node_modules\yaeti\lib\EventTarget.js:107:17)
    at W3CWebSocket.onMessage (D:\Dev\kaspa\rusty-kaspa\wasm\examples\node_modules\websocket\lib\W3CWebSocket.js:246:18)
    at WebSocketConnection.<anonymous> (D:\Dev\kaspa\rusty-kaspa\wasm\examples\node_modules\websocket\lib\W3CWebSocket.js:205:19)
    at WebSocketConnection.emit (node:events:518:28)

```

### DIFF
--- a/wasm/Cargo.toml
+++ b/wasm/Cargo.toml
@@ -68,7 +68,8 @@ targets = [
 ]
 folder = "setup"
 
-
+# Workaround for Rust 1.87.0
+# https://github.com/rust-lang/rust/issues/141048
 [package.metadata.wasm-pack.profile.release]
 # This optimization configuration requires "-Ctarget-cpu=mvp" Rustflag (only enable W3C MVP feature set)
 # This flag is automatically used either via (release optimized) build script at root


D-Stacks:
kind of only a half-review comment, but just something i realized re-reading the changes, in #691 

```
# Workaround for Rust 1.87.0
# https://github.com/rust-lang/rust/issues/141048
export RUSTFLAGS=-Ctarget-cpu=mvp
```

was added as comments over the rust flags, maybe we should re-instate them as well. that being said, maybe this caused issues for builds using that rust version, nothing in my PR addressed that fact, perhaps simply moving the minimal rust version above 1.87 could solve the issue at hand (simply as a hunch).

wasm/.cargo/config.toml
```diff
@@ -0,0 +1,2 @@
+[build]
+rustflags = ["-Ctarget-cpu=mvp"]
```
D-Stacks, line 2: not sure, but i think setting this should suffice? not sure we need to export rust flags on the build scripts additionally. but might be better to be safe then sorry, just thinking it might also complicate debugging further down the road, if one overrides the other.

D-Stacks:
I also had changes in `wasm/build/docs` and `wasm/examples/nodejs/typescript/tsconfig.json` just wanted to point that out in case we want to go for a full revert.

wasm/.cargo/config.toml
IzioDev: It allows environment that either cannot or doesn't want to go through the scripts (build-*) to compile without issue using wasm-pack directly.
Basically it adds the flag when executing cargo build from this directory (doesn't harm)

IzioDev:
> I also had changes in `wasm/build/docs` and `wasm/examples/nodejs/typescript/tsconfig.json` just wanted to point that out in case we want to go for a full revert.

I saw, but I thought this was general improvements regardless, what do you think?

D-Stacks:
> > I also had changes in `wasm/build/docs` and `wasm/examples/nodejs/typescript/tsconfig.json` just wanted to point that out in case we want to go for a full revert.
> 
> I saw, but I thought this was general improvements regardless, what do you think?

I'll trust you on that, I don't have great affinity for, or knowledge of, node.js (I originally just added them, as i needed these for the build scripts to work after the update).

wasm/.cargo/config.toml
D-Stacks: I agree, it shouldn't harm, I guess that is the main thing.

D-Stacks:
> kind of only a half-review comment, but just something i realized re-reading the changes, in #691
> 
> ```
> # Workaround for Rust 1.87.0
> # https://github.com/rust-lang/rust/issues/141048
> export RUSTFLAGS=-Ctarget-cpu=mvp
> ```
> 
> was added as comments over the rust flags, maybe we should re-instate them as well. that being said, maybe this caused issues for builds using that rust version, nothing in my PR addressed that fact, perhaps simply moving the minimal rust version above 1.87 could solve the issue at hand (simply as a hunch).

I think i want this back in though, then also for the new `config.toml`, and the `wasm/Cargo.toml` had this in regards to `--enable-bulk-memory`. version 1.87 is still over our minimal required rust version, and i see no reason why this work-around no longer applies. I had the assumption that with updating wasm-bindgen this was no longer relevant (and I had to change this for the update to work anyway), but this might never actually have been the case (who knows), but if we revert everything back, I guess there is no reason to remove these comments.

PR #831, opened 2026-01-20
Rk stratum2.0

LiveLaughLove13:
# Bridge Configuration Options

## RKStratum 2.0 — What’s New (vs CustomIdentifier)

This is a user-facing summary of the main upgrades that landed in the `RKStratum2.0` branch compared to `CustomIdentifier`.

## Web UI & API

- **Built-in web dashboard**
  - Served directly by the bridge (no external UI service required).
  - Includes a main dashboard and a raw view.
- **New HTTP API endpoints for monitoring**
  - `/api/status` for basic node/bridge status.
  - `/api/stats` for summarized worker/block/share stats.
  - `/metrics` for Prometheus scraping.

## CLI & Configuration Quality-of-Life

- **Full CLI support for config values**
  - Most `config.yaml` settings can now be overridden via CLI flags.
- **`--config <path>` support + smarter config discovery**
  - The bridge can locate config relative to the executable in common deployment layouts.
- **More automation-friendly argument parsing**
  - Boolean parsing supports common forms (`true/false`, `1/0`, `yes/no`, `on/off`).
  - Port inputs accept `:PORT` or `HOST:PORT`.

## Multi-Instance Improvements

- **Improved multi-instance configuration & overrides**
  - Adds `--instance` syntax for defining multiple stratum listeners from the CLI.
  - Adds validation for duplicate ports and missing required instance fields.

## Observability & Metrics

- **Expanded Prometheus metrics + better dashboard data**
  - More complete worker/share/block instrumentation and better aggregation.
- **Optional global Web UI bind**
  - A single port can expose aggregated stats across all instances.

## Reliability / Operator Experience

- **Graceful shutdown**
  - Coordinated shutdown with a drain window and forced-exit behavior on repeated interrupts.
- **Logging fixes and improvements**
  - Log files are no longer created empty.
  - Log file naming standardized to `RKStratum_*.log`.
  - Improved console formatting (including better readability of periodic stats output).
- **Health check server**
  - Optional lightweight health endpoint via `health_check_port`.

## Optional / Advanced Features

- **Internal CPU miner (feature-gated)**
  - Optional built-in CPU miner mode controlled via CLI flags when compiled with the `rkstratum_cpu_miner` feature.

## Packaging & Deployment

- **Stratum bridge Docker support**
  - dedicated `Dockerfile.stratum-bridge`.
- **Release packaging includes the bridge binary**
  - Linux/Windows/macOS release assets build and ship `stratum-bridge` alongside other binaries.

## Quick Start

- **Config file**
  - Default: `config.yaml` (current directory).
  - Repository example: `bridge/config.yaml`.
- **Run (external node)**
  - `stratum-bridge --config bridge/config.yaml --node-mode external`
- **Run (in-process node)**
  - `stratum-bridge --config bridge/config.yaml --node-mode inprocess -- <kaspad args...>`
- **Dashboard (optional)**
  - Set `web_port` (or use `--web-port`) and open: `http://127.0.0.1:<PORT>/`

## Global Settings (shared by all instances)

| Setting | Type | Default | Description |
|---|---|---|---|
| `kaspad_address` | String | `"localhost:16110"` | Kaspa node gRPC address. All instances use the same node. Format: `"HOST:PORT"` or `"grpc://HOST:PORT"`. |
| `block_wait_time` | Integer (milliseconds) | `1000` | How long to wait between checking for new block templates. |
| `print_stats` | Boolean | `true` | Print mining statistics to the console. |
| `log_to_file` | Boolean | `true` | Default log-to-file setting (can be overridden per-instance). |
| `health_check_port` | String | `""` (disabled) | Global health check server port. Leave empty to disable. |
| `web_port` | String | `""` (disabled) | Optional global Web UI / aggregated metrics port. Examples: `":3030"`, `"0.0.0.0:3030"`. |
| `var_diff` | Boolean | `true` | Enable variable difficulty (can be overridden per-instance). |
| `shares_per_min` | Integer | `20` | Target shares per minute for variable difficulty (can be overridden per-instance). |
| `var_diff_stats` | Boolean | `false` | Print variable difficulty statistics (can be overridden per-instance). |
| `extranonce_size` | Integer | `0` | Extranonce size (auto-detected per client; this is for backward compatibility). |
| `pow2_clamp` | Boolean | `false` | Enable power-of-2 difficulty clamping (can be overridden per-instance). |
| `coinbase_tag_suffix` | String | `""` (empty) | Optional suffix appended to the fixed base coinbase tag `"RK-Stratum"`. Results in `"RK-Stratum/<suffix>"`. |

---

## Instance Settings (each instance can have its own)

| Setting | Type | Default | Description |
|---|---|---|---|
| `stratum_port` | String | `":5555"` | **Required.** Stratum port for this instance. Must be unique across instances. Can be `":PORT"` or `"HOST:PORT"`. |
| `min_share_diff` | Integer | `8192` | **Required.** Minimum share difficulty for this instance. |
| `prom_port` | String | `None` (disabled) | Optional Prometheus port for this instance. Can be `":PORT"` or `"HOST:PORT"`. |
| `log_to_file` | Boolean | `None` (inherits global) | Optional per-instance log-to-file setting. If not set, uses the global `log_to_file`. |
| `block_wait_time` | Integer (milliseconds) | `None` (inherits global) | Optional per-instance block template polling interval. |
| `extranonce_size` | Integer | `None` (inherits global) | Optional per-instance extranonce size override. |
| `var_diff` | Boolean | `None` (inherits global) | Optional per-instance variable difficulty override. |
| `shares_per_min` | Integer | `None` (inherits global) | Optional per-instance target shares per minute override. |
| `var_diff_stats` | Boolean | `None` (inherits global) | Optional per-instance variable difficulty statistics override. |
| `pow2_clamp` | Boolean | `None` (inherits global) | Optional per-instance power-of-2 difficulty clamping override. |

---

## Notes and Behavior

- **Multi-instance mode**: If the `instances` array exists in `config.yaml`, the bridge runs in multi-instance mode. Otherwise, it runs in single-instance mode using the global settings.
- **Port formatting**: `stratum_port` and `prom_port` can be specified as `":PORT"` or `"HOST:PORT"`. The bridge will prepend `"0.0.0.0"` if only a port is provided.
- **Coinbase tag**: The base tag is always `"RK-Stratum"`. You can only append a suffix via `coinbase_tag_suffix`. The suffix is sanitized to alphanumeric characters, `.`, `_`, and `-`, and is limited to 64 characters.
- **Variable difficulty**: When enabled, the bridge adjusts the difficulty for miners based on their hashrate to target `shares_per_min`.
- **Logging**: If `log_to_file` is enabled, logs are written under:
  - Windows: `%LOCALAPPDATA%\kaspa-stratum-bridge\logs\RKStratum_<timestamp>.log`
  - Linux/macOS: `~/.kaspa-stratum-bridge/logs/RKStratum_<timestamp>.log`

---

## Configuration Options with CLI Arguments

| Setting | Type | Default | CLI Argument | TODO Status | Description |
|---|---|---|---|---|---|
| `kaspad_address` | String | `"localhost:16110"` | `--kaspad-address <HOST:PORT>` | - | Kaspa node gRPC address. All instances use the same node. |
| `block_wait_time` | Integer (ms) | `1000` | `--block-wait-time <MILLISECONDS>` | - | How long to wait between checking for new block templates. |
| `print_stats` | Boolean | `true` | `--print-stats <true\|false>` | - | Print mining statistics to the console. |
| `log_to_file` | Boolean | `true` | `--log-to-file <true\|false>` | - | Default log-to-file setting (can be overridden per-instance). |
| `health_check_port` | String | `""` (disabled) | `--health-check-port <PORT>` | - | Global health check server port. Leave empty to disable. |
| `web_port` | String | `""` (disabled) | `--web-port <HOST:PORT>` | - | Global Web UI / aggregated metrics server port (optional). |
| `var_diff` | Boolean | `true` | `--var-diff <true\|false>` | - | Enable variable difficulty (can be overridden per-instance). |
| `shares_per_min` | Integer | `20` | `--shares-per-min <COUNT>` | - | Target shares per minute for variable difficulty. |
| `var_diff_stats` | Boolean | `false` | `--var-diff-stats <true\|false>` | - | Print variable difficulty statistics. |
| `extranonce_size` | Integer | `0` | `--extranonce-size <SIZE>` | - | Extranonce size (auto-detected per client; backward compatibility). |
| `pow2_clamp` | Boolean | `false` | `--pow2-clamp <true\|false>` | - | Enable power-of-2 difficulty clamping. |
| `coinbase_tag_suffix` | String | `""` (empty) | `--coinbase-tag-suffix <SUFFIX>` | **COMPLETED** | Optional suffix appended to `"RK-Stratum"` coinbase tag. |
| `stratum_port` | String | `":5555"` | `--stratum-port <HOST:PORT>` | - | **Required.** Stratum port for this instance. Must be unique. |
| `min_share_diff` | Integer | `8192` | `--min-share-diff <DIFFICULTY>` | - | **Required.** Minimum share difficulty for this instance. |
| `prom_port` | String | `None` (disabled) | `--prom-port <HOST:PORT>` | - | Optional Prometheus port for this instance. |
| `instance_log_to_file` | Boolean | `None` (inherits global) | `--instance-log-to-file <true\|false>` | - | Optional per-instance log-to-file setting. |
| `instance_var_diff` | Boolean | `None` (inherits global) | `--instance-var-diff <true\|false>` | - | Optional per-instance variable difficulty override. |
| `instance_shares_per_min` | Integer | `None` (inherits global) | `--instance-shares-per-min <COUNT>` | - | Optional per-instance target shares per minute override. |
| `instance_var_diff_stats` | Boolean | `None` (inherits global) | `--instance-var-diff-stats <true\|false>` | - | Optional per-instance variable difficulty statistics override. |
| `instance_pow2_clamp` | Boolean | `None` (inherits global) | `--instance-pow2-clamp <true\|false>` | - | Optional per-instance power-of-2 difficulty clamping override. |

---

## CLI Notes

- **Booleans**: Accept `true/false`, `1/0`, `yes/no`, `on/off`.
- **Ports**: Accept both `:PORT` and `HOST:PORT` formats.
- **Config override**: CLI flags override values from `config.yaml`.
- **Multi-instance via CLI**:
  - Use one or more `--instance` specs (example: `--instance "port=:5555,diff=8192"`).
  - When using `--instance`, do not use the single-instance flags `--stratum-port/--prom-port/--instance-*`.

---

# Notes

- For the most up-to-date CLI reference, run `stratum-bridge --help`.


utils/alloc/Cargo.toml
```diff
@@ -9,11 +9,14 @@ edition.workspace = true
 include.workspace = true
 repository.workspace = true
 
-[target.'cfg(not(target_os = "macos"))'.dependencies]
+[target.'cfg(all(not(target_os = "macos"), not(target_env = "musl")))'.dependencies]
```
coderofstuff: Why is this change necessary? Same question for the one below it

LiveLaughLove13: It was failing build for, the bridge alloc without it.

LiveLaughLove13: Specifically the Linux build when building the package release

bridge/src/rkstratum_cpu_miner.rs
```diff
@@ -0,0 +1,255 @@
...
+    fn wait_for_update(&self, last_seen: u64) -> (u64, Option<Work>) {
+        let mut slot = self.slot.lock();
+        while slot.version == last_seen {
+            self.cv.wait(&mut slot);
+        }
+        (slot.version, slot.work.as_ref().map(|w| Work { id: w.id, block: w.block.clone(), pow_state: Arc::clone(&w.pow_state) }))
+    }
+}
+
+pub fn spawn_internal_cpu_miner(
```
IzioDev, line 81: https://github.com/kaspanet/rusty-kaspa/blob/d44e31aee830b012bd65e7fe43031f9686b7e099/daemon/src/cpu_miner/mod.rs#L153
It seems RK already have a way to spawn new CPU miner instance. It seems to be correlated to this miner https://github.com/elichai/kaspa-miner

Question:
- Can stratum re-use this method of relying on external program rather than re-defining everything here?

LiveLaughLove13: The current method is to maintain all in once single binary. So it stays as one simpler operation and maintain consistency.

LiveLaughLove13: The objective, is operational simplicity, and ease of deployment for users. Having to bundle / have users need multiple processes for example not having to also bundle another miner imo detracts from a simplistic approach.
Utilizing a shared memory, simple startup, graceful shutdown, unified logging is important in my opinion for maintaining one application, in opposed to needing to rely on more external programs different block timing etc. The objective of this as a whole is utilizing one scope.

IzioDev: (this only is my personal opinion, treat it as such)
I think using CPU miner should be considered a "test" case where only a very small percentage of users would actually runs the CPU miner. For example, myself, I don't have any ASIC machine, and I would use the CPU miner only to test the bridge

LiveLaughLove13: It, makes it simpler to just run in the in process mode for testing as well. Since you do not need to open separate stratum. Which actually makes, testing a test net build much simpler. You just run the stratum-bridge. 

But, this  was the thought. 

I know. Appreciate your input and opinion. Just sharing the reasoning and thought process. 

Was also trying to keep future proof, and option open. Better to have, then not, and want later...

coderofstuff: Izio's just pointing out that a CPU miner code already exists as he linked in https://github.com/kaspanet/rusty-kaspa/blob/d44e31aee830b012bd65e7fe43031f9686b7e099/daemon/src/cpu_miner/mod.rs#L153

You can simply use it instead of rewriting that bit. He's not asking to include or use the miner from the external repo.

utils/alloc/Cargo.toml
coderofstuff: The bridge is the only thing that needs this specific change, so instead of applying this change here you can probably apply it in the Cargo.toml in `bridge/Cargo.toml` so that only that specific binary is affected and the rest is not (especially the node).

So remove this update here, and apply the change in `bridge/Cargo.toml` instead

bridge/docs/README.md
```diff
@@ -2,7 +2,7 @@
```
IzioDev, line 2: I recommend bringing this file at `/bridge` instead in a nested `docs` folder.

This will display the readme when someone clicks in `bridge` from Github.

bridge/src/rkstratum_cpu_miner.rs
LiveLaughLove13: roger that

bridge/docs/README.md
```diff
@@ -39,12 +39,50 @@ cargo run --release --bin stratum-bridge -- --help
...
+cargo build -p kaspa-stratum-bridge --release --features rkstratum_cpu_miner
+```
+
+Run (external node mode + internal CPU miner enabled):
+
+```bash
+cargo run -p kaspa-stratum-bridge --release --features rkstratum_cpu_miner --bin stratum-bridge -- --config bridge/config.yaml --node-mode external --internal-cpu-miner --internal-cpu-miner-address kaspa:YOUR_WALLET_ADDRESS --internal-cpu-miner-threads 1
+```
+
+### Running two bridges at once (two dashboards)
```
IzioDev, line 86: Is this section necessary?
What is the use-case of running two bridges (and in different mode)?

bridge/src/app_config.rs
```diff
@@ -25,6 +28,7 @@ pub(crate) struct GlobalConfig {
     pub(crate) print_stats: bool,
     pub(crate) log_to_file: bool, // Default for instances that don't specify
     pub(crate) health_check_port: String,
+    pub(crate) web_port: String,
```
IzioDev: proposal rename `web_port` -> `web_dashboard_port`

bridge/src/app_config.rs
IzioDev, line 144: Instead of:
```
if multiple then
  // multiple handling
else
  // single handling
end
```

I'd recommend:
```
if single then
  // single handling
  return Ok(...)
end

// multiple handling
```

IzioDev, line 83: Instead of doing manual conversion, have we evaluated the usage of `serde_yaml`?

bridge/src/cli.rs
```diff
@@ -0,0 +1,332 @@
...
+    Inprocess,
+}
+
+fn parse_bool(s: &str) -> Result<bool, anyhow::Error> {
+    match s.trim().to_ascii_lowercase().as_str() {
+        "true" | "1" | "yes" | "y" | "on" => Ok(true),
+        "false" | "0" | "no" | "n" | "off" => Ok(false),
+        _ => Err(anyhow::anyhow!("invalid boolean value: {s}")),
+    }
+}
```
IzioDev: `BoolishValueParser` from clap::builder does exactly this, I think we could re-use it.

bridge/src/cli.rs
```diff
@@ -0,0 +1,332 @@
...
+            ));
+        }
+    }
+
+    Ok(instance)
+}
+
+#[derive(Debug, Clone, Parser)]
+#[command(author, version, about)]
+pub struct Cli {
```
IzioDev, line 108: The current CLI format doesn't look very practicable UX-wise compared to the YAML configuration.

I wonder if we shouldn't rethink the configuration definition with more opinions, potentially dissociating configuration into two groups (one static: changes not often vs one dynamic: changes could be often).

bridge/docs/README.md
```diff
@@ -39,12 +39,50 @@ cargo run --release --bin stratum-bridge -- --help
...
+- `http://127.0.0.1:3030/` for the in-process bridge
+- `http://127.0.0.1:3031/` for the external bridge
+
 ### Run (external node)
 
 Terminal A (node):
 
 ```bash
-cargo run --release --bin kaspad -- --utxoindex --rpclisten=127.0.0.1:16110 --rpclisten-borsh=127.0.0.1:17110
+cargo run --release --bin kaspad -- --utxoindex --rpclisten=127.0.0.1:16110 --rpclisten-borsh=127.0.0.1:17110 --rpclisten-json=127.0.0.1:18110
```
LiveLaughLove13: Mainly used for example of all of the flags. Not necessary no.

bridge/src/app_config.rs
LiveLaughLove13: Any particular reason?

IzioDev: for clarity. even though it's not that ambiguous

bridge/docs/README.md
IzioDev: I suggest to restrict the arguments to the bare minimum to avoid confusion and stay concise

bridge/src/main.rs
```diff
@@ -1,40 +1,98 @@
...
+#[cfg(windows)]
+fn install_windows_ctrl_handler(shutdown_tx: watch::Sender<bool>) -> Result<(), anyhow::Error> {
+    let _ = CTRL_HANDLER_STATE.set(CtrlHandlerState { presses: AtomicUsize::new(0), last_event_ms: AtomicU64::new(0), shutdown_tx });
+
+    let ok = unsafe { SetConsoleCtrlHandler(Some(console_ctrl_handler), 1) };
+    if ok == 0 {
+        return Err(anyhow::anyhow!("failed to install Windows console control handler"));
+    }
+    Ok(())
+}
```
IzioDev, line 87: Any reason to use this instead of tokio's native ctrl_c?
https://docs.rs/tokio/latest/tokio/signal/windows/fn.ctrl_c.html

bridge/docs/README.md
LiveLaughLove13: Yes, people run multiple bridge modes for different reasons

bridge/src/main.rs
```diff
@@ -173,14 +250,20 @@ async fn main() -> Result<(), anyhow::Error> {
...
+        // Install our handler after the embedded node starts so we run first (Windows calls handlers LIFO).
+        // This prevents the embedded node's ctrl handler from consuming Ctrl+C and bypassing our graceful shutdown.
+        #[cfg(windows)]
+        install_windows_ctrl_handler(shutdown_tx.clone())?;
+    } else {
+        // In external mode on Windows, tokio's Ctrl+C handling is usually fine, but we install our handler
+        // anyway to keep behavior consistent across modes.
+        #[cfg(windows)]
+        install_windows_ctrl_handler(shutdown_tx.clone())?;
+    }
```
IzioDev: Behavior can be grouped:
```
if then
  // register windows
else
  // register windows
end
```

to:
```
if then
  // ...
end

// register windows
```

bridge/docs/README.md
LiveLaughLove13: Not sure what you mean here. Want to confirm. So essentially not have the docs folder and just have it indside /bridge  directly right

bridge/src/client_handler.rs
```diff
@@ -155,6 +155,19 @@ impl ClientHandler {
...
+            let guard = self.clients.lock();
+            guard.values().cloned().collect::<Vec<_>>()
+        };
+
+        for client in clients {
+            client.disconnect();
+        }
+
+        self.clients.lock().clear();
+    }
```
IzioDev, line 169: in the `.disconnect` function of a `StratumContext` (shouldn't it be called `StratumClient`?), is this behavior required, cannot it only log on info?
```rs
if is_pre_handshake {
  // log as debug
} else {
  // log as info
}
```
Note: we're acquiring locks + clone 3 variables only to log client information (that we're disconnecting)

Last thing in there:
```rs
let write_half_opt = {
    let mut write_guard = self.write_half.lock();
    write_guard.take()
};

if let Some(mut write_half) = write_half_opt {
    tokio::spawn(async move {
        let _ = write_half.shutdown().await;
    });
}
```
This seems like an anti-pattern to me, but I'm no Rust expert. cc @coderofstuff

bridge/src/kaspaapi.rs
```diff
@@ -139,6 +140,15 @@ pub struct KaspaApi {
 impl KaspaApi {
     /// Create a new Kaspa API client
     pub async fn new(address: String, _block_wait_time: Duration, coinbase_tag_suffix: Option<String>) -> Result<Arc<Self>> {
+        Self::new_with_shutdown(address, _block_wait_time, coinbase_tag_suffix, None).await
+    }
```
IzioDev: This seems unused, maybe we should make `new_with_shutdown` the new `new`

bridge/src/kaspaapi.rs
```diff
@@ -139,6 +140,15 @@ pub struct KaspaApi {
...
+
+    pub async fn new_with_shutdown(
+        address: String,
+        _block_wait_time: Duration,
```
IzioDev: If `_block_wait_time` is unused, let's remove it from the fn signature

bridge/src/cli.rs
LiveLaughLove13: Working on this , so more like

pub(crate) fn parse_bool(s: &str) -> Result<bool, anyhow::Error> {
    match s.trim().to_ascii_lowercase().as_str() {
        "true" | "1" | "yes" | "y" | "on" | "enable" | "enabled" => Ok(true),
        "false" | "0" | "no" | "n" | "off" | "disable" | "disabled" => Ok(false),
        _ => Err(anyhow::anyhow!("invalid boolean value: {s}")),
    }
}

bridge/src/kaspaapi.rs
```diff
@@ -139,6 +140,15 @@ pub struct KaspaApi {
...
+        coinbase_tag_suffix: Option<String>,
+        shutdown_rx: Option<watch::Receiver<bool>>,
```
IzioDev: This shouldn't be an option according to current usage

bridge/src/kaspaapi.rs
```diff
@@ -161,24 +175,99 @@ impl KaspaApi {
...
-        client
-            .start_notify(ListenerId::default(), NewBlockTemplateScope {}.into())
-            .await
-            .context("Failed to subscribe to block template notifications")?;
+        // Some nodes may take time to accept notification subscriptions; retry until it succeeds.
+        let mut attempt: u64 = 0;
+        let mut backoff_ms: u64 = 250;
+        loop {
+            attempt += 1;
+            let notify_fut = client.start_notify(ListenerId::default(), NewBlockTemplateScope {}.into());
```
IzioDev, line 221: I'm surprised this is required to safely start getting notifications, what was the error you were getting without this exponential backoff strategy?

bridge/src/kaspaapi.rs
IzioDev: this function is always called with verbose = `false`, which makes the parameter (and its control flow) useless.

note: verbosity in this context should come from log level (since it's used to decide what log level to use)

bridge/src/kaspaapi.rs
```diff
@@ -516,6 +633,53 @@ impl KaspaApi {
         Ok(())
     }
 
+    pub async fn wait_for_sync_with_shutdown(&self, verbose: bool, mut shutdown_rx: watch::Receiver<bool>) -> Result<()> {
```
IzioDev: this function is always called with verbose = `true`

bridge/docs/README.md
IzioDev: Yes, but I just found out the reason is now invalid because Github now understands the `docs` sub-folder and go seek there.
Anyway, if the `docs` folder only contains a README.md, it probably isn't a bad idea to make it lives at `/bridge` instead of `/bridge/docs`

bridge/docs/README.md
IzioDev: Could you please describe what the reasons could be?

utils/alloc/Cargo.toml
biryukovmaxim: In rk ci we build node with musl toolchain that uses mimalloc.
https://github.com/kaspanet/rusty-kaspa/blob/d44e31aee830b012bd65e7fe43031f9686b7e099/.github/workflows/ci.yaml#L405

bridge/docs/README.md
LiveLaughLove13: Should we leave as, incase further documentations in future. Or test information etc as example

bridge/docs/README.md
LiveLaughLove13: There are many reasons why miners may run 2 bridges. Same reason some users use multiple pools, and set back up pools etc

bridge/src/main.rs
LiveLaughLove13: Did not test tokio native.  The custom handler is to maintain ordering for shutdown.

bridge/src/cli.rs
IzioDev: Nevermind, I shouldn't have mentioned that. The `instances` property is a `Vec<String>`, I thought it was typed to receive the instances object

utils/alloc/Cargo.toml
LiveLaughLove13: This is what I run into if I use the alloc as is


error: linking with `/home/runner/x-tools/x86_64-multilib-linux-musl/bin/x86_64-multilib-linux-musl-gcc` failed: exit status: 1
  |
  = note:  "/home/runner/x-tools/x86_64-multilib-linux-musl/bin/x86_64-multilib-linux-musl-gcc" "-m64" "<sysroot>/lib/rustlib/x86_64-unknown-linux-musl/lib/self-contained/rcrt1.o" "<sysroot>/lib/rustlib/x86_64-unknown-linux-musl/lib/self-contained/crti.o" "<sysroot>/lib/rustlib/x86_64-unknown-linux-musl/lib/self-contained/crtbeginS.o" "<1572 object files omitted>" "-Wl,--as-needed" "-Wl,-Bstatic" "/tmp/rustcXYcI95/liblibrocksdb_sys-37d8dd8c82aea7e3.rlib" "-lstdc++" "/tmp/rustcXYcI95/{libzstd_sys-a7d6bf6bfcbbf931,liblz4_sys-e7152eaad850f377,liblibz_sys-91762ba69da6d051,libbzip2_sys-d3e6141a738e140a,libring-e4925e473b304f3e,libsecp256k1_sys-74a0c1fa80730134,libkaspa_hashes-762a04baa682ef83,liblibmimalloc_sys-3793f9816548f352}.rlib" "-lunwind" "-lc" "<sysroot>/lib/rustlib/x86_64-unknown-linux-musl/lib/libcompiler_builtins-*.rlib" "-L" "/tmp/rustcXYcI95/raw-dylibs" "-Wl,-Bdynamic" "-Wl,--eh-frame-hdr" "-Wl,-z,noexecstack" "-nostartfiles" "-L" "/home/runner/work/rusty-kaspa/rusty-kaspa/target/x86_64-unknown-linux-musl/release/build/libmimalloc-sys-23151fd9632964f9/out" "-L" "/home/runner/work/rusty-kaspa/rusty-kaspa/target/x86_64-unknown-linux-musl/release/build/kaspa-hashes-116057262fb32336/out" "-L" "/home/runner/work/rusty-kaspa/rusty-kaspa/target/x86_64-unknown-linux-musl/release/build/secp256k1-sys-1cc22ba3af33fb00/out" "-L" "/home/runner/work/rusty-kaspa/rusty-kaspa/target/x86_64-unknown-linux-musl/release/build/ring-5babe6629ed1bbee/out" "-L" "/home/runner/work/rusty-kaspa/rusty-kaspa/target/x86_64-unknown-linux-musl/release/build/librocksdb-sys-37b8fc5003b10aac/out" "-L" "/home/runner/work/rusty-kaspa/rusty-kaspa/target/x86_64-unknown-linux-musl/release/build/bzip2-sys-d417c05ec2a83027/out/lib" "-L" "/home/runner/work/rusty-kaspa/rusty-kaspa/target/x86_64-unknown-linux-musl/release/build/libz-sys-b0e2b3a8ad423e4e/out/lib" "-L" "/home/runner/work/rusty-kaspa/rusty-kaspa/target/x86_64-unknown-linux-musl/release/build/lz4-sys-3bd1b3b7b849b227/out" "-L" "/home/runner/work/rusty-kaspa/rusty-kaspa/target/x86_64-unknown-linux-musl/release/build/zstd-sys-4803fe0729df773a/out" "-L" "<sysroot>/lib/rustlib/x86_64-unknown-linux-musl/lib/self-contained" "-L" "<sysroot>/lib/rustlib/x86_64-unknown-linux-musl/lib" "-o" "/home/runner/work/rusty-kaspa/rusty-kaspa/target/x86_64-unknown-linux-musl/release/deps/stratum_bridge-2d303546e3048068" "-Wl,--gc-sections" "-static-pie" "-Wl,-z,relro,-z,now" "-Wl,-O1" "-Wl,--strip-all" "-nodefaultlibs" "-static" "<sysroot>/lib/rustlib/x86_64-unknown-linux-musl/lib/self-contained/crtendS.o" "<sysroot>/lib/rustlib/x86_64-unknown-linux-musl/lib/self-contained/crtn.o"
  = note: some arguments are omitted. use `--verbose` to show all linker arguments
  = note: /home/runner/x-tools/x86_64-multilib-linux-musl/lib/gcc/x86_64-multilib-linux-musl/13.2.0/../../../../x86_64-multilib-linux-musl/bin/ld.bfd: /tmp/rustcXYcI95/liblibmimalloc_sys-3793f9816548f352.rlib(ad82a007e87f09df-static.o): in function `mi_free':
          static.c:(.text.mi_free+0x0): multiple definition of `operator delete[](void*)'; /home/runner/x-tools/x86_64-multilib-linux-musl/lib/gcc/x86_64-multilib-linux-musl/13.2.0/../../../../x86_64-multilib-linux-musl/lib/../lib64/libstdc++.a(del_opv.o):/home/runner/work/rusty-kaspa/rusty-kaspa/musl-toolchain/crosstool-ng/.build/x86_64-multilib-linux-musl/src/gcc/libstdc++-v3/libsupc++/del_opv.cc:34: first defined here
          /home/runner/x-tools/x86_64-multilib-linux-musl/lib/gcc/x86_64-multilib-linux-musl/13.2.0/../../../../x86_64-multilib-linux-musl/bin/ld.bfd: /tmp/rustcXYcI95/liblibmimalloc_sys-3793f9816548f352.rlib(ad82a007e87f09df-static.o): in function `mi_free':
          static.c:(.text.mi_free+0x0): multiple definition of `operator delete(void*)'; /home/runner/x-tools/x86_64-multilib-linux-musl/lib/gcc/x86_64-multilib-linux-musl/13.2.0/../../../../x86_64-multilib-linux-musl/lib/../lib64/libstdc++.a(del_op.o):/home/runner/work/rusty-kaspa/rusty-kaspa/musl-toolchain/crosstool-ng/.build/x86_64-multilib-linux-musl/src/gcc/libstdc++-v3/libsupc++/del_op.cc:48: first defined here
          /home/runner/x-tools/x86_64-multilib-linux-musl/lib/gcc/x86_64-multilib-linux-musl/13.2.0/../../../../x86_64-multilib-linux-musl/bin/ld.bfd: /tmp/rustcXYcI95/liblibmimalloc_sys-3793f9816548f352.rlib(ad82a007e87f09df-static.o): in function `mi_free_size':
          static.c:(.text.mi_free_size+0x0): multiple definition of `operator delete(void*, unsigned long)'; /home/runner/x-tools/x86_64-multilib-linux-musl/lib/gcc/x86_64-multilib-linux-musl/13.2.0/../../../../x86_64-multilib-linux-musl/lib/../lib64/libstdc++.a(del_ops.o):/home/runner/work/rusty-kaspa/rusty-kaspa/musl-toolchain/crosstool-ng/.build/x86_64-multilib-linux-musl/src/gcc/libstdc++-v3/libsupc++/del_ops.cc:32: first defined here
          /home/runner/x-tools/x86_64-multilib-linux-musl/lib/gcc/x86_64-multilib-linux-musl/13.2.0/../../../../x86_64-multilib-linux-musl/bin/ld.bfd: /tmp/rustcXYcI95/liblibmimalloc_sys-3793f9816548f352.rlib(ad82a007e87f09df-static.o): in function `mi_free_size':
          static.c:(.text.mi_free_size+0x0): multiple definition of `operator delete[](void*, unsigned long)'; /home/runner/x-tools/x86_64-multilib-linux-musl/lib/gcc/x86_64-multilib-linux-musl/13.2.0/../../../../x86_64-multilib-linux-musl/lib/../lib64/libstdc++.a(del_opvs.o):/home/runner/work/rusty-kaspa/rusty-kaspa/musl-toolchain/crosstool-ng/.build/x86_64-multilib-linux-musl/src/gcc/libstdc++-v3/libsupc++/del_opvs.cc:32: first defined here
          /home/runner/x-tools/x86_64-multilib-linux-musl/lib/gcc/x86_64-multilib-linux-musl/13.2.0/../../../../x86_64-multilib-linux-musl/bin/ld.bfd: /tmp/rustcXYcI95/liblibmimalloc_sys-3793f9816548f352.rlib(ad82a007e87f09df-static.o): in function `operator delete(void*, unsigned long, std::align_val_t)':
          static.c:(.text._ZdlPvmSt11align_val_t+0x0): multiple definition of `operator delete(void*, unsigned long, std::align_val_t)'; /home/runner/x-tools/x86_64-multilib-linux-musl/lib/gcc/x86_64-multilib-linux-musl/13.2.0/../../../../x86_64-multilib-linux-musl/lib/../lib64/libstdc++.a(del_opsa.o):/home/runner/work/rusty-kaspa/rusty-kaspa/musl-toolchain/crosstool-ng/.build/x86_64-multilib-linux-musl/src/gcc/libstdc++-v3/libsupc++/del_opsa.cc:31: first defined here
          /home/runner/x-tools/x86_64-multilib-linux-musl/lib/gcc/x86_64-multilib-linux-musl/13.2.0/../../../../x86_64-multilib-linux-musl/bin/ld.bfd: /tmp/rustcXYcI95/liblibmimalloc_sys-3793f9816548f352.rlib(ad82a007e87f09df-static.o): in function `operator delete[](void*, unsigned long, std::align_val_t)':
          static.c:(.text._ZdaPvmSt11align_val_t+0x0): multiple definition of `operator delete[](void*, unsigned long, std::align_val_t)'; /home/runner/x-tools/x86_64-multilib-linux-musl/lib/gcc/x86_64-multilib-linux-musl/13.2.0/../../../../x86_64-multilib-linux-musl/lib/../lib64/libstdc++.a(del_opvsa.o):/home/runner/work/rusty-kaspa/rusty-kaspa/musl-toolchain/crosstool-ng/.build/x86_64-multilib-linux-musl/src/gcc/libstdc++-v3/libsupc++/del_opvsa.cc:31: first defined here
          /home/runner/x-tools/x86_64-multilib-linux-musl/lib/gcc/x86_64-multilib-linux-musl/13.2.0/../../../../x86_64-multilib-linux-musl/bin/ld.bfd: /tmp/rustcXYcI95/liblibmimalloc_sys-3793f9816548f352.rlib(ad82a007e87f09df-static.o): in function `operator delete(void*, std::align_val_t)':
          static.c:(.text._ZdlPvSt11align_val_t+0x0): multiple definition of `operator delete(void*, std::align_val_t)'; /home/runner/x-tools/x86_64-multilib-linux-musl/lib/gcc/x86_64-multilib-linux-musl/13.2.0/../../../../x86_64-multilib-linux-musl/lib/../lib64/libstdc++.a(del_opa.o):/home/runner/work/rusty-kaspa/rusty-kaspa/musl-toolchain/crosstool-ng/.build/x86_64-multilib-linux-musl/src/gcc/libstdc++-v3/libsupc++/del_opa.cc:48: first defined here
          /home/runner/x-tools/x86_64-multilib-linux-musl/lib/gcc/x86_64-multilib-linux-musl/13.2.0/../../../../x86_64-multilib-linux-musl/bin/ld.bfd: /tmp/rustcXYcI95/liblibmimalloc_sys-3793f9816548f352.rlib(ad82a007e87f09df-static.o): in function `operator delete[](void*, std::align_val_t)':
          static.c:(.text._ZdaPvSt11align_val_t+0x0): multiple definition of `operator delete[](void*, std::align_val_t)'; /home/runner/x-tools/x86_64-multilib-linux-musl/lib/gcc/x86_64-multilib-linux-musl/13.2.0/../../../../x86_64-multilib-linux-musl/lib/../lib64/libstdc++.a(del_opva.o):/home/runner/work/rusty-kaspa/rusty-kaspa/musl-toolchain/crosstool-ng/.build/x86_64-multilib-linux-musl/src/gcc/libstdc++-v3/libsupc++/del_opva.cc:34: first defined here
          /home/runner/x-tools/x86_64-multilib-linux-musl/lib/gcc/x86_64-multilib-linux-musl/13.2.0/../../../../x86_64-multilib-linux-musl/bin/ld.bfd: /tmp/rustcXYcI95/liblibmimalloc_sys-3793f9816548f352.rlib(ad82a007e87f09df-static.o): in function `mi_new':
          static.c:(.text.mi_new+0x0): multiple definition of `operator new[](unsigned long)'; /home/runner/x-tools/x86_64-multilib-linux-musl/lib/gcc/x86_64-multilib-linux-musl/13.2.0/../../../../x86_64-multilib-linux-musl/lib/../lib64/libstdc++.a(new_opv.o):/home/runner/work/rusty-kaspa/rusty-kaspa/musl-toolchain/crosstool-ng/.build/x86_64-multilib-linux-musl/src/gcc/libstdc++-v3/libsupc++/new_opv.cc:31: first defined here
          /home/runner/x-tools/x86_64-multilib-linux-musl/lib/gcc/x86_64-multilib-linux-musl/13.2.0/../../../../x86_64-multilib-linux-musl/bin/ld.bfd: /tmp/rustcXYcI95/liblibmimalloc_sys-3793f9816548f352.rlib(ad82a007e87f09df-static.o): in function `mi_new':
          static.c:(.text.mi_new+0x0): multiple definition of `operator new(unsigned long)'; /home/runner/x-tools/x86_64-multilib-linux-musl/lib/gcc/x86_64-multilib-linux-musl/13.2.0/../../../../x86_64-multilib-linux-musl/lib/../lib64/libstdc++.a(new_op.o):/home/runner/work/rusty-kaspa/rusty-kaspa/musl-toolchain/crosstool-ng/.build/x86_64-multilib-linux-musl/src/gcc/libstdc++-v3/libsupc++/new_op.cc:43: first defined here
          /home/runner/x-tools/x86_64-multilib-linux-musl/lib/gcc/x86_64-multilib-linux-musl/13.2.0/../../../../x86_64-multilib-linux-musl/bin/ld.bfd: /tmp/rustcXYcI95/liblibmimalloc_sys-3793f9816548f352.rlib(ad82a007e87f09df-static.o): in function `operator new(unsigned long, std::nothrow_t const&)':
          static.c:(.text._ZnwmRKSt9nothrow_t+0x0): multiple definition of `operator new(unsigned long, std::nothrow_t const&)'; /home/runner/x-tools/x86_64-multilib-linux-musl/lib/gcc/x86_64-multilib-linux-musl/13.2.0/../../../../x86_64-multilib-linux-musl/lib/../lib64/libstdc++.a(new_opnt.o):/home/runner/work/rusty-kaspa/rusty-kaspa/musl-toolchain/crosstool-ng/.build/x86_64-multilib-linux-musl/src/gcc/libstdc++-v3/libsupc++/new_opnt.cc:33: first defined here
          /home/runner/x-tools/x86_64-multilib-linux-musl/lib/gcc/x86_64-multilib-linux-musl/13.2.0/../../../../x86_64-multilib-linux-musl/bin/ld.bfd: /tmp/rustcXYcI95/liblibmimalloc_sys-3793f9816548f352.rlib(ad82a007e87f09df-static.o): in function `operator new[](unsigned long, std::nothrow_t const&)':
          static.c:(.text._ZnamRKSt9nothrow_t+0x0): multiple definition of `operator new[](unsigned long, std::nothrow_t const&)'; /home/runner/x-tools/x86_64-multilib-linux-musl/lib/gcc/x86_64-multilib-linux-musl/13.2.0/../../../../x86_64-multilib-linux-musl/lib/../lib64/libstdc++.a(new_opvnt.o):/home/runner/work/rusty-kaspa/rusty-kaspa/musl-toolchain/crosstool-ng/.build/x86_64-multilib-linux-musl/src/gcc/libstdc++-v3/libsupc++/new_opvnt.cc:32: first defined here
          /home/runner/x-tools/x86_64-multilib-linux-musl/lib/gcc/x86_64-multilib-linux-musl/13.2.0/../../../../x86_64-multilib-linux-musl/bin/ld.bfd: /tmp/rustcXYcI95/liblibmimalloc_sys-3793f9816548f352.rlib(ad82a007e87f09df-static.o): in function `mi_new_aligned':
          static.c:(.text.mi_new_aligned+0x0): multiple definition of `operator new[](unsigned long, std::align_val_t)'; /home/runner/x-tools/x86_64-multilib-linux-musl/lib/gcc/x86_64-multilib-linux-musl/13.2.0/../../../../x86_64-multilib-linux-musl/lib/../lib64/libstdc++.a(new_opva.o):/home/runner/work/rusty-kaspa/rusty-kaspa/musl-toolchain/crosstool-ng/.build/x86_64-multilib-linux-musl/src/gcc/libstdc++-v3/libsupc++/new_opva.cc:31: first defined here
          /home/runner/x-tools/x86_64-multilib-linux-musl/lib/gcc/x86_64-multilib-linux-musl/13.2.0/../../../../x86_64-multilib-linux-musl/bin/ld.bfd: /tmp/rustcXYcI95/liblibmimalloc_sys-3793f9816548f352.rlib(ad82a007e87f09df-static.o): in function `mi_new_aligned':
          static.c:(.text.mi_new_aligned+0x0): multiple definition of `operator new(unsigned long, std::align_val_t)'; /home/runner/x-tools/x86_64-multilib-linux-musl/lib/gcc/x86_64-multilib-linux-musl/13.2.0/../../../../x86_64-multilib-linux-musl/lib/../lib64/libstdc++.a(new_opa.o):/home/runner/work/rusty-kaspa/rusty-kaspa/musl-toolchain/crosstool-ng/.build/x86_64-multilib-linux-musl/src/gcc/libstdc++-v3/libsupc++/new_opa.cc:119: first defined here
          collect2: error: ld returned 1 exit status
          

error: could not compile `kaspa-stratum-bridge` (bin "stratum-bridge") due to 1 previous error

Caused by:
  process didn't exit successfully: `/home/runner/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/bin/rustc --crate-name stratum_bridge --edition=2024 bridge/src/main.rs --error-format=json --json=diagnostic-rendered-ansi,artifacts,future-incompat --crate-type bin --emit=dep-info,link -C opt-level=3 -C lto=thin '--allow=clippy::uninlined_format_args' '--allow=clippy::empty_docs' -C overflow-checks=on --check-cfg 'cfg(docsrs,test)' --check-cfg 'cfg(feature, values("rkstratum_cpu_miner"))' -C metadata=cc7317ac6ba6ca1a -C extra-filename=-2d303546e3048068 --out-dir /home/runner/work/rusty-kaspa/rusty-kaspa/target/x86_64-unknown-linux-musl/release/deps --target x86_64-unknown-linux-musl -C linker=/home/runner/x-tools/x86_64-multilib-linux-musl/bin/x86_64-multilib-linux-musl-gcc -C strip=symbols -L dependency=/home/runner/work/rusty-kaspa/rusty-kaspa/target/x86_64-unknown-linux-musl/release/deps -L dependency=/home/runner/work/rusty-kaspa/rusty-kaspa/target/release/deps --extern anyhow=/home/runner/work/rusty-kaspa/rusty-kaspa/target/x86_64-unknown-linux-musl/release/deps/libanyhow-2512457d4e99f0aa.rlib --extern async_trait=/home/runner/work/rusty-kaspa/rusty-kaspa/target/release/deps/libasync_trait-833197e17e20fcd1.so --extern blake2=/home/runner/work/rusty-kaspa/rusty-kaspa/target/x86_64-unknown-linux-musl/release/deps/libblake2-271f83cb38c051d7.rlib --extern chrono=/home/runner/work/rusty-kaspa/rusty-kaspa/target/x86_64-unknown-linux-musl/release/deps/libchrono-6ef22f494db33bb3.rlib --extern clap=/home/runner/work/rusty-kaspa/rusty-kaspa/target/x86_64-unknown-linux-musl/release/deps/libclap-bb4b880527a7d16e.rlib --extern dirs=/home/runner/work/rusty-kaspa/rusty-kaspa/target/x86_64-unknown-linux-musl/release/deps/libdirs-c375a0e497571b37.rlib --extern faster_hex=/home/runner/work/rusty-kaspa/rusty-kaspa/target/x86_64-unknown-linux-musl/release/deps/libfaster_hex-07a30fed9d8daddc.rlib --extern futures_util=/home/runner/work/rusty-kaspa/rusty-kaspa/target/x86_64-unknown-linux-musl/release/deps/libfutures_util-0ab4a3fc3a09e4c8.rlib --extern hex=/home/runner/work/rusty-kaspa/rusty-kaspa/target/x86_64-unknown-linux-musl/release/deps/libhex-f9c0d888eb5db755.rlib --extern include_dir=/home/runner/work/rusty-kaspa/rusty-kaspa/target/x86_64-unknown-linux-musl/release/deps/libinclude_dir-b5301f69b4fad4ba.rlib --extern kaspa_addresses=/home/runner/work/rusty-kaspa/rusty-kaspa/target/x86_64-unknown-linux-musl/release/deps/libkaspa_addresses-4b66cc9b74158d96.rlib --extern kaspa_alloc=/home/runner/work/rusty-kaspa/rusty-kaspa/target/x86_64-unknown-linux-musl/release/deps/libkaspa_alloc-04dd0a1755a5d6b3.rlib --extern kaspa_consensus_core=/home/runner/work/rusty-kaspa/rusty-kaspa/target/x86_64-unknown-linux-musl/release/deps/libkaspa_consensus_core-b3204db723b51526.rlib --extern kaspa_core=/home/runner/work/rusty-kaspa/rusty-kaspa/target/x86_64-unknown-linux-musl/release/deps/libkaspa_core-5998004ff78f1a67.rlib --extern kaspa_grpc_client=/home/runner/work/rusty-kaspa/rusty-kaspa/target/x86_64-unknown-linux-musl/release/deps/libkaspa_grpc_client-658bf9ec90a2de23.rlib --extern kaspa_hashes=/home/runner/work/rusty-kaspa/rusty-kaspa/target/x86_64-unknown-linux-musl/release/deps/libkaspa_hashes-762a04baa682ef83.rlib --extern kaspa_notify=/home/runner/work/rusty-kaspa/rusty-kaspa/target/x86_64-unknown-linux-musl/release/deps/libkaspa_notify-e6f457878fa37100.rlib --extern kaspa_pow=/home/runner/work/rusty-kaspa/rusty-kaspa/target/x86_64-unknown-linux-musl/release/deps/libkaspa_pow-89e4ea2499629630.rlib --extern kaspa_rpc_core=/home/runner/work/rusty-kaspa/rusty-kaspa/target/x86_64-unknown-linux-musl/release/deps/libkaspa_rpc_core-461edd8e35aa9def.rlib --extern kaspa_rpc_service=/home/runner/work/rusty-kaspa/rusty-kaspa/target/x86_64-unknown-linux-musl/release/deps/libkaspa_rpc_service-31b4152e2b20938b.rlib --extern kaspa_stratum_bridge=/home/runner/work/rusty-kaspa/rusty-kaspa/target/x86_64-unknown-linux-musl/release/deps/libkaspa_stratum_bridge-6d75bc167172b569.rlib --extern kaspa_utils=/home/runner/work/rusty-kaspa/rusty-kaspa/target/x86_64-unknown-linux-musl/release/deps/libkaspa_utils-928c0e57d67ad48a.rlib --extern kaspad_lib=/home/runner/work/rusty-kaspa/rusty-kaspa/target/x86_64-unknown-linux-musl/release/deps/libkaspad_lib.rlib --extern num_bigint=/home/runner/work/rusty-kaspa/rusty-kaspa/target/x86_64-unknown-linux-musl/release/deps/libnum_bigint-ae015251399cbd15.rlib --extern num_traits=/home/runner/work/rusty-kaspa/rusty-kaspa/target/x86_64-unknown-linux-musl/release/deps/libnum_traits-4d2c269311b41777.rlib --extern once_cell=/home/runner/work/rusty-kaspa/rusty-kaspa/target/x86_64-unknown-linux-musl/release/deps/libonce_cell-20163947d285d5dc.rlib --extern parking_lot=/home/runner/work/rusty-kaspa/rusty-kaspa/target/x86_64-unknown-linux-musl/release/deps/libparking_lot-046bcb58eead9d4a.rlib --extern prometheus=/home/runner/work/rusty-kaspa/rusty-kaspa/target/x86_64-unknown-linux-musl/release/deps/libprometheus-cbe1fd6cf67d2a03.rlib --extern regex=/home/runner/work/rusty-kaspa/rusty-kaspa/target/x86_64-unknown-linux-musl/release/deps/libregex-723764362eecea9b.rlib --extern serde=/home/runner/work/rusty-kaspa/rusty-kaspa/target/x86_64-unknown-linux-musl/release/deps/libserde-f8b239d57400c909.rlib --extern serde_json=/home/runner/work/rusty-kaspa/rusty-kaspa/target/x86_64-unknown-linux-musl/release/deps/libserde_json-effe48db55a724d8.rlib --extern serde_yaml=/home/runner/work/rusty-kaspa/rusty-kaspa/target/x86_64-unknown-linux-musl/release/deps/libserde_yaml-309e98686f142864.rlib --extern thiserror=/home/runner/work/rusty-kaspa/rusty-kaspa/target/x86_64-unknown-linux-musl/release/deps/libthiserror-255e7b9ad427953e.rlib --extern tokio=/home/runner/work/rusty-kaspa/rusty-kaspa/target/x86_64-unknown-linux-musl/release/deps/libtokio-93d48090f0701add.rlib --extern tracing=/home/runner/work/rusty-kaspa/rusty-kaspa/target/x86_64-unknown-linux-musl/release/deps/libtracing-9c35f3808fb305ac.rlib --extern tracing_appender=/home/runner/work/rusty-kaspa/rusty-kaspa/target/x86_64-unknown-linux-musl/release/deps/libtracing_appender-e75bc626864252cd.rlib --extern tracing_log=/home/runner/work/rusty-kaspa/rusty-kaspa/target/x86_64-unknown-linux-musl/release/deps/libtracing_log-ce9539f22854582b.rlib --extern tracing_subscriber=/home/runner/work/rusty-kaspa/rusty-kaspa/target/x86_64-unknown-linux-musl/release/deps/libtracing_subscriber-fcd18e7147871526.rlib --extern uuid=/home/runner/work/rusty-kaspa/rusty-kaspa/target/x86_64-unknown-linux-musl/release/deps/libuuid-71efb341511e3bd6.rlib -C link-arg=-static -L native=/home/runner/work/rusty-kaspa/rusty-kaspa/target/x86_64-unknown-linux-musl/release/build/libmimalloc-sys-23151fd9632964f9/out -L native=/home/runner/work/rusty-kaspa/rusty-kaspa/target/x86_64-unknown-linux-musl/release/build/kaspa-hashes-116057262fb32336/out -L native=/home/runner/work/rusty-kaspa/rusty-kaspa/target/x86_64-unknown-linux-musl/release/build/secp256k1-sys-1cc22ba3af33fb00/out -L native=/home/runner/work/rusty-kaspa/rusty-kaspa/target/x86_64-unknown-linux-musl/release/build/ring-5babe6629ed1bbee/out -L native=/home/runner/work/rusty-kaspa/rusty-kaspa/target/x86_64-unknown-linux-musl/release/build/librocksdb-sys-37b8fc5003b10aac/out -L native=/home/runner/work/rusty-kaspa/rusty-kaspa/target/x86_64-unknown-linux-musl/release/build/librocksdb-sys-37b8fc5003b10aac/out -L native=/home/runner/work/rusty-kaspa/rusty-kaspa/target/x86_64-unknown-linux-musl/release/build/bzip2-sys-d417c05ec2a83027/out/lib -L native=/home/runner/work/rusty-kaspa/rusty-kaspa/target/x86_64-unknown-linux-musl/release/build/libz-sys-b0e2b3a8ad423e4e/out/lib -L native=/home/runner/work/rusty-kaspa/rusty-kaspa/target/x86_64-unknown-linux-musl/release/build/libz-sys-b0e2b3a8ad423e4e/out/lib -L native=/home/runner/work/rusty-kaspa/rusty-kaspa/target/x86_64-unknown-linux-musl/release/build/lz4-sys-3bd1b3b7b849b227/out -L native=/home/runner/work/rusty-kaspa/rusty-kaspa/target/x86_64-unknown-linux-musl/release/build/zstd-sys-4803fe0729df773a/out` (exit status: 1)
Error: Process completed with exit code 101.

bridge/src/main.rs
IzioDev: The tokio signal should handle that already, it also avoid defining unsafe Rust code

bridge/docs/README.md
IzioDev: I'm just tying to understand the use-case because the support of multi-instances add a non null complexity to configuration (and code handling it).

If the need purely is backup, it should be running on a dedicated node / stratum, otherwise if the node goes down both instances go down

LiveLaughLove13: External, mode a user can point at another running node.  In process is the node included. A user may run a node on a personal system, and also may run one on a server etc and want to utilize the external on the server. 

These are things miners do. It may seem a little foreign. But, for miners being able to have connections to node = their miners working. Making funds. Its a VERY important thing that their is redundancy and options.

IzioDev: Redundancy on the same process doesn't seems like redundancy to me.
It the process die, all instances die.

A redundancy setup would be multiple Kaspa node running + multiple stratum running on dedicated processes and servers.

I'd love to talk with miners running multiple instances on the same process for redundancy reasons

LiveLaughLove13: External and INPROCESS are different. UNLESS you are pointing your external at your in process node. Then yes if in process dies your external dies. 

The external running at same time is reference to utilizing another kaspad instance, which may be located else where. Meaning that say you accidentally turn off your node. The external is still running pointed at the node you designated it to. Since it does not have its own. The point is redundancy. 

The same way a data server, has a fall back server.

LiveLaughLove13: Good chance we are talking about completely different things as well. 
Most likely I am misinterpreting your question.

IzioDev: What about letting the bridge the ability to reference multiple node IPs?

This way, each of your stratum instances (not running on the same process) can point to many kaspad instances. In case one kaspad dies, stratum rolls to another one within the list. (note: a resolver ideally could be used too)

With this setup, I don't envision a possible need to support multi stratum instances (on the same process)

LiveLaughLove13: like khost , pnn resolver? Like how that works?

LiveLaughLove13: Could be a future, future change. I do not think it is pertinent at present moment. But, its a solid idea @IzioDev

bridge/src/client_handler.rs
LiveLaughLove13: disconnect_all() snapshots the Arc<StratumContext> values into a Vec specifically to avoid holding the clients map lock while calling disconnect(), since disconnect() takes additional locks internally and may spawn async work; this reduces lock contention and potential deadlock risk and is not “just for logging.” Also, the requested logging behavior is already present in StratumContext::disconnect() via the is_pre_handshake check (pre-handshake logged at debug, otherwise at info). Finally, the write_half.take() followed by tokio::spawn(async move { write_half.shutdown().await }) is a deliberate pattern because disconnect() is synchronous and cannot await; we take() to avoid holding a mutex across an await and spawn a short task to attempt a graceful shutdown (and even if the task doesn’t run during shutdown, taking/dropping the halves still closes the connection).

bridge/src/kaspaapi.rs
LiveLaughLove13: We added the retry/backoff because start_notify can fail transiently right after connect on some nodes (not ready to accept subscriptions yet); retrying with bounded exponential backoff makes notification startup reliable and avoids tight-looping/spam.

bridge/docs/README.md
IzioDev: PNN Resolver is one network constituted using the Resolver. One could build a private network composed of multiple Kaspa nodes (and reverse proxy).

Even without going that far, I emphasis the fact that the multi instance feature is adding a non null complexity to the codebase (at different levels). I do think a better tradeoff would be to allow the one instance to accept a list of kaspad IPs, this will enable true redundancy (both in term of `n` stratum running as different processes and `m` kaspad)

bridge/src/client_handler.rs
IzioDev: yes, the 3 variables I mentioned are only used for logging purposes, see the attached image:
image

I can expand on the other points if necessary

LiveLaughLove13: This , was placed due to Goldshell noise issue. To reduce the extra noise that get sent from goldshell when they connect. They cause a trailing IP and so was placed so that would not get all the spammed noise from them.

LiveLaughLove13: Will try tidy it up per your request to see if can be more optimally written per your suggestion

LiveLaughLove13: @coderofstuff  thoughts on alternative to reduce the GS noise ?

PR #837, opened 2026-01-22
More Covenant ID related opcodes

someone235:


crypto/txscript/errors/src/lib.rs
```diff
@@ -83,7 +84,7 @@ pub enum TxScriptError {
     InvalidIndex(i32),
 
     #[error("{0} is not a valid covenant output index for input {1} with {2} covenant outputs")]
```
michaelsutton:
```suggestion
    #[error("{0} is not a valid covenant output index for input {1} with {2} authorized outputs")]
```

crypto/txscript/errors/src/lib.rs
```diff
@@ -83,7 +84,7 @@ pub enum TxScriptError {
...
-    InvalidCovOutIndex(usize, usize, usize),
+    InvalidInputCovOutIndex(usize, usize, usize),
```
michaelsutton:
```suggestion
    InvalidAuthCovOutIndex(usize, usize, usize),
```

michaelsutton: perhaps move down near all cov errors

crypto/txscript/src/covenants.rs
michaelsutton: unused for now 

```suggestion
    pub _covenant_id: Hash,
```

PR #839, opened 2026-01-23
Fix Covenant RPC serialization + feat(rothschild) which adds covenant id and new tx version

IzioDev:
Add two new params:
- `--enable-covenant-id` (default: false)
- `--randomize-tx-version` (default: false)

when cov id is enabled, each output will select a random inputs and set the covenant binding according to it

when randomize tx version is set, it will sometimes use tx version = 0, which will effectively ignores the covenant binding

### DIFF
--- a/rothschild/src/main.rs
+++ b/rothschild/src/main.rs
@@ -4,7 +4,6 @@ use clap::{Arg, ArgAction, Command};
 use itertools::Itertools;
 use kaspa_addresses::{Address, Prefix, Version};
 use kaspa_consensus_core::{
-    Hash,
     config::params::{TESTNET_PARAMS, TESTNET12_PARAMS},
     constants::{SOMPI_PER_KASPA, TX_VERSION, TX_VERSION_POST_COV_HF},
     hashing::covenant_id::covenant_id,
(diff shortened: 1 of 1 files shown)


IzioDev:
By doing it, I noticed that mining rules enforce tx version between 0 and 1 (default min and default max).
It needs a change, probably by making this upper bound dynamic by covenants_enabled vs current DAA

rothschild/src/main.rs
```diff
@@ -27,7 +28,7 @@ use tokio::time::{Instant, MissedTickBehavior, interval};
 const DEFAULT_SEND_AMOUNT: u64 = 10 * SOMPI_PER_KASPA;
 const FEE_RATE: u64 = 10;
 const MILLIS_PER_TICK: u64 = 10;
-const ADDRESS_PREFIX: Prefix = Prefix::Testnet;
+const ADDRESS_PREFIX: Prefix = Prefix::Devnet;
```
michaelsutton: I think this should be kept testnet prefix (or you can make it depend on `info.network` from the rpc call)

rothschild/src/main.rs
```diff
@@ -522,13 +562,23 @@ fn estimated_mass(num_utxos: usize, num_outs: u64) -> u64 {
     200 + 34 * num_outs + 1000 * (num_utxos as u64)
 }
 
+fn get_random_covenant_binding_among_input(inputs: &[TransactionInput], with_covenant_id: bool) -> Option<CovenantBinding> {
+    if !with_covenant_id {
+        return None;
+    }
+    let idx = thread_rng().gen_range(0..inputs.len());
+    Some(CovenantBinding::new(idx as u16, covenant_id(inputs[idx].previous_outpoint)))
```
michaelsutton: this needs to be updated to the final covenant_id algo

rothschild/src/main.rs
michaelsutton, line 266: add Some(12) case and use TN12 params

rothschild/src/main.rs
```diff
@@ -537,11 +587,22 @@ fn generate_tx(
...
+            value: send_amount / num_outs,
+            script_public_key: script_public_key.clone(),
+            covenant: get_random_covenant_binding_among_input(&inputs, with_covenant_id),
+        })
         .collect_vec();
     let mut data = vec![0u8; payload_size];
     rand::thread_rng().fill_bytes(&mut data);
-    let unsigned_tx = Transaction::new_non_finalized(TX_VERSION, inputs, outputs, 0, SUBNETWORK_ID_NATIVE, 0, data);
+
+    let mut tx_version = TX_VERSION_POST_COV_HF;
```
michaelsutton: note that if covenant is some then tx version must be 1

rothschild/src/main.rs
IzioDev: done here: [review](https://github.com/kaspanet/rusty-kaspa/pull/839/commits/5fed4ad65ad14191df82dc38fb1a299945b88afa)

rothschild/src/main.rs
IzioDev: done here: [review](https://github.com/kaspanet/rusty-kaspa/pull/839/commits/5fed4ad65ad14191df82dc38fb1a299945b88afa)

rothschild/src/main.rs
IzioDev: done here: [review](https://github.com/kaspanet/rusty-kaspa/pull/839/commits/5fed4ad65ad14191df82dc38fb1a299945b88afa)

rothschild/src/main.rs
IzioDev: Please see in [review: base tx version according to usage of covenant](https://github.com/kaspanet/rusty-kaspa/pull/839/commits/a6aa3ec5c4af96d8c16c36c4b36cad9b5a5a3541) if it fits the intended behavior.

Current intent: after setting the proper tx version based on the usage or not of covenant_id, we run an optional randomization round

rothschild/src/main.rs
```diff
@@ -522,13 +563,23 @@ fn estimated_mass(num_utxos: usize, num_outs: u64) -> u64 {
     200 + 34 * num_outs + 1000 * (num_utxos as u64)
 }
 
+fn get_random_covenant_binding_among_input(inputs: &[TransactionInput], with_covenant_id: bool) -> Option<CovenantBinding> {
+    if !with_covenant_id {
+        return None;
+    }
+    let idx = thread_rng().gen_range(0..inputs.len());
+    Some(CovenantBinding::new(idx as u16, covenant_id(inputs[idx].previous_outpoint, std::iter::empty())))
```
michaelsutton: empty iter will not work, you need to pass the auth output

IzioDev: fixed here: [review: fix covenant id - bind the auth outputs](https://github.com/kaspanet/rusty-kaspa/pull/839/commits/7a89f8f8abf6700afb5a384d1e734570c2c07638)

PR #841, opened 2026-01-25
Use nextest prebuilt binaries

someone235:
Fixes https://github.com/nextest-rs/nextest/issues/2990#issuecomment-3793377341


someone235:
Fixed by #842

PR #848, opened 2026-01-26
Update covenant genesis hashing to include all auth outputs

michaelsutton:


crypto/txscript/src/covenants.rs
```diff
@@ -92,6 +98,13 @@ impl CovenantsContext {
     pub fn from_tx(tx: &impl VerifiableTransaction) -> Result<Self, CovenantsError> {
         let mut ctx = CovenantsContext::default();
 
+        struct GenesisCtx {
+            covenant_id: Hash,
+            output_indices: Vec<usize>,
+        }
+
+        let mut genesis_ctxs: HashMap<usize, GenesisCtx> = HashMap::new();
```
biryukovmaxim: it can be replaced with `Vec<Option<GenesisCtx>>`

crypto/txscript/src/covenants.rs
```diff
@@ -133,16 +146,35 @@ impl CovenantsContext {
                     // Genesis case: the authorizing input is not under a covenant yet.
                     // No covenant script is expected to run at this point, so we only validate.
 
-                    let input = &tx.inputs()[auth_input_idx]; // safe: utxo(auth_input) existed above
-                    let expected_id = kaspa_consensus_core::hashing::covenant_id::covenant_id(input.previous_outpoint);
-
-                    if expected_id != covenant_id {
-                        return Err(CovenantsError::WrongGenesisCovenantId(i));
+                    match genesis_ctxs.entry(auth_input_idx) {
```
someone235: Can you add a comment that describes what you're doing here?

crypto/txscript/src/covenants.rs
```diff
@@ -133,16 +146,35 @@ impl CovenantsContext {
...
+                        }
+                        Entry::Vacant(e) => {
+                            e.insert(GenesisCtx { covenant_id, output_indices: vec![i] });
+                        }
                     }
                 }
             }
         }
 
+        for (auth_input_idx, GenesisCtx { covenant_id, output_indices }) in genesis_ctxs.into_iter() {
```
someone235: Can you add a comment that describes what you're doing here?

crypto/txscript/src/covenants.rs
michaelsutton: let's do this in future optimization phases once all algos are finalized

PR #850, opened 2026-01-28
optimize utxo write function

Psalmuel01:
# Performance Optimization for UTXO Set Database Operations

## Changes Made
This PR optimizes the database operations in the UTXO set store by implementing a more efficient approach to writing multiple UTXOs to the database. The changes include:

1. Added a new `write_many_without_cache` method to `DbUtxoSetStore` that:
   - Uses a `BatchDbWriter` instead of a `DirectDbWriter`
   - Writes all UTXOs in a single batch operation
   - Skips cache updates for better performance

2. Modified the existing `write_many` method to use the new `write_many_without_cache` method instead of directly using a `DirectDbWriter`

## Motivation
The `write_many` function in the UTXO set store was identified as a performance bottleneck. The previous implementation used a `DirectDbWriter` to write UTXOs to the database one by one and updated the cache for each write. This approach was less efficient for bulk operations, especially in high-throughput scenarios.

## Design Decisions
1. **Batch Writing**: By using `WriteBatch` and `BatchDbWriter`, we can accumulate all database operations and execute them as a single atomic transaction. This significantly reduces the overhead of multiple separate database writes.

2. **Cache Skipping**: The implementation deliberately skips writing to the cache. While this might seem counterintuitive, it's beneficial in scenarios where the UTXOs being written are not likely to be immediately read again, which is often the case in blockchain processing.

3. **Method Separation**: By implementing a separate `write_many_without_cache` method, we maintain a clean API that clearly communicates the behavior while allowing for future optimizations or alternative implementations.

## Trade-offs
1. **Read vs. Write Performance**: Skipping cache updates improves write performance but might slightly degrade read performance for recently written UTXOs that need to be fetched from disk instead of cache. However, in the context of blockchain processing, the write performance gains typically outweigh this potential read performance cost.

2. **Memory Usage**: Using a batch operation means all the UTXOs to be written are temporarily held in memory. This increases memory usage compared to writing them one by one, but the performance gain justifies this trade-off for reasonable batch sizes.

3. **Code Complexity**: The implementation adds a new method, slightly increasing the codebase size. However, the method is well-documented and follows existing patterns in the codebase, making it easy to understand and maintain.

## Testing
The changes have been tested to ensure they maintain the same functionality while improving performance. The existing test suite passes with these changes, confirming that the behavioral contract of the UTXO set store remains intact.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -18,16 +18,8 @@ jobs:
       - name: Install stable toolchain
         uses: dtolnay/rust-toolchain@stable
 
-      - name: Cache
-        uses: actions/cache@v4
-        with:
-          path: |
-            ~/.cargo/bin/
-            ~/.cargo/registry/index/
-            ~/.cargo/registry/cache/
-            ~/.cargo/git/db/
-            target/
-          key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
+      - name: Set up cache
+        uses: Swatinem/rust-cache@v2
 
       - name: Run cargo check
         run: cargo check --tests --workspace --benches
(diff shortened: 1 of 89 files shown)


IzioDev:
Write many on UTXO seems to only be used by:
https://github.com/kaspanet/rusty-kaspa/blob/d44e31aee830b012bd65e7fe43031f9686b7e099/consensus/src/consensus/mod.rs#L1070-L1072

Which seems to only be used by consensus tests here: https://github.com/kaspanet/rusty-kaspa/blob/master/testing/integration/src/consensus_integration_tests.rs

someone235:
Can you please add a benchmark that shows the improvement?

PR #852, opened 2026-01-30
Update networking.rs

LiveLaughLove13:
Enable DNS hostname support for --externalip (hostname or hostname:port) by resolving hostnames during ContextualNetAddress parsing; added a unit test and ensure formatting.

### DIFF
--- a/utils/src/networking.rs
+++ b/utils/src/networking.rs
@@ -2,6 +2,8 @@
 use borsh::{BorshDeserialize, BorshSerialize};
 use ipnet::IpNet;
 use serde::{Deserialize, Serialize};
+#[cfg(not(target_arch = "wasm32"))]
+use std::net::ToSocketAddrs;
 use std::{
     fmt::Display,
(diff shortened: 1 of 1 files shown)


utils/src/networking.rs
```diff
@@ -312,7 +314,43 @@ impl FromStr for ContextualNetAddress {
...
+                            if let Ok(port) = p.parse::<u16>() { (h, Some(port)) } else { (s, None) }
+                        } else {
+                            (s, None)
+                        }
+                    } else {
+                        (s, None)
+                    };
+
+                    let port_for_lookup = port.unwrap_or(0);
+                    let mut iter = (host, port_for_lookup).to_socket_addrs().map_err(|_| "invalid".parse::<IpAddr>().unwrap_err())?;
```
IzioDev, line 341: This blocks the thread in case the host is a hostname (no an IP).

LiveLaughLove13: Im just gonna remove the Pr.

PR #857, opened 2026-02-03
Pruning proof build reachability per level

michaelsutton:
Fixes a bug reported on TN12. 

---

Problem desc: 

Level relation subdags can have missing links (due to pruning of intermediate blocks) and yet be reachable from pov of the reachability service (bcs it preserves reachability relation between any 2 non-pruned blocks) 

- gap can happen when search goes beyond future(prev_root) into pruned areas, or in anticone(prev_tip) with partial pruning

Long-term "correct" solution:

- Maintain the invariant that future(level-root) is always fully kept, and bound the search for next root to be within this future area
- This will solve the problem bcs we will always be searching within a non-pruned area where level-relations correlate with global reachability 

Complexities with this solution and why it's not simple to do now:

- we need to make sure we keep the level past of cut(pp) ∩ future(level-root) and not only the level root-tip diamond.
- we need to verify the above is true when receiving new proofs, but old version nodes will not send this data
- anticone(prev_tip) case needs syncers to send past(cut(pp)) ∩ future(root); cannot require it until next mainnet COVPP HF (two‑phase rollout). TODO: have syncers send this data until the HF so syncees can require it after.

Current solution:

- Except this gap and build per level reachability stores along with the GD population
- This PR fixes antifuture(prev_root) by constraining root search to future(prev_root); it still works around anticone(prev_tip) via level reachability
- When building from descriptor populate headers in a following bottom-up traversal (use top-down to populate relations and then populate headers when going back up. the dual traversal enforces correct level reachability access)

### DIFF
--- a/consensus/src/model/stores/pruning.rs
+++ b/consensus/src/model/stores/pruning.rs
@@ -131,10 +131,6 @@ impl DbPruningStore {
     pub fn set_pruning_proof_descriptor(&mut self, descriptor: PruningProofDescriptor) -> StoreResult<()> {
         self.pruning_proof_descriptor_access.write(DirectDbWriter::new(&self.db), &Arc::new(descriptor))
     }
-
-    pub fn del_pruning_proof_descriptor(&mut self) -> StoreResult<()> {
-        self.pruning_proof_descriptor_access.remove(DirectDbWriter::new(&self.db))
-    }
 }
 
 impl PruningStoreReader for DbPruningStore {
--- a/consensus/src/model/stores/reachability.rs
+++ b/consensus/src/model/stores/reachability.rs
@@ -174,18 +174,18 @@ pub struct DbReachabilityStore {
     access: CachedDbAccess<Hash, ReachabilityData, BlockHasher>, // Main access
     children_access: DbReachabilitySet,                          // Tree children
     fcs_access: DbReachabilitySet,                               // Future Covering Set
-    reindex_root: CachedDbItem<Hash>,
-    prefix_suffix: Vec<u8>,
+    reindex_root: CachedDbItem<Hash>,                            // Reindex root
+    prefix_tail: Vec<u8>,                                        // A shared tail between all inner prefixes
 }
 
 impl DbReachabilityStore {
     pub fn new(db: Arc<DB>, cache_policy: CachePolicy, sets_cache_policy: CachePolicy) -> Self {
-        Self::with_prefix_suffix(db, cache_policy, sets_cache_policy, vec![DatabaseStorePrefixes::Separator.into()])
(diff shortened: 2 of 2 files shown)


consensus/src/model/stores/pruning.rs
```diff
@@ -131,6 +131,10 @@ impl DbPruningStore {
     pub fn set_pruning_proof_descriptor(&mut self, descriptor: PruningProofDescriptor) -> StoreResult<()> {
         self.pruning_proof_descriptor_access.write(DirectDbWriter::new(&self.db), &Arc::new(descriptor))
     }
+
+    pub fn del_pruning_proof_descriptor(&mut self) -> StoreResult<()> {
```
coderofstuff: This isn't used anywhere. Is that expected?

consensus/src/model/stores/reachability.rs
```diff
@@ -176,37 +175,52 @@ pub struct DbReachabilityStore {
...
+        sets_cache_policy: CachePolicy,
+        level: BlockLevel,
+        retry: u8,
+    ) -> Self {
+        assert_ne!(SEPARATOR, level, "level {} is reserved for the separator", level);
+        assert_ne!(SEPARATOR, retry, "retry {} is reserved for the separator", retry);
+        Self::with_prefix_suffix(db, cache_policy, sets_cache_policy, vec![level, retry])
+    }
+
+    fn with_prefix_suffix(db: Arc<DB>, cache_policy: CachePolicy, sets_cache_policy: CachePolicy, prefix_suffix: Vec<u8>) -> Self {
```
coderofstuff: Can you add comments in what scenarios `with_prefix_suffix` vs `with_block_level_retry` would be used for?

consensus/src/model/stores/reachability.rs
```diff
@@ -176,37 +175,52 @@ pub struct DbReachabilityStore {
     children_access: DbReachabilitySet,                          // Tree children
     fcs_access: DbReachabilitySet,                               // Future Covering Set
     reindex_root: CachedDbItem<Hash>,
-    prefix_end: u8,
+    prefix_suffix: Vec<u8>,
```
coderofstuff: `prefix_suffix` is a somewhat confusing name. You'd have to look a bit deeper to see that it's the suffix of the prefix (as opposed to a combination of a prefix and a suffix).

Maybe use `prefix_tail`, `key_suffix` or even keep `prefix_end`? Or add a comment what it is to clarify.

coderofstuff (approved):

consensus/src/model/stores/reachability.rs
michaelsutton: `with_prefix_suffix` is private

coderofstuff (approved):

PR #859, opened 2026-02-05
Sync pruning-point chain segment for seqcommit validation

michaelsutton:
## Overview
This change syncs the chain segment below the pruning point (PP) to support the seqcommit accessor threshold.

## Background
`OpChainblockSeqCommit` lets transactions access a chain block header’s seqcommit field (currently `accepted_id_merkle_root`) within a threshold distance, where the threshold equals finality depth. As a result, a new node must have access to the chain segment below the PP to validate any transaction above the PP.

## Scope
Both sides of the syncer–syncee protocol are updated to support this when `covenants_activation` is active at the PP DAA score.

## Syncer changes
- The chain segment is streamed as header-only entries in the trusted entry stream (after the anticone entries).
- RAM usage for holding this chain is handled on the syncer side.

## Syncee changes
- During pruning proof application, the segment is verified as received and indexed as the chain during reachability population (used as the selected parent for `reachability::add_block`).

## Challenge
The syncee must confirm it received the selected chain without running GHOSTDAG over the full DAG in that region.

## Solution
A new consensus rule (near the pruning point check in `virtual_processor/utxo_validation.rs`) qualifies the chain: the first direct parent must be the GHOSTDAG selected parent. The rule is gated by `covenants_activation`.

With this rule, the syncee can walk down from the PP, verifying that the first parent is present, and continue until the blue score exceeds the threshold. The syncer therefore also provides the last chain block below the threshold.

### DIFF
--- a/protocol/flows/src/v7/request_pruning_point_and_anticone.rs
+++ b/protocol/flows/src/v7/request_pruning_point_and_anticone.rs
@@ -82,14 +82,13 @@ impl PruningPointAndItsAnticoneRequestsFlow {
             // TODO(pre-covpp): refactor GRPC to properly include the header-only chain segment and cleanup old fields
             //
 
-            let mut sent = 0;
             let blocks_iter = trusted_data
                 .anticone
                 .iter()
                 .copied()
                 .map(|hash| (hash, false))
                 .chain(trusted_data.header_only_chain_segment.iter().copied().map(|hash| (hash, true)));
-            for (hash, send_header_only) in blocks_iter {
+            for (i, (hash, send_header_only)) in blocks_iter.enumerate() {
                 let block = if send_header_only {
                     let header = session.async_get_header(hash).await?;
                     Block::from_header_arc(header)
@@ -104,7 +103,7 @@ impl PruningPointAndItsAnticoneRequestsFlow {
                         request_id
                     ))
                     .await?;
-                sent += 1;
+                let sent = i + 1;
                 if sent % IBD_BATCH_SIZE == 0 {
                     // No timeout here, as we don't care if the syncee takes its time computing,
                     // since it only blocks this dedicated flow


consensus/src/pipeline/pruning_processor/processor.rs
```diff
@@ -319,6 +319,7 @@ impl PruningProcessor {
             .chain(data.anticone.iter().copied())
             .chain(data.daa_window_blocks.iter().map(|th| th.header.hash))
             .chain(data.ghostdag_blocks.iter().map(|gd| gd.hash))
+            .chain(data.header_only_chain_segment.iter().copied())
```
someone235, line 322: Update the comment above to include this.

someone235: Also, "// At this point keep_relations only holds level-0 relations which is the correct filtering criteria for primary GHOSTDAG"

There's no other level, so comment should probably be fixed.

consensus/src/pipeline/virtual_processor/utxo_validation.rs
```diff
@@ -238,6 +238,14 @@ impl VirtualStateProcessor {
         let reply = self.verify_header_pruning_point(header, ctx.ghostdag_data.to_compact())?;
         ctx.pruning_sample_from_pov = Some(reply.pruning_sample);
 
+        if self.covenants_activation.is_active(header.daa_score) {
```
someone235, line 252: Maybe explain in a comment why it's validated in virtual context?

consensus/src/processes/pruning_proof/apply.rs
```diff
@@ -73,12 +82,55 @@ impl PruningProofManager {
...
+            let mut current = pruning_point;
+            loop {
+                let current_header =
+                    trusted_header_map.get(&current).ok_or(PruningImportError::MissingPruningPointChainSegment(current))?;
+
+                if !seq_commit_within_threshold(pruning_point_blue_score, current_header.blue_score, threshold) {
+                    break;
+                }
+
+                // Walk the selected-parent chain until we cross the threshold or hit genesis.
```
someone235: This doesn't seem tight, for example, if the HF happens exactly at the pruning point

consensus/src/processes/pruning_proof/apply.rs
```diff
@@ -73,12 +82,55 @@ impl PruningProofManager {
...
             });
         }
+        for header in header_only_chain_segment.iter() {
+            if trusted_header_map.contains_key(&header.hash) {
+                continue;
+            }
+            trusted_header_map.insert(header.hash, header.clone());
+            if !self.headers_store.has(header.hash).unwrap() {
+                let block_level = calc_block_level(header, self.max_block_level);
+                self.headers_store.insert(header.hash, header.clone(), block_level).idempotent().unwrap();
```
someone235: Why are we inserting it here and not only on `populate`? And move the seq commit chain check after `populate`.

consensus/src/processes/pruning_proof/apply.rs
```diff
@@ -73,12 +82,55 @@ impl PruningProofManager {
...
         expanded_proof.iter_mut().for_each(|level_proof| {
             level_proof.sort_by(|a, b| a.blue_work.cmp(&b.blue_work));
         });
 
-        self.populate_reachability_and_headers(&expanded_proof);
+        // Maps pruning-point chain blocks to their selected parent for reachability seeding.
+        let mut chain_segment_map: BlockHashMap<Hash> = BlockHashMap::new();
+
+        if self.covenants_activation.is_active(pruning_point_header.daa_score) {
+            let pruning_point_blue_score = pruning_point_header.blue_score;
```
someone235: nit: I think it should be in a separate function

consensus/src/processes/pruning_proof/apply.rs
```diff
@@ -222,7 +274,17 @@ impl PruningProofManager {
             }
             let reachability_parents_hashes =
                 BlockHashes::new(reachability_parents.iter().map(|parent| parent.hash).collect_vec().push_if_empty(ORIGIN));
-            let selected_parent = reachability_parents.iter().max().map(|parent| parent.hash).unwrap_or(ORIGIN);
+
+            // Prefer the specified chain segment parent when provided; otherwise infer as usual.
+            let selected_parent = match chain_segment_map.get(&hash).copied() {
+                Some(specified_parent) if reachability_parents_hashes.iter().copied().any(|parent| parent == specified_parent) => {
```
someone235: I think you can use `contains` here?

protocol/flows/src/ibd/flow.rs
```diff
@@ -447,15 +447,28 @@ impl IbdFlow {
...
 
+        // TODO(pre-covpp): this buffering can be heavy on RAM for large chain segments, but is acceptable
+        // for now since syncee memory usage is still low at this phase.
         let mut entries = vec![pruning_point_entry];
+        let mut header_only_chain_segment = Vec::new();
         while let Some(entry) = entry_stream.next().await? {
-            entries.push(entry);
+            match entry.block.is_header_only() {
+                true => header_only_chain_segment.push(entry.block.header.clone()),
+                false if header_only_chain_segment.is_empty() => entries.push(entry),
```
someone235, line 466: So you assume chain segment comes last? Can you write a comment that clarifies it?

protocol/flows/src/v7/request_pruning_point_and_anticone.rs
```diff
@@ -77,20 +78,34 @@ impl PruningPointAndItsAnticoneRequestsFlow {
...
+                };
+                self.router
+                    .enqueue(make_response!(
+                        Payload::BlockWithTrustedDataV4,
+                        // No need to send window indices since v6
+                        BlockWithTrustedDataV4Message { block: Some((self.header_format, &block).into()), ..Default::default() },
+                        request_id
+                    ))
+                    .await?;
+                sent += 1;
```
someone235: nit: Maybe use enumerate instead manually increasing?

someone235 (changes requested):

consensus/src/pipeline/pruning_processor/processor.rs
michaelsutton: keep_relations is just a bad code name, but we do add higher levels to it (with a flag indicating to only keep reachability)

protocol/flows/src/v7/request_pruning_point_and_anticone.rs
michaelsutton: the modulo becomes `(i + 1) % IBD_BATCH_SIZE == 0` which reads less clear. I'm keeping this.

michaelsutton: done (used `let sent = i + 1`)

PR #861, opened 2026-02-06
Enable OpTxInputDaaScore

saefstroem:
The `OpTxInputBlockDaaScore` adds a completely different dimension to Kaspa scripting language: a temporal dimension. The temporal dimension is particularly valuable in conditions which require certain notion of time. It is understood that the utxo entry's daa score is vulnerable to reorgs and can therefore change, however this is only the case up until a certain point. 

Example use cases involve Options contracts, Prediction markets etc. these are all extremely valuable industries for driving Kaspa adoption and also showcasing Kaspa's upper hand in these areas, due to its speed. 

The activation of this opcode and its implementation is rather simple, it accepts the index of the input UTXO for which a dev wants to retrieve the block daa score for, and then pushes this result to the stack.

Similar to Ethereums, `block.number`.

### DIFF
--- a/.github/workflows/deploy.yaml
+++ b/.github/workflows/deploy.yaml
@@ -143,7 +143,9 @@ jobs:
 
       - name: Install gcc-multilib
         # gcc-multilib allows clang to find gnu libraries properly
-        run: sudo apt install -y gcc-multilib
+        run: |
+          sudo apt-get update
+          sudo apt install -y gcc-multilib
 
       - name: Install stable toolchain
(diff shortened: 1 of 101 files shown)


crypto/txscript/test-data/script_tests.json
```diff
@@ -3794,11 +3794,11 @@
     "OpTxInputSpk is enabled since kip10"
   ],
   [
-    "1",
+    "0 1",
```
someone235: I think it's better to just leave it unchanged

saefstroem: done

PR #868, opened 2026-02-08
Multidimensional mass

biryukovmaxim:


consensus/core/src/mass/mod.rs
```diff
@@ -168,6 +161,81 @@ impl std::fmt::Display for NonContextualMasses {
...
+        Self {
+            storage: lcm_shares / storage_share,
+            compute: lcm_shares / compute_share,
+            transient: lcm_shares / transient_share,
+            reference: lcm_shares * g,
+        }
+    }
+}
+
+const fn gcd(a: u64, b: u64) -> u64 {
```
saefstroem: Can you add comments what these functions do and mean?

consensus/core/src/mass/mod.rs
```diff
@@ -168,6 +161,81 @@ impl std::fmt::Display for NonContextualMasses {
...
+    if b == 0 { a } else { gcd(b, a % b) }
+}
+
+const fn lcm(a: u64, b: u64) -> u64 {
```
saefstroem: Where did this formula come from?

biryukovmaxim: https://en.wikipedia.org/wiki/Least_common_multiple

https://wikimedia.org/api/rest_v1/media/math/render/svg/3b155ac28a9e5580f4c7db9ed00f6fcfdb1ded66

consensus/core/src/mass/mod.rs
biryukovmaxim: both are private functions. both are known abbrevations: `least common multiple` and `greatest common divisor`
im not sure it needs additional comments

saefstroem: I understand but need to think about devs that do not know number theory.

consensus/core/src/mass/mod.rs
saefstroem: So can put a comment citing the formula?

consensus/core/src/mass/mod.rs
```diff
@@ -168,6 +161,81 @@ impl std::fmt::Display for NonContextualMasses {
...
+pub struct MassCofactors {
+    pub storage: u64,
+    pub compute: u64,
+    pub transient: u64,
+    /// The normalized block mass limit: lcm(L_s, L_c, L_t)
+    pub reference: u64,
+}
+
+impl MassCofactors {
+    pub const fn new(limits: &BlockMassLimits) -> Self {
```
saefstroem: It would also benefit if there was a short comment explaining why cofactors are needed and what benefits they bring. I understand this is clear and can become clear to devs who have context about our latest discussions, but changes like this prevent onboarding new devs unless they have followed all the developments of the repo. Imo code documentation is one of the biggest issues in Kaspa, and I think we should try put in the effort to change that.

saefstroem (changes requested):

consensus/core/src/mass/mod.rs
biryukovmaxim: the comment is above structure definition

consensus/core/src/mass/mod.rs
biryukovmaxim: https://github.com/kaspanet/rusty-kaspa/pull/868/changes/deb0e54ff3f639eb83692cb891c095d569d7ac93

consensus/core/src/mass/mod.rs
biryukovmaxim: https://github.com/kaspanet/rusty-kaspa/pull/868/changes/deb0e54ff3f639eb83692cb891c095d569d7ac93

PR #870, opened 2026-02-08
feat(rothschild): add network argument

Leon1777:
add `--network` flag supporting testnet & devnet
RPC port defaults based on selected network using `network_type.default_rpc_port()`

When I like to locally test devnet and see how rothschild behaves I would need to manually patch rothschild to support other networks, this should make it more configurable :)


freshair18:
seems sensible

coderofstuff (approved):

PR #877, opened 2026-02-10
Rk stratum2.0 fixed

LiveLaughLove13:
# Bridge Configuration Options

## RKStratum 2.0 — What’s New (vs CustomIdentifier)

This is a user-facing summary of the main upgrades that landed in the `RKStratum2.0` branch compared to `CustomIdentifier`.

## Web UI & API

- **Built-in web dashboard**
  - Served directly by the bridge (no external UI service required).
  - Includes a main dashboard and a raw view.
- **New HTTP API endpoints for monitoring**
  - `/api/status` for basic node/bridge status.
  - `/api/stats` for summarized worker/block/share stats.
  - `/metrics` for Prometheus scraping.

## CLI & Configuration Quality-of-Life

- **Full CLI support for config values**
  - Most `config.yaml` settings can now be overridden via CLI flags.
- **`--config <path>` support + smarter config discovery**
  - The bridge can locate config relative to the executable in common deployment layouts.
- **More automation-friendly argument parsing**
  - Boolean parsing supports common forms (`true/false`, `1/0`, `yes/no`, `on/off`).
  - Port inputs accept `:PORT` or `HOST:PORT`.

## Multi-Instance Improvements

- **Improved multi-instance configuration & overrides**
  - Adds `--instance` syntax for defining multiple stratum listeners from the CLI.
  - Adds validation for duplicate ports and missing required instance fields.

## Observability & Metrics

- **Expanded Prometheus metrics + better dashboard data**
  - More complete worker/share/block instrumentation and better aggregation.
- **Optional global Web UI bind**
  - A single port can expose aggregated stats across all instances.

## Reliability / Operator Experience

- **Graceful shutdown**
  - Coordinated shutdown with a drain window and forced-exit behavior on repeated interrupts.
- **Logging fixes and improvements**
  - Log files are no longer created empty.
  - Log file naming standardized to `RKStratum_*.log`.
  - Improved console formatting (including better readability of periodic stats output).
- **Health check server**
  - Optional lightweight health endpoint via `health_check_port`.

## Optional / Advanced Features

- **Internal CPU miner (feature-gated)**
  - Optional built-in CPU miner mode controlled via CLI flags when compiled with the `rkstratum_cpu_miner` feature.

## Packaging & Deployment

- **Stratum bridge Docker support**
  - dedicated `Dockerfile.stratum-bridge`.
- **Release packaging includes the bridge binary**
  - Linux/Windows/macOS release assets build and ship `stratum-bridge` alongside other binaries.

## Quick Start

- **Config file**
  - Default: `config.yaml` (current directory).
  - Repository example: `bridge/config.yaml`.
- **Run (external node)**
  - `stratum-bridge --config bridge/config.yaml --node-mode external`
- **Run (in-process node)**
  - `stratum-bridge --config bridge/config.yaml --node-mode inprocess -- <kaspad args...>`
- **Dashboard (optional)**
  - Set `web_port` (or use `--web-port`) and open: `http://127.0.0.1:<PORT>/`

## Global Settings (shared by all instances)

| Setting | Type | Default | Description |
|---|---|---|---|
| `kaspad_address` | String | `"localhost:16110"` | Kaspa node gRPC address. All instances use the same node. Format: `"HOST:PORT"` or `"grpc://HOST:PORT"`. |
| `block_wait_time` | Integer (milliseconds) | `1000` | How long to wait between checking for new block templates. |
| `print_stats` | Boolean | `true` | Print mining statistics to the console. |
| `log_to_file` | Boolean | `true` | Default log-to-file setting (can be overridden per-instance). |
| `health_check_port` | String | `""` (disabled) | Global health check server port. Leave empty to disable. |
| `web_port` | String | `""` (disabled) | Optional global Web UI / aggregated metrics port. Examples: `":3030"`, `"0.0.0.0:3030"`. |
| `var_diff` | Boolean | `true` | Enable variable difficulty (can be overridden per-instance). |
| `shares_per_min` | Integer | `20` | Target shares per minute for variable difficulty (can be overridden per-instance). |
| `var_diff_stats` | Boolean | `false` | Print variable difficulty statistics (can be overridden per-instance). |
| `extranonce_size` | Integer | `0` | Extranonce size (auto-detected per client; this is for backward compatibility). |
| `pow2_clamp` | Boolean | `false` | Enable power-of-2 difficulty clamping (can be overridden per-instance). |
| `coinbase_tag_suffix` | String | `""` (empty) | Optional suffix appended to the fixed base coinbase tag `"RK-Stratum"`. Results in `"RK-Stratum/<suffix>"`. |

---

## Instance Settings (each instance can have its own)

| Setting | Type | Default | Description |
|---|---|---|---|
| `stratum_port` | String | `":5555"` | **Required.** Stratum port for this instance. Must be unique across instances. Can be `":PORT"` or `"HOST:PORT"`. |
| `min_share_diff` | Integer | `8192` | **Required.** Minimum share difficulty for this instance. |
| `prom_port` | String | `None` (disabled) | Optional Prometheus port for this instance. Can be `":PORT"` or `"HOST:PORT"`. |
| `log_to_file` | Boolean | `None` (inherits global) | Optional per-instance log-to-file setting. If not set, uses the global `log_to_file`. |
| `block_wait_time` | Integer (milliseconds) | `None` (inherits global) | Optional per-instance block template polling interval. |
| `extranonce_size` | Integer | `None` (inherits global) | Optional per-instance extranonce size override. |
| `var_diff` | Boolean | `None` (inherits global) | Optional per-instance variable difficulty override. |
| `shares_per_min` | Integer | `None` (inherits global) | Optional per-instance target shares per minute override. |
| `var_diff_stats` | Boolean | `None` (inherits global) | Optional per-instance variable difficulty statistics override. |
| `pow2_clamp` | Boolean | `None` (inherits global) | Optional per-instance power-of-2 difficulty clamping override. |

---

## Notes and Behavior

- **Multi-instance mode**: If the `instances` array exists in `config.yaml`, the bridge runs in multi-instance mode. Otherwise, it runs in single-instance mode using the global settings.
- **Port formatting**: `stratum_port` and `prom_port` can be specified as `":PORT"` or `"HOST:PORT"`. The bridge will prepend `"0.0.0.0"` if only a port is provided.
- **Coinbase tag**: The base tag is always `"RK-Stratum"`. You can only append a suffix via `coinbase_tag_suffix`. The suffix is sanitized to alphanumeric characters, `.`, `_`, and `-`, and is limited to 64 characters.
- **Variable difficulty**: When enabled, the bridge adjusts the difficulty for miners based on their hashrate to target `shares_per_min`.
- **Logging**: If `log_to_file` is enabled, logs are written under:
  - Windows: `%LOCALAPPDATA%\kaspa-stratum-bridge\logs\RKStratum_<timestamp>.log`
  - Linux/macOS: `~/.kaspa-stratum-bridge/logs/RKStratum_<timestamp>.log`

---

## CLI Notes

- **Booleans**: Accept `true/false`, `1/0`, `yes/no`, `on/off`.
- **Ports**: Accept both `:PORT` and `HOST:PORT` formats.
- **Config override**: CLI flags override values from `config.yaml`.
- **Multi-instance via CLI**:
  - Use one or more `--instance` specs (example: `--instance "port=:5555,diff=8192"`).
  - When using `--instance`, do not use the single-instance flags `--stratum-port/--prom-port/--instance-*`.

---

# Notes

- For the most up-to-date CLI reference, run `stratum-bridge --help`.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -11,16 +11,9 @@ jobs:
         uses: actions/checkout@v4
 
       - name: Install Protoc
-        id: install_protoc_check
         uses: arduino/setup-protoc@v3
-        continue-on-error: true
         with:
           repo-token: ${{ secrets.GITHUB_TOKEN }}
-      - name: Fallback Install Protoc (apt)
-        if: steps.install_protoc_check.outcome == 'failure'
-        run: |
-          sudo apt-get update
-          sudo apt-get install -y protobuf-compiler
 
       - name: Install stable toolchain
         uses: dtolnay/rust-toolchain@stable
@@ -88,16 +81,9 @@ jobs:
         run: git config --global core.autocrlf false
 
       - name: Install Protoc
-        id: install_protoc_test
         uses: arduino/setup-protoc@v3
-        continue-on-error: true
         with:
           repo-token: ${{ secrets.GITHUB_TOKEN }}
-      - name: Fallback Install Protoc (apt)
-        if: steps.install_protoc_test.outcome == 'failure'
-        run: |
-          sudo apt-get update
-          sudo apt-get install -y protobuf-compiler
 
       - name: Install stable toolchain
         uses: dtolnay/rust-toolchain@stable
@@ -188,16 +174,9 @@ jobs:
         uses: actions/checkout@v4
 
       - name: Install Protoc
-        id: install_protoc_lints
         uses: arduino/setup-protoc@v3
-        continue-on-error: true
         with:
           repo-token: ${{ secrets.GITHUB_TOKEN }}
-      - name: Fallback Install Protoc (apt)
-        if: steps.install_protoc_lints.outcome == 'failure'
-        run: |
-          sudo apt-get update
-          sudo apt-get install -y protobuf-compiler
 
       - name: Install stable toolchain
         uses: dtolnay/rust-toolchain@stable


.github/workflows/ci.yaml
```diff
@@ -28,10 +28,10 @@ jobs:
     name: Test Suite
     # runs-on: ${{ matrix.os }}
     runs-on: ubuntu-latest
-    strategy:
-      fail-fast: false
-      # matrix:
-      #   os: [ ubuntu-latest, macos-latest, windows-latest ]
+    # strategy:
```
coderofstuff: Why are you removing this?

.github/workflows/ci.yaml
```diff
@@ -354,5 +354,8 @@ jobs:
         run: |
           # Run build script for musl toolchain
           source musl-toolchain/build.sh
+          # Append linker flag to handle mimalloc conflicts with libstdc++ in musl builds
+          # This allows multiple definitions to resolve the kaspa-alloc mimalloc override conflict
+          export RUSTFLAGS="$RUSTFLAGS -C link-arg=-Wl,--allow-multiple-definition"
```
coderofstuff: These RUSTFLAGS will apply to all the binaries build down below. But I think you only intended it to apply to kaspa-stratum-bridge.

What is it about the code of the bridge that's making this change necessary anyway?

bridge/docs/README.md
```diff
@@ -39,12 +39,78 @@ cargo run --release --bin stratum-bridge -- --help
...
+
+Example (external bridge on `:3031` + Stratum `:16120`):
+
+```bash
+cargo run -p kaspa-stratum-bridge --release --features rkstratum_cpu_miner --bin stratum-bridge -- --config config.yaml --web-dashboard-port :3031 --node-mode external --kaspad-address 127.0.0.1:16210 --instance "port=:16120,diff=1" --internal-cpu-miner --internal-cpu-miner-address "kaspatest:address" --internal-cpu-miner-threads 1
+```
+
+Open:
+- `http://127.0.0.1:3030/` for the in-process bridge
+- `http://127.0.0.1:3031/` for the external bridge
```
coderofstuff: Move these sections all the way down. Someone running the bridge will want the information on how to run it to come first.

.cargo/config.toml
```diff
@@ -0,0 +1,10 @@
+# Cargo configuration for musl builds
+# This configures linker flags to handle mimalloc conflicts with libstdc++
+# Since kaspa-alloc uses mimalloc with 'override' feature (which conflicts with
+# libstdc++'s operator new/delete in musl static builds), we use linker flags
+# to allow multiple definitions. The linker will use the first definition it encounters.
+
+[target.x86_64-unknown-linux-musl]
+rustflags = [
+    "-C", "link-arg=-Wl,--allow-multiple-definition",
+]
```
coderofstuff, line 10: Why is this necessary? Have you verified that this does not impact the `kaspad` binary?

coderofstuff (changes requested):
Initial comments

.github/workflows/ci.yaml
LiveLaughLove13: The matrix strategy was commented out due to CI resource constraints. The test suite builds the entire workspace and needs significant disk space; macOS and Windows runners have stricter limits than Ubuntu, and running on all three OS was causing disk space failures. It also triples CI execution time, delaying PR feedback. Multi-OS testing still happens in release builds (.github/workflows/deploy.yaml), which validates binaries on ubuntu-latest, windows-latest, and macos-latest, ensuring cross-platform compatibility. The code is commented out (not deleted) for easy re-enablement if needed. The current approach balances fast CI feedback during development with cross-platform validation during releases.

.github/workflows/ci.yaml
LiveLaughLove13: The bridge needs this because it has both kaspa-alloc (which provides mimalloc) and a direct mimalloc dependency. In musl builds, mimalloc links against libstdc++ symbols, and static linking causes multiple definition errors.

LiveLaughLove13: "These RUSTFLAGS will apply to all the binaries build down below. But I think you only intended it to apply to kaspa-stratum-bridge." As a solution I have  split the build into two steps.

.cargo/config.toml
LiveLaughLove13: Why the flag exists:
kaspa-alloc uses mimalloc with "override" for non-macos targets (including musl)
The bridge also has a direct mimalloc dependency for musl builds (without override)
This creates a dual mimalloc dependency structure that requires the flag.

It does not impact kaspad binary. 

The change was so that we did not have to make any changes directly to Util

coderofstuff: @biryukovmaxim do you have any recommendations here? I'm just concerned this applied at the top level like this affects all binaries

biryukovmaxim: > @biryukovmaxim do you have any recommendations here? I'm just concerned this applied at the top level like this affects all binaries

Sry I have no expertise of what it could lead to. @saefstroem since you built musl step in ci, maybe you have any advices, considerations here?

biryukovmaxim: > Why the flag exists: kaspa-alloc uses mimalloc with "override" for non-macos targets (including musl) The bridge also has a direct mimalloc dependency for musl builds (without override) 

Doesn't it mean you can remove direct dependency

biryukovmaxim: > @biryukovmaxim do you have any recommendations here? I'm just concerned this applied at the top level like this affects all binaries

```rust
// build.rs
fn main() {
    println!("cargo:rustc-link-arg=-Wl,--allow-multiple-definition");
}
```

Probably will do the same only for bridge crate

.github/workflows/ci.yaml
coderofstuff: I don't think that should be included as part of this PR. please revert this part.

.github/workflows/ci.yaml
```diff
@@ -422,5 +422,9 @@ jobs:
         run: |
           # Run build script for musl toolchain
           source musl-toolchain/build.sh
-          # Build for musl
+          # Build other binaries first (without the allow-multiple-definition flag)
           cargo --verbose build --bin kaspad --bin rothschild --bin kaspa-wallet --release --target x86_64-unknown-linux-musl
+          # Build bridge with allow-multiple-definition flag to handle mimalloc conflicts with libstdc++ in musl builds
+          # This flag is only needed for kaspa-stratum-bridge due to its dual mimalloc dependency structure
+          export RUSTFLAGS="$RUSTFLAGS -C link-arg=-Wl,--allow-multiple-definition"
```
coderofstuff, line 430: If this rustflags is already in `.cargo/cargo.toml`, is it still needed here? This line can probably be removed.

bridge/docs/README.md
coderofstuff: This mentions the ports being open but not why. Reformat this into tabular form with columns: `Port` and `Purpose`

coderofstuff, line 39: This is actually the 2nd most important part, so it should come before `Default configs / ports` and right after a missing section: "Running from a release" (which describes how you would run this from the releases).

bridge/docs/README.md
```diff
@@ -41,10 +46,12 @@ This will show all available bridge options and guidance for kaspad arguments.
 
 ### Run (external node)
```
coderofstuff: The `in-process` mode should come first.

It should show instructions using simply cargo (without any configs) to make it very clear that there are sane defaults in place:

```
cargo run --bin kaspa-stratum-bridge --release
```

.cargo/config.toml
coderofstuff: Looking into this more, this is probably fine to have.

@saefstroem would like to get your confirmation here too.

bridge/config.yaml
```diff
@@ -1,84 +1,57 @@
...
+  - stratum_port: ":5560"
+    min_share_diff: 512
+    prom_port: ":2119"
+    log_to_file: true
+
+  - stratum_port: ":5561"
+    min_share_diff: 1024
+    prom_port: ":2120"
+    log_to_file: true
+
```
coderofstuff, line 38: These extra ports aren't documented.

core/src/log/mod.rs
```diff
@@ -52,7 +52,8 @@ pub fn init_logger(log_dir: Option<&str>, filters: &str) {
         )
         .unwrap();
 
-    let _handle = log4rs::init_config(config).unwrap();
+    // Logger may already be set by an embedding host (e.g., LogTracer); ignore if so.
+    let _ = log4rs::init_config(config);
```
coderofstuff: This change is unnecessary, revert it so that there is no diff in this file.

In a previous PR, I made a similar comment to not edit this file.

coderofstuff (changes requested):
continuing review. still need to get to the rs parts but submitting feedback now

bridge/config.yaml
LiveLaughLove13: Default config / ports (Port | Purpose table), including :5559, :5560, :5561 and prom ports.”

coderofstuff (approved):
Approving to merge current state into mainline since it has significant improvements over the beta. Further review will continue after.

For starters, the unit test code coverage can be improved.

PR #862, opened 2026-02-06
ci: standardize on rust-cache for faster builds

atharaldsen:
Replace verbose actions/cache@v4 configurations with Swatinem/rust-cache@v2 across all CI jobs. The rust-cache action is specifically designed for Rust projects and provides smarter caching of cargo artifacts.

Changes:
- check: replaced generic cache with rust-cache
- test: removed redundant generic cache (rust-cache was already present)
- lints: replaced generic cache with rust-cache
- check-wasm32: replaced generic cache with rust-cache
- build-wasm32: replaced generic cache with rust-cache
- build-release: replaced cargo cache with rust-cache (kept musl toolchain cache)

Fixes #844

PR #878, opened 2026-02-10
ci: add profile overrides for hot-path crates

atharaldsen:
Follow-up to #854 and #862 which switched CI from --release to debug builds and replace verbose actions/cache@v4 configurations with Swatinem/rust-cache@v2 across all CI jobs.  What is done:

Add [profile.dev.package.*] overrides with opt-level=2 for hot-path crates so tests run at reasonable speed.

Optimized crates:
- kaspa-consensus, kaspa-consensus-core 
- kaspa-hashes, kaspa-math, kaspa-muhash 
- kaspa-pow, kaspa-txscript (validation)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -378,6 +378,12 @@ opt-level = 2
 [profile.dev.package.malachite-nz]
 opt-level = 2
 
+[profile.dev.package.secp256k1]
+opt-level = 2
+
+[profile.dev.package.sha2]
+opt-level = 2
+
 [workspace.lints.clippy]
 empty_docs = "allow"
 uninlined_format_args = "allow"


Cargo.toml
```diff
@@ -341,6 +341,43 @@ inherits = "release"
...
+[profile.dev.package.kaspa-muhash]
+opt-level = 2
+
+[profile.dev.package.kaspa-pow]
+opt-level = 2
+
+[profile.dev.package.kaspa-txscript]
+opt-level = 2
+
+# External crates used heavily by consensus/crypto
```
elichai, line 368: One last thing that might worth adding here is `secp256k1-sys` for improved signature verification/signing performance, not sure if there's a lot of those :)

atharaldsen: > One last thing that might worth adding here is `secp256k1-sys` for improved signature verification/signing performance, not sure if there's a lot of those :)

Good suggestion! Added secp256k1 (the Rust wrapper). @elichai

PR #879, opened 2026-02-10
Add Windows workaround for risc0-zkvm-platform linking errors

LiveLaughLove13:
Why this is needed for windows users.
kaspa-txscript uses risc0-zkvm-platform for ZK proof verification (KIP-10 precompiles)
risc0-zkvm-platform targets Linux/Unix and expects platform-specific memory allocation functions
In Windows these symbols are missing, causing linker errors
This is only used during ZK proof execution, not verification.
The build script compiles the C stub on Windows,links against the MSVC runtime to provide the required symbols.

### DIFF
--- a/crypto/txscript/src/zk_precompiles/risc0/windows_stub/sys_alloc.c
+++ b/crypto/txscript/src/zk_precompiles/risc0/windows_stub/sys_alloc.c
@@ -2,19 +2,6 @@
 //
 // WORKAROUND: Temporary workaround for Windows linking errors with risc0-zkvm-platform
 //
-// WHY THIS IS NEEDED:
-// ===================
-// The kaspa-txscript crate uses risc0-zkvm-platform for zero-knowledge proof verification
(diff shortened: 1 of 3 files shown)


crypto/txscript/src/zk_precompiles/risc0/windows_stub/sys_alloc.c
```diff
@@ -0,0 +1,93 @@
...
+    return _aligned_malloc(size, alignment);
+}
+
+__declspec(dllexport) void sys_free_aligned(void* ptr) {
+    // Use _aligned_free on Windows (provided by MSVC runtime)
+    if (ptr) {
+        _aligned_free(ptr);
+    }
+}
+#endif  // _WIN32
```
IzioDev: I don't see this symbol being used within risc0

michaelsutton (approved):
LGTM. Tested on windows and r0 opcode verify both builds and passes verification

PR #882, opened 2026-02-23
IBD: Batch UTXO set writes

### LINKED ISSUE #768: IBD: Batch UTXO set writes

RossKU:
Replace `DirectDbWriter` with `WriteBatch`/`BatchDbWriter` in IBD and pruning UTXO write paths (`write_many`, `write_from_iterator_without_cache`, `write_diff`). Also adds `write_many_without_cache` for the IBD import path where cache population is unnecessary.

Closes #768

### DIFF
--- a/consensus/src/consensus/mod.rs
+++ b/consensus/src/consensus/mod.rs
@@ -26,7 +26,7 @@ use crate::{
             selected_chain::SelectedChainStore,
             statuses::StatusesStoreReader,
             tips::{TipsStore, TipsStoreReader},
-            utxo_set::{UtxoSetStore, UtxoSetStoreReader},
+            utxo_set::UtxoSetStoreReader,
             virtual_state::VirtualState,
         },
     },


freshair18:
Did you run practical tests regarding the effect of this? can you share the results?

freshair18:
also how does this compare with https://github.com/kaspanet/rusty-kaspa/pull/781/changes ?

@LiveLaughLove13 @coderofstuff

freshair18:
I think https://github.com/kaspanet/rusty-kaspa/pull/850 also revolves around the same aspects. 
Probably a good time to review and incorporate these PRs.

RossKU:
I interpreted "the effect of this" as the performance difference between `DirectDbWriter` (one `db.put()` per
  entry) and `BatchDbWriter` (all writes collected into a `WriteBatch`, then flushed in a single `db.write()`
  call), so I ran a criterion benchmark comparing the two approaches with chunk sizes representative of the IBD
  UTXO import path.

  **Results** (Windows 11, release build):

  | Chunk size | DirectDbWriter (old) | BatchDbWriter (new) | Speedup |
  |---|---|---|---|
  | 1,000 UTXOs | 11.64 ms | 7.23 ms | ~1.6x |
  | 10,000 UTXOs | 56.75 ms | 14.20 ms | ~4.0x |

PR #885, opened 2026-02-25
Add API for covenant genesis ID computation & output population

smartgoo:
## Summary:

This PR adds a standard process and helper function for covenant ID computation and population on genesis outputs:

- A standard process to compute covenant ID and populate on genesis outputs. Available to Rust and WASM SDK. 
- A WASM SDK-exposed wrapper function around `kaspa-consensus-core`'s [`covenant_id()` function](https://github.com/kaspanet/rusty-kaspa/blob/5e2fd7bd14bb9783e9ead229e89eddf9df35834c/consensus/core/src/hashing/covenant_id.rs#L16). 

Combined, these streamline the covenant genesis process, reducing lift for developers.

**Note**: this PR originally started as WASM SDK `covenant_id()` function only. Scope changed to include functionality that further streamline the process.

## Why:

This codebase should provide a streamlined process and required building blocks for the covenant genesis process. Given the need to 1.) compute covenant ID for genesis; and .2) populate the covenant ID on genesis outputs. 

This added functionality simplifies the process for Rust & WASM SDK developers.

## Notable Callouts
- All WASM related code in this PR follows existing WASM patters as closely as possible.
- WASM bindings exposing `GenesisCovenantGroup` was added using wrapper approach `consensus-client`, to avoid adding more to consensus core. There seemed to be some precedent for both this approach and exposing consensus core struct directly.
- `populate_genesis_covenants()` on consensus client `Transaction` probably needs a good look to ensure this is done correctly. Took this approach as Transactions's `finalize()` fn does something similar.

### DIFF
--- a/consensus/client/src/covenant.rs
+++ b/consensus/client/src/covenant.rs
@@ -6,9 +6,96 @@
 
 use crate::imports::*;
 use crate::result::Result as ClientResult;
+use borsh::{BorshDeserialize, BorshSerialize};
 use kaspa_consensus_core::errors::tx::PopulateGenesisCovenantsError;
-use kaspa_consensus_core::tx::GenesisCovenantGroup as CoreGenesisCovenantGroup;
+use kaspa_consensus_core::tx::{CovenantBinding as CoreCovenantBinding, GenesisCovenantGroup as CoreGenesisCovenantGroup};
+use kaspa_hashes::Hash;
 use kaspa_wasm_core::types::NumberArray;
+use workflow_wasm::error::Error as WasmError;
+
+#[wasm_bindgen(typescript_custom_section)]
+const TS_COVENANT_BINDING: &'static str = r#"
+/**
+ * A covenant binding binds a transaction output to the covenant and input authorizing its creation.
+ *
+ * @category Consensus
+ */
+export interface ICovenantBinding {
+    authorizingInput: number;
+    covenantId: HexString;
+}
+"#;
+
+#[derive(Debug, Serialize, Deserialize, Clone, PartialEq, Eq, Copy, BorshSerialize, BorshDeserialize, CastFromJs)]
+#[serde(rename_all = "camelCase")]
+#[wasm_bindgen(inspectable)]
+pub struct CovenantBinding {
+    inner: CoreCovenantBinding,
+}
+
+#[wasm_bindgen]
+impl CovenantBinding {
+    #[wasm_bindgen(constructor)]
+    pub fn new(authorizing_input: u16, covenant_id: Hash) -> Self {
+        Self { inner: CoreCovenantBinding::new(authorizing_input, covenant_id) }
+    }
+
+    #[wasm_bindgen(setter, js_name = authorizingInput)]
+    pub fn set_authorizing_input(&mut self, v: u16) {
+        self.inner.authorizing_input = v;
+    }
+
+    #[wasm_bindgen(getter, js_name = authorizingInput)]
+    pub fn get_authorizing_input(&self) -> u16 {
+        self.inner.authorizing_input
+    }
+
+    #[wasm_bindgen(setter, js_name = covenantId)]
+    pub fn set_covenant_id(&mut self, v: Hash) {
+        self.inner.covenant_id = v;
+    }
+
+    #[wasm_bindgen(getter, js_name = covenantId)]
+    pub fn get_covenant_id(&self) -> Hash {
+        self.inner.covenant_id
+    }
+
+    #[wasm_bindgen(js_name = toJSON)]
+    pub fn to_js_object(&self) -> Result<Object, WasmError> {
+        let obj = Object::new();
+        obj.set("authorizingInput", &self.get_authorizing_input().into())?;
+        obj.set("covenantId", &self.get_covenant_id().to_string().into())?;
+        Ok(obj)
+    }
+}
+
+impl From<CoreCovenantBinding> for CovenantBinding {
+    fn from(value: CoreCovenantBinding) -> Self {
+        Self { inner: value }
+    }
+}
+
+impl From<CovenantBinding> for CoreCovenantBinding {
+    fn from(value: CovenantBinding) -> Self {
+        value.inner
+    }
+}
+
+impl TryCastFromJs for CovenantBinding {
+    type Error = WasmError;
+
+    fn try_cast_from<'a, R>(value: &'a R) -> Result<Cast<'a, Self>, Self::Error>
+    where
+        R: AsRef<JsValue> + 'a,
+    {
+        Self::resolve(value, || {
+            let Some(object) = Object::try_from(value.as_ref()) else { return Err(Self::Error::NotAnObject) };
+            let authorizing_input = object.get_u16("authorizingInput")?;
+            let covenant_id = object.get_value("covenantId")?.try_into_owned()?;
+            Ok(Self { inner: CoreCovenantBinding::new(authorizing_input, covenant_id) })
+        })
+    }
+}
 
 #[wasm_bindgen(typescript_custom_section)]
 const TS_GENESIS_COVENANT_GROUP: &'static str = r#"
@@ -106,7 +193,7 @@ impl GenesisCovenantGroup {
     }
 
     #[wasm_bindgen(js_name = "toJSON")]
-    pub fn to_js_object(&self) -> Result<Object, workflow_wasm::error::Error> {
+    pub fn to_js_object(&self) -> Result<Object, WasmError> {
         let obj = Object::new();
         obj.set("authorizingInput", &self.inner.authorizing_input.into())?;
         obj.set("outputs", &js_sys::Array::from_iter(self.inner.outputs.iter().map(|&v| JsValue::from(v))))?;
@@ -114,13 +201,13 @@ impl GenesisCovenantGroup {
     }
 
     #[wasm_bindgen(js_name = "toString")]
-    pub fn js_to_string(&self) -> Result<js_sys::JsString, workflow_wasm::error::Error> {
+    pub fn js_to_string(&self) -> Result<js_sys::JsString, WasmError> {
         Ok(js_sys::JSON::stringify(&self.to_js_object()?.into())?)
     }
 }
 
 impl TryCastFromJs for GenesisCovenantGroup {
-    type Error = workflow_wasm::error::Error;
+    type Error = WasmError;
 
     fn try_cast_from<'a, R>(value: &'a R) -> Result<Cast<'a, Self>, Self::Error>
     where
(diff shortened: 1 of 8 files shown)


consensus/wasm/src/hashing.rs
```diff
@@ -0,0 +1,35 @@
+use crate::error::Error;
```
IzioDev: I think it's fair to move the content to here: `/consensus/client/src/hash.rs` considering what it has at the moment

What do you think?

consensus/wasm/src/hashing.rs
```diff
@@ -0,0 +1,35 @@
...
+/// `genesis_outpoint` may be a [`TransactionOutpoint`] instance or a
+/// compatible plain object: `{ transactionId: HexString, index: number }`.
+///
+/// `auth_outputs` is a JS array of objects, each with:
+/// - `index: number` — position of this output in the transaction's output array
+/// - `output: TransactionOutput | ITransactionOutput` — the authorized output
+///
+/// @category Consensus
+#[wasm_bindgen(js_name = covenantId)]
+pub fn covenant_id_js(genesis_outpoint: &TransactionOutpointT, auth_outputs: JsValue) -> Result<Hash> {
```
IzioDev: Why not using `Vec<TransactionOutput>` as the second argument here? I don't remember the exact behavior with `Vec` Js value parsing, but TransactionOutput does have a dedicated deserializer.

consensus/wasm/src/lib.rs
```diff
@@ -4,8 +4,10 @@ pub mod error;
 
 cfg_if! {
     if #[cfg(feature = "wasm32-sdk")] {
+        pub mod hashing;
         pub mod result;
         mod utils;
+        pub use hashing::*;
```
IzioDev: Unrelated, but I'm sharing here in case :)
This has for effect to "spread" / "share" all of the public hashing members to the current crate, so within `consensus/wasm` you can do `use crate::covenant_id_js;` instead of `use crate::hashing::covenant_id_js;`

I'm not sure this was intended

consensus/wasm/src/hashing.rs
IzioDev: I'm not sure what's the general recommendation to enforce in term of naming, but many (if not all?) members of `wallet/core/wasm` define `js` (/wasm) functions as `js_my_function` (as a prefix)

consensus/client/src/outpoint.rs
```diff
@@ -22,6 +22,14 @@ export interface ITransactionOutpoint {
 }
 "#;
 
+#[wasm_bindgen]
+extern "C" {
+    /// WASM (TypeScript) type representing `ITransactionOutpoint | TransactionOutpoint`
+    /// @category Consensus
```
IzioDev, line 28: I'm not sure it's relevant to precise it's a wasm typescript type, end-users most likely don't care

smartgoo:
@IzioDev TY for feedback (& the quick response)! Will get back to you on these in the next day or so.

consensus/client/src/outpoint.rs
smartgoo: @IzioDev Admittedly I just copy/pasted this comment from other extern WASM blocks and changed out the actual type. This comment structure seems to be standard throughout WASM consensus client extern blocks - [transaction input](https://github.com/kaspanet/rusty-kaspa/blob/c01841ce353a92445cda679a59546160e8627c70/consensus/client/src/input.rs#L42), [transaction output](https://github.com/kaspanet/rusty-kaspa/blob/c01841ce353a92445cda679a59546160e8627c70/consensus/client/src/output.rs#L37), [transaction](https://github.com/kaspanet/rusty-kaspa/blob/c01841ce353a92445cda679a59546160e8627c70/consensus/client/src/transaction.rs#L60) for example use this same comment structure.

That said, happy to change it if you want!

consensus/wasm/src/hashing.rs
smartgoo: @IzioDev Looks like a mixed bag of `js_` prefix and `_js` suffix throughout the codebase.

A quick (very incomplete) list to give a feel for it:

`js_` prefix is used on functions [`js_sign_message`](https://github.com/kaspanet/rusty-kaspa/blob/c01841ce353a92445cda679a59546160e8627c70/wallet/core/src/wasm/message.rs#L30), [`js_sign_verify_message`](https://github.com/kaspanet/rusty-kaspa/blob/c01841ce353a92445cda679a59546160e8627c70/wallet/core/src/wasm/message.rs#L70), [`js_sign_transaction`](https://github.com/kaspanet/rusty-kaspa/blob/c01841ce353a92445cda679a59546160e8627c70/wallet/core/src/wasm/signer.rs#L37), everything in [encryption.rs](https://github.com/kaspanet/rusty-kaspa/blob/master/wallet/core/src/wasm/encryption.rs).

`_js` suffix is used on functions [`create_address_js`](https://github.com/kaspanet/rusty-kaspa/blob/c01841ce353a92445cda679a59546160e8627c70/wallet/core/src/derivation.rs#L460), [`create_multisig_address_js`](https://github.com/kaspanet/rusty-kaspa/blob/c01841ce353a92445cda679a59546160e8627c70/wallet/core/src/derivation.rs#L478), [`create_transaction_js`](https://github.com/kaspanet/rusty-kaspa/blob/c01841ce353a92445cda679a59546160e8627c70/wallet/core/src/wasm/tx/utils.rs#L14), [`create_transactions_js`](https://github.com/kaspanet/rusty-kaspa/blob/c01841ce353a92445cda679a59546160e8627c70/wallet/core/src/wasm/tx/utils.rs#L86), [`estimate_transactions_js`](https://github.com/kaspanet/rusty-kaspa/blob/c01841ce353a92445cda679a59546160e8627c70/wallet/core/src/wasm/tx/utils.rs#L118), etc.

By quantity, it looks like a slight skew towards `js_` prefix. So I have updated to use js prefix in [a29c544](https://github.com/kaspanet/rusty-kaspa/pull/885/commits/a29c54438585ea76246c88dd1f3e45a050410934). Thanks!

smartgoo: @IzioDev great question, didn't think of trying to use `Vec<_>` here.

So each object in `auth_outputs` array needs to include an `index` in addition to the TransactionOutput instance/opbject. e.g.

```js
[
    { index: 0, output: { value: 10000, scriptPublicKey: spk } },
    ...
]
```

For that reason param type would need to be something like `Vec<(u32, TransactionOutput)>`.

But with that param type, compile error:
```
trait bound `(u32, kaspa_consensus_client::TransactionOutpoint): JsObject` is not satisfied
```

So my current thought is - not worth it to impl that for this use case. However, curious what you think. Also if you have any other thoughts on alternative approaches?

consensus/wasm/src/lib.rs
smartgoo: Ah yeah good call out. Used that approach since it was already used for `pub use utils::*;` a couple lines down. I intend to move it to other file per your other comment. So this will go away!

consensus/wasm/src/hashing.rs
smartgoo: @IzioDev makes sense - this is done, moved to `/consensus/client/src/hash.rs` in [63aea0a](https://github.com/kaspanet/rusty-kaspa/pull/885/commits/63aea0ae0f1414014f1b92ba30b007cde7e93fda).

consensus/wasm/src/hashing.rs
smartgoo: @IzioDev Actually, even `Vec<(u32, TransactionOutput)>` is wrong.

I think it would be `Vec<JsValue>`, where JsValue looks like or would require some new struct e.g. `AuthOutputs` to use as `Vec<AuthOutputs>`. Where `AuthOutputs` probably looks like this:

```rust
struct AuthOutputs {
    index: u32,
    output: TransactionOutput
}
```

smartgoo: @IzioDev Ok I changed it to `Vec<JsValue>` in [2495111](https://github.com/kaspanet/rusty-kaspa/pull/885/commits/2495111e435dc3ebed40faa0b16c60084f2ff91c).

I feel this is a step in the right direction - it simplifies internals slightly and makes fn signature a little more explicit. Taking it a step further would be a custom struct a la the comment above.

However, I'm not sure how `Vec<JsValue>` plays with TypeScript. There are some ArrayT patterns used elsewhere for parameter types in WASM functions. Like [`PrivateKeyArrayT`](https://github.com/smartgoo/rusty-kaspa/blob/c01841ce3/wallet/core/src/wasm/signer.rs#L14-L19), [`PublicKeyArrayT`]( https://github.com/smartgoo/rusty-kaspa/blob/c01841ce3/wallet/keys/src/publickey.rs#L150-L154), [`TransactionInputArrayAsArgT`](https://github.com/smartgoo/rusty-kaspa/blob/c01841ce3/consensus/client/src/input.rs#L47-L50), [`TransactionOutputArrayAsArgT`](https://github.com/smartgoo/rusty-kaspa/blob/c01841ce3/consensus/client/src/output.rs#L42-L45). Can explore this is you think it makes sense.

Sorry for yet another comment here.

smartgoo:
Updated title & description to reflect expanded scope. Converted to draft status as this is WIP. Only a few remaining TODOs.

consensus/client/src/covenant.rs
```diff
@@ -0,0 +1,127 @@
...
+    ///
+    /// @category Consensus
+    #[wasm_bindgen(extends = js_sys::Array, typescript_type = "GenesisCovenantGroup[]")]
+    pub type GenesisCovenantGroupArrayT;
+}
+
+impl TryFrom<&GenesisCovenantGroupArrayT> for Vec<CoreGenesisCovenantGroup> {
+    type Error = PopulateGenesisCovenantsError;
+    fn try_from(value: &GenesisCovenantGroupArrayT) -> Result<Self, Self::Error> {
+        if value.is_array() {
```
IzioDev, line 125: Since we extend "js_sys::Array", I don't think this is necessary.
If we think it is, we should apply this logic elsewhere too (see js_covenant_id)

IzioDev: Can be ignored, it works as is, it may be doing more than needed, but that's fine.
Other places have similar handling, let's keep as is until we find the desired direction to take

IzioDev (approved):

consensus/core/src/tx.rs
```diff
@@ -179,6 +180,51 @@ impl CovenantBinding {
     pub fn new(authorizing_input: u16, covenant_id: Hash) -> Self {
         Self { authorizing_input, covenant_id }
     }
+
+    #[wasm_bindgen(setter, js_name = authorizingInput)]
```
michaelsutton: it seems like for most other tx-related structs we have such `wasm_bindgen` defined elsewhere on a wrapper object. eg https://github.com/smartgoo/rusty-kaspa/blob/01191f8715127d2fbc812e6755536bd750286a43/consensus/client/src/output.rs#L88
can/should we do the same for `CovenantBinding` ?

smartgoo: Makes sense.

@IzioDev recommended same thing with `GenesisCovenantGroup`. I had originally added `wasm_bindgen` to `GenesisCovenantGroup` in consensus core before moving to a wrapper in consensus client at his recommendation.

The new [covenant module in consensus client](https://github.com/smartgoo/rusty-kaspa/blob/01191f8715127d2fbc812e6755536bd750286a43/consensus/client/src/covenant.rs)  feels like a logical home for a `CovenantBinding` wrapper. Will add wrapper with wasm glue there and add wasm tests.

smartgoo: `CovenantBinding` WASM bindings have been moved to `consensus/client/src/covenant.rs` in [b5826b3](https://github.com/kaspanet/rusty-kaspa/pull/885/commits/b5826b38ce5e059b5bc4accd7f60d0b7da97ea1e) @michaelsutton @IzioDev

consensus/client/src/covenant.rs
```diff
@@ -6,9 +6,96 @@
...
+    }
+
+    #[wasm_bindgen(js_name = toJSON)]
+    pub fn to_js_object(&self) -> Result<Object, WasmError> {
+        let obj = Object::new();
+        obj.set("authorizingInput", &self.get_authorizing_input().into())?;
+        obj.set("covenantId", &self.get_covenant_id().to_string().into())?;
+        Ok(obj)
+    }
+}
```
IzioDev, line 70: Is the goal of this block to "bypass" `inner` property, because in theory `inspectable` exposes this API, right?

More generally, is this something we envision as necessary (to bypass `inner`), if so it raises the same question to other similar wasm wrappers (out of scope, just a thought sharing)

smartgoo: @IzioDev I think I know what you're getting at here. Here are some notes.

With this block (`to_js_object` fn, all getter/setter fns)... `console.log(covenantBindingInstance)` yields:
```json
{
  authorizingInput: 0,
  covenantId: '4d0f8fd4428f043de5dec28b21487286c8fde96414448811f90934c4146332eb'
}
```

Without `to_js_object` (no other changes)... `console.log(covenantBindingInstance)` yields:
```json
{ covenantId: Hash { __wbg_ptr: 4063368 }, authorizingInput: 0 }
```

Removing all getters/setters, removing `to_js_object` fn, and leaving the constructor fn in place... `console.log(covenantBindingInstance)` yields the following. This is because field `inner` is private.
```
{}
```

Continuing the same scenario, you are also unable to access the fields from js like so:
```js
console.log(covenantBindingInstance.authorizingInput`);
// outputs: undefined

console.log(covenantBindingInstance.covenantId`);
// outputs: undefined
```

Removing all getters/setters & `to_js_objet`, and changing field `inner` to pub will not compile because field `inner` is core `CovenantBinding` which is not decorated with `wasm_bindgen` attribute.

I hope this is going down the right path!

IzioDev: Super clear, thanks for the detailed breakdown.

michaelsutton (approved):

PR #888, opened 2026-02-26
feat(p2p): implement --disable-relay-tx flag

### LINKED ISSUE #228: Implement disable_relay_tx
Yes - a user that only wants to use the node as an API does not need to receive tx messages. Not having to deal with txs will reduce mempool use.

Implement the `disable_relay_tx` flag. When enabled, this lets a node's peers that it doesn't want to receive transactions. The peers will not send txs to this node.

N/A

**Additional context**

One consideration is the issue of starvation: What if a node that wants to receive txs is only connected to peers that have `DisableRelayTx=false`?

**Notes**
- Rust code: in flow_context.rs the line `// TODO: disable_relay_tx from config/cmd` is where this would go
- Go code: this was `BlockOnly` in config, and `DisableRelayTx` in the rest of the codebase.

RossKU:
## Summary
Closes #228.

Implements the `--disable-relay-tx` CLI flag, allowing API-only nodes to opt out of transaction relay, reducing mempool overhead and bandwidth.

### What it does
- Adds `--disable-relay-tx` CLI flag (also configurable via `KASPAD_DISABLE_RELAY_TX` env var or config file)
- Sets `disableRelayTx` in the version handshake message so peers know not to send us transactions
- Filters outbound tx broadcasts to skip peers that have set `disable_relay_tx` in their version handshake
- Drains incoming tx inv messages without acting on them when our node has relay disabled

### Changes by file
- **`kaspad/src/args.rs`** — CLI arg, Args struct field, parsing, apply_to_config
- **`consensus/core/src/config/mod.rs`** — Config struct field + default
- **`protocol/flows/src/flow_context.rs`** — Populate `disable_relay_tx` in version message (resolves the TODO at line 716)
- **`protocol/flows/src/flowcontext/transactions.rs`** — Use `broadcast_filtered` / `broadcast_to_some_peers_filtered` to skip `disable_relay_tx` peers
- **`protocol/p2p/src/core/hub.rs`** — Add `broadcast_filtered()` and `broadcast_to_some_peers_filtered()` methods
- **`protocol/flows/src/v7/txrelay/flow.rs`** — Drain inbound tx inv when relay is disabled (v8 reuses v7 txrelay, so both versions are covered)

### Design decisions
- **Both directions**: We don't broadcast tx inv to peers who set the flag, AND we don't request txs if we set the flag ourselves
- **Drain, don't unregister**: The relay flow is still registered (to avoid breaking protocol expectations) but drains inv messages without requesting transactions
- **Mining safe**: Mining uses RPC + mempool, not P2P relay. Nodes with `--disable-relay-tx` can still accept transactions via RPC

### Out of scope: starvation mitigation
As noted by @coderofstuff in #228, if a node wants txs but all its peers have `disable_relay_tx=true`, no transactions arrive. Mitigation (peer eviction + reconnection logic) is deferred to a follow-up issue — in practice, `disable_relay_tx` nodes will be a small minority (API-only), so this scenario is unlikely.

## Test plan
- [x] `cargo check --workspace` passes
- [x] `cargo test --workspace` passes (all tests)
- [ ] Manual: start kaspad with `--disable-relay-tx`, verify no tx inv messages are processed
- [ ] Manual: connect two nodes, one with flag set, verify txs are not relayed to it

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/protocol/flows/src/v7/txrelay/flow.rs
+++ b/protocol/flows/src/v7/txrelay/flow.rs
@@ -90,11 +90,14 @@ impl RelayTransactionsFlow {
     }
 
     async fn start_impl(&mut self) -> Result<(), ProtocolError> {
-        // If our node has disabled transaction relay, drain inv messages without acting on them
+        // If our node has disabled transaction relay, reject inv messages from peers.
+        // The disable_relay_tx flag is communicated during the handshake, so well-behaved
+        // peers should never send us tx invs. If they do, it's a protocol violation.
         if self.ctx.config.disable_relay_tx {
-            loop {
-                let _ = dequeue!(self.invs_route, Payload::InvTransactions)?;
-            }
+            let _ = dequeue!(self.invs_route, Payload::InvTransactions)?;
+            return Err(ProtocolError::Other(
+                "peer sent tx inv despite our disable_relay_tx flag in the handshake",
+            ));
         }
 
         // trace!("Starting relay transactions flow with {}", self.router.identity());


coderofstuff:
wdyt about the possible starvation issue @someone235 @michaelsutton?

D-Stacks:
`Drains incoming tx inv messages without acting on them when our node has relay disabled`

wouldn't it be better to communicate the flag with peers during the handshake, and actually disable the bandwidth? also issue a protocol error if a peer ignores this (might depend on bumping the p2p protocol version as well).

RossKU:
Thanks! updated to return a ProtocolError on first inv instead of silently draining.

PR #889, opened 2026-02-27
Feat: Kaspa-Math 32 bit malachite support

hmoog:
This PR fixes a compilation failure when kaspa-math is used in a workspace alongside risc0-based crates.

**Problem**

kaspa-math's mod_inverse passes its internal [u64; N] limb array directly to malachite's Natural::from_limbs_asc(&[Limb]), assuming Limb = u64. This breaks when malachite's 32_bit_limbs feature is enabled, causing a type
 mismatch (expected &[u32], found &[u64; N]).

**Root cause: Cargo feature unification**

malachite-nz has an optional 32_bit_limbs feature that switches Limb from u64 to u32. By default it is disabled, so Limb = u64 and kaspa-math compiles fine in the rusty-kaspa workspace.

However, Cargo unifies features across an entire workspace. In the vprogs workspace, risc0-circuit-rv32im (RISC Zero zkVM) depends on malachite with:

`features = ["naturals_and_integers", "32_bit_limbs"]`

This activates 32_bit_limbs globally - including for kaspa-math - making Limb = u32 and breaking mod_inverse.

**Fix**

Make mod_inverse generic over the Limb type by converting between the internal u64 words and malachite Limbs at the boundary:

  - u64 → Limb (input): When Limb is 32-bit, each u64 word is split into two u32 limbs (low, high) in little-endian order. When Limb is 64-bit, the words pass through directly.
  - Limb → u64 (output): The inverse operation - pairs of u32 limbs are reassembled into u64 words.

The size_of::<Limb>() check is a compile-time constant, so the compiler eliminates the unused branch entirely - zero runtime cost.

### DIFF
--- a/math/src/uint.rs
+++ b/math/src/uint.rs
@@ -349,10 +349,14 @@ macro_rules! construct_uint {
                 use $crate::uint::malachite_nz::platform::Limb;
                 use $crate::uint::malachite_base::num::arithmetic::traits::ModInverse;
 
+                // Malachite's Limb is u64 on 64-bit platforms but u32 on 32-bit
+                // (e.g. risc0 riscv32im). Convert our u64 words to limbs, splitting
+                // each word into two 32-bit halves when necessary.
                 let to_limbs = |words: &[u64]| -> Vec<Limb> {
                     if core::mem::size_of::<Limb>() == 8 {
                         words.iter().map(|&w| w as Limb).collect()
                     } else {
+                        // Split each u64 into [low_u32, high_u32] (little-endian limb order).
                         words.iter().flat_map(|&w| {
                             [(w & 0xFFFF_FFFF) as Limb, ((w >> 32) & 0xFFFF_FFFF) as Limb]
                         }).collect()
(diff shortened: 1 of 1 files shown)


math/src/uint.rs
```diff
@@ -346,16 +346,45 @@ macro_rules! construct_uint {
...
                 use $crate::uint::malachite_base::num::arithmetic::traits::ModInverse;
 
-                let x = Natural::from_limbs_asc(&self.0);
-                let p = Natural::from_limbs_asc(&prime.0);
-                let mod_inv = x.mod_inverse(p);
+                // Malachite's Limb type changes from u64 to u32 when the `32_bit_limbs`
+                // feature is enabled (e.g. by risc0 via Cargo feature unification).
+                // On little-endian, [u64; N] and [Limb; N * 8/sizeof(Limb)] have
+                // identical byte layouts, so we reinterpret directly without copying.
+                #[cfg(not(target_endian = "little"))]
```
michaelsutton, line 356: we can allow `not(target_endian = "little")` if `core::mem::size_of::<u64>() == core::mem::size_of::<Limb>()`

biryukovmaxim:
https://github.com/biryukovmaxim/rusty-kaspa/blob/e77acc2afb1950ffdffb568e467cd4f6005761d3/Cargo.toml#L222-L223
I had the same issue and solved it by enforcing different major version of malachite. I still consider it as a workaround but at least it doesn't force 32 bit on host/kaspad which I would see wrong

math/src/uint.rs
```diff
@@ -346,16 +346,45 @@ macro_rules! construct_uint {
...
+                compile_error!("mod_inverse assumes little-endian byte order");
```
elichai, line 357: We could reverse the words if we cared about big-endian

math/src/uint.rs
```diff
@@ -346,16 +346,45 @@ macro_rules! construct_uint {
...
+
+                const LIMBS_PER_WORD: usize =
+                    core::mem::size_of::<u64>() / core::mem::size_of::<Limb>();
```
elichai, line 360: Maybe add `const { assert!(core::mem::size_of::<Limb>() <= core::mem::size_of::<u64>()) }`

elichai (approved):
Looks good and correct

Issue #256, opened 2023-08-26
Implement IndexDb-backed transaction storage in the wallet core


aspect:
Wallet core transaction storage is backed by a regular File I/O in the Rust native and NodeJs platforms.

The current web wallet transaction storage remains a placeholder (transaction history is not currently stored when the wallet operates in the web browser).

We need to implement an `IndexDb` storage interface for transaction storage.

### References:
- File I/O storage (for reference):
https://github.com/kaspanet/rusty-kaspa/blob/master/wallet/core/src/storage/local/transaction/fsio.rs

- Index Db storage (scaffolding):
https://github.com/kaspanet/rusty-kaspa/blob/master/wallet/core/src/storage/local/transaction/indexdb.rs

### Notes:
- Transaction records are stored within multiple named stores (databases) simultaneously.  Each wallet account should have a separate database store.
- Review and consider using https://crates.io/crates/idb as an async interface layer (as opposed to `web_sys`)

PR #892, opened 2026-03-05
feat(wallet): CSV export of transaction history

### LINKED ISSUE #267: Implement CSV export of the wallet account transaction history
We need to be able to export transactions in a CSV format. 

This functionality would be a replica of the history command:

https://github.com/kaspanet/rusty-kaspa/blob/master/cli/src/modules/history.rs

where the `TransactionRecord` needs to be taught to produce a CSV record (line), similarly to current format output:

https://github.com/kaspanet/rusty-kaspa/blob/master/wallet/core/src/storage/transaction.rs#L318

The accumulated data then needs to be collected and depending on the environment either made to be downloadable (KOS/web) via the download data handler:

https://github.com/workflow-rs/workflow-rs/blob/master/dom/src/download.rs

or (CLI) stored as a CSV file:

https://github.com/workflow-rs/workflow-rs/blob/master/store/src/fs.rs

The filename should be supplied via the cli command arguments.

This issue is dependent on #256 to be functional in the web environment.

The actual content generation should be made available as a part of the wallet Account API

https://github.com/kaspanet/rusty-kaspa/blob/master/wallet/core/src/runtime/account/mod.rs

So that it can be exposed to WASM SDK (which is currently WIP):

https://github.com/kaspanet/rusty-kaspa/blob/master/wallet/core/src/wasm/wallet/account.rs

RossKU:
Closes #267

Add CSV export capability to `TransactionRecord` and a `history export` CLI subcommand.

## Changes

**wallet/core** (`record.rs`):
- `TransactionRecord::csv_header()` — static 9-column header
- `TransactionRecord::to_csv_row()` — formats one record as a CSV line
- `unixtime_msec_to_iso8601()` — pure-arithmetic UTC formatter (Howard Hinnant algorithm, no chrono dep)
- `csv_escape()` — RFC 4180 field quoting
- WASM bindings: `toCsvRow()` / `csvHeader()` for JS/WASM SDK consumers
- Unit tests for ISO 8601 conversion, CSV escaping, and column count

**cli** (`history.rs`):
- `history export [<filename>]` subcommand
- Default filename: `kaspa-tx-{account}.csv`
- Streams all records via `transaction_data_iter`, writes CSV to file

## CSV Format

```csv
id,timestamp,type,value_sompi,fees_sompi,block_daa_score,is_coinbase,note,metadata
```

- Timestamps: ISO 8601 with milliseconds (`2024-01-15T12:30:45.123Z`)
- Values: sompi integers (no KAS conversion)
- 1 transaction = 1 row

## Not in this PR

- Browser/KOS file download (depends on #256)
- No new crate dependencies

---

@aspect Could you review the CSV column layout? Current columns: `id, timestamp, type, value_sompi, fees_sompi, block_daa_score, is_coinbase, note, metadata`. Let me know if you'd like any columns added, removed, or renamed. Thanks!

### DIFF
--- a/cli/src/modules/history.rs
+++ b/cli/src/modules/history.rs
@@ -64,6 +64,52 @@ impl History {
                 let last = if argv.is_empty() { None } else { argv[0].parse::<usize>().ok() };
                 (last, true)
             }
+            "export" => {
+                let filename = if argv.is_empty() { format!("kaspa-tx-{}.csv", account.name_or_id()) } else { argv.remove(0) };
+
+                let store = ctx.wallet().store().as_transaction_record_store()?;
+                let mut records = match store.transaction_data_iter(&binding, &network_id).await {
+                    Ok(r) => r,
+                    Err(err) => {
+                        if matches!(err, WalletError::NoRecordsFound) {
+                            tprintln!(ctx);
+                            tprintln!(ctx, "No transactions found for this account.");
+                            tprintln!(ctx);
+                        } else {
+                            terrorln!(ctx, "{err}");
+                        }
+                        return Ok(());
+                    }
+                };
+
+                let mut csv = String::from(kaspa_wallet_core::storage::TransactionRecord::csv_header());
+                csv.push('\n');
+
+                let mut count = 0u64;
+                while let Some(record) = records.try_next().await? {
+                    csv.push_str(&record.to_csv_row());
+                    csv.push('\n');
+                    count += 1;
+                }
+
+                #[cfg(not(target_arch = "wasm32"))]
+                {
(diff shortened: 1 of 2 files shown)


IzioDev:
Hello :)
Could you please check #855 that seems to achieve the same goal? (and probably not the same way considering number of file touched)

RossKU:
Thanks for pointing that out! I checked #855 and it fully covers the same functionality (and more). Happy to close this one in favor of it.

That said, this PR takes a more minimal approach — only 2 files changed — so it may be easier to review. Let me know if that's useful.

Issue #332, opened 2023-11-22
`UtxoEntry` from RPC and wasm class has different field name


Cryptok777:
**Describe the bug**
From the RPC call `getUtxosByAddresses` it's returning `{utxoEntry: {...}}`, but in WASM class `UtxoEntry` has the field  `entry` instead of `utxoEntry`, [code](https://github.com/kaspanet/rusty-kaspa/blob/f674ad5e071f79d294c6964ee114e8feea9d29ff/rpc/core/src/wasm/mod.rs#L8) - would be good to keep it consistent and so that the interface can be reused

**To Reproduce**
```
const resp await this.rpc.getUtxosByAddresses({ addresses: [address] })
resp.utxoEntry // This is valid and returned by RPC

// In kaspa_wasm.d.ts, the UtxoEntry class is 

export class UtxoEntry {
/**
** Return copy of self without private attributes.
*/
  toJSON(): Object;
/**
* Return stringified version of self.
*/
  toString(): string;
  free(): void;
/**
*/
  address?: Address;
/**
*/
  entry: TxUtxoEntry;
/**
*/
  outpoint: TransactionOutpoint;
}
`

**Expected behavior**
Would be ideal to keep these consistent so we can reuse the interface

**Comment by IzioDev** [comment]
RPC schema and SDK schema can be different, and in this case I tend to prefer the SDK API:
```ts
// in place
const utxoEntry = utxo.entry;

// the proposal
const utxoEntry = utxo.utxoEntry;
```

It feels redundant to have `utxo` prefix qualifier here.

PR #893, opened 2026-03-05
feat(wasm): add utxoEntry getter alias on UtxoEntryReference

RossKU:
Addresses #332

The RPC `getUtxosByAddresses` response serializes the field as `utxoEntry` (camelCase), but the WASM `UtxoEntryReference` class only exposes `.entry`. This adds a `.utxoEntry` getter alias so the WASM SDK property name matches the RPC JSON field name.

## Change

- Add `utxoEntry` getter on `UtxoEntryReference` (returns same `UtxoEntry` as `.entry`)
- **Non-breaking**: existing `.entry` getter is preserved

## File

`consensus/client/src/utxo.rs` — 5 lines added

### DIFF
--- a/consensus/client/src/utxo.rs
+++ b/consensus/client/src/utxo.rs
@@ -161,6 +161,11 @@ impl UtxoEntryReference {
         self.as_ref().clone()
     }
 
+    #[wasm_bindgen(getter, js_name = "utxoEntry")]
+    pub fn utxo_entry(&self) -> UtxoEntry {
+        self.as_ref().clone()
+    }
+
     #[wasm_bindgen(getter)]
     pub fn outpoint(&self) -> TransactionOutpoint {
         self.utxo.outpoint.clone()


IzioDev:
RPC schema and SDK schema can be different, and in this case I tend to prefer the SDK API:
```ts
// in place
const utxoEntry = response.utxo.entry;

// your proposal
const utxoEntry = response.utxo.utxoEntry;
```

It feels redundant to have `utxo` prefix qualifier here, what do you think?

RossKU:
That's a fair point — `utxo.utxoEntry` is indeed redundant. I just noticed the issue was still open, so I assumed there was still demand for aligning the naming. Happy to close this if the current SDK API is preferred.

IzioDev:
Well, personally I don't see any issues here unless I miss something.

I'll check initial issue, thank you for raising it again 👍

RossKU:
IzioDev's feedback makes sense — `utxo.utxoEntry` is redundant. Closing this.

PR #894, opened 2026-03-05
fix(wallet/generator): use consensus storage mass calculatation on with_change

### LINKED ISSUE #701: SDK - Generator - Mass Exceed on a single utxo set evaluation round should not exit if there are other utxos that could fit
**Describe the bug**
Under certain conditions (more below), the Generator could have picked a valid (according to KIP-9 rules) UTXO to spend according to the request conditions, but has failed while supposedly only attempting one of them.

Method: `generator.next()`

Input parameters:
* payload
  ```
  636970685f6d73673a01b09eadce2f05f00289a1de1603383e2b8283bc8e74c37bed75a0ce0db369628860d5d2e6b1243f9685809f5889d0f8cab14313432b19059baa0621e6a5d8e9888415b5884e7ba44e483ebd72e089d617aab093bdeb590e38ec1b30d6bf0083f4780e0b351caab91af44d11fd85b0291fcff75b5d87445ba1709858cc89ea121b86304249e63b8311283c6fa8be683e1ee1c786629e79fd4d469634072006d3f8cdfc74b4d6732b41a76919071e4ebb488900297d757abe73fb8d6d8b012fa2a038414c6cea7a14cd2e07ad3bd9083e0b1976b0c59cb3b64566250455ab4d37562a839e1f6ee0e665b535255a233a0ade5d9205e52cc5ab6cb4fe74cb1d736c33619e1401c8e1cd75e519e138db4b7b101d509d1fbf7820a4a3223129038e1f3bc2064dae148f36dac0b4d53ab82fb1c76d2b9add63574662169cba7b30797b896f7aa1c79d9ed5c1be48112e1c2ac4ab0fe84b6194aa3c94ff4cdf6798f3c217c8
  ```
* address: `kaspatest:qqfxqnutdqycengg2w3ch36djg20y9ws5kyvjhzflgxw5ledzwt9ulkhlny0x` (as an `Address`)
* priorityFee: `0n`
* outputs: `[new PaymentOutput(destinationAddress, 20000000n)]` (0.2 TKAS)

UTXO Context: (sorry about the format)
```json
{"utxos":[{"entry":{"address":{"version":"PubKey","prefix":"kaspatest","payload":"qzadn9tyvela5tc9rpn5tnrlkmdq9988wu8nh57pa4d62nw92pspsd740qlkv"},"outpoint":{"transactionId":"3017aa20bfe904407a214d81997468b8defef0617cef9007c7fa12b322a22c05","index":0},"amount":"29624618","scriptPublicKey":{"version":0,"script":"20bad99564667fda2f05186745cc7fb6da0294e7770f3bd3c1ed5ba54dc5506018ac"},"blockDaaScore":"183065193","isCoinbase":false},"outpoint":{"transactionId":"3017aa20bfe904407a214d81997468b8defef0617cef9007c7fa12b322a22c05","index":0},"address":{"version":"PubKey","prefix":"kaspatest","payload":"qzadn9tyvela5tc9rpn5tnrlkmdq9988wu8nh57pa4d62nw92pspsd740qlkv"},"amount":"29624618","isCoinbase":false,"blockDaaScore":"183065193","scriptPublicKey":{"version":0,"script":"20bad99564667fda2f05186745cc7fb6da0294e7770f3bd3c1ed5ba54dc5506018ac"}},{"entry":{"address":{"version":"PubKey","prefix":"kaspatest","payload":"qzadn9tyvela5tc9rpn5tnrlkmdq9988wu8nh57pa4d62nw92pspsd740qlkv"},"outpoint":{"transactionId":"7a973447dd13e0c1375f98e669794f87c7421f9976d54a09e1bcf8443bbe961a","index":0},"amount":"50000000","scriptPublicKey":{"version":0,"script":"20bad99564667fda2f05186745cc7fb6da0294e7770f3bd3c1ed5ba54dc5506018ac"},"blockDaaScore":"185277525","isCoinbase":false},"outpoint":{"transactionId":"7a973447dd13e0c1375f98e669794f87c7421f9976d54a09e1bcf8443bbe961a","index":0},"address":{"version":"PubKey","prefix":"kaspatest","payload":"qzadn9tyvela5tc9rpn5tnrlkmdq9988wu8nh57pa4d62nw92pspsd740qlkv"},"amount":"50000000","isCoinbase":false,"blockDaaScore":"185277525","scriptPublicKey":{"version":0,"script":"20bad99564667fda2f05186745cc7fb6da0294e7770f3bd3c1ed5ba54dc5506018ac"}}]}
```

To make it simpler, here are the UTXO:
Image

console error: `Storage mass exceeds maximum`

**Expected behavior**
Generator should potentially fail on the utxo that has `0.29624618 TKAS` but should eventually try another combination for the request, in this case the utxo that has `0.5 TKAS`.

**SDK Information**
 - wasm web kaspa v1.0.0, taken from built wasm package on the releases page.

IzioDev:
- Use consensus storage mass calculation for `storage_mass_with_change` in `calculate_mass` while trying this UTXO candidate.
- (needs validation): absorb change to true when `storage_mass_with_change` is exceeding `MAXIMUM_STANDARD_TRANSACTION_MASS` (existing check: and that `storage_mass` is not exceeding it)
- In rare case where the output value is close to input value, try adding the next utxo input if available
- Remove a wrong double count of transaction fees in case it decides to absorb fees
- Small refacto to better understand the conditions deciding between with change vs without change

Fixes #701

### DIFF
--- a/wallet/core/src/tx/generator/test.rs
+++ b/wallet/core/src/tx/generator/test.rs
@@ -826,14 +826,16 @@ fn test_generator_fan_out_1() -> Result<()> {
 
 #[test]
 fn test_generator_should_pick_valid_alternative_utxo_after_storage_mass_rejection() -> Result<()> {
-    let generator =
-        generator(test_network_id(), &[0.21, 0.5], &[], None, Fees::sender(Sompi(0)), [(output_address, Kaspa(0.2))].as_slice())?;
-
-    let pending = generator.generate_transaction()?.expect("expected final transaction");
-    assert!(pending.is_final(), "expected final transaction");
-    assert_eq!(pending.utxo_entries().len(), 2, "expected to aggregate another input");
-    assert!(pending.change_output_index().is_some(), "expected explicit change output (no fee absorption)");
-    assert!(pending.change_value() > 0, "expected non-zero change output (no fee absorption)");
+    generator(test_network_id(), &[0.21, 0.5], &[], None, Fees::sender(Sompi(0)), [(output_address, Kaspa(0.2))].as_slice())?
+        .harness()
+        .fetch(&Expected {
+            is_final: true,
+            input_count: 2,
+            aggregate_input_value: Kaspa(0.71),
+            output_count: 2,
+            priority_fees: FeesExpected::None,
+        })
+        .finalize();
 
     Ok(())
 }


wallet/core/src/tx/generator/generator.rs
```diff
@@ -692,9 +692,9 @@ impl Generator {
         }
     }
 
-    /// Test if the current state has additional UTXOs. Use with caution as this
-    /// function polls the iterator and relocates UTXO into UTXO stash.
-    fn has_utxo_entries(&self, context: &mut Context, stage: &mut Stage) -> bool {
+    /// Pull the next available UTXO, if any, and stash it for the next accumulation attempt.
+    /// Use with caution: this advances one of the underlying UTXO sources and mutates `utxo_stash`.
+    fn try_stash_next_utxo(&self, context: &mut Context, stage: &mut Stage) -> bool {
```
coderofstuff: `try_` usually means you'll return a `Result` or `Option`. This returns a `bool`.

Probably better to just rename to `stash_next_utxo` to behave similar to [HashSet::insert](https://doc.rust-lang.org/src/std/collections/hash/set.rs.html#976)

wallet/core/src/tx/generator/generator.rs
```diff
@@ -868,7 +878,6 @@ impl Generator {
 
                 data.aggregate_mass = calc.combine_mass(compute_mass, storage_mass);
 
-                transaction_fees += change_output_value;
```
coderofstuff, line 871: good catch

wallet/core/src/tx/generator/generator.rs
```diff
@@ -930,19 +943,20 @@ impl Generator {
                     0
                 } else {
                     let storage_mass_no_change = self.calc_storage_mass(data, self.inner.final_transaction_outputs_harmonic);
-                    if storage_mass_with_change < storage_mass_no_change {
+                    let fees_with_change = calc.calc_fee_for_mass(storage_mass_with_change);
+                    let fees_no_change = calc.calc_fee_for_mass(storage_mass_no_change);
+                    let difference = fees_with_change.saturating_sub(fees_no_change);
+                    let should_absorb_change = difference > change_value;
+                    let can_keep_change = storage_mass_with_change <= MAXIMUM_STANDARD_TRANSACTION_MASS;
+                    let prefer_with_change = storage_mass_with_change < storage_mass_no_change;
```
coderofstuff: In what scenario will `prefer_with_change` be true? Based on KIP9's formula I think this will always be false (storage mass of something with outputs N will always be less than the storage mas of something with outputs N+1)

wallet/core/src/tx/generator/test.rs
```diff
@@ -823,3 +823,71 @@ fn test_generator_fan_out_1() -> Result<()> {
...
+#[test]
+fn test_generator_should_pick_valid_alternative_utxo_after_storage_mass_rejection() -> Result<()> {
+    let generator =
+        generator(test_network_id(), &[0.21, 0.5], &[], None, Fees::sender(Sompi(0)), [(output_address, Kaspa(0.2))].as_slice())?;
+
+    let pending = generator.generate_transaction()?.expect("expected final transaction");
+    assert!(pending.is_final(), "expected final transaction");
+    assert_eq!(pending.utxo_entries().len(), 2, "expected to aggregate another input");
+    assert!(pending.change_output_index().is_some(), "expected explicit change output (no fee absorption)");
+    assert!(pending.change_value() > 0, "expected non-zero change output (no fee absorption)");
```
coderofstuff: You can assert the exact value of the change, as well as assert that fees remained the same (hence did not get absorbed)

IzioDev: Should this test know the expected fees value in advance?
Otherwise, I don't know how the test adds extra confidence (verifies total_in - fees - 0.2 = change_output)

wallet/core/src/tx/generator/generator.rs
IzioDev: I think it's true it will never happens. And nor the reader of the wallet framework knows about formula details, so I mostly added this for clarity of no undefined behaviors. I'm happy to remove this line and switch to a comment if you find it more appropriate

coderofstuff: This is going to cause more confusion than clarity if it's always false. I suggest removing it, and collapsing the conditions that depend on it.

wallet/core/src/tx/generator/test.rs
coderofstuff: My concern really is that checking that `.change_value() > 0` and `.change_output_index().is_some()` only implies that there's a change. I was wrong in my earlier statement here - the fee would actually change because a new UTXO is added in the input. I think it's worth checking that the amount that fee increased  is proportional to the utxo being added and was not absorbing the change partially (unlikely, but that's my motivation).

You can treat this comment as optional.

IzioDev: I realized the file had good utils for that (`Expected`, `expect()`...), changed the test to that.

It's also looking more compacted: [review: hardened test](https://github.com/kaspanet/rusty-kaspa/pull/894/commits/ad72faf675aa1f7977940d5d501b0dfb4c15c2d9)

wallet/core/src/tx/generator/generator.rs
IzioDev: [review: remove unecessary condition](https://github.com/kaspanet/rusty-kaspa/pull/894/commits/461698f8632bb08101e9349a08d2513813356b91)

wallet/core/src/tx/generator/generator.rs
IzioDev: [review: rename stash next utxo](https://github.com/kaspanet/rusty-kaspa/pull/894/commits/f998fd4c29be6e6a605b2ed36da62d5d0749efdb)

Issue #514, opened 2024-08-05
Implement a way for a node operator to include their own transactions in blocks


coderofstuff:
**Use case:** A node operator (such as a pool) may want to choose to always include transactions that they submit to their own node in blocks they mine. A specific example is when a pool may want to send its payout transaction in blocks that it mines on its own.

**Status:** Requires Design

Kaspa-francophone:
next week , we will be solo mining, a lot of us . If i found a block , it will be nice to send it "internaly"t straigh o the wallet attached to my node without paying fees to other miners :)

PR #902, opened 2026-03-06
feat(rpc): add SubmitLocalTransaction (#514)

RossKU:
## Summary

Implements [#514](https://github.com/kaspanet/rusty-kaspa/issues/514) — a new RPC method `SubmitLocalTransaction` that lets node operators include their own transactions in blocks they mine, bypassing the relay fee while preserving all other standard checks.

- **Local TX store** in `MiningManager` — separate `HashMap`, not in mempool
- **`LocalFirstSelector`** wraps frontier selector, yields local TXs first in block templates
- **Full validation** — consensus (sigs, UTXO, locktime) + standard checks (dust, mass, script class). Only relay fee is skipped via `check_fee: bool` parameter on existing `check_transaction_standard_in_context`
- **Not broadcast** — local TXs are never sent to peers
- **Auto-cleanup** — removed from store when included in a mined block

## Changes (24 files, +883 -16)

| Area | Files | What |
|------|-------|------|
| Mining | `selector.rs`, `manager.rs`, `check_transaction_standard.rs`, `validate_and_insert_transaction.rs` | `LocalFirstSelector`, local store, `check_fee` param |
| RPC Core | `ops.rs`, `rpc.rs`, `message.rs`, `wasm/message.rs` | Op enum, trait, request/response types, TS bindings |
| RPC Service | `service.rs` | Handler implementation |
| Protocol | `flow_context.rs` | `submit_local_transaction()` (no broadcast) |
| gRPC | `proto/*.proto`, `convert/`, `ops.rs`, `factory.rs`, `client/lib.rs` | Full gRPC plumbing |
| wRPC | `client.rs`, `router.rs`, `wasm/client.rs` | wRPC + WASM bindings |
| Tests | `manager_tests.rs` (12 unit), `rpc_tests.rs` (4 E2E) | Acceptance, rejection, template, zero-fee, mempool isolation |
| Mocks | `rpc_core_mock.rs` ×2 | Mock implementations |

## Security

- **Relay fee is the only check skipped.** Dust threshold, mass limits, script standardness, and all consensus validation rules are enforced.
- **Access control**: Same model as existing RPC — infrastructure-level protection (firewall/VPN). A `TODO` comment notes potential future CLI flag gating.
- **No new attack surface beyond free TX inclusion**: An unauthorized caller can include transactions without paying fees, but all other protections (dust, mass, signatures) remain.

### DIFF
--- a/mining/src/block_template/selector.rs
+++ b/mining/src/block_template/selector.rs
@@ -1,6 +1,6 @@
 use kaspa_core::{time::Stopwatch, trace};
 use rand::Rng;
-use std::collections::HashMap;
+use std::collections::{HashMap, HashSet};
 
 use crate::model::candidate_tx::CandidateTransaction;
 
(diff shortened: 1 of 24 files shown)


RossKU:
**Security note**: `submit_local_transaction_call` currently has no access guard. Since local TXs get priority in block templates (via `LocalFirstSelector`), an attacker reaching the RPC port could submit zero-fee TXs that displace fee-paying ones — effectively free-riding on block space.

Proposed fix: block when `--unsaferpc` is ON (reverse of the usual pattern). Operators using `--unsaferpc` have more methods exposed, increasing abuse risk. Operators without it typically run behind localhost/firewall — the intended use case for this feature.

```rust
if self.config.unsafe_rpc {
    return Err(RpcError::UnavailableInSafeMode);
}
```

Issue #162, opened 2023-04-12
Implement RPC authentication


D-Stacks:
I would like to transfer the following issue to the `rusty-kaspa` repo:

> Currently, many nodes have RPC ports potentially open to the world
> Some manner of authentication should be implemented to mitigate it
> 
> Suggested solution: gRPC supports authentication out-of-the-box: https://grpc.io/docs/guides/auth/

Source: https://github.com/kaspanet/kaspad/issues/1893

Only the auth method should probably ideally support wRPC, gRPC and all potential future protocols in a generic manner.

D-Stacks:
As per the original issue; it is preferable if the auth of a connection can be handled on a per call / notfification basis.

KaffinPX:
It should be handled by proxies, reference sofwares should be simple as much as possible i think

RoboticWF:
Starting to figure this out now

freshair18:
Was there been progress on this issue since?

Issue #105, opened 2023-01-26
Create the API request that could return the free disc space on the node host PC, for a remote monitoring purposes


Aleoami:

KaffinPX:
I think its too sensitive info so shouldnt be on node

KaffinPX:
if we can control allowed endpoints on config, why not?

freshair18:
@Aleoami @KaffinPX  did you guys converge on this?

RossKU:
Implementation draft completed on local branch `feat/rpc-auth-162` (built on top of #162).

**Changes (7 files):**
- `StorageMetrics` extended with `disk_free_bytes` and `disk_total_bytes`
- Uses `sysinfo::Disks` to query the partition containing `data_dir`
- Borsh serialization versioned (v1 → v2) for backward compat
- Proto, gRPC conversion, and mock updated accordingly

No PR at this time — waiting for #162 to land first so this can be rebased cleanly onto master.

PR #903, opened 2026-03-07
feat(rpc): cookie-based RPC authentication (#162)

RossKU:
## Summary

Implements opt-in RPC authentication for kaspad, addressing #162. Node operators exposing RPC (especially `--unsaferpc`) over a network can now require a shared-secret cookie token.

- Adds `--auth` CLI flag with three modes: `unsafe`, `all`, `unsafe,-Method1,-Method2`
- Generates a random 256-bit cookie at startup, written to `<datadir>/<network>/rpc.cookie`
- Cookie is deleted on clean shutdown
- gRPC clients authenticate via `Authorization: Bearer <token>` metadata
- Unauthenticated calls to protected methods return `AuthenticationRequired` error

### Why cookie-based, not TLS client certificates?

Client certificates require a PKI (CA setup, cert distribution, revocation). For a single-node monitoring scenario, this is heavy. A cookie file is simple: read the file, pass the hex string. The file is local to the machine, so access control is delegated to filesystem permissions (`0600` on Unix).

### Auth granularity: per-method, not per-connection-type

Rather than a blanket "all gRPC is authenticated", auth is checked **per RPC method call** via `require_unsafe()` and `require_any_auth()`. This allows:

- `--auth=unsafe` — only dangerous methods (Shutdown, Ban, AddPeer, ResolveFinalityConflict) require the token. Read-only methods like GetInfo remain open. Requires `--unsaferpc`.
- `--auth=all` — every RPC call requires the token.
- `--auth=unsafe,-Ban,-Unban` — unsafe methods except Ban/Unban require auth (exclusion list).

New unsafe methods added in the future **default to requiring auth** (fail-safe: exclusion list, not inclusion list).

### Connection-level auth state

Authentication is checked **once at connection time** for gRPC (token extracted from metadata in `connection_handler.rs`). The `is_authenticated` flag is stored on the `Connection` struct and queried per-call. This avoids repeated token parsing and is consistent with gRPC's connection-oriented model.

For wRPC (WebSocket), `is_authenticated` currently returns `false` — wRPC has no metadata mechanism for token passing. This is a known limitation; wRPC auth is deferred to a follow-up.

### Macro-level enforcement

The gRPC server uses procedural macros (`rpc/macros/src/grpc/server.rs`) to generate request handlers. Auth checks are injected **inside the macro expansion**, so every handler gets auth enforcement automatically. This prevents the failure mode where a developer adds a new RPC method and forgets to add an auth check.

### Token generation and comparison

- 256-bit random secret via `rand::thread_rng().fill_bytes()`
- Stored as hex in the cookie file (64 characters)
- Comparison uses direct byte equality (not constant-time) — timing attacks are not practical for this threat model (local file, localhost or LAN)

## Files changed (18 files, +430 / -33)

| Area | Files | Purpose |
|------|-------|---------|
| Core auth module | `rpc/core/src/auth.rs` (new) | `AuthMode` enum, `RpcAuthConfig`, token verification, 15 unit tests |
| Error/trait plumbing | `rpc/core/src/{error,lib,api/connection}.rs` | `AuthenticationRequired` error variant, `is_authenticated()` trait method |
| CLI integration | `kaspad/src/{args,daemon}.rs` | `--auth` flag, cookie generation, arg validation |
| RPC service | `rpc/service/src/service.rs` | `require_unsafe()` / `require_any_auth()` helpers, cookie cleanup on stop |
| gRPC server | `rpc/grpc/server/src/{connection,connection_handler,service,adaptor}.rs` | Token extraction from metadata, `authenticated` field on Connection |
| wRPC server | `rpc/wrpc/server/src/{connection,server}.rs` | `RpcConnection` impl (stub), `core_service()` accessor |
| Macros | `rpc/macros/src/{grpc,wrpc}/server.rs` | Auth checks injected into generated handlers, connection forwarding |
| Config | `consensus/core/src/errors/config.rs`, `kaspad/Cargo.toml` | `ConfigError::General`, `faster-hex` dep |
| Tests | `rpc/grpc/server/src/tests/client_server.rs` | Pass `None` auth_config to test adaptor |

## CI note

Two pre-existing CI failures unrelated to this PR:
- **Lints**: `unstable_name_collisions` in `utils/src/as_slice.rs` (toolchain upgrade issue)
- **Tests**: `daemon_utxos_propagation_test` timeout (flaky in CI)

### DIFF
--- a/rpc/core/src/auth.rs
+++ b/rpc/core/src/auth.rs
@@ -65,6 +65,10 @@ impl RpcAuthConfig {
 mod tests {
     use super::*;
 
+    // =========================================================================
+    // A. AuthMode parsing
+    // =========================================================================
+
     #[test]
     fn test_parse_unsafe() {
         let mode = AuthMode::parse("unsafe").unwrap();
@@ -85,7 +89,7 @@ mod tests {
             assert!(exclusions.contains("Unban"));
             assert!(!exclusions.contains("Shutdown"));
         } else {
-            panic!("Expected Admin mode");
+            panic!("Expected Unsafe mode");
         }
     }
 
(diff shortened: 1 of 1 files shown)


RossKU:
Once this lands, I'll follow up with a PR for #105 (disk free/total bytes in StorageMetrics). Implementation is ready locally.

PR #908, opened 2026-03-08
fix(wasm): use JsValue marshalling for subscribe/unsubscribe UTXOs changed addresses (#443)

### LINKED ISSUE #443: subscribeUtxosChanged Method Invalid Types Issue: Throws RuntimeError with Account Object
**Describe the bug**
When attempting to use the subscribeUtxosChanged method with an Account object, despite the documentation suggesting it accepts either an account object or a string, a runtime error is thrown indicating a memory access out of bounds. This issue occurs when running the method with a new Account(string) as an argument.

**To Reproduce**
Instantiate an Account object using a string identifier.
` derivedAddresses.push(new Address(address))`
Call the subscribeUtxosChanged method with the newly created Account object.
`await this.wsRpc.subscribeUtxosChanged(addresses)`
**Error:**
```
RuntimeError: memory access out of bounds
    at null.<anonymous> (wasm://wasm/01be8a16:1:2782986)
    at null.<anonymous> (wasm://wasm/01be8a16:1:4883531)

```
**Expected behavior**
The subscribeUtxosChanged method should accept an Account object as its documentation indicates, without throwing a runtime error.

**Desktop (please complete the following information):**
OS: macOS M2
Environment: WebAssembly version 0.13.5

RossKU:
## Summary
- Fix `subscribeUtxosChanged` / `unsubscribeUtxosChanged` WASM methods to use `JsValue` marshalling instead of `wasm_bindgen` ABI for Address arrays, matching the pattern used elsewhere in the codebase
- Add `wasm_bindgen_test` verifying Address array JsValue roundtrip

Closes #443

### DIFF
--- a/crypto/addresses/src/wasm.rs
+++ b/crypto/addresses/src/wasm.rs
@@ -195,8 +195,7 @@ mod tests {
         array.push(&JsValue::from(addr1.clone()));
         array.push(&JsValue::from(addr2.clone()));
 
-        let addresses_result: Result<Vec<Address>, _> =
-            AddressOrStringArrayT::from(JsValue::from(array)).try_into();
+        let addresses_result: Result<Vec<Address>, _> = AddressOrStringArrayT::from(JsValue::from(array)).try_into();
 
         assert!(addresses_result.is_ok(), "direct try_into() failed: {:?}", addresses_result.err());
         let addresses = addresses_result.unwrap();


IzioDev:
This seems to be the proper fix indeed. Have you tried reproducing the error prior changing the implementation? Did it failed with the same error we had on the wrpc request argument (address)? (and that you fixed too in another PR)

RossKU:
Rebased on top of #916 and added a comparative test that exercises both the old `try_into()` pattern and the new `JsValue::from()` marshalling pattern side by side.

CI result: both patterns pass — https://github.com/RossKU/rusty-kaspa/actions/runs/23085069068

This means the original conversion path was already working correctly, and the issue reported in #443 was likely not caused by this code path. Given aspect's comment on #443 (the reporter was passing `Account` objects instead of `Address`), the original error was probably API misuse rather than a marshalling bug.

Happy to close this PR if you agree.

RossKU:
Rebased on #916. To answer your question about reproduction — I added a comparative test that exercises both the old `try_into()` and the new `JsValue::from()` patterns. Both pass on CI: https://github.com/RossKU/rusty-kaspa/actions/runs/23085069068

The conversion path wasn't the problem. #443 is likely closeable as well. Happy to close this PR.

PR #910, opened 2026-03-08
fix(wallet): use actual P2PK script size instead of SCRIPT_VECTOR_SIZE (#733)

### LINKED ISSUE #733: Why `SCRIPT_VECTOR_SIZE` is 36?

RossKU:
## Summary
- `transaction_standard_output_serialized_byte_size()` was using `SCRIPT_VECTOR_SIZE` (35) instead of the actual standard P2PK script size (34 bytes: 1 OP_DATA_32 + 32 pubkey + 1 OP_CHECKSIG)
- `SCRIPT_VECTOR_SIZE` is the SmallVec stack allocation size, not the serialized script size
- This caused the function to return 53 instead of the correct 52

Closes #733

### DIFF
--- a/wallet/core/src/tx/mass.rs
+++ b/wallet/core/src/tx/mass.rs
@@ -7,7 +7,7 @@ use crate::result::Result;
 use kaspa_consensus_client as kcc;
 use kaspa_consensus_client::UtxoEntryReference;
 use kaspa_consensus_core::mass::calc_storage_mass as consensus_calc_storage_mass;
-use kaspa_consensus_core::tx::{SCRIPT_VECTOR_SIZE, Transaction, TransactionInput, TransactionOutput};
+use kaspa_consensus_core::tx::{Transaction, TransactionInput, TransactionOutput};
 use kaspa_consensus_core::{config::params::Params, constants::*, subnets::SUBNETWORK_ID_SIZE};
 use kaspa_hashes::HASH_SIZE;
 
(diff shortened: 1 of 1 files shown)


IzioDev:
ECDSA pubkeys are 33-bytes length, adding op_data and checksig, it totals in 35 bytes.
I think we should keep the current declaration, used by mass calculator to decides whether or not this is a dust amount.

edit: the function name says standard, so I assume it was intended to be designed specifically for Schnorr. Maybe assuming 32 bytes is desired.

RossKU:
Right, your edit is spot on — the function is about `standard` outputs, which means Schnorr P2PK.

From the codebase itself, `script_public_key.rs` L24:
```
/// Used as the underlying type for script public key data, optimized for the common p2pk script size (34).
```

And in `standard.rs`, `pay_to_pub_key` asserts a 32-byte payload → `OpData32 (1) + pubkey (32) + OpCheckSig (1) = 34 bytes`. ECDSA's `pay_to_pub_key_ecdsa` is 35 bytes but that's the non-standard path.

`SCRIPT_VECTOR_SIZE = 35` is the SmallVec inline capacity (sized to fit ECDSA), not the standard script size. So using it here was a mixup.

RossKU:
`MassCalculator::is_dust()` does use this constant, so the threshold shifts slightly (by 1 byte worth). I see this as a correction rather than a loosening — would like to hear your thoughts.

PR #913, opened 2026-03-08
Disable minimal data encoding policy after covenants activation

someone235:
Since the covpp HF already disabled minimal encoding checks for numbers, and Kaspa has different ways to deal with transaction malleability (by excluding the signatures from the transaction ID digest), it seems like the minimal push data encoding is no longer relevant and introduces an unneeded hurdle for ecosystem developers. This is especially relevant for advanced covenants that require the script to be aware in advance about push data size, in a way that's independent from the actual value of the pushed data.

### DIFF
--- a/crypto/txscript/src/script_builder.rs
+++ b/crypto/txscript/src/script_builder.rs
@@ -139,21 +139,21 @@ impl ScriptBuilder {
     /// the length of the data. A zero length buffer will lead to a push of empty
     /// data onto the stack (OP_0). No data limits are enforced with this function.
     fn add_raw_data(&mut self, data: &[u8]) -> &mut Self {
-        let data_len = data.len();
-
-        // When the data consists of a single number that can be represented
-        // by one of the "small integer" opcodes, use that opcode instead of
-        // a data push opcode followed by the number.
-        if data_len == 1 && OP_SMALL_INT_MIN_VAL <= data[0] && data[0] <= OP_SMALL_INT_MAX_VAL {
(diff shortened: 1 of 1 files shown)


crypto/txscript/src/script_builder.rs
```diff
@@ -144,34 +144,16 @@ impl ScriptBuilder {
...
-            self.script.extend(once(OpPushData2).chain((data_len as u16).to_le_bytes()));
-        } else {
-            self.script.extend(once(OpPushData4).chain((data_len as u32).to_le_bytes()));
-        }
-
-        // Append the actual data.
-        self.script.extend(data);
-        self
+        // For all other data, choose the appropriate push opcode based on data length.
+        self.add_raw_data_with_data_opcode(data)
```
biryukovmaxim:
```suggestion
        match data {
            [] => {
                self.script.push(Op0);
            }
            [OP_1_NEGATE_VAL] => {
                self.script.push(Op1Negate);
            }
            // When the data consists of a single number that can be represented
            // by one of the "small integer" opcodes, use that opcode instead of
            // a data push opcode followed by the number.
            [OP_SMALL_INT_MIN_VAL..=OP_SMALL_INT_MAX_VAL] => {
                self.script.push((Op1 - 1) + data[0]);
            }
            // For all other data, choose the appropriate push opcode based on data length.
            _ => {
                self.add_raw_data_with_data_opcode(data);
            }
        }
        self
```

PR #916, opened 2026-03-11
feat(crypto/addresses): no std support and segregate wasm usage

IzioDev:
* allow usage in no_std environment: `cargo check -p kaspa-addresses --no-default-features --lib --target thumbv7em-none-eabi`
* allow disabling wasm glue completely by disabling default features and enabling `std` only: useful for mobile targets

The PR decides to keep the default behavior of having `wasm` feature enabled; this should be reconsidered at a greater scale in the future (intuitively, disabled by default).

A small refactoring of pure wasm parts has been proposed to be moved in a new `wasm.rs` files, this file is feature gated from `lib.rs`.

Downside that should be evaluated more globally:
- thiserror only support no_std since v1.81+, it has been proposed to only upgrade in `crypto/addresses` for now to reduce impacts. Ideally unified version should live at workspace-level in the future.
- borsh and serde have to be re-pinned from `crypto/addresses` to disable default features, ideally it should be disabled at workspace-level and granularly enabled at crate-level


crypto/addresses/src/lib.rs
```diff
@@ -1,21 +1,40 @@
...
 //!
 //! Kaspa [`Address`] implementation.
 //!
 //! In it's string form, the Kaspa [`Address`] is represented by a `bech32`-encoded
-//! address string combined with a network type.  The `bech32` string encoding is
+//! address string combined with a network type. The `bech32` string encoding is
 //! comprised of a public key, the public key version and the resulting checksum.
 //!
 
+extern crate alloc;
```
biryukovmaxim, line 11: u also need to extern core

crypto/addresses/src/lib.rs
```diff
@@ -1,21 +1,40 @@
...
-use std::fmt::{Display, Formatter};
 use thiserror::Error;
-use wasm_bindgen::prelude::*;
-use workflow_wasm::{
-    convert::{Cast, CastFromJs, TryCastFromJs},
-    extensions::object::*,
-};
+
+#[cfg(feature = "wasm32-sdk")]
+use wasm_bindgen::prelude::wasm_bindgen;
```
biryukovmaxim, line 30: actually wasm bindgen works in no-std. so i prefer to have different features for workflow and wasm-bindgen whenever its possible

crypto/addresses/src/lib.rs
```diff
@@ -202,19 +205,20 @@ pub type PayloadVec = SmallVec<[u8; PAYLOAD_VECTOR_SIZE]>;
 /// Kaspa [`Address`] struct that serializes to and from an address format string: `kaspa:qz0s...t8cv`.
 ///
 /// @category Address
-#[derive(PartialEq, Eq, PartialOrd, Ord, Clone, Hash, CastFromJs)]
-#[wasm_bindgen(inspectable)]
+#[derive(PartialEq, Eq, PartialOrd, Ord, Clone, Hash)]
+#[cfg_attr(feature = "wasm32-sdk", derive(CastFromJs))]
+#[cfg_attr(feature = "wasm32-sdk", wasm_bindgen(inspectable))]
```
biryukovmaxim: you can unify it to single line

crypto/addresses/src/lib.rs
```diff
@@ -354,63 +319,69 @@ impl<'de> Deserialize<'de> for Address {
...
+                #[cfg(all(feature = "wasm32-sdk", target_arch = "wasm32"))]
                 {
                     write!(formatter, "string-type: string, str; bytes-type: slice of bytes, vec of bytes; map; number-type - pointer")
                 }
-                #[cfg(not(target_arch = "wasm32"))]
+                #[cfg(not(all(feature = "wasm32-sdk", target_arch = "wasm32")))]
                 {
                     write!(formatter, "string-type: string, str; bytes-type: slice of bytes, vec of bytes; map")
                 }
             }
```
biryukovmaxim: i dont get why do we need gates here

crypto/addresses/Cargo.toml
```diff
@@ -9,15 +10,27 @@ include.workspace = true
...
-workflow-wasm.workspace = true
-workflow-log.workspace = true
+# these two crates defines same version as workspace, but disable default features (that enable std)
+# long term: disable at workspace level and enable granularily where required
+borsh = { version = "1.5.1", default-features = false, features = ["derive", "rc"] }
+serde = { version = "1.0.190", default-features = false, features = ["alloc", "derive", "rc"] }
+
+# newer version thiserror version (compared to workspace) allows disabling std
+# long term: upgrade thiserror at workspace level
+thiserror = { version = "2.0.18", default-features = false }
```
biryukovmaxim, line 27: there are fine to me, but if you cherry pick(https://github.com/kaspanet/rusty-kaspa/commit/ca853f7728e31f658b747d005baa6858e0ace856) it will be fixed there already

biryukovmaxim:
consider adding https://github.com/hobofan/cargo-nono to cicd or at least to local testing pipeline to check if crate is nostd compatible
@IzioDev

crypto/addresses/src/lib.rs
IzioDev: it seems to work fine on my nRF52810, any specific reasons to add it?

crypto/addresses/src/lib.rs
IzioDev: Is there a place where i'd like one without the other in this crate?

crypto/addresses/src/lib.rs
IzioDev: me neither, but it was there, should I remove?

crypto/addresses/src/lib.rs
biryukovmaxim: no need, since in Rust 2018 edition and later, core is implicitly available in all crates

crypto/addresses/src/lib.rs
biryukovmaxim: unlikely

IzioDev:
> consider adding https://github.com/hobofan/cargo-nono to cicd or at least to local testing pipeline to check if crate is nostd compatible @IzioDev

this doesn't seems compatible on windows, so local testing will not work in all cases, i'm adding a CI job for this one.

crypto/addresses/src/lib.rs
IzioDev: done [review: one cfg attr for two attrs](https://github.com/kaspanet/rusty-kaspa/pull/916/commits/2869ff9b54c2d926315054d81555d98aa831c409)

crypto/addresses/src/lib.rs
IzioDev: done: [review: serde visitor expecting unified for all features](https://github.com/kaspanet/rusty-kaspa/pull/916/commits/6e32184522bd6c4db8bffe4662e465b2ee8b64f3)

IzioDev:
> > consider adding https://github.com/hobofan/cargo-nono to cicd or at least to local testing pipeline to check if crate is nostd compatible @IzioDev
> 
> this doesn't seems compatible on windows, so local testing will not work in all cases, i'm adding a CI job for this one.

as per discussed in DM, we are trying another leaner approach which just is to use cargo check on a no std target. added to the CI here: [ci: add no std check](https://github.com/kaspanet/rusty-kaspa/pull/916/commits/f2d8cd28946326dcadd92081195bcac0cd30fd94)

PR #918, opened 2026-03-15
Add IBD benchmark

someone235:
This pull request introduces a new benchmarking tool for measuring Initial Block Download (IBD) performance using a simpa-generated database, and updates the `Daemon` resource management logic to allow for more flexible temporary directory handling. The main addition is the `ibd_bench` benchmark, which enables reproducible performance testing of the sync process.


biryukovmaxim (commented):
Seems fine to me. The only question I have should not it be able to work against any db, like I have a running devnet. Why not clone it's dir and measure against it. Probably only word "simpa" would be changed

biryukovmaxim (approved):
Im definitely not against to have such tool, but from my perspective client and node must be 2 separate binaries. Id like to use it not rebuilding client everytime, since

PR #919, opened 2026-03-16
fix(wasm): temporary patch to freeze wasm-related crates

IzioDev:
Consumers of rk crates defining wasm attributes were hitting a bug related to a breaking change in js-sys.

In order to fix the root cause and upgrade, workflow-rs needs a patch. In the meantime, this freeze all wasm-related crate previous to the introduced breaking change.

explanation: https://github.com/workflow-rs/workflow-rs/issues/16

Note: Despite the PR title and the ambiguous description: in a pure Rust project, using kaspa-wallet-core for example result in a compilation error


intricous:
This patch resolves the cause of the "trait not implemented" error that occurs when building a bin crate that includes kaspa-wallet-core v1.1.0 as a dependency.

PR #920, opened 2026-03-16
Add intropsections ops for tx covenant fields

someone235:
This pull request adds two new opcodes—`OpOutputCovenantId` and `OpOutputAuthorizingInput`—to the transaction script engine, enabling covenant-related queries on transaction outputs. It also updates the behavior of `OpInputCovenantId` to consistently return a zero hash when no covenant is present, and provides comprehensive test coverage for these changes. The changes ensure that covenant-related opcodes behave predictably and are correctly enabled when covenant flags are set.

**New covenant output opcodes:**

* Added `OpOutputCovenantId` (opcode `0xd5`): Returns the covenant ID of a specified transaction output, or the zero hash if none exists. [[1]](diffhunk://#diff-0220766fcee8670d6f6d8807998eb9d5bfb38b0fc1a9526ccf76cc49f13ab791L1530-R1564) [[2]](diffhunk://#diff-de7381394db57cf3590e8033ff9318190d26f680f0353b8fd5374ae401a9a0b2L252-R252)
* Added `OpOutputAuthorizingInput` (opcode `0xd6`): Returns the index of the authorizing input for a specified transaction output's covenant, or `-1` if none exists.

**Behavioral improvements and consistency:**

* Updated `OpInputCovenantId` to return the zero hash instead of `false` when no covenant ID is present, ensuring consistent return types.


crypto/txscript/src/opcodes/mod.rs
```diff
@@ -1435,10 +1436,8 @@ opcode_list! {
                     let [idx]: [i32; 1] = vm.dstack.pop_items()?;
                     let idx = i32_to_usize(idx)?;
                     let utxo = tx.utxo(idx).ok_or_else(|| TxScriptError::InvalidInputIndex(idx as i32, tx.inputs().len()))?;
-                    match utxo.covenant_id{
-                        None => vm.dstack.push_item(false),
-                        Some(covenant_id) => push_data(covenant_id.as_bytes().into(), vm),
-                    }
+                    let covenant_id = utxo.covenant_id.unwrap_or(ZERO_HASH);
```
biryukovmaxim, line 1439: iiuc false leads to deserialization error when we use it as hash, zero hash will lead to error if only used for OpCodes that use it as covenantid.  I think fail earlier is better here

crypto/txscript/src/opcodes/mod.rs
```diff
@@ -1527,8 +1526,42 @@ opcode_list! {
...
-    opcode OpUnknown214<0xd6, 1>(self, vm) Err(TxScriptError::InvalidOpcode(format!("{self:?}")))
+
+    opcode OpOutputCovenantId<0xd5, 1>(self, vm) {
+        if vm.flags.covenants_enabled {
+            match vm.script_source {
+                ScriptSource::TxInput{tx, ..} => {
+                    let [idx]: [i32; 1] = vm.dstack.pop_items()?;
+                    let idx = i32_to_usize(idx)?;
+                    let output = tx.outputs().get(idx).ok_or_else(|| TxScriptError::InvalidOutputIndex(idx as i32, tx.inputs().len()))?;
+                    let covenant_id = output.covenant.map(|c| c.covenant_id).unwrap_or(ZERO_HASH);
```
biryukovmaxim, line 1537: same comment regarding zerohash vs false

PR #926, opened 2026-03-25
Implement OpCheckSigFromStack and OpCheckSigFromStackECDSA

someone235:


crypto/txscript/src/lib.rs
```diff
@@ -663,73 +664,81 @@ impl<'a, T: VerifiableTransaction, Reused: SigHashReusedValues> TxScriptEngine<'
...
-                        }
+                self.check_schnorr_signature_for_msg_hash(sig_hash, key, sig)
+            }
+            _ => Err(TxScriptError::NotATransactionInput),
+        }
+    }
+
+    fn check_schnorr_signature_for_msg_hash(&mut self, msg_hash: Hash, key: &[u8], sig: &[u8]) -> Result<bool, TxScriptError> {
+        self.runtime_sig_op_counter.consume_sig_op()?;
+        if sig.len() != 64 {
```
biryukovmaxim: When we make signature from slice the same check is applied below, the error just needs to be moved below

crypto/txscript/src/lib.rs
```diff
@@ -663,73 +664,81 @@ impl<'a, T: VerifiableTransaction, Reused: SigHashReusedValues> TxScriptEngine<'
...
+            return Err(TxScriptError::SigLength(sig.len()));
+        }
+        Self::check_pub_key_encoding(key)?;
```
biryukovmaxim: Xonlypubkey from slice checks exactly the len, so it also may be removed

crypto/txscript/src/lib.rs
```diff
@@ -663,73 +664,81 @@ impl<'a, T: VerifiableTransaction, Reused: SigHashReusedValues> TxScriptEngine<'
...
+        let pk = secp256k1::XOnlyPublicKey::from_slice(key).map_err(TxScriptError::InvalidSignature)?;
+        let sig = secp256k1::schnorr::Signature::from_slice(sig).map_err(TxScriptError::InvalidSignature)?;
+        let secp_msg = secp256k1::Message::from_digest_slice(msg_hash.as_bytes().as_slice()).unwrap();
```
biryukovmaxim: There's a method from_digest, so unwrap may be omitted

crypto/txscript/src/lib.rs
```diff
@@ -663,73 +664,81 @@ impl<'a, T: VerifiableTransaction, Reused: SigHashReusedValues> TxScriptEngine<'
...
+            _ => Err(TxScriptError::NotATransactionInput),
+        }
+    }
+
+    fn check_ecdsa_signature_for_msg_hash(&mut self, msg_hash: Hash, key: &[u8], sig: &[u8]) -> Result<bool, TxScriptError> {
+        self.runtime_sig_op_counter.consume_sig_op()?;
+        if sig.len() != 64 {
+            return Err(TxScriptError::SigLength(sig.len()));
+        }
+        Self::check_pub_key_encoding_ecdsa(key)?;
```
biryukovmaxim, line 751: The same check exists below, not exactly the same but length bounds are considered during from_slice conversion

crypto/txscript/src/lib.rs
```diff
@@ -663,73 +664,81 @@ impl<'a, T: VerifiableTransaction, Reused: SigHashReusedValues> TxScriptEngine<'
...
+        let pk = secp256k1::PublicKey::from_slice(key).map_err(TxScriptError::InvalidSignature)?;
+        let sig = secp256k1::ecdsa::Signature::from_compact(sig).map_err(TxScriptError::InvalidSignature)?;
```
biryukovmaxim, line 753: It makes the check of length of the sig above pointless

crypto/txscript/src/opcodes/mod.rs
```diff
@@ -1563,8 +1563,29 @@ opcode_list! {
             Err(TxScriptError::InvalidOpcode(format!("{self:?}")))
         }
     }
-    opcode OpUnknown215<0xd7, 1>(self, vm) Err(TxScriptError::InvalidOpcode(format!("{self:?}")))
-    opcode OpUnknown216<0xd8, 1>(self, vm) Err(TxScriptError::InvalidOpcode(format!("{self:?}")))
+
+    opcode OpCheckSigFromStack<0xd7, 1>(self, vm) {
```
IzioDev, line 1567: unrelated: we can now use `/// my doc on how the stack is expected to be shaped`

crypto/txscript/src/lib.rs
```diff
@@ -675,13 +668,9 @@ impl<'a, T: VerifiableTransaction, Reused: SigHashReusedValues> TxScriptEngine<'
 
     fn check_schnorr_signature_for_msg_hash(&mut self, msg_hash: Hash, key: &[u8], sig: &[u8]) -> Result<bool, TxScriptError> {
         self.runtime_sig_op_counter.consume_sig_op()?;
-        if sig.len() != 64 {
-            return Err(TxScriptError::SigLength(sig.len()));
-        }
-        Self::check_pub_key_encoding(key)?;
         let pk = secp256k1::XOnlyPublicKey::from_slice(key).map_err(TxScriptError::InvalidSignature)?;
```
biryukovmaxim: That  should be another error

PR #936, opened 2026-03-28
Linux AppImage: RKStratum packaging, double-click terminal + XDG config; release as .tar.gz

LiveLaughLove13:
Double click functionality to run bridge/node on Linux. With terminal option.


WhosHaris:
Dg.mm.xm
Njzxrmnf xkk mkki...k
  A,j

Issue #929, opened 2026-03-25
decrypt_mnemonic uses String::from_utf8_unchecked on decrypted bytes


9292945:
In wallet/core/src/compat/gen1.rs (decrypt_mnemonic), the code converts AEAD-decrypted bytes into a String using unsafe { String::from_utf8_unchecked(decrypted) }.

This is Undefined Behavior if the decrypted plaintext is not valid UTF-8, because String requires a UTF-8 invariant. The code assumes the plaintext is always a UTF-8 mnemonic, but that invariant is not enforced at this boundary (it’s raw bytes coming from decryption of external wallet data). If the decrypted bytes ever contain invalid UTF-8 (corrupted file, unexpected format, or crafted input that still decrypts successfully), the program can exhibit UB (crash/memory corruption) before any later validation runs.

zelenevn:
@biryukovmaxim close plz

PR #941, opened 2026-03-31
fix(wallet): replace unsafe UTF-8 conversion in decrypt_mnemonic

zelenevn:
Description:
  ## Summary
  - Replace `String::from_utf8_unchecked` with `String::from_utf8` in `decrypt_mnemonic`
    to eliminate undefined behavior on invalid UTF-8 input
  - Invalid UTF-8 in decrypted bytes (e.g. corrupted wallet file, wrong password)
    now returns an error instead of triggering UB

  ## Motivation
  `from_utf8_unchecked` violates Rust's safety invariants if the decrypted payload
  is not valid UTF-8. This can happen with a corrupted wallet file or a wrong
  decryption key, turning a recoverable error into undefined behavior.


zelenevn:
Closes #929

zelenevn:
@biryukovmaxim Hi, please take a look. As an alternative, we could perform a preliminary validity check using `std::str::from_utf8`, but I think it's better to use the safe option.

biryukovmaxim (approved):

IzioDev (approved):
nice one (two really) 👏

PR #943, opened 2026-04-02
Kip21 impl

biryukovmaxim:


consensus/smt-store/src/processor.rs
```diff
@@ -0,0 +1,433 @@
...
+    /// read is happening — it drives `get_at`'s seek so non-canonical future
+    /// versions are skipped in O(log n) rather than scanned linearly.
+    pub fn get_node(
+        &self,
+        entity: BranchEntity,
+        target_blue_score: u64,
+        min_blue_score: u64,
+        mut is_canonical: impl FnMut(Hash) -> bool,
+    ) -> Option<Verified<Option<Node>>> {
+        if let Some((score, block_hash, value)) =
```
michaelsutton, line 116: consider adding a comment like this:

```
// Safe to return directly from cache: in the incremental-processing path every
// flushed version is also inserted into the cache, and eviction removes entries
// by lowest blue_score first. Therefore, for a given entity, no cached version
// can remain while a strictly newer-score version is missing from cache.
//
// Within a single blue_score bucket, eviction may drop some same-score siblings
// and keep others. This does not affect correctness: for a chain POV, at most
// one version at a given blue_score can be canonical, so the first canonical
// cached hit cannot skip a newer canonical version.
```

consensus/smt-store/src/processor.rs
```diff
@@ -0,0 +1,433 @@
...
+    /// Find the latest canonical lane version in `[min_blue_score, target_blue_score]`,
+    /// checking cache first then DB.
+    pub fn get_lane(
+        &self,
+        lane_key: LaneKey,
+        target_blue_score: u64,
+        min_blue_score: u64,
+        mut is_canonical: impl FnMut(Hash) -> bool,
+    ) -> Option<Verified<LaneTipHash>> {
+        if let Some((score, block_hash, value)) =
```
michaelsutton, line 138: copy or refer to the comment at `get_node`

consensus/smt-store/DESIGN.md
michaelsutton, line 1: The doc is out of date so should probably be deleted or rewritten

docs/kip21_ibd_smt_chunking.md
michaelsutton, line 1: should be deleted or rewritten and moved to an inner folder around the implemented logic

consensus/smt-store/src/streaming_import/mod.rs
```diff
@@ -0,0 +1,202 @@
...
+
+fn write_lane_versions(
+    stores: &SmtStores,
+    block_hash: BlockHash,
+    chunk: &[ImportLane],
+    lane_batch: &mut WriteBatch,
+    lane_batch_count: &mut usize,
+) -> Result<(), StreamError<StoreError>> {
+    for lane in chunk {
+        stores
```
michaelsutton, line 193: consider a comment (perhaps with better wording):

```
// This bypasses the lane cache on purpose. Cache hits in `SmtStores::get_lane`
// are treated as authoritative, so this is safe only because IBD SMT import
// begins after `SmtStores::clear_all()` has emptied both the DB stores and the
// caches. Thus there can be no stale cached entries disagreeing with the
// imported DB state. After import the caches remain cold, and reads fall back
// to DB until later incremental writes repopulate them.
```

---
Alternatively, the cache subtleties can be explained in a doc and all comments I suggested around this topic point to the doc.

consensus/smt-store/src/streaming_import/db_sink.rs
```diff
@@ -0,0 +1,106 @@
...
+    nodes_written: usize,
+}
+
+impl<'a> DbSink<'a> {
+    pub(crate) fn new(db: &'a DB, stores: &'a SmtStores, blue_score: u64, block_hash: BlockHash, max_batch_entries: usize) -> Self {
+        Self { db, stores, batch: WriteBatch::default(), batch_count: 0, max_batch_entries, blue_score, block_hash, nodes_written: 0 }
+    }
+
+    pub(crate) fn write_node(&mut self, bk: BranchKey, node: Node) -> StoreResult<()> {
+        self.stores.branch_version.put(
```
michaelsutton, line 39: same, add a comment or point to a doc:

```
// This bypasses the branch cache on purpose. Cache hits in `SmtStores::get_node`
// are treated as authoritative, so this is safe only because IBD SMT import
// begins after `SmtStores::clear_all()` has emptied both the DB stores and the
// caches. Thus there can be no stale cached branch versions disagreeing with
// the imported DB state. After import the caches remain cold, and reads fall
// back to DB until later incremental writes repopulate them.
```

consensus/smt-store/src/cache.rs
```diff
@@ -0,0 +1,311 @@
...
+}
+
+use crate::values::LaneTipHash;
+use kaspa_smt::store::Node;
+
+pub type BranchVersionCache = VersionedCache<BranchEntity, Option<Node>>;
+pub type LaneVersionCache = VersionedCache<Hash, LaneTipHash>;
+
+#[cfg(test)]
+mod tests {
```
michaelsutton, line 237: The cache fast path in `SmtStores::get_lane/get_node` assumes that, for each entity, retained cache entries always form a newest-score suffix of the written history. The current tests cover ordering and simple eviction, but I don’t see a test that really pins down that suffix invariant under cross-entity capacity pressure. It would be good to add a test that inserts interleaved versions for multiple entities, forces global lowest-score evictions, and verifies that each entity’s remaining cached versions are still exactly a newest suffix (so an authoritative cache hit cannot skip a newer canonical version).

michaelsutton (commented):
Sharing a review batch around the SMT cache area.

The cache logic itself is well-designed and solid. After tracing the write paths and eviction behavior, the core fast-path argument is sound: in the normal incremental-processing path, flushed versions are mirrored into the cache, and score-ordered eviction preserves the “newest relevant suffix” behavior needed for authoritative cache hits. The equal-score/canonical nuance is ok as well.

That said, this area is subtle enough that I think it wants a bit more help for future readers. In particular, the safety story depends on non-obvious invariants around cache population, eviction order, canonical lookup, and the import path bypassing caches after `clear_all()`. I’d like to see a bit more coverage here in comments/docs/tests, especially one test that more explicitly pins down the per-entity newest-suffix invariant under cross-entity capacity pressure.

utils/src/serde_bytes/de.rs
```diff
@@ -57,8 +57,9 @@ impl<'de, T: FromHex> serde::de::Visitor<'de> for FromHexVisitor<'de, T> {
         FromHex::from_hex(v).map_err(serde::de::Error::custom)
     }
 
+    #[cfg(feature = "std")]
     #[inline]
-    fn visit_string<E>(self, v: String) -> Result<Self::Value, E>
+    fn visit_string<E>(self, v: std::string::String) -> Result<Self::Value, E>
```
michaelsutton: seems like this can be `alloc` if we add `alloc = ["serde/alloc"]` under cargo features

```suggestion
    #[cfg(any(feature = "std", feature = "alloc"))]
    #[inline]
    fn visit_string<E>(self, v: alloc::string::String) -> Result<Self::Value, E>
```

utils/src/serde_bytes/de.rs
```diff
@@ -83,8 +84,9 @@ impl<'de, T: FromHex> serde::de::Visitor<'de> for FromHexVisitor<'de, T> {
         FromHex::from_hex(str).map_err(serde::de::Error::custom)
     }
 
+    #[cfg(feature = "std")]
     #[inline]
-    fn visit_byte_buf<E>(self, v: Vec<u8>) -> Result<Self::Value, E>
+    fn visit_byte_buf<E>(self, v: std::vec::Vec<u8>) -> Result<Self::Value, E>
```
michaelsutton: seems like this can be `alloc` if we add `alloc = ["serde/alloc"]` under cargo features

```suggestion
    #[cfg(any(feature = "std", feature = "alloc"))]
    #[inline]
    fn visit_byte_buf<E>(self, v: alloc::vec::Vec<u8>) -> Result<Self::Value, E>
```

crypto/hashes/src/lib.rs
```diff
@@ -1,19 +1,32 @@
...
+    hash::{Hash as CoreHash, Hasher as CoreHasher},
     str::{self, FromStr},
 };
+use kaspa_utils::{
+    hex::{FromHex, ToHex},
+    serde_impl_deser_fixed_bytes_ref, serde_impl_ser_fixed_bytes_ref,
+};
 use wasm_bindgen::prelude::*;
+
+// export libraries so it's possible to use macroses
```
michaelsutton:
```suggestion
// export libraries so it's possible to use macros
```

michaelsutton (commented):
small review batch on nostd changes

consensus/src/processes/transaction_validator/tx_validation_in_isolation.rs
```diff
@@ -155,7 +155,16 @@ fn check_transaction_output_value_ranges(tx: &Transaction) -> TxResult<()> {
...
 fn check_transaction_subnetwork(tx: &Transaction) -> TxResult<()> {
-    if tx.is_coinbase() || tx.subnetwork_id.is_native() { Ok(()) } else { Err(TxRuleError::SubnetworksDisabled(tx.subnetwork_id)) }
+    const ZEROES: &[u8; 19] = &[0; 19];
+
+    match tx.subnetwork_id.as_bytes() {
+        // Native and coinbase (reserved) subnetwork IDs are always allowed
+        [NativeSubnetwork::FIRST_BYTE, rest @ ..] | [CoinbaseSubnetwork::FIRST_BYTE, rest @ ..] if rest == ZEROES => Ok(()),
+        // Post-covenants HF (KIP-21): any non-reserved subnetwork ID is allowed as a lane selector.
+        // Reserved IDs (`x || 0x00..0x00`) are excluded - they are for system usage by consensus.
+        [_x, rest @ ..] if tx.version >= TX_VERSION_POST_COV_HF && rest != ZEROES => Ok(()),
```
michaelsutton: wdyt about allowing only 4 bytes here and forcing last 16 bytes to be zeros? ~`2^32` seems enough as a subnet namespace

consensus/seq-commit/src/types.rs
```diff
@@ -0,0 +1,55 @@
...
+    pub payload: &'a [u8],
+}
+
+/// Input for computing an SMT leaf hash for an active lane.
+///
+/// `leaf_payload = lane_key(32) || lane_tip_hash(32) || le_u64(blue_score)`.
+#[derive(Clone, Copy, Debug)]
+pub struct SmtLeafInput<'a> {
+    pub lane_key: &'a Hash,
+    pub lane_tip: &'a Hash,
```
michaelsutton, line 41: `lane_tip` is a direct hash of lane key. perhaps we can avoid it here in the leaf

consensus/seq-commit/src/hashing.rs
```diff
@@ -0,0 +1,401 @@
...
+    hasher.finalize()
+}
+
+/// Compute the activity digest for a lane's transactions in a mergeset block.
+///
+/// Merkle root over the activity leaves using `H_seq` (SeqCommitMerkleBranch).
+/// For a single leaf, hashes `H_seq(leaf || ZERO_HASH)`.
+#[inline]
+pub fn activity_digest_lane(leaves: impl ExactSizeIterator<Item = Hash>) -> Hash {
+    calc_merkle_root_with_hasher::<SeqCommitMerkleBranch, true>(leaves)
```
michaelsutton: not totally sure about this comment, but since all activity trees (including miner payloads) are per domain, it seems ok to go back to normal Merkle convention and represent a tree with one leaf as the leaf itself

(which means the generic `USE_BRANCH_HASHER_FOR_SINGLE` arg can be removed altogether)

consensus/src/pipeline/pruning_processor/processor.rs
```diff
@@ -577,6 +578,17 @@ impl PruningProcessor {
...
 
+        // Prune SMT lane/branch version stores and score index.
+        // Cutoff = pp.blue_score − finality_depth: the lowest score any future
+        // SMT lookup could reference (all lookups use min_blue_score = current − finality_depth,
+        // and all future blocks are above the pruning point).
+        {
+            let pp_header = self.headers_store.get_header(new_pruning_point).unwrap();
+            let smt_cutoff = pp_header.blue_score.saturating_sub(self.config.params.finality_depth());
+            info!("SMT pruning: cutoff_blue_score={}", smt_cutoff);
+            self.smt_stores.prune(&self.db, smt_cutoff);
```
michaelsutton, line 592: off-by-one: prune deletes the cutoff blue score itself although it's part of the pp active window

consensus/src/pipeline/virtual_processor/processor.rs
```diff
@@ -547,29 +575,163 @@ impl VirtualStateProcessor {
...
+                continue;
+            }
+            for lk in entry.data().iter() {
+                // Check if this lane has a newer canonical version within [curr_min, parent].
+                // target=parent filters anticone entries at (parent, current] at the seek level.
+                let has_newer = self.smt_stores.get_lane(*lk, read_bounds, |bh| self.is_smt_canonical(bh, selected_parent)).is_some();
+
+                if !has_newer {
+                    proc.expire_lane(*lk);
+                    expired += 1;
```
michaelsutton, line 797: if lane L was touched twice in `[parent-F, current-F-1]` it will get counted twice here (so metadata will be wrong). 

you should either have `proc.expire_lane` return a boolean indicating a new hit, or maintain a local set (in which case you can avoid querying `has_newer` for the second item)

protocol/flows/src/flow_context.rs
```diff
@@ -739,7 +739,8 @@ impl ConnectionInitializer for FlowContext {
 
         // Register all flows according to version
         let (flows, applied_protocol_version) = match peer_version.protocol_version {
-            v if v >= PROTOCOL_VERSION => (v8::register(self.clone(), router.clone(), PROTOCOL_VERSION), PROTOCOL_VERSION),
+            v if v >= PROTOCOL_VERSION => (v9::register(self.clone(), router.clone(), PROTOCOL_VERSION), PROTOCOL_VERSION),
+            9 => (v9::register(self.clone(), router.clone(), 9), 9),
```
michaelsutton: protocol version is already 9 on this branch, so this line is redundant. In general, I expect the P2P flow to be upgraded before mainnet (cleanup + pass chain segment [P-F, P] in a more correct way). So we should keep that in mind

consensus/src/consensus/storage.rs
```diff
@@ -224,6 +230,13 @@ impl ConsensusStorage {
         let virtual_stores =
             Arc::new(RwLock::new(VirtualStores::new(db.clone(), lkg_virtual_state.clone(), utxo_set_builder.build())));
 
+        // SMT stores (KIP-21)
+        // TODO: make cache capacities configurable via consensus params
+        let smt_stores = Arc::new(SmtStores::new(db.clone(), 500_000, 50_000));
```
michaelsutton, line 235: Any good reason to think that branches require 10x cache?

consensus/src/model/stores/virtual_state.rs
```diff
@@ -55,15 +57,76 @@ impl VirtualState {
...
+            accepted_id_digests,
             mergeset_rewards,
             mergeset_non_daa,
         }
     }
 
     pub fn from_genesis(genesis: &GenesisBlock, ghostdag_data: GhostdagData, covenants_activation: ForkActivation) -> Self {
-        let accepted_tx_digests = if covenants_activation.is_active(genesis.daa_score) {
-            genesis.build_genesis_transactions().iter().map(|tx| tx.seq_commit_digest()).collect()
+        let accepted_id_digests = if covenants_activation.is_active(genesis.daa_score) {
```
michaelsutton: non critical comment: it feels wrong that we repeat all this logic here. `VirtualState::from_genesis` is only called form virtual processor. I think it makes more sense to prep the seqcommit/accepted_id_digests in the processor, using existing methods, and pass them here directly

consensus/src/consensus/storage.rs
michaelsutton: Also, the TODO should probably be addressed now, it's an easy one

biryukovmaxim: Every lane update can lead up to 256 updates of branches. As well as sibling update(activation or expiration) may change branches but doesn't affect not changed leaves

consensus/seq-commit/src/hashing.rs
```diff
@@ -0,0 +1,398 @@
...
+    hasher.update(payload);
+    hasher.finalize()
+}
+
+/// Compute a miner payload leaf: `H_miner_payload_leaf(block_hash || blue_work_bytes || H_miner_payload(payload))`.
+#[inline]
+pub fn miner_payload_leaf(input: &MinerPayloadLeafInput<'_>) -> Hash {
+    let payload_h = miner_payload_hash(input.payload);
+    let mut hasher = SeqCommitMinerPayloadLeaf::new();
+    hasher.update(input.block_hash).update(input.blue_work_bytes).update(payload_h);
```
michaelsutton: change `input.blue_work_bytes` from bytes to `blue_work: &'a BlueWorkType` and
replace `.update(input.blue_work_bytes)` with `.write_blue_work(..)`

the reason is that current `BlueWorkType` (=Uint192) is not a consensus limit. Header blue work can be with arbitrary bigint length (it counts the total amount of pow the net had; you don't want a hard cap on that). So seqcommit hashing should be compatible with the same flexibility.

protocol/flows/src/v9/request_pruning_point_smt_state.rs
```diff
@@ -0,0 +1,146 @@
...
+        }?;
+        drop(session_for_reader);
+        let reader_handle = tokio::task::spawn_blocking(move || -> Result<u64, ConsensusError> {
+            let mut count: u64 = 0;
+            let mut batch: Vec<ImportLane> = Vec::with_capacity(SMT_CHUNK_SIZE);
+            for item in stream {
+                batch.push(item?);
+                if batch.len() == SMT_CHUNK_SIZE {
+                    count += batch.len() as u64;
+                    if tx.blocking_send(std::mem::take(&mut batch)).is_err() {
```
michaelsutton, line 87:
```suggestion
                    if tx.blocking_send(std::mem::replace(&mut batch, Vec::with_capacity(SMT_CHUNK_SIZE))).is_err() {
```

(currently capacity only helps the first chunk)

michaelsutton: actually you already call `batch.reserve(SMT_CHUNK_SIZE)` below, so it's just a mater of taste

consensus/src/consensus/mod.rs
```diff
@@ -1123,6 +1130,115 @@ impl ConsensusApi for Consensus {
...
+        lanes_root: Hash,
+        payload_and_ctx_digest: Hash,
+        expected_lane_count: u64,
+        lane_batches: ImportLaneBatchIterator,
+    ) -> PruningImportResult<()> {
+        use kaspa_hashes::ZERO_HASH;
+        use kaspa_smt_store::streaming_import::streaming_import;
+
+        let pp_header = self.storage.headers_store.get_header(new_pruning_point).unwrap();
+        let result = streaming_import(
```
michaelsutton: `streaming_import` uses a `par_iter` internally, so it's better to call it under `self.virtual_processor.install`
(see `append_imported_pruning_point_utxos` above where we do the same). 

The reason is that we try avoiding the creation of the default rayon thread-pool (bcs we create 2 other ones anyway). Virtual processor isn't active yet on pp import stages, so mind as well use it

michaelsutton (approved):
LGTM. This is a seriously impressive piece of work @biryukovmaxim.

It implements a complicated and powerful component, but the code remains sound and well designed throughout, with well-structured layering, correct crate-boundary choices, and overall elegant, beautiful work.

PR #944, opened 2026-04-05
Fix non-deterministic tests

someone235:
This PR hardens a set of flaky test paths that were failing under repeated workspace runs and parallel integration execution.

  The changes focus on:

  - reducing file descriptor pressure in integration tests
  - making test notification waits more tolerant of slower environments
  - avoiding daemon port allocation races in integration tests
  - preventing a debug-only subscription counter underflow from crashing tests

  ## Changes

  - Make integration tests use TEST_FD_LIMIT instead of using `fd_budget::limit()`
  - cap daemon FD usage in testing/integration/src/rpc_tests.rs and testing/integration/src/tasks/daemon.rs so one test daemon cannot exhaust the shared FD pool
  - reserve integration daemon ports through an in-process registry in testing/integration/src/common/daemon.rs to reduce collisions between concurrently started test daemons
  - increase notifier-related test wait windows in notify/src/notifier.rs and testing/integration/src/common/utils.rs for slower CI environments
  - relax the RwLock reentrance test timing in utils/src/sync/rwlock.rs to reduce scheduler-related flakiness
  - make UtxosChangedSubscription drop accounting in notify/src/subscription/single.rs avoid underflowing the debug counter during teardown-heavy test runs

  ## Why

  Repeated `cargo test --workspace` runs exposed failures caused by shared-resource contention rather than product logic regressions:

  - integration tests do not run with the `test` feature on, so `fd_budget::limit()` skips the `if #[cfg(test)]` branch and ends up returning the whole OS capacity.
  - daemon startup could race on randomly assigned ports
  - notification-based assertions were too aggressive for slower runs
  - debug-only subscription accounting could panic during teardown

  This PR makes those test paths more deterministic without changing production behavior.

### DIFF
--- a/testing/integration/src/common/daemon.rs
+++ b/testing/integration/src/common/daemon.rs
@@ -6,7 +6,7 @@ use kaspa_grpc_server::service::GrpcService;
 use kaspa_notify::subscription::context::SubscriptionContext;
 use kaspa_rpc_core::notify::mode::NotificationMode;
 use kaspa_rpc_service::service::RpcCoreService;
-use kaspa_utils::triggers::Listener;
+use kaspa_utils::{networking::ContextualNetAddress, triggers::Listener};
 use kaspad_lib::{args::Args, daemon::create_core_with_runtime};
 use parking_lot::RwLock;
 use std::{ops::Deref, sync::Arc, time::Duration};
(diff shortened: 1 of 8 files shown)


testing/integration/src/common/daemon.rs
```diff
@@ -13,6 +13,9 @@ use std::{ops::Deref, sync::Arc, time::Duration};
 use tempfile::TempDir;
 
 use kaspa_grpc_client::ClientPool;
+use std::{collections::HashSet, sync::LazyLock};
+
+static RESERVED_PORTS: LazyLock<parking_lot::Mutex<HashSet<u16>>> = LazyLock::new(|| parking_lot::Mutex::new(HashSet::new()));
```
biryukovmaxim: nextest runs tests in many binaries in parallel, so it needs to track unused port via system info or shared memory. also lazylock seems redundant imo, static mutex will do the same

utils/src/fd_budget.rs
```diff
@@ -61,10 +61,14 @@ pub fn try_set_fd_limit(limit: u64) -> std::io::Result<u64> {
     }
 }
 
+// Many tests can be run in parallel, and each of them may acquire some FDs, so we set a lower limit for tests to avoid hitting the actual OS limit.
+// Note: Integration tests need to explicitly use this constant and not `limit()`, since they set `#[cfg(test)]` to false.
+pub const TEST_FD_LIMIT: i32 = 100;
```
biryukovmaxim: Iirc mac has 256 fd by default if not changed

biryukovmaxim (approved):

PR #945, opened 2026-04-06
Covpp reset2 blake

biryukovmaxim:


crypto/txscript/src/opcodes/mod.rs
```diff
@@ -3407,6 +3441,116 @@ mod test {
...
+            vec![TestCase {
+                code: opcodes::OpBlake2bWithKey::empty().expect("Should accept empty"),
+                init: vec![b"test data".to_vec(), vec![]],
+                dstack: vec![b"\xea\xb9\x49\x77\xa1\x77\x91\xd0\xc0\x89\xfe\x9e\x39\x32\x61\xb3\xab\x66\x7c\xf0\xe8\x45\x66\x32\xa8\x42\xd9\x05\xc4\x68\xcf\x65".to_vec()],
+            }],
+            flags,
+        );
+
+        // 1-byte key
+        run_success_test_cases_with_flags(
```
someone235: I think we should also add a success test for key of max length

someone235 (changes requested):
In addition to the below comment, I also think that it can be much more compact and readable to move `test_opblake2b_with_key` etc to `crypto/txscript/test-data/script_tests_covenants.json`

someone235 (approved):

PR #951, opened 2026-04-10
fix(wasm): mempool entries request args are required

IzioDev:
- adjusted wasm sdk types; IGetMempoolEntriesRequest require the two bool properties to be present
- added mempool entries wasm example

### DIFF
--- a/rpc/core/src/wasm/message.rs
+++ b/rpc/core/src/wasm/message.rs
@@ -1043,13 +1043,13 @@ declare! {
     IGetMempoolEntriesRequest,
     r#"
     /**
-     *
-     *
      * @category Node RPC
      */
     export interface IGetMempoolEntriesRequest {
-        includeOrphanPool? : boolean;
-        filterTransactionPool? : boolean;
+        /// Whether or not to include the orphan pool (transactions which inputs are not known at this time)
+        includeOrphanPool: boolean;
+        /// Whether or not to filter out the transaction pool
+        filterTransactionPool: boolean;
     }
     "#,
 }
--- a/rpc/core/src/wasm/message.rs
+++ b/rpc/core/src/wasm/message.rs
@@ -1081,13 +1081,13 @@ declare! {
     IGetMempoolEntriesRequest,
     r#"
     /**
-     *
-     *
      * @category Node RPC
      */
     export interface IGetMempoolEntriesRequest {
-        includeOrphanPool? : boolean;
-        filterTransactionPool? : boolean;
+        /// Whether or not to include the orphan pool (transactions which inputs are not known at this time)
+        includeOrphanPool: boolean;
+        /// Whether or not to filter out the transaction pool
+        filterTransactionPool: boolean;
     }
     "#,
 }
--- a/rpc/core/src/wasm/message.rs
+++ b/rpc/core/src/wasm/message.rs
@@ -1084,9 +1084,9 @@ declare! {
      * @category Node RPC
      */
     export interface IGetMempoolEntriesRequest {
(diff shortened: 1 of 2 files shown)


demisrael:
Thanks — the substance of the fix is right. The Rust [`GetMempoolEntriesRequest:L514-L518`](https://github.com/kaspanet/rusty-kaspa/blob/3e465f017ee26aa5614a3339d1a4895b79b486e8/rpc/core/src/model/message.rs#L514-L518) struct holds both fields as plain `bool` with no serde default, so the previous `?`-optional TS declaration was misleading: any caller that omitted them got a serde `missing field` error at runtime. Tightening the TS type matches the actual contract.

A few things worth folding in before merge:

1. The same misdeclaration sits one block down at [`IGetMempoolEntriesByAddressesRequest:L1090-L1091`](https://github.com/kaspanet/rusty-kaspa/blob/3e465f017ee26aa5614a3339d1a4895b79b486e8/rpc/core/src/wasm/message.rs#L1090-L1091). The Rust target [`GetMempoolEntriesByAddressesRequest:L1670-L1675`](https://github.com/kaspanet/rusty-kaspa/blob/3e465f017ee26aa5614a3339d1a4895b79b486e8/rpc/core/src/model/message.rs#L1670-L1675) has the same plain-`bool` shape, so the same `?` removal applies. Same diff, same justification — easier to fix here than as a follow-up.

2. The new field comments use `///`, which TypeScript reads as a regular line comment, not as a doc comment (triple-slash is only meaningful as a directive at the top of a file). Editors won't surface them on hover. The rest of `rpc/core/src/wasm/message.rs` uses JSDoc `/** … */` — `IGetMempoolEntriesRequest`'s own outer block does too. Switching the new field comments to JSDoc keeps the docs visible:

   ```ts
   export interface IGetMempoolEntriesRequest {
       /** Include the orphan pool (transactions whose inputs are not yet known). */
       includeOrphanPool: boolean;
       /** Filter the transaction pool out of the response. */
       filterTransactionPool: boolean;
   }
   ```

3. Tiny typo in the example: "transations" → "transactions" ([`mempool-entries.js:L24`](https://github.com/kaspanet/rusty-kaspa/blob/3e465f017ee26aa5614a3339d1a4895b79b486e8/wasm/examples/nodejs/javascript/general/mempool-entries.js#L24)).

4. Optional — the example logs `Connecting to ${rpc.url}` after `await rpc.connect()` returns. Same misorder exists in [`get-virtual-chain-v2.js:L21`](https://github.com/kaspanet/rusty-kaspa/blob/3e465f017ee26aa5614a3339d1a4895b79b486e8/wasm/examples/nodejs/javascript/general/get-virtual-chain-v2.js#L21) and [`get-balances-by-addresses.js:L28`](https://github.com/kaspanet/rusty-kaspa/blob/3e465f017ee26aa5614a3339d1a4895b79b486e8/wasm/examples/nodejs/javascript/general/get-balances-by-addresses.js#L28), so it's consistent with the repo, but it reads oddly in a tutorial example. Either log "Connecting to …" before `connect()` or rename the post-connect log to "Connected to …".

5. Optional — worth noting in the PR description that this is a TS-types-only breaking change. Runtime was already broken when fields were omitted, so no JS code is newly affected, but TS consumers will see new `tsc` errors after upgrading.

coderofstuff:
The comments by @demisrael look correct. Up to you if you want to address all instances of the other incorrect optionals for includeOrphanPool and filterTransactionPool. Do fix the typo in the sample you introduced.

michaelsutton (approved):

PR #953, opened 2026-04-12
Zk sdk

saefstroem:
The goal with this is PR is to reduce the knowledge requirement factor when an actor needs to deal with UTXO based ZK systems that are largely non-existent in the same relative scale as others. This is achieved by introducing a a wrapper around the script builder: `R0ScriptBuilder` which allows one to convert any R0 proof whether this is a SuccinctReceipt or Groth16Receipt into a native Kaspa script, reducing the workload for protocols to migrate onto Kaspa and barrier for adoption.

### R0 Script Builder

The off-chain SDK that generates the **redeem script** (lock -> P2SH address) and **sig script** (spend) for locking a UTXO behind a RISC Zero proof. Wraps `ScriptBuilder` so callers never need to touch raw opcodes and bind them to certain opcodes. As you cannot just simply invoke the OpZkPrecompile opcode. 

### Design
- **Typestate machine**: `Unbounded -> commit_to_{groth16|succinct} -> Bounded -> finalize_with_proof -> FinalizedR0Script{sig_script, redeem_script}`. 
- **commit vs finalize = P2SH commit/reveal**: commit builds the lock, finalize builds the spend and returns both.
- **Groth16**: redeem script reconstructs R0's `ReceiptClaim` digest in raw opcodes from image_id + journal, with R0's fixed VK. **Succinct**: commit just pushes image_id/control_id/hashfn.
- Finalize takes native `Groth16Receipt`/`SuccinctReceipt` (borsh-encoded in WASM).

### API

|          | Native (Rust)                                                                 | WASM (JS)                                                                 |
| -------- | ----------------------------------------------------------------------------- | ------------------------------------------------------------------------ |
| construct | `new()` / `with_flags(EngineFlags)`                                          | `new R0ScriptBuilder({flags})`                                           |
| commit   | `commit_to_groth16(image_id)` · `commit_to_succinct(image_id, control_id, hash_fn?)` | `commitToGroth16(imageId)` · `commitToSuccinct(imageId, controlId, hashFnId?)` |
| finalize | `finalize_with_proof(receipt, journal) -> FinalizedR0Script`                  | `finalizeWithGroth16Proof(...)` · `finalizeWithSuccinctProof(...)`        |
| shared   | `script()` · `script_mut()` · `drain()`                                       | `script()` · `drain()`; `FinalizedR0Script.sigScript`/`.redeemScript`     |

### Expected Flow
commit -> hash `redeemScript` into P2SH address (fund it) -> finalize -> set `sigScript` on redeem-tx input -> submit.


IzioDev:
Nice! Should we start investigating wasm bridging for r0 script builder? Since the SDK will mostly (ark seems no-std compat) depends on types (and no native calls?), it should be doable, what do you think?

saefstroem:
> Nice! Should we start investigating wasm bridging for r0 script builder? Since the SDK will mostly (ark seems no-std compat) depends on types (and no native calls?), it should be doable, what do you think?

hey, yeah I think this is a good idea. let me check on this on my next pass on this

crypto/txscript/src/zk_precompiles/points/mod.rs
```diff
@@ -0,0 +1,57 @@
...
+
+impl<'input> PointFromBytes<'input> for G1 {
+    type Input = Vec<Vec<u8>>;
+
+    /// Deserialize an element over the G1 group from bytes in big-endian format
+    fn from_bytes(bytes: &Self::Input) -> Result<G1, PointError> {
+        if bytes.len() != 2 {
+            return Err(PointError::MalformedG1);
+        }
+        let g1_affine: Vec<u8> = bytes[0].iter().rev().chain(bytes[1].iter().rev()).cloned().collect();
```
biryukovmaxim, line 21: why shout it be 2 dimensional  reversed vectors?

crypto/txscript/src/zk_precompiles/points/mod.rs
```diff
@@ -0,0 +1,57 @@
...
+            return Err(PointError::MalformedG2);
+        }
+        let g2_affine: Vec<u8> = bytes[0][1]
+            .iter()
+            .rev()
+            .chain(bytes[0][0].iter().rev())
+            .chain(bytes[1][1].iter().rev())
+            .chain(bytes[1][0].iter().rev())
+            .cloned()
+            .collect();
```
biryukovmaxim, line 41: why shout it be 2 dimensional  reversed vectors?

crypto/txscript/src/zk_precompiles/risc0/zk_to_script/groth16/mod.rs
```diff
@@ -0,0 +1,77 @@
...
+    /// Converts a Groth16Receipt into a Kaspa script.
+    /// This script unlocks the UTXO if the verification of the receipt
+    /// succeeds.
+    pub fn from_groth<Claim: Digestible + Clone>(receipt: &Groth16Receipt<Claim>) -> Result<ScriptBuilder> {
+        let mut params = Groth16ReceiptVerifierParameters::default();
+        let seal = &receipt.seal;
+        let digested_claim = receipt.claim.digest::<sha::Impl>();
+        let (a0, a1) = split_digest_bytes(params.control_root);
+        let (c0, c1) = split_digest_bytes(digested_claim);
+        let id_bn254 = to_fixed_array(params.bn254_control_id.as_bytes());
```
biryukovmaxim, line 48: digest impls `Into` into [u8;32]. method seems redundant

crypto/txscript/src/zk_precompiles/risc0/zk_to_script/groth16/vk.rs
```diff
@@ -0,0 +1,67 @@
...
+        G1::from_bytes(&vec![from_u256(IC0_X)?, from_u256(IC0_Y)?])?.0,
+        G1::from_bytes(&vec![from_u256(IC1_X)?, from_u256(IC1_Y)?])?.0,
+        G1::from_bytes(&vec![from_u256(IC2_X)?, from_u256(IC2_Y)?])?.0,
+        G1::from_bytes(&vec![from_u256(IC3_X)?, from_u256(IC3_Y)?])?.0,
+        G1::from_bytes(&vec![from_u256(IC4_X)?, from_u256(IC4_Y)?])?.0,
+        G1::from_bytes(&vec![from_u256(IC5_X)?, from_u256(IC5_Y)?])?.0,
+    ];
+
+    Ok(ark_groth16::VerifyingKey::<Bn254> { alpha_g1, beta_g2, gamma_g2, delta_g2, gamma_abc_g1 })
+}
```
biryukovmaxim, line 67: the key is constant, why is it calculated in runtime. why does it require vector allocations and conversions from string

biryukovmaxim, line 67: there is no test that ensures its the same calculatuon of verifying key as upstream crate has

crypto/txscript/src/zk_precompiles/risc0/rcpt.rs
```diff
@@ -45,6 +45,19 @@ impl TryFrom<u8> for HashFnId {
     }
 }
 
+impl TryFrom<&String> for HashFnId {
```
biryukovmaxim, line 48: &str

biryukovmaxim, line 48: fromStr seems a better trait candidate than tryfrom

biryukovmaxim (changes requested):

crypto/txscript/zk-sdk/src/points/mod.rs
```diff
@@ -0,0 +1,57 @@
...
+
+impl<'input> PointFromBytes<'input> for G1 {
+    type Input = Vec<Vec<u8>>;
+
+    /// Deserialize an element over the G1 group from bytes in big-endian format
+    fn from_bytes(bytes: &Self::Input) -> Result<G1, PointError> {
+        if bytes.len() != 2 {
+            return Err(PointError::MalformedG1);
+        }
+        let g1_affine: Vec<u8> = bytes[0].iter().rev().chain(bytes[1].iter().rev()).cloned().collect();
```
saefstroem: r0 stores in BE order whilst ark expects LE, but I agree here that the fn name should be more specific to r0, not generic `from_bytes`

crypto/txscript/zk-sdk/src/points/mod.rs
```diff
@@ -0,0 +1,57 @@
...
+            return Err(PointError::MalformedG2);
+        }
+        let g2_affine: Vec<u8> = bytes[0][1]
+            .iter()
+            .rev()
+            .chain(bytes[0][0].iter().rev())
+            .chain(bytes[1][1].iter().rev())
+            .chain(bytes[1][0].iter().rev())
+            .cloned()
+            .collect();
```
saefstroem: see above

crypto/txscript/src/zk_precompiles/risc0/zk_to_script/groth16/mod.rs
saefstroem: fair point

crypto/txscript/src/zk_precompiles/risc0/zk_to_script/groth16/vk.rs
saefstroem: There is no way to extract inner verifying key from r0 repo its only public to the crate see: `pub struct VerifyingKey(pub(crate) VerifyingKey<Bn<Config>>)` in `risc0_groth16::verifier`

saefstroem: There is an indirect test, there is an r0 proof that verifies using this key.

saefstroem: But ill make it better

saefstroem: Agreed, let me fix this

crypto/txscript/src/zk_precompiles/risc0/rcpt.rs
saefstroem: why what if you pass incorrect string? then everything will panic

saefstroem: Agreed

saefstroem:
There are some elements which I have identified to be missing, most importantly the generation of safe bounded Groth16 script by image id.

Cargo.toml
```diff
@@ -75,7 +75,7 @@ license = "ISC"
 repository = "https://github.com/kaspanet/rusty-kaspa"
 edition = "2024"
 include = [
-    "src/**/*.rs",
+    "src/*risc0-zkvm.workspace*/*.rs",
```
michaelsutton: is this intentional or a leftover? why does that path need special treatment here?

saefstroem: You are right must be a leftover, fixing.

crypto/txscript/zk-sdk/src/zk_to_script/wasm/mod.rs
```diff
@@ -0,0 +1,97 @@
...
+            InnerState::BoundedGroth16(b) => b.drain(),
+            InnerState::BoundedSuccinct(b) => b.drain(),
+            InnerState::Taken => Vec::new(),
+        }
+    }
+}
+
+pub(super) fn into_array_32(bytes: Vec<u8>, name: &'static str) -> Result<[u8; 32]> {
+    bytes.as_slice().try_into().map_err(|_| Error::custom(format!("{name} must be 32 bytes")))
+}
```
IzioDev, line 74: we can use `Hash::try_cast_from(JSValue)` instead of defining a function. (or use the reverse `js_value.try_into_owned()`

example in code:
```rust
let transaction_id: TransactionId = object.get_value("transactionId")?.try_into_owned()?;
```

note: it may only supports `HexString` out of the box, but that's how we deal with hashes in general anyway.

crypto/txscript/zk-sdk/src/zk_to_script/wasm/mod.rs
```diff
@@ -0,0 +1,97 @@
...
+/// express the native compile-time type-state transitions.
+///
+/// Flow:
+///   1. `new()` — unbounded.
+///   2. `commitToGroth16(imageId)` *or* `commitToSuccinct(imageId, controlId, hashFnId?)` — bounded.
+///   3. `finalizeWithGroth16Proof(receipt, journalHash)` *or* `finalizeWithSuccinctProof(receipt, journal)` — finalized hex bytes.
+///
+/// Calling a method in the wrong state returns an error.
+#[wasm_bindgen(inspectable)]
+pub struct R0ScriptBuilder {
```
IzioDev: Should the name be more generic than R0 at calling surface, like `ZkScriptBuilder`?

My main concern is creating a breaking change (or need to maintain a legacy) if we support building scripts outside of `R0`. If we agree with this, I also suggest to name `new` -> `new_r0` or other.

wasm/examples/nodejs/javascript/zkproof/groth16.js
```diff
@@ -1,238 +1,219 @@
...
-const PRIVATE_KEY =
-  "b99d75736a0fd0ae2da658959813d680474f5a740a9c970a7da867141596178f";
+const { PrivateKey, RpcClient, ScriptBuilder, Opcodes,
+    payToScriptHashScript, addressFromScriptPublicKey,
+    createTransaction, signTransaction, 
+    Encoding} = require('./kaspa');
+
+// Configuration  
+const NETWORK_ID = 'devnet';
+const RPC_URL = 'ws://127.0.0.1:17610';
```
IzioDev, line 8: I suggest to change to either `testnet` or `mainnet` (small preference for testnet), and remove the URL, let's the `RpcClient` uses `Resolver`

wasm/examples/nodejs/javascript/zkproof/groth16.js
IzioDev, line 1: I'm questioning the need for this example. Shouldn't examples only showcase the `Builder` path instead? What do you think?

wasm/examples/nodejs/javascript/zkproof/data/direct/succinct.image.hex
IzioDev, line 1: Is there a reason why we have both hex files flat at `wasm/examples/nodejs/javascript/zkproof/` level, and files in `wasm/examples/nodejs/javascript/zkproof/builder_data`?

I'd suggest to put them all in sub-dir to increase clarity

crypto/txscript/zk-sdk/src/zk_to_script/wasm/mod.rs
saefstroem: sure agreed

wasm/examples/nodejs/javascript/zkproof/groth16.js
saefstroem: These examples were done on devnet, ive had issues with testnet before so it was more convenient for me to do them via devnet.

wasm/examples/nodejs/javascript/zkproof/groth16.js
saefstroem: This example existed before the zk script builder existed and I think it is important to showcase both, advanced users might want to see how to use the raw internals

Cargo.lock
```diff
@@ -172,15 +172,49 @@ dependencies = [
  "password-hash",
 ]
 
+[[package]]
+name = "ark-bn254"
+version = "0.5.0"
```
michaelsutton, line 177: seems like the new dep on r0 g16 also introduces another version of ark
any way to avoid this? cc @biryukovmaxim

biryukovmaxim: not really. currently lock has 2 different versions of ark-bn254.
the only way to switch to 0.6.0 - use fork or ask risc0 to bump deps

crypto/txscript/zk-sdk/src/zk_to_script/fragments/groth16.rs
```diff
@@ -0,0 +1,213 @@
...
+
+/// Pushes the r0 groth16 witness material into a caller-owned builder.
+///
+/// Pre-stack:  `[..., journal_hash]`
+/// Post-stack: `[..., journal_hash, compressed_proof]`
+///
+/// Only the compressed proof is pushed, the `journal_hash` is caller-owned and
+/// must already be on the stack (the caller decides how it gets there: a
+/// constant for a one-time covenant, or a runtime in-script computation).
+pub fn push_r0_groth16_witness<Claim: Digestible + Clone>(builder: &mut ScriptBuilder, receipt: Groth16Receipt<Claim>) -> Result<()> {
```
michaelsutton: nit/api clarity: maybe this should be named `push_r0_groth16_proof` rather than `push_r0_groth16_witness`?

more importantly, having `journal_hash` on the stack is not part of the contract of pushing the proof. this helper maps an r0 `Groth16Receipt` into the compressed groth16 proof bytes expected by the kaspa verifier, and pushes only that proof. in the common p2sh flow this is called while building the sigscript, while `journal_hash` is caller/script-owned and may be produced later by the redeem script / covenant logic before the verifier fragment runs.

Suggested doc:

```
/// Converts an R0 Groth16 receipt into the compressed proof bytes expected by
/// the Kaspa Groth16 verifier and pushes them onto the provided builder.
///
/// This is typically called while building a signature script. The script that
/// invokes `append_r0_groth16_verifier` is responsible for producing or placing
/// `journal_hash` under this proof before the verifier runs.
///
/// Pre-stack:  `[...]`
/// Post-stack: `[..., compressed_proof]`
///
/// When paired with `append_r0_groth16_verifier`, execution should reach the
/// verifier with:
///
/// Pre-stack:  `[..., journal_hash, compressed_proof]`
/// Post-stack: `[..., true]`
```

crypto/txscript/zk-sdk/src/zk_to_script/wasm/fragments.rs
```diff
@@ -0,0 +1,174 @@
...
+    /// script from fragments.
+    #[wasm_bindgen(js_name = "addData")]
+    pub fn add_data(&mut self, data: BinaryT) -> Result<()> {
+        let data = data.try_as_vec_u8()?;
+        self.inner.builder_mut()?.add_data(&data)?;
+        Ok(())
+    }
+
+    /// Pushes the r0 groth16 witness material.
+    /// The caller-owned `journal_hash` must already be on the stack beneath it.
```
michaelsutton: same here: the wasm doc should not make `journal_hash` part of the contract of `pushR0Groth16Witness`.

this method only pushes the compressed proof derived from the r0 `Groth16Receipt`. the caller/script just needs to arrange `[..., journal_hash, compressed_proof]` before `appendR0Groth16Verifier` runs.

crypto/txscript/zk-sdk/src/zk_to_script/fragments/groth16.rs
```diff
@@ -0,0 +1,213 @@
...
+/// setup as per Risc0, but the verification itself is done by the generic
+/// Arkworks implementation.
+pub fn append_r0_groth16_verifier(builder: &mut ScriptBuilder, image_id: [u8; 32]) -> Result<()> {
+    let params = Groth16ReceiptVerifierParameters::default();
+    let (a0, a1) = split_digest_bytes(params.control_root);
+    let id_bn254: [u8; 32] = params.bn254_control_id.into();
+
+    // Bake the VK in. R0's Groth16 VK is fixed, but pinning it here means
+    // a malicious spender can't swap it for a VK they control.
+    let mut serialized_vk = Vec::new();
```
michaelsutton: nit: `serialized_vk` can itself be a const (your call since this happens offchain)

michaelsutton: the above + other small stack optimization suggestions can be found here: https://github.com/michaelsutton/rusty-kaspa/commit/43b444bb73d97225b1af70a1bf974873ea435045

michaelsutton:
`git diff --check master...HEAD` shows a few trailing spaces in js files, perhaps clean that up

crypto/txscript/src/zk_precompiles/risc0/rcpt.rs
```diff
@@ -69,7 +69,6 @@ pub struct SuccinctReceipt {
     /// The control ID of this receipt, identifying the recursion program that was run (e.g. lift,
     /// join, or resolve).
     control_id: Digest,
-
```
michaelsutton, line 72: leftover line removal

saefstroem:
I agree with the latest round of feedback, one minor note from my end is that I removed the vk submodule since it is unused in txscript now and moved it to the zk-sdk crate, as well as updated the tests for the r0 upstream VK verification.

crypto/txscript/zk-sdk/src/zk_to_script/fragments/groth16.rs
```diff
@@ -30,15 +26,73 @@ const POST_DIGEST_HALTED_ZERO: [u8; 32] = [
...
+    0x00,
+];
+
+/// u32_le(sys_exit = 0) || u32_le(user_exit = 0) || u16_le(4).
+const RECEIPT_CLAIM_SUFFIX: [u8; 10] = [0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x04, 0x00];
+
+/// Ark-serialized compressed Groth16 VK for the fixed R0 circuit. Precomputed
+/// offchain so the verifier fragment doesn't have to deserialize/serialize the
+/// key at build time. `compressed_vk_matches_uncompressed_vk` pins it to the
+/// canonical [`R0_SERIALIZED_UNCOMPRESSED_VK`][crate::zk_to_script::builder::R0_SERIALIZED_UNCOMPRESSED_VK].
```
michaelsutton:
```suggestion
/// Ark-serialized compressed Groth16 VK for the fixed R0 circuit. Precomputed
/// offchain so the verifier fragment can embed it directly in the script.
/// `verify_upstream_vk_roundtrip` pins this constant to
/// `risc0_groth16::verifying_key()`.
```

biryukovmaxim (dismissed):

Cargo.lock
michaelsutton: I made sure that kaspad's dep is the toml-pinned v0.6.0

michaelsutton (approved):

PR #955, opened 2026-04-12
Refactor txscript runtime metering into versioned resource meters

michaelsutton:
Refactors txscript runtime metering so the engine holds a single versioned meter object instead of separate top-level accounting fields.

- introduces `RuntimeResourceMeter` with `Sigops` and `ScriptUnits` variants
- keeps the legacy runtime sigop counter path as is
- moves script-unit accounting state into a dedicated `RuntimeScriptUnitMeter`
- renames `runtime_sig_op_counter.rs` to `runtime_resource_meter.rs`

This is intended as a structural refactor only, to make metering mode selection explicit and reduce repeated `covenants_enabled` branching in execution-time metering code.


crypto/txscript/src/runtime_resource_meter.rs
```diff
@@ -0,0 +1,180 @@
...
+    sigop_script_units: ScriptUnits,
+    allowed_script_units: ScriptUnits,
+    remaining_script_units: ScriptUnits,
+}
+
+impl RuntimeScriptUnitMeter {
+    pub fn new(sigop_script_units: ScriptUnits, allowed_script_units: ScriptUnits) -> Self {
+        Self {
+            used_sig_ops: 0,
+            used_script_units: ScriptUnits(0),
```
someone235: Maybe just calc it as `allowed - remaining`

crypto/txscript/src/runtime_resource_meter.rs
```diff
@@ -0,0 +1,180 @@
...
+            accounted_pushed_bytes: 0,
+            sigop_script_units,
+            allowed_script_units,
```
someone235: nit, maybe call it `script_units_limit`?

someone235 (changes requested):

someone235 (dismissed):

someone235 (approved):

PR #956, opened 2026-04-14
aDBptation

biryukovmaxim:
## Summary

Adds Toccata-hardfork backward compatibility for DB-persisted consensus data so a post-Toccata node can still read databases written by pre-Toccata nodes.

## What's done

- **`UtxoEntry` serde split out** into `consensus/core/src/tx/utxo_entry.rs` with a manual `Deserialize` that reads the four pre-Toccata fields and treats EOF on the trailing `covenant_id` Option tag as `None`. Human-readable (JSON) formats keep the camelCase shape via a shadow struct.
- **`Transaction` serde** moved into `consensus/core/src/tx/serde_impl.rs`; fixtures refreshed.
- **Version-aware `CachedDbAccess`**: opt-in `version_suffix: Option<u8>` lets a store whose on-disk value layout changed across a hardfork keep decoding legacy rows while writing new ones under a versioned key. Reads use a `rocksdb::PrefixRange` scan on `[store_prefix || logical_key]` that matches either layout and dispatches on returned key length. A new `TLegacy: Into<TData>` generic (defaults to `TData`) decodes legacy rows. Deletes clear both layouts; cache stays keyed on the logical `TKey`.
- **`DbUtxoDiffsStore` opts in** with `POST_TOCCATA_UTXO_DIFFS_VERSION = 1` and `PreToccataUtxoDiff` (in `consensus/core/src/utxo/pre_toccata.rs`) as the legacy shadow type, mirroring the pre-Toccata four-field `UtxoEntry` layout. Its derived `Deserialize` composes cleanly inside `HashMap`, sidestepping the EOF-based covenant detection that fails for nested entries. `From<PreToccataUtxoDiff>` converts to `Arc<UtxoDiff>` with `covenant_id = None`.
- **Compact UTXO entry** gains an EOF-tolerant decode path for pre-Toccata rows.

## Tests

- Frozen pre-Toccata bincode fixtures for `UtxoEntry`, `Transaction`, and `BlockBody` payload in `consensus/core/tests/db_compat.rs`, decoded through post-Toccata types and round-tripped with new covenant / `TxInputMass::ComputeBudget` variants.
- Seven `database/src/access.rs` unit tests covering versioned round trip, legacy decode via `TLegacy`, suffix/length validation, delete-both, and the `has` guard that upholds "at most one row per logical key".
- Four `consensus/src/model/stores/utxo_diffs.rs` integration tests against real RocksDB: planted pre-Toccata bytes decode, post-Toccata writes land at the versioned key, `insert` refuses to overwrite a legacy row, `delete` clears both layouts.


consensus/core/src/tx/serde_impl.rs
```diff
@@ -0,0 +1,450 @@
...
+use kaspa_utils::serde_bytes;
+use serde::{
+    Deserialize, Deserializer, Serialize, Serializer,
+    de::{self, DeserializeSeed, IntoDeserializer, MapAccess, SeqAccess, Visitor},
+    ser::{SerializeSeq, SerializeStruct},
+};
+use serde_value::Value as BufferedValue;
+
+#[repr(transparent)]
+struct SerdeBytesRef<'a>(&'a [u8]);
```
someone235: Maybe it should move to `utils/src/serde_bytes`? We can also call it `serde_bytes::Bytes` to remain similar to https://docs.rs/serde_bytes/latest/serde_bytes/index.html

consensus/core/src/tx/serde_impl.rs
```diff
@@ -0,0 +1,450 @@
...
+
+impl Serialize for SerdeBytesRef<'_> {
+    fn serialize<S: Serializer>(&self, serializer: S) -> Result<S::Ok, S::Error> {
+        serde_bytes::serialize(&self.0, serializer)
+    }
+}
+
+#[derive(Deserialize)]
+#[repr(transparent)]
+struct SerdeBytesOwned(#[serde(with = "serde_bytes")] Vec<u8>);
```
someone235: Same here (consider calling it ByteBuf)

consensus/core/src/tx/serde_impl.rs
```diff
@@ -0,0 +1,450 @@
...
+    Payload,
+    Mass,
+    Id,
+}
+
+impl<'de> Deserialize<'de> for Field {
+    fn deserialize<D: Deserializer<'de>>(deserializer: D) -> Result<Self, D::Error> {
+        #[derive(Deserialize)]
+        #[serde(field_identifier, rename_all = "camelCase")]
+        enum FieldInner {
```
someone235: Why do we need `FieldInner`? Can't we write

```
        #[derive(Deserialize)]
        #[serde(field_identifier, rename_all = "camelCase")]
        enum Field{...}
```

consensus/core/src/tx/utxo_entry.rs
```diff
@@ -0,0 +1,171 @@
...
+        if deserializer.is_human_readable() {
+            UtxoEntryHumanReadable::deserialize(deserializer).map(Into::into)
+        } else {
+            struct UtxoEntryVisitor;
+
+            impl<'de> Visitor<'de> for UtxoEntryVisitor {
+                type Value = UtxoEntry;
+
+                fn expecting(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+                    f.write_str("a UtxoEntry tuple (amount, script_public_key, block_daa_score, is_coinbase[, covenant_id])")
```
someone235: Why tuple and not struct? In `TransactionVisitor` you wrote "struct Transaction"

consensus/core/src/utxo/utxo_entry.rs
```diff
@@ -0,0 +1,171 @@
...
+    }
+}
+
+impl MemSizeEstimator for UtxoEntry {}
+
+/// Shadow struct used on the human-readable serde path so JSON/RPC consumers
+/// keep seeing a camelCase object shape with `covenant_id` optional.
+#[derive(Deserialize)]
+#[serde(rename_all = "camelCase")]
+struct UtxoEntryHumanReadable {
```
someone235, line 55: Can you add a comment that explains why this doesn't need special deserialization? i.e. because we expect it to be used in a JSON-like formats?

consensus/core/src/tx/utxo_entry.rs
```diff
@@ -0,0 +1,171 @@
...
+mod tests {
+    use super::*;
+    use crate::tx::ScriptPublicKey;
+
+    /// Bincode encoding produced by the pre-Toccata `UtxoEntry` (4-field layout,
+    /// no trailing `covenant_id` Option tag).
+    const PRE_TOCCATA_HEX: &str = "efcdab89674523010000060000000000000076a9140102032a0000000000000001";
+    const PRE_TOCCATA_AMOUNT: u64 = 0x0123_4567_89ab_cdef;
+    const PRE_TOCCATA_DAA_SCORE: u64 = 42;
+    const PRE_TOCCATA_SPK_SCRIPT: &[u8] = &[0x76, 0xa9, 0x14, 0x01, 0x02, 0x03];
```
someone235: No need to say `SPK_SCRIPT`, SPK is enough :)

consensus/core/src/utxo/utxo_entry.rs
someone235, line 1: Isn't it better to put it inside `consensus/core/src/utxo/utxo_entry.rs`?

consensus/core/src/utxo/pre_toccata.rs
someone235, line 1: Since you keep this struct, maybe also add some tests to check we can serialize it, and then deser it back to post-toccata without any problem?

I think you can optionally implement serialization like this

```
#[cfg_attr(test, derive(Serialize))]
pub struct PreToccataUtxoEntry {
...
}
```

consensus/core/src/utxo/utxo_entry.rs
```diff
@@ -0,0 +1,171 @@
...
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::tx::ScriptPublicKey;
+
+    /// Bincode encoding produced by the pre-Toccata `UtxoEntry` (4-field layout,
+    /// no trailing `covenant_id` Option tag).
+    const PRE_TOCCATA_HEX: &str = "efcdab89674523010000060000000000000076a9140102032a0000000000000001";
```
someone235, line 119: Maybe add some sanity test to check this is deserializable to pre-toccata entry?

consensus/core/tests/fixtures/pre_toccata_db_compat_hex
someone235, line 1: Since this is not pure hex, I prefer not to give it the file extension hex. Maybe call it `consensus/core/tests/fixtures/pre_toccata_db_compat_hex`

database/src/access.rs
```diff
@@ -45,28 +94,156 @@ where
...
+                scan_len,
+                scan_len + 1,
+            )));
+        };
+
+        self.cache.insert(key, data.clone());
+        Ok(data)
+    }
+
+    /// Note: `has_with_fallback` and [`Self::read_with_fallback`] are the
```
someone235, line 247: I think that if we write a note here it's better to also explain better what `has_with_fallback` does (i.e. used to access legacy stores).

database/src/access.rs
```diff
@@ -195,8 +380,39 @@ where
...
+    /// Deletes both the versioned and legacy rows for a logical key. Delete
+    /// of a non-existent row is a no-op in RocksDB, so non-versioned stores
+    /// still issue exactly one delete and versioned stores issue one extra
+    /// no-op delete during the migration window.
+    fn delete_both_layouts(&self, writer: &mut impl DbWriter, key: TKey) -> Result<(), StoreError>
+    where
+        TKey: Clone + AsRef<[u8]>,
+    {
+        if let Some(suffix) = self.version_suffix {
+            let mut versioned_key = DbKey::new(&self.prefix, key.clone());
```
someone235: Maybe use `build_write_key`?

indexes/core/src/indexed_utxos.rs
```diff
@@ -46,6 +48,56 @@ impl From<UtxoEntry> for CompactUtxoEntry {
...
+        Self { amount: e.amount, block_daa_score: e.block_daa_score, is_coinbase: e.is_coinbase, covenant_id: e.covenant_id }
+    }
+}
+
+impl<'de> Deserialize<'de> for CompactUtxoEntry {
+    fn deserialize<D: Deserializer<'de>>(deserializer: D) -> Result<Self, D::Error> {
+        if deserializer.is_human_readable() {
+            CompactUtxoEntryHumanReadable::deserialize(deserializer).map(Into::into)
+        } else {
+            struct CompactUtxoEntryVisitor;
```
someone235: Same comment about `else`

consensus/core/tests/db_compat.rs
```diff
@@ -0,0 +1,237 @@
...
+const EXPECTED_COINBASE_PAYLOAD: &[u8] = &[0xc0, 0xff, 0xee];
+const EXPECTED_REGULAR_IN1_SEQ: u64 = 0x00ff_ff00_ff00_ff00;
+
+#[test]
+fn decode_pre_toccata_coinbase_transaction() {
+    let bytes = fixture("transaction_v0_coinbase");
+
+    let decoded: Transaction = bincode::deserialize(bytes).expect("decode pre-Toccata coinbase Transaction");
+
+    assert_eq!(decoded.version, 0, "version");
```
someone235: `Transaction` has `#[derive(Eq)]`, so maybe it's better to compare to `expected_tx`? (Might apply to other places as well)

someone235 (changes requested):

consensus/core/src/utxo/pre_toccata.rs
```diff
@@ -38,7 +40,8 @@ use crate::utxo::utxo_diff::UtxoDiff;
 /// `covenant_id` tag. Deserializing this type through a derived
 /// `Deserialize` consumes exactly the pre-Toccata byte sequence and
 /// composes inside container types such as `HashMap`.
-#[derive(Debug, Clone, Deserialize)]
+#[derive(Debug, Clone, PartialEq, Eq, Deserialize)]
+#[cfg_attr(test, derive(Serialize))]
```
someone235: nit, but it feels cleaner to write

`#[cfg_attr(test, derive(serde::Serialize))]`

and then delete

```
#[cfg(test)]
use serde::Serialize;
```

Do whatever you prefer.

consensus/core/src/utxo/utxo_entry.rs
```diff
@@ -111,30 +116,77 @@ mod tests {
...
+    /// Post-Toccata wire for the same logical entry with `covenant_id = None`
+    /// (4 fields + Option tag `0x00`).
+    const POST_TOCCATA_NONE_HEX: &str = "efcdab89674523010000060000000000000076a9140102032a000000000000000100";
+    /// Post-Toccata wire for the same logical entry with `covenant_id = Some(Hash::from_bytes([0x5a; 32]))`
+    /// (4 fields + Option tag `0x01` + 32-byte hash).
+    const POST_TOCCATA_SOME_HEX: &str = "efcdab89674523010000060000000000000076a9140102032a000000000000000101\
+                                         5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a";
+    const SHARED_AMOUNT: u64 = 0x0123_4567_89ab_cdef;
+    const SHARED_DAA_SCORE: u64 = 42;
+    const SHARED_SPK_SCRIPT: &[u8] = &[0x76, 0xa9, 0x14, 0x01, 0x02, 0x03];
```
someone235: Can be renamed to `SHARED_SPK`

consensus/core/src/utxo/utxo_entry.rs
```diff
@@ -0,0 +1,223 @@
...
+}
+
+impl<'de> Deserialize<'de> for UtxoEntry {
+    fn deserialize<D: Deserializer<'de>>(deserializer: D) -> Result<Self, D::Error> {
+        if deserializer.is_human_readable() {
+            UtxoEntryHumanReadable::deserialize(deserializer).map(Into::into)
+        } else {
+            struct UtxoEntryVisitor;
+
+            impl<'de> Visitor<'de> for UtxoEntryVisitor {
```
someone235: I think to improve readability it's better to put the impl in the file top-level and not inside the else-branch

someone235 (changes requested):

consensus/core/src/utxo/utxo_entry.rs
```diff
@@ -75,33 +75,32 @@ impl From<UtxoEntryHumanReadable> for UtxoEntry {
...
+                let is_coinbase: bool = seq.next_element()?.ok_or_else(|| DeError::invalid_length(3, &self))?;
+                // Pre-Toccata entries have no trailing Option tag; the
+                // bincode reader hits EOF here and we treat that as None.
+                let covenant_id: Option<Hash> = match seq.next_element::<Option<Hash>>() {
+                    Ok(Some(value)) => value,
+                    Ok(None) | Err(_) => None,
+                };
+                Ok(UtxoEntry { amount, script_public_key, block_daa_score, is_coinbase, covenant_id })
+            }
+        }
```
someone235, line 101: Maybe add space here?

someone235 (changes requested):

someone235 (approved):

PR #962, opened 2026-04-22
fix(notification): reject utxo change sub when no utxo index

### LINKED ISSUE #960: Reject UtxosChanged subscriptions when utxoindex is disabled
UtxosChanged subscriptions are still accepted when `utxoindex` is disabled, as long as the request contains a non-empty address list.

That looks like a bug. In this configuration the node cannot actually serve UTXO-change notifications, but it still accepts the request and mutates subscription / tracker state. A remote client can keep sending large sets of unique addresses and force the node to retain useless state in memory.

From reading the code, only the blanket case is rejected:
- rpc/service/src/service.rs:1313

Relevant paths:
- rpc/service/src/service.rs:165-205
- notify/src/scope.rs:157
- notify/src/subscription/context.rs:27
- notify/src/address/tracker.rs:247
- rpc/wrpc/server/src/service.rs:16

Expected behavior:
If `utxoindex` is disabled, all `UtxosChanged` subscriptions should be rejected before notifier / tracker state is changed.

Suggested fix:
1. Reject all `UtxosChanged` subscriptions when `utxoindex` is disabled.
2. Add a hard cap on addresses per request and per listener.
3. Deduplicate addresses before registration.
4. Add a regression test that verifies rejected subscriptions do not mutate tracker state.

IzioDev:
Fixes #960

When subscribing to `UtxosChanged` notification, it now rejects early if `utxoindex` is not enabled on the node

### DIFF
--- a/rpc/service/src/service.rs
+++ b/rpc/service/src/service.rs
@@ -1312,6 +1312,7 @@ NOTE: This error usually indicates an RPC conversion error between the node and 
     /// Start sending notifications of some type to a listener.
     async fn start_notify(&self, id: ListenerId, scope: Scope) -> RpcResult<()> {
         match scope {
+            Scope::UtxosChanged(_) if !self.config.utxoindex => Err(RpcError::NoUtxoIndex),
             Scope::UtxosChanged(ref utxos_changed_scope) if !self.config.unsafe_rpc && utxos_changed_scope.addresses.is_empty() => {
                 // The subscription to blanket UtxosChanged notifications is restricted to unsafe mode only
                 // since the notifications yielded are highly resource intensive.


demisrael:
Thanks for the quick fix. I traced the dispatch end-to-end against `780f3417` and I'm not sure this actually closes #960 — the bug is reachable from gRPC and wRPC clients, but neither of those subscription paths goes through `RpcCoreService::start_notify` where the new guard sits.

**gRPC side.** Notify-prefixed methods are macro-generated at `rpc/macros/src/grpc/server.rs:104`, and the Notify branch routes them to `server_ctx.notifier.execute_subscribe_command(...)`. That `notifier` is the gRPC server's own `Notifier` (`connection_handler.rs:56,108`), not the `RpcApi` trait — the non-Notify branch in the same macro does take the trait path, but the Notify branch deliberately bypasses it. There's no `replace_method` override for `NotifyUtxosChanged` either; the only override is `NotifyFinalityConflict`, and it routes through the same notifier.

**wRPC side.** `router.rs:86` calls `Server::start_notify` (`rpc/wrpc/server/src/server.rs:175`), which goes straight into `wrpc_notifier.try_start_notify(...)` on line 190. The Subscriber set up at lines 80–89 propagates upstream into the rpc-core `Notifier` (a `SubscriptionManager` per `notify/src/notifier.rs:213`), so again the trait method on `RpcCoreService` isn't on the path. And the rpc-core `Notifier` itself carries no `utxoindex` flag, so it can't gate at that layer.

Net of that: against a node started without `--utxoindex`, a remote subscribe to `UtxosChanged` with a non-empty address list still registers and grows `SubscriptionContext::address_tracker` — the exact pattern the issue describes.

The patch isn't wasted, though. In-process consumers that hold `Arc<dyn RpcApi>` directly do hit the new guard — the embedded wallet at `wallet/core/src/utxo/processor.rs:224` is the obvious one — and that's a genuine improvement on its own. So I'd suggest keeping this guard and adding the two missing pieces:

1. wRPC: early return at the top of `Server::start_notify` (`rpc/wrpc/server/src/server.rs:175`) before `wrpc_notifier.try_start_notify`.
2. gRPC: either special-case `NotifyUtxosChanged` in the macro, or add a `replace_method` override in `rpc/grpc/server/src/request_handler/factory.rs` modelled on the existing `NotifyFinalityConflict` block at lines 100–138.

And the regression test the issue asks for would help lock this in: daemon started with `utxoindex=false`, subscribe to `UtxosChanged` with both empty and non-empty address lists, expect `NoUtxoIndex` back, and assert `SubscriptionContext::address_tracker.len()` is unchanged after the rejected call.

If I missed a path that wires the trait method back into the public dispatch I'd love a pointer — entirely possible I'm wrong about that.

PR #963, opened 2026-04-22
Add Toccata lane/gas limits to consensus and implement corresponding mempool policies 

michaelsutton:
**Motivation / background:**

The lane-per-block limit bounds the amount of seqcommit active-set churn a single block can cause. Without it, worst-case mergesets can touch thousands of lanes, forcing large active-lane updates, SMT writes/proofs, and corresponding template-construction overhead. The goal is not to limit long-term lane activity: the system can still support many lanes over time, but per-block fanout should remain bounded.

The gas-per-lane limit gives each app/lane a way to regulate its own compute throughput under a fixed global consensus envelope. Consensus only enforces the per-lane gas budget; each app/lane can choose what a gas unit means internally and scale its own execution model accordingly. Related background: https://research.kas.pa/t/fees-and-throughput-regulation-dynamics/247

**Main implementation points, tradeoffs, and follow-up questions:**

1. Consensus rule
   - block has at most N distinct non-coinbase lanes
   - per lane accumulated gas <= G
   - gas allowed only for post-Toccata non-system lanes
   - pre-Toccata harmless: valid txs are native + zero gas

2. Shared params
   - BlockLaneLimits added next to BlockMassLimits
   - params carry lane/gas limits into consensus + mining
   - current defaults: lanes=50, gas_per_lane=500k

3. Mempool std policy
   - single tx gas must be <= gas_per_lane
   - this is policy, not extra consensus
   - open q: maybe std gas should be stricter, e.g. G/5, since gas scarcity is lane-local and not fully priced

4. Template selection policy
   - all selector paths enforce same lane/gas limits
   - selector weight remains mostly fee/mass
   - gas is enforced greedily during selection, not priced into global weight
   - sample-inplace now has a bounded mass-gap compensation path for large txs: it may sample a little past the nominal target if the greedy pack would otherwise leave a large unused mass gap

5. Where this is good enough
   - if active mempool lanes <= limit: behavior is basically unchanged
   - if active mempool lanes > limit: first chosen lanes still come from fee/mass sampling
   - after lanes are fixed, intra-lane continuation is efficient via per-lane ordered sets / tree paths
   - block validity path is protected: templates should not exceed consensus lane/gas limits

6. Known suboptimalities / intentional compromises
   - high-feerate small lanes can occupy slots and crowd out lower-feerate high-volume lanes
   - gas-heavy txs can consume lane gas greedily because gas is not in global weight
   - TakeAll / MutatingTree do not retry lane/gas-skipped txs after later rejection frees resources
   - SequenceSelector has better retry behavior and is the stressed-mempool path
   - no sophisticated lane allocation/preselection yet
   - no native-lane reservation yet

7. Why acceptable for first version
   - limits protect node/consensus worst cases now
   - miner policy remains simple and fast
   - hard cases are economic optimality, not consensus safety
   - room for later heuristics: native lane reservation, per-lane aggregate stats, stricter std gas, better gas-aware scoring

8. Selector implementation cleanup
   - the mutating-tree selector replaces the rebalancing selector by cloning the frontier tree and removing selected/skipped candidates directly
   - this gives us simpler log-time mutation instead of the old lazy deletion/rebuild model
   - the old rebalancing selector is kept only temporarily for benchmark comparison and can be deleted once we are done comparing it to the mutating-tree selector

9. Testing / validation angle
   - consensus unit tests for lane/gas rejection
   - tx isolation tests for gas/subnet rules
   - selector tests for lane/gas enforcement
   - stdness test for tx gas > lane limit
   - perf baseline: selector remains in acceptable range after lane-aware paths


consensus/src/pipeline/body_processor/body_validation_in_isolation.rs
```diff
@@ -64,6 +68,7 @@ impl BlockBodyProcessor {
         let mut total_compute_mass: u64 = 0;
         let mut total_transient_mass: u64 = 0;
         let mut total_storage_mass: u64 = 0;
+        let mut lanes = HashMap::<SubnetworkId, u64>::new();
```
IzioDev, line 71: should we use with_capacity with `block_lane_limits.lanes_per_block`?

consensus/src/processes/transaction_validator/tx_validation_in_isolation.rs
```diff
@@ -116,16 +116,24 @@ fn check_duplicate_transaction_inputs(tx: &Transaction) -> TxResult<()> {
...
-    if tx.gas > 0 {
-        return Err(TxRuleError::TxHasGas);
+    if tx.gas == 0 {
+        return Ok(());
     }
-    Ok(())
+
+    // Only post-Toccata non-system lanes may carry gas; reserved system lanes
+    // have a 19-byte zero suffix and must remain gas-free.
+    if tx.version >= TX_VERSION_TOCCATA && !matches!(tx.subnetwork_id.as_bytes(), [_x, rest @ ..] if rest == ZEROES_19) {
```
IzioDev, line 128: it may be intended to not delegate subnetid strict check to a SubnetworkId impl (hardcoded rule), just noticing.

question: code often use native/system/builtin subnet. Is there redundancy across them, or do they all reflects different meaning?

consensus/src/pipeline/body_processor/body_validation_in_isolation.rs
michaelsutton: hard to say. can be a premature allocation

consensus/src/processes/transaction_validator/tx_validation_in_isolation.rs
michaelsutton: it seems more correct to put nuanced validation rules in consensus code and not in the type definition. native/system/builtin have slightly different meaning. builtin is somewhat outdated (includes a registry entry which was never activated and is now part of the "system reserved" space).

mining/src/mempool/model/frontier/selectors.rs
```diff
@@ -160,3 +238,182 @@ impl TemplateTransactionSelector for TakeAllSelector {
...
+
+            if !self.lanes.try_select(&self.policy, lane, tx.gas) {
+                // For now we compromise on retry optimality in this less-congested path:
+                // LPB/gas-skipped candidates are not reconsidered after later rejections
+                // (tx rejections are expected to be rare here).
+                self.tree.remove(&candidate);
+                continue;
+            }
+
+            self.tree.remove(&candidate);
```
biryukovmaxim: both try_select branches lead to removal. sweep-bptree supports `remove_by_argumen` so we can use it instread of another walk for removal

mining/src/mempool/model/frontier/selectors.rs
```diff
@@ -160,3 +238,182 @@ impl TemplateTransactionSelector for TakeAllSelector {
...
+        self.selected_vec.clear();
+        self.selected_map = None;
+    }
+}
+
+impl TemplateTransactionSelector for MutatingTreeSelector {
+    fn select_transactions(&mut self) -> Vec<Transaction> {
+        self.reset_selection();
+        let mut rng = rand::thread_rng();
+        let mut transactions = Vec::new();
```
biryukovmaxim, line 317: isnt it better to reuse vector from previous attempts (if they were done)

mining/src/mempool/model/frontier.rs
biryukovmaxim, line 232: i think it makes sense to add to documentation how it selects transactions regarding lanes. That it freezes when it  reaches when it reaches the limit

mining/src/mempool/model/frontier.rs
```diff
@@ -174,14 +267,74 @@ impl Frontier {
                 }
                 item
             };
+            if lanes.occupied.len() < policy.lanes_per_block_limit {
```
biryukovmaxim, line 274: should we add couple more lanes since its only sampler, `SequenceSelector` will handle lanes nicely if we give some more to him. just in case of rejections we will be able to read couple more lanes to fill the block.

biryukovmaxim (changes requested):

mining/src/mempool/model/frontier/selectors.rs
michaelsutton: the code above breaks without removal if `next_total_mass` is above the limit (the item is kept in the tree in case future tx rejections clear up sufficient mass space)

however, looking at that area, it does trigger the thinking that maybe we should:
- remove also in that case (which will allow using your suggestion)
- try a few more iterations in case we find smaller mass items
- keep those large mass items in a side collection and reinsert them if txs are rejected and enough space is cleared

consensus/core/src/config/constants.rs
```diff
@@ -89,6 +89,13 @@ pub mod consensus {
 
     /// Coinbase maturity in seconds
     pub const COINBASE_MATURITY_SECONDS: u64 = 100;
+
+    //
+    // ~~~~~~~~~~~~~~~~~~~ Toccata lane limits ~~~~~~~~~~~~~~~~~~~
+    //
+
+    pub const DEFAULT_LANES_PER_BLOCK_LIMIT: usize = 50;
+    pub const DEFAULT_GAS_PER_LANE_LIMIT: u64 = 500_000;
```
someone235, line 98: Maybe we want some higher number? To allow the underlying subnetworks to have a lot of granularity (I would set it to least 1B)

consensus/src/pipeline/body_processor/body_validation_in_isolation.rs
```diff
@@ -88,6 +93,31 @@ impl BlockBodyProcessor {
             if total_storage_mass > self.block_mass_limits.storage {
                 return Err(RuleError::ExceedsStorageMassLimit(total_storage_mass, self.block_mass_limits.storage));
             }
+
+            // Pre-Toccata valid blocks contain only native non-coinbase txs with zero gas,
+            // so applying these lane/gas limits unconditionally is harmless before activation.
+            if !tx.is_coinbase() {
+                let occupied_lanes = lanes.len();
+                let gas = match lanes.entry(tx.subnetwork_id) {
```
someone235, line 101: I think calling it `lane_gas` will be a bit clearer

consensus/src/pipeline/body_processor/body_validation_in_isolation.rs
```diff
@@ -88,6 +93,31 @@ impl BlockBodyProcessor {
...
+                    Entry::Occupied(mut entry) => {
+                        let gas = entry.get_mut();
```
someone235, line 103: Here as well

consensus/src/processes/transaction_validator/tx_validation_in_isolation.rs
someone235, line 128: nit: I think it'll be clearer to have some reusable `is_reserved` function.

mining/src/mempool/model/frontier/selectors.rs
michaelsutton: not sure what you mean, we return the vec to the caller

mining/src/mempool/model/frontier.rs
michaelsutton: it might backfire bcs the selector will have less txs to actually choose from

mining/src/block_template/policy.rs
```diff
@@ -1,14 +1,24 @@
...
 /// the generation of block templates. See the documentation for
 /// NewBlockTemplate for more details on how each of these parameters are used.
 #[derive(Clone)]
 pub struct Policy {
     /// max_block_mass is the maximum block mass to be used when generating a block template.
     pub(crate) max_block_mass: u64,
+    /// lanes_per_block_limit is the maximum number of distinct subnet lanes a block template may include.
+    pub(crate) lanes_per_block_limit: usize,
+    /// gas_per_lane_limit is the maximum total gas per lane in a block template.
+    pub(crate) gas_per_lane_limit: u64,
```
someone235, line 13: Maybe it should be a constant, so we won't have to carry it around everywhere? Once we set a high enough number, I don't see any reason to change it in the future.

mining/src/mempool/model/frontier/selectors.rs
michaelsutton: done

mining/src/block_template/selector.rs
```diff
@@ -175,6 +190,8 @@ impl RebalancingWeightedTransactionSelector {
                 *gas_usage = next_gas_usage.unwrap();
             }
 
+            *lane_entry.or_default() += 1;
```
someone235: I think logically it belongs below the comment

mining/src/mempool/model/frontier.rs
michaelsutton: done

mining/src/mempool/model/frontier/feerate_key.rs
```diff
@@ -24,6 +25,9 @@ impl FeerateTransactionKey {
         // NOTE: any change to the way this weight is calculated (such as scaling by some factor)
         // requires a reversed update to total_weight in `Frontier::build_feerate_estimator`. This
         // is because the math methods in FeeEstimator assume this specific weight function.
+        //
+        // Gas is intentionally not folded into this global weight: gas capacity is lane-local and
+        // is enforced by selectors during block template construction.
```
someone235, line 30: Maybe add some comment/TODO that talks about the possibility of introducing some kind of multi-dimensional weight system?

mining/src/mempool/check_transaction_standard.rs
```diff
@@ -440,6 +446,22 @@ mod tests {
...
             },
+            Test {
+                name: "Transaction gas exceeds the per-lane limit",
+                mtx: new_mtx(
+                    Transaction::new(
+                        TX_VERSION,
+                        vec![dummy_tx_input.clone()],
+                        vec![dummy_tx_out.clone()],
+                        0,
+                        SUBNETWORK_ID_NATIVE,
```
someone235, line 457: It makes more sense, at least optically, to put here some non-native lane, no?

biryukovmaxim (dismissed):

michaelsutton:
Deleted the rebalancing selector since the mutating-tree selector consistently outperformed it by about `1.8x-2.3x` across the relevant `300-2000` range, including under `BENCH_ACTIVE_LANES=1000`, while also giving us a simpler implementation model

biryukovmaxim (approved):

PR #964, opened 2026-04-23
fix(indexes/utxoindex): spk key pre-allocation

IzioDev:
allocation should've been prefix + dyn data, where prefix is version (2) + length (8), not version only

### DIFF
--- a/indexes/utxoindex/src/stores/indexed_utxos.rs
+++ b/indexes/utxoindex/src/stores/indexed_utxos.rs
@@ -16,13 +16,14 @@ use std::sync::Arc;
 pub const VERSION_TYPE_SIZE: usize = size_of::<ScriptPublicKeyVersion>(); // Const since we need to re-use this a few times.
 
 /// [`ScriptPublicKeyBucket`].
-/// Consists of 2 bytes of little endian [VersionType] bytes, followed by a variable size of [ScriptVec].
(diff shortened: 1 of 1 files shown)


indexes/utxoindex/src/stores/indexed_utxos.rs
```diff
@@ -101,7 +102,7 @@ impl Display for UtxoEntryFullAccessKey {
 impl UtxoEntryFullAccessKey {
     /// Creates a new [UtxoEntryFullAccessKey] from a [ScriptPublicKeyBucket] and [TransactionOutpointKey].
     pub fn new(script_public_key_bucket: ScriptPublicKeyBucket, transaction_outpoint_key: TransactionOutpointKey) -> Self {
-        let mut bytes = Vec::with_capacity(TRANSACTION_OUTPOINT_KEY_SIZE + script_public_key_bucket.as_ref().len());
+        let mut bytes = Vec::with_capacity(script_public_key_bucket.as_ref().len() + TRANSACTION_OUTPOINT_KEY_SIZE);
```
someone235, line 105: Why is it changed?

someone235 (changes requested):

someone235 (dismissed):

someone235 (approved):

PR #965, opened 2026-04-25
Use kaspanet workflow-perf-monitor

saefstroem:
This PR changes the workflow-perf-monitor crate to use the updated from `kaspanet` in order to use the latest version of bindgen. Given the short time until HF, it is important that we resolve such upstream crate issues as fast as possible. I have tried contacting @aspect , without any response in order to try and PR it into the original crate.


Cargo.toml
```diff
@@ -302,7 +302,7 @@ indexed_db_futures = "0.5.0"
 # workflow dependencies that are not a part of core libraries
 
 # workflow-perf-monitor = { path = "../../../workflow-perf-monitor-rs" }
-workflow-perf-monitor = "0.0.2"
+workflow-perf-monitor = { git='https://github.com/kaspanet/workflow-perf-monitor-rs.git' }
```
michaelsutton: pls specify the `v0.0.3` tag

michaelsutton (approved):

PR #966, opened 2026-04-27
SMT: fix incorrect blue score during ibd (kip21-impl followup)

biryukovmaxim:


consensus/smt-store/src/streaming_import/mod.rs
```diff
@@ -162,7 +165,16 @@ fn write_score_index(
         *batch_count += 1;
     }
 
-    // Structural: all lanes at the pruning point's blue_score (tree is built at this point)
+    // Structural: all lanes at the pruning point's blue_score.
```
michaelsutton: it seems like the branch entries with `max_bs` will be deleted by `prune_chunk` in the following lines:
```
        let mut branch_keys: BTreeSet<(BranchKey, u64, Hash)> = BTreeSet::new();
        for e in entries {
            for depth in 0..=255u8 {
                branch_keys.insert((BranchKey::new(depth, &e.lane_key), e.blue_score, e.block_hash));
            }
        }
```
(bcs there will be a lane entry with `max_bs` by def)

assuming this understanding is correct, it's not clear what these Structural `pp_blue_score`-stamped entries are useful for

michaelsutton (commented):
Note: it seems like the prior SLO promotion fix makes this bug nearly impossible to reproduce with high level tests such as the new simpa-based smt-repro. I managed to reproduce it only when reversing the SLO fix

michaelsutton (approved):

michaelsutton:
Indeed, following the SLO promotion fix this is no longer a bug fix in practice, but rather a semantic correction. It is more correct for streaming import to persist branch entries with the live-equivalent blue score (= max blue score of children).

The important consequence of this bookkeeping is that imported branches are now pruned through the same lane-derived score-index entries as live branches, so `Structural` entries are no longer needed in the import path. Conceptually, `Structural` entries are only needed to record branch changes caused by lane expirations, but at the imported pruning point there are no expiration events to record: all imported branch entries are derived from active lanes.

PR #970, opened 2026-04-28
chore(deps): address cargo audit/deny advisories

biryukovmaxim:
Replace unmaintained async-std and derivative; drop intertrait for downcast; bump tokio, rustls, igd-next, log4rs, prometheus, rv


michaelsutton (approved):
Approved. 

Worth noting that the optional `Vec<u8>` JSON fields (`payload` / `signatureScript`) now serialize as hex strings instead of byte arrays. That aligns `RpcOptionalTransaction` with the existing non-optional `RpcTransaction` wire format, so this looks like a format correction rather than an arbitrary breaking change, but it should be noted, fyi @smartgoo @IzioDev.

Note: async-std still remains transitively via workflow-* crates, to be handled in follow-up PRs.

PR #971, opened 2026-04-29
Add script pricing benchmarks and price hash opcodes

michaelsutton:
## Summary

This PR adds a focused `kaspa-txscript` Criterion benchmark suite for script-pricing workloads and uses it to tune hash opcode pricing under the new compute-budget model.

The suite distinguishes between:
- multi-transaction block-shaped workloads, where Rayon parallelism is available
- single-transaction / single-input workloads, where there is no available script parallelism

The important takeaway is that real multi-tx/script-app workloads remain cheap with parallel validation, while adversarial single-tx hash and stack-data workloads define the pricing boundries.

## Key Results

| Workload | Shape | Selected timing | Compute mass | Notes |
|---|---:|---:|---:|---|
| `schnorr_2in1` | 182 tx / 364 inputs, 8 threads | ~2.10 ms | 499,772 | signature-heavy normal tx workload |
| `introspection_cat_substr_math` | 159 tx / 159 inputs, 8 threads | ~0.74 ms | 497,511 | covenant/introspection-style workload |
| `blake2b_storm` | 24 tx / 24 inputs, 8 threads | ~0.76 ms | 389,136 | parallel hash storm stays cheap |
| `blake2b_storm_single_tx` | 1 tx / 1 input, single thread | ~9.97 ms | 499,798 | BLAKE2b is the heavier hash case |
| `blake3_storm_single_tx` | 1 tx / 1 input, single thread | ~5.92 ms | 499,880 | byte pricing keeps BLAKE3 bounded |
| `sha256_storm_single_tx` | 1 tx / 1 input, single thread | ~5.18 ms | 499,880 | byte pricing keeps SHA256 bounded |
| `large_push_dup_cat` | 1 tx / 1 input, single thread | ~14.47 ms | 499,936 | max-stack-depth large data edge case |
| `single_stark` | 1 tx / 1 input, single thread | ~9.76 ms | 473,148 | baseline cost for single input verification |

## Hash Opcode Pricing

The benchmark exposed that unpriced per-byte hashing allowed a single transaction to pack too much work into a block.

This PR prices hash input bytes as:

- BLAKE2b / BLAKE2b-with-key: `2` units per byte
- BLAKE3 / BLAKE3-with-key: `1` unit per byte
- SHA256: `1` unit per byte

Observed before/after numbers:

- BLAKE2b single-tx storm: ~30 ms before byte pricing, now ~10 ms
- BLAKE3 single-tx storm with zero byte pricing: ~12 ms, now ~5.9 ms
- SHA256 single-tx storm: ~10.5 ms before byte pricing, now ~5.2 ms

BLAKE2b was materially heavier than BLAKE3/SHA256, hence the higher byte scale.

## Data / Stack Edge Case

`large_push_dup_cat` was tuned to stress stack data usage: it expands one large pushed value via CAT, then DUPs it up to `MAX_STACK_SIZE`.

That workload now lands around ~14.5 ms, making it the largest single-tx no-parallelism edge in this suite. This raises a follow-up pricing question: stack byte movement may deserve `2` units per byte rather than `1`. I tend to think this is ok as an extreme edge case and that we can keep `1`.

## Application Signal

Real Silverscript-style applications, including the chess benchmark, perform well and are not the bottleneck (checked via a branch on silverscript). The challenging cases are adversarial single-tx constructions that remove parallelism and maximize hashing or stack-data movement.

## Mergeset Parallelism 

Note that future non-consensus changes can introduce parallelism over the mergeset (by analyzing the tx dep relations in the mergeset upfront) if we ever get to a point where single-tx heavy-blocks are the dominant factor.  

---

Additional coverage: the suite now includes an ECDSA 2-input workload comparable to `schnorr_2in1`, plus Groth16 cases with three total verifications per block. In the long run, `groth16_3tx` measured ~7.06 ms single-thread, ~4.67 ms at 2 threads, ~2.91 ms at 4 threads, and ~3.23 ms at 8 threads; the single-input `groth16_3tags_single_tx` case measured ~7.03 ms. `ecdsa_2in1` measured ~8.68 ms single-thread and ~2.11 ms at 8 threads, broadly tracking the existing signature-validation workload shape.


consensus/src/processes/transaction_validator/tx_validation_in_isolation.rs
```diff
@@ -458,16 +462,16 @@ mod tests {
...
+            (SUBNETWORK_ID_NATIVE, TX_VERSION_TOCCATA, false),
+            (SUBNETWORK_ID_COINBASE, TX_VERSION_TOCCATA, false),
+            (reserved_shape, TX_VERSION_TOCCATA, false),
+            (user_lane, TX_VERSION, false),
+            (user_lane, TX_VERSION_TOCCATA, true),
         ] {
             let mut tx = tx_with(subnetwork_id, version);
             tx.gas = 1;
-            assert_eq!(super::check_gas(&tx), expected);
+            assert!(matches!((super::check_gas(&tx), gas_allowed), (Ok(()), true) | (Err(TxRuleError::TxHasGas(_)), false)));
```
biryukovmaxim: May it use assert_march as on 367th lane

biryukovmaxim (dismissed):

someone235 (dismissed):

someone235 (approved):

jdanthenat:
"Been going through your recent commit series on the pricing branch — the Groth16 benchmark additions are directly relevant to something I'm building.
USIL is a cross-chain state commitment protocol targeting Kaspa as a native vProg. The core operation is SHA256(chain_id || block_height || state_root) — 44 input bytes + 32 output = 76 script units per commitment verification under the new pricing model.
A few things from your commits that directly inform the design:
The groth16_3tx vs groth16_3tags_single_tx benchmark split is exactly the parallelism tradeoff USIL needs to understand — each bridge transaction generates one Groth16 proof, so the multi-tx parallel path is the realistic production shape.
The DEFAULT_GAS_PER_LANE_LIMIT raise to 1B gives USIL's user-lane commitment transactions enough gas headroom for fine-grained per-proof pricing without hitting the ceiling.
SHA-256 coming in as the cheapest hash opcode at 1 unit/byte makes the commitment format economically sound at L1.
Spec and running simulation (Ghost→Shadow→Live + CDAG layer) here if useful context: github.com/jdanthenat/USIL — also posted a draft on research.kas.pa under L1/L2 protocols."  https://github.com/jdanthenat/USIL

PR #972, opened 2026-04-29
SMT: bound prune branch deletes by per-block max touched depth

biryukovmaxim:


consensus/smt-store/src/streaming_import/mod.rs
```diff
@@ -138,31 +162,114 @@ pub fn streaming_import(
...
+        db,
+        stores,
+        block_hash,
+        &mut pending,
+        &mut sink,
+        &mut score_batch,
+        &mut score_batch_count,
+        max_batch_entries,
+    )?;
+    debug_assert!(pending.is_empty(), "post-finalize: every pending score-index entry must be resolvable");
```
michaelsutton: shouldn't this be a full inconsistency error? 

eg
```
if !pending.is_empty() {
    return Err(StreamError::Sink(StoreError::DataInconsistency(format!(
        "streaming import finished with {} unresolved score-index entries",
        pending.len()
    ))));
}
```

crypto/smt/src/streaming.rs
```diff
@@ -90,6 +90,15 @@ pub trait MergeSink {
 
     /// Write a collapsed single-leaf subtree at the given position.
     fn write_collapsed(&mut self, branch_key: BranchKey, leaf: CollapsedLeaf, blue_score: u64) -> Result<(), Self::Error>;
+
+    /// Called when a leaf's collapsed-leaf state transitions into a merge —
+    /// once per leaf, at the moment its first/only `merge_pair` runs.
```
michaelsutton: the comment doesn't mention `chain_up` so it's slightly misleading

crypto/smt/src/streaming.rs
```diff
@@ -90,6 +90,15 @@ pub trait MergeSink {
...
+    /// `seal_depth` is `parent_depth + 1` clamped to `DEPTH - 1`.
+    ///
+    /// Default no-op so test sinks need not implement it.
+    fn record_seal(&mut self, _lane_key: Hash, _blue_score: u64, _seal_depth: u8) -> Result<(), Self::Error> {
```
michaelsutton: do we really need to return an err here? I'd remove it until actually needed

michaelsutton (dismissed):

michaelsutton (approved):

PR #981, opened 2026-05-03
Add configurable user agent admission rules

michaelsutton:
PR for configurable P2P user-agent admission rules.

This adds `--ua-rule`, supporting allow/reject rules over either regexes or parsed semantic versions:

```bash
--ua-rule='allow;regex:(^|/)(kaspad|kaspa-dnsseeder):'
--ua-rule='reject;ver:kaspad<1.1.1'
```

Policy order is:

```text
1. If allow rules exist and none match, reject.
2. If any reject rule matches, reject.
3. Otherwise accept.
```

A practical TN12 example is running only:

```bash
--ua-rule='reject;ver:kaspad<1.1.1'
```

With no allow rules configured, the default admission policy remains open. The rule only rejects user-agent segments named exactly `kaspad` whose version is below `1.1.1`, which effectively filters older kaspad nodes from previous TN12 launches. It does not reject other entities such as DNS seeders or crawlers, for example `/kaspa-dnsseeder:.../`, because those do not have the `kaspad` user-agent name.

Version rules use semver parsing, but compare only `major.minor.patch`, so suffixes like `1.1.1-toc.1` compare as `1.1.1`.

Log example:
```
2026-05-03 13:21:14.118+03:00 [INFO ] Rejecting peer <IP>:16311 because user agent matched rejection rule `reject;ver:kaspad<1.1.1`: /kaspad:1.1.0-rc.3/kaspad:1.1.0-rc.3/
```


protocol/flows/src/user_agent_rule.rs
```diff
@@ -0,0 +1,241 @@
...
+        matches!(
+            (self, cmp),
+            (VersionOp::Lt, Ordering::Less)
+                | (VersionOp::Lte, Ordering::Less | Ordering::Equal)
+                | (VersionOp::Gt, Ordering::Greater)
+                | (VersionOp::Gte, Ordering::Greater | Ordering::Equal)
+                | (VersionOp::Eq, Ordering::Equal)
+                | (VersionOp::Ne, Ordering::Less | Ordering::Greater)
+        )
+    }
```
IzioDev: we could use semver notation, to re-use semver [internal ops](https://docs.rs/semver/latest/semver/enum.Op.html) and built-in [comparison](https://docs.rs/semver/latest/semver/struct.Version.html#method.cmp_precedence)

protocol/flows/src/user_agent_rule.rs
```diff
@@ -0,0 +1,241 @@
...
+    pub fn empty() -> Self {
+        Self { allow: vec![], reject: vec![] }
+    }
+
+    pub fn parse_lossy(rules: &[String]) -> Self {
+        let mut rule_set = Self::empty();
+        for rule in rules {
+            match UserAgentRule::parse(rule) {
+                Ok(rule) => rule_set.add(rule),
+                Err(err) => log::warn!("Ignoring invalid user agent rule `{}`: {}", rule, err),
```
someone235: Why not return an error?

someone235: We usually use `warn!` over `log::warn!`

protocol/flows/src/user_agent_rule.rs
```diff
@@ -0,0 +1,241 @@
...
+            return None;
+        }
+
+        let version = version.split_once('(').map_or(version, |(version, _)| version).trim();
+        Version::parse(version).ok()
+    })
+}
+
+#[cfg(test)]
+mod tests {
```
someone235, line 182: You should also test some version with parenthesis

someone235 (changes requested):

protocol/flows/src/user_agent_rule.rs
michaelsutton: cli kaspad errs on rule parsing. Here we only make one more check for cases where daemon was created programmatically. It's called from `FlowContext::new` which has no err path.

someone235 (dismissed):

protocol/flows/src/user_agent_rule.rs
michaelsutton: from the PR desc: Version rules use semver parsing, but compare only `major.minor.patch`, so suffixes like `1.1.1-toc.1` compare as `1.1.1`.

I tend towards keeping this, bcs intra-version comparison feels wrong to me here (and yet can always be achieved via the regex option)

michaelsutton: still - checking if reuse is relevant regardless

michaelsutton: done https://github.com/kaspanet/rusty-kaspa/pull/981/changes/f4ee1144fcaed665bc6a07c6027dc6274c2153f4

IzioDev (approved):

PR #984, opened 2026-05-04
chore(docker): bump alpine rust version to 1.91

IzioDev:
built image here: https://hub.docker.com/layers/kkluster/rusty-kaspa/toccata-1/images/sha256-e2285c87d11e49a41f7fd81cbe3ab9d040fe7fba8dbd1689d7323798b6f55fd4

### DIFF
--- a/docker/Dockerfile.kaspa-wallet
+++ b/docker/Dockerfile.kaspa-wallet
@@ -1,5 +1,5 @@
 # ---------------------------------------- Chef image -------------------------------------------
-FROM rust:1.90-alpine AS chef
+FROM rust:1.91-alpine AS chef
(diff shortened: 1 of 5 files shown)


IzioDev:
maybe the builder and the runner should have same pinned version too (3.23 here), I'm unsure how alpine versioning system works but i'll look into it.
so far, no runtime issues observed.

coderofstuff:
is this still a relevant PR? it's pointed at tn12

IzioDev:
> is this still a relevant PR? it's pointed at tn12

Yes, without it, docker builds fail.

someone235 (approved):

Issue #986, opened 2026-05-05
connectionmanager: `handle_connection_requests` holds `connection_requests` lock across N `connect_peer().await` calls


demisrael:
## Summary

`ConnectionManager::handle_connection_requests` (`components/connectionmanager/src/lib.rs:116-160` at master `4969c6c3`) acquires `connection_requests.lock().await` and holds it across N sequential `p2p_adaptor.connect_peer(address.to_string()).await` calls — one per pending request.

While the dial loop is in flight, every other consumer of `connection_requests.lock()` blocks. With N pending requests and a typical `connect_peer` cost dominated by TCP-handshake + p2p-handshake latency, the lock is held for tens of seconds under realistic conditions; under degraded-network or unresponsive-peer conditions, longer.

## Operator impact

The lock is shared by these consumers (citations against `kaspanet/rusty-kaspa@4969c6c3` in `components/connectionmanager/src/lib.rs`):

- [`add_connection_request`](https://github.com/kaspanet/rusty-kaspa/blob/4969c6c3/components/connectionmanager/src/lib.rs#L106-L112) (decl ~106; lock acquired ~108) — RPC `AddPeer` registration writes here
- [`is_permanent`](https://github.com/kaspanet/rusty-kaspa/blob/4969c6c3/components/connectionmanager/src/lib.rs#L330-L332) (decl ~330; lock acquired ~331) — `connection_requests.contains_key(address)` read path
- [`is_banned`](https://github.com/kaspanet/rusty-kaspa/blob/4969c6c3/components/connectionmanager/src/lib.rs#L325-L328) (decl ~325) — calls `is_permanent`, so blocks transitively on the same lock
- [`ip_has_permanent_connection`](https://github.com/kaspanet/rusty-kaspa/blob/4969c6c3/components/connectionmanager/src/lib.rs#L335-L337) (decl ~335; lock acquired ~336) — iterates the map for the permanent-IP scan

Operator-perceived effect during a dial cycle:

- An RPC `AddPeer` issued mid-cycle stalls until all dials in the snapshot finish.
- Inbound connections held while ban-check waits behind the dial cycle (delayed accept, not denied — but delayed).
- A new permanent peer registration cannot promote the IP-permanent map until the cycle finishes.

[...]

Issue #985, opened 2026-05-05
Flaky integration test: daemon_utxos_propagation_test panics with `Elapsed` under CI load


demisrael:
## Summary

`daemon_integration_tests::daemon_utxos_propagation_test` intermittently
panics under CI load with:

```
thread 'daemon_integration_tests::daemon_utxos_propagation_test' (...) panicked at testing/integration/src/common/utils.rs:207:117:
called `Result::unwrap()` on an `Err` value: Elapsed(())
```

The test passes locally and on lightly-loaded runners; it flakes on busy
GitHub-hosted runners. Two parallel CI runs at the same head SHA have been
observed where one PASSed and the other FAILed at this exact line — the
canonical signature of a timing flake rather than a deterministic regression.

## Root cause

`testing/integration/src/common/utils.rs:195-214` waits for a `BlockAdded`
(then a `VirtualDaaScoreChanged`) notification across the
`consensus → protocol → grpc → async_channel` pipeline using a single-shot
`tokio::time::timeout(...)` with `Duration::from_millis(10_000)`:

https://github.com/kaspanet/rusty-kaspa/blob/master/testing/integration/src/common/utils.rs#L195-L214

```rust
let timeout_duration = Duration::from_millis(10_000);

for client in listening_clients.iter() {
    let block_daa_score: u64 =
        match timeout(timeout_duration, client.block_added_listener().unwrap().receiver.recv()).await.unwrap().unwrap() {
            Notification::BlockAdded(BlockAddedNotification { block }) => {
                assert_eq!(block.header.hash, block_hash);
                block.header.daa_score
            }
            _ => panic!("wrong notification type"),
        };
    match timeout(timeout_duration, client.virtual_daa_score_changed_listener().unwrap().receiver.recv()).await.unwrap().unwrap() {
        Notification::VirtualDaaScoreChanged(VirtualDaaScoreChangedNotification { virtual_daa_score }) => {
            assert_eq!(virtual_daa_score, block_daa_score + 1);
        }
        _ => panic!("wrong notification type"),
    }
}
```

Two `.unwrap()`s are stacked on each timeout site:

| Layer | Type | Panic mode |
|---|---|---|
| outer `.await.unwrap()` | `Result<_, tokio::time::error::Elapsed>` | timeout elapsed before the inner future resolved |
| inner `.unwrap()` | `Result<Notification, async_channel::RecvError>` | channel closed (teardown) |

[...]

PR #988, opened 2026-05-06
DNS-resolved peer addresses

### LINKED ISSUE #947: kaspad - addpeer argument should accept hostnames
When passing hostnames in `--addpeer`, the program stops with "invalid IP address syntax", this is an issue especially in cloud native environments.

I'd like the program to resolves IP at runtime

Resolve IP as part of a pre-start step:
```sh
peer_ip="$(getent ahostsv4 devnet-kaspad-0.devnet-kaspad.kaspa.svc.cluster.local | awk 'NR==1 { print $1 }')"
```

demisrael:
Closes #947

## Summary

Adds hostname support to `--addpeer` / `--connect` and the `AddPeer` RPC,
mirroring Bitcoin Core's tolerant `addnode` semantics. Hostnames are
resolved at dial time (not parse time), tolerated when transient DNS
failures occur, and re-resolved periodically in the background so a
running node tracks DNS changes (round-robin rotation, IP failover) for
the lifetime of the configured peer. kaspad starts cleanly even when
every configured peer is unresolvable; the RPC `AddPeer` returns
`Ok(AddPeerResponse {})` on unresolvable hosts and registers the host for
periodic retry. IP-literal callers see no change to the wire format or
the on-the-wire behaviour: a new client emits a v1 Borsh frame for
IP-literal `AddPeer` requests, byte-identical to what every previously-
shipped client emitted, so an old server in a mixed-version cluster
keeps decoding new clients without re-handshake.

## Motivation

`--addpeer` and the `AddPeer` RPC accept only IP literals today, which
forces operators to either (a) hard-code peer IPs that may rotate via
DNS, (b) resolve hostnames out-of-band before each kaspad invocation, or
(c) sit at startup waiting for DNS. None of these match the operator
ergonomics of Bitcoin Core, where `-addnode=host` accepts hostnames,
`-addnode` is tolerant of unresolvable hosts at startup, and the daemon
re-resolves hostnames periodically as part of its connection
maintenance loop.

This PR closes that gap, taking Bitcoin Core's `ThreadOpenAddedConnections`
behaviour as the parity reference. See *Bitcoin parity rationale* below
for the source citations.

## Bitcoin parity rationale

Reference: `bitcoin/bitcoin/src/net.cpp:ThreadOpenAddedConnections`
(~line 2974, HEAD `8f4a3ba`).

| Bitcoin Core behaviour | This PR |
|---|---|
| `-addnode=host` accepts hostnames | `--addpeer=host[:port]` accepts hostnames + RFC 1123 validation |
| `-addnode` tolerates unresolvable hosts at startup | kaspad starts even when every configured peer is unresolvable; resolution failures are non-fatal |
| `addnode` re-resolves hostnames inside `ThreadOpenAddedConnections` (per dial attempt; loop sleeps 60 s after activity, 2 s when idle) | New connection-manager **periodic** refresh task at a configurable cadence (default 600 s). Re-resolve also fires on dial failure for IPv4 / IPv6 path symmetry. The fixed periodic cadence is a kaspa-specific design choice rather than strict parity |
| Periodic refresh is silent at the log level (operators monitor metrics, not log noise) | Periodic refresh failure path emits no log macro; observable via `peer_hostname_resolutions_total{status="failed",trigger="periodic"}` counter increment + per-entry saturating `refresh_failures` counter |
| Bitcoin's `addnode` never drops a configured peer entry on resolution failure (operator's choice always wins) | This PR preserves the same invariant: hostname entries are removed only by explicit `mark_stale`, never by failure paths |
| Bitcoin's `addnode` wire format is never broken without a protocol-version bump | Same here: IP-literal `AddPeer` Borsh frames stay v1 byte-identical; v2 is emitted **only** when the payload is a hostname |

The logging discipline is a **strict superset** of Bitcoin Core's:
this PR additionally emits a one-shot `info!` on initial-resolve success
and a one-shot `warn!` on initial-resolve failure (so the operator sees
the boot-time outcome), then goes silent for steady-state periodic
ticks (matching Bitcoin Core).

## Commits walk-through

This branch is 7 commits ahead of `master`, organised by component for
ease of review. **Suggested merge strategy: squash-merge** — keeps the
per-component story visible in the PR diff while landing a single
bisectable commit on `master`.

| # | Subject | Scope |
|---|---|---|
| 1 | `kaspa-utils: add PeerEndpoint, RFC 1123 validator` | New `PeerEndpoint` enum (Address vs Hostname variants) + RFC 1123 hostname validator + `resolve_with_timeout` wrapper around `tokio::time::timeout` for the resolver path |
| 2 | `rpc-core: AddPeerRequest borsh v1<->v2 dispatch` | Borsh wire-format dispatch for `AddPeerRequest`. Encode side: `Address` payloads emit v1 (byte-identical to the legacy frame); `Hostname` payloads emit v2 (canonical Display string). Decode side: server accepts both versions and dispatches on the version tag. Mixed-version clusters interoperate without coordination |
| 3 | `connectionmanager: hostname registry + refresh + reconciliation` | New `HostnameRegistry` + `HostnameResolver` (test seam) + `refresh_all` reconciliation; periodic-refresh task spawning. `last_refresh: Option<tokio::time::Instant>` (monotonic; `None` is the explicit refresh-ASAP sentinel set by `mark_stale`) |
| 4 | `kaspad: wire hostname endpoints + refresh interval + metrics` | CLI flag `--hostname-refresh-interval=<seconds>` (env `KASPAD_HOSTNAME_REFRESH_INTERVAL_SEC`, default 600); kaspad startup wires endpoints and spawns the refresh task; `HostnameMetrics` counter / gauge surface; `pending_refreshes` returns per-entry `ResolveTrigger` so a wakeup arm that aggregates cadence-elapsed entries with mark-stale entries records each under its own eligibility reason |
| 5 | `testing-integration: hostname endpoint integration tests` | Integration tests for IP-literal regression + hostname add + periodic refresh observed + zero-interval disables refresh + cross-version Borsh wire round-trip |
| 6 | `connectionmanager: tolerate unresolvable hosts at addpeer registration` | Tolerant-registration: unresolvable hostnames register with an empty resolved set; the periodic refresh task picks them up. Drops the resolve-failure error variant from the public API. Introduces `kaspad_lib::daemon::DaemonOverrides` test seam (the resolver injection point) and removes the resolver field from the operator-facing `Args` struct |
| 7 | `kaspa-utils: gate tokio dep on non-wasm32 targets` | `PeerEndpoint::resolve()` and the `resolve_with_timeout` helper are the only tokio-dependent surface in `kaspa-utils`, and both are `#[cfg(not(target_arch = "wasm32"))]`-gated. The whole tokio dep moves under `[target.'cfg(not(target_arch = "wasm32"))'.dependencies]` and inherits `workspace = true`. The wasm32 build of utils declares no tokio dep at all and so cannot inherit any wasm32-incompatible features (`net`, `rt-multi-thread`) from the workspace tokio entry. No member Cargo.toml carries a direct `tokio = "X.Y.Z"` pin; the workspace tokio entry is unchanged. The wasm32 build of `kaspa-wrpc-wasm` (transitive dependency on `kaspa-utils`) compiles cleanly |

## Backward compatibility

| Surface | Pre-PR behaviour | Post-PR behaviour | Compat |
|---|---|---|---|
| `--addpeer=<ip[:port]>` (IP literal) | Accepted; eager resolve | **Unchanged** — same parse, same dial path | ✅ |
| `--addpeer=<host[:port]>` (hostname) | **Rejected** at parse | Accepted; resolved at dial time; periodic refresh | New surface |
| `--connect=<host[:port]>` | **Rejected** at parse | Accepted | New surface |
| `RPC AddPeer { peer_address, is_permanent }` (IP literal) | Returns `Ok` | **Unchanged** | ✅ |
| `RPC AddPeer { peer_address, is_permanent }` (hostname literal) | Returns `Err` | **Returns `Ok`** even when the host is unresolvable; the host is registered for periodic retry | **Behavioural change** — operator-facing |
| Borsh `AddPeerRequest` wire (IP-literal payload) | v1 (IP-only fixed layout) | **v1 byte-identical** — encoded as `(u16=1, RpcContextualPeerAddress, bool)`; an old v1 server decodes a new client's IP-literal request unchanged | ✅ |
| Borsh `AddPeerRequest` wire (hostname payload) | (rejected at parse pre-PR) | v2 — encoded as `(u16=2, String, bool)` where the string is the canonical `PeerEndpoint::Display` form; only emitted when the payload is a hostname (which a v1 server cannot accept anyway, since hostnames need the new resolver loop on the server side) | New surface |
| Borsh `AddPeerRequest` wire (decoder) | v1 only | **v1 ↔ v2 accepted** by version-tag dispatch; new server decodes old client unchanged | ✅ |
| gRPC `AddPeerRequestMessage` schema | `address: string`, `isPermanent: bool` (IP-only contract) | Same fields; the `address` contract is widened to also accept RFC 1123 hostnames, parsed via `RpcPeerEndpoint::from_str` at decode time | Old gRPC servers parse hostnames as IPs and reject — clients sending hostnames must target a server that runs a kaspad with the new parser; IP-literal traffic is unchanged in either direction |
| Address manager keying / ban list keying | IP-keyed | **Unchanged** — no hostname leakage into gossip economy | ✅ |

The only operator-facing breaking change is the RPC `AddPeer` `Err` →
`Ok` transition for unresolvable hosts. Callers that previously
inspected the `Err` to fail-fast will now see `Ok`; the recommended
migration is to monitor the new
`peer_hostname_resolutions_total{status="failed",trigger="initial"}`
counter for boot-time resolution health, or the new `peer_hostname_active`
gauge to confirm hostnames have at least one resolved address.

### Configuration

| Surface | Default | Range | Disable |
|---|---|---|---|
| CLI `--hostname-refresh-interval=<seconds>` | 600 | u64 seconds | `0` disables the periodic refresh task entirely (initial resolve + dial-failure-triggered re-resolve still work) |
| Env `KASPAD_HOSTNAME_REFRESH_INTERVAL_SEC` | 600 | u64 seconds | `0` (same) |

### `--addpeer` / `--connect` permanence semantics

Endpoints supplied via `--addpeer` and `--connect` on the kaspad CLI are
**always registered with `is_permanent: true`** in the connection
manager — matching the operator intent of "this is a configured peer,
keep trying to connect" and matching Bitcoin Core's `-addnode` /
`-connect` semantics. The `is_permanent: false` path is reserved for
the RPC `AddPeer` caller, which exposes the flag explicitly on the
request. Re-registration of a hostname already present in the registry
is **first-write-wins** (Bitcoin Core parity with `CConnman::AddNode`):
a subsequent `AddPeer { is_permanent: true }` against a hostname already
registered by `--addpeer` (no flag change required) returns `Ok` without
mutating the existing entry; operators who need to flip the
permanence flag must remove and re-register the host.

**Variant asymmetry note.** First-write-wins is the contract for the **Hostname** arm of `add_endpoint_request`. The **Address** arm delegates to the existing `add_connection_request`, which retains its prior **last-write-wins** semantics: a second `--addpeer=<ip>` with a different `is_permanent` flag mutates the existing entry's permanence. Operators who rely on first-write-wins should pass hostnames where the contract holds, or guard against duplicate IP-literal `--addpeer` invocations at the operator-tooling layer.
### Metrics (new, surfaced via `kaspa-cli getmetrics`)

| Metric | Type | Labels | Purpose |
|---|---|---|---|
| `peer_hostname_resolutions_total` | counter | `status` ∈ {ok, failed}, `trigger` ∈ {initial, initial_retry, dial_failure, periodic} | Per-trigger × per-status resolution outcomes; load-bearing observable for periodic refresh health |
| `peer_hostname_active` | gauge | — | Number of hostname entries currently registered |
| `peer_hostname_resolved_addrs` | gauge | — | Total resolved socket addresses across all hostnames |

The `peer_hostname_resolutions_total` counter is the canonical
observable for periodic refresh health (matching Bitcoin Core's
silent-on-periodic discipline). Operators monitoring connectivity
SHOULD scrape this counter rather than the kaspad log. The `trigger`
label reflects the per-entry eligibility reason — `Periodic` for
cadence-elapsed entries that resolved successfully at least once,
`DialFailure` for entries the dial loop flagged after a dial against
a previously-resolved socket failed, and `InitialRetry` for entries
that have never resolved successfully (registered with an empty /
errored initial resolve, retried at the next refresh tick) —
independent of which wakeup arm woke the refresh task.

`PeerHostnameMetrics` wire-flat field names (gRPC `.proto` field
names + JSON-RPC keys) drop the redundant `peerHostname` /
`peer_hostname_` prefix; the type name `PeerHostnameMetrics` already
carries the group, consistent with `BandwidthMetrics`,
`ConnectionMetrics`, `ProcessMetrics`, etc. The Prometheus-canonical
metric NAMES above remain unchanged at the spec / KIP layer; a
future Prometheus exporter prepends the type-group prefix at
presentation time.

### Unit tests

`components/connectionmanager/src/hostname.rs` — registry + metrics
+ refresh-cadence + per-entry-trigger predicate (25 tests).

`utils/src/networking.rs` — `PeerEndpoint` parse + RFC 1123 validation
+ `resolve_with_timeout` wrapper (timeout firing under
`tokio::time::pause`, payload propagation, io-error mapping).

`rpc/core/src/model/message.rs` — Borsh wire-format tests including
the v1 byte fixture decode, the new client's `Address`-variant emit
being byte-identical to the v1 fixture, and the v2 string-payload
layout for hostname-variant requests.

All unit tests run without real DNS and without real time.

### Integration tests

16 integration tests across two files (11 + 5):

`testing/integration/src/daemon_integration_tests.rs` (11 tests):
- `kaspad_addpeer_ipv4_unchanged` / `_ipv6_unchanged` — IP-literal regression guards
- `kaspad_addpeer_hostname_localhost_starts` — kaspad with `--addpeer=localhost` boots and registers the resolved socket
- `kaspad_addpeer_hostname_unresolvable_keeps_running` — kaspad with `--addpeer=nonexistent.invalid` boots cleanly and stays alive
- `kaspad_addpeer_hostname_resolve_timeout_metric` — synthetic `PeerEndpointResolveError::Timeout` injected via `FakeHostnameResolver::set_timeout`; asserts `initial_failed >= 1` post-registration and `initial_retry_failed >= 1` after a cadence tick
- `kaspad_periodic_refresh_observed` — counter `trigger="periodic"` ≥ 2 over a controlled window
- `kaspad_periodic_refresh_disabled_with_zero_interval` — `--hostname-refresh-interval=0` keeps the periodic-counter at zero
- `kaspad_dial_failure_re_resolves` — dial-failure path triggers a re-resolve
- `dns_volatility::kaspad_unresolvable_periodic_no_log_spam` — single-arm DNS-failure suite locking silent-on-no-delta + per-tick periodic-counter increments under the new `initial_retry_failed` label
- `dns_volatility::kaspad_unresolvable_to_resolvable_toggle` — 4-arm cycle (unresolvable → resolvable → unresolvable → resolvable) locking the intra-arm-silence contract and the metric interleave (`initial_retry_*` on the seeded-unresolvable arm, `periodic_*` from then on)
- `dns_volatility::kaspad_resolvable_to_unresolvable_toggle` — 4-arm cycle from the opposite phase (resolvable seed) locking the same intra-arm-silence contract and the `last_resolved`-preserved-on-failure invariant

`testing/integration/src/rpc_tests.rs` (5 tests):
- `rpc_add_peer_ipv4_unchanged` — IP-literal regression
- `rpc_add_peer_v1_wire_compat` — old v1 client buffer decodes through the new server's `AddPeer` wRPC path
- `rpc_add_peer_v2_emit_default` — new client's hostname-variant request emits v2 over the wire
- `rpc_add_peer_hostname_localhost_success` — happy-path RPC `AddPeer` with a hostname
- `rpc_add_peer_hostname_unresolvable_accepts` — RPC `AddPeer` with an unresolvable hostname returns `Ok` and registers the entry for periodic retry

### CI gates (verified at HEAD)

- `cargo fmt --all -- --check` — clean
- `cargo clippy --workspace --tests --benches --examples --no-deps -- -D warnings` — clean
- `cargo nextest run` — touched-crate sweep PASS — 238 passed, 1 skipped. The 1-skipped is `kaspa_utils::expiring_cache::tests::test_expiring_cache`, a pre-existing `#[ignore]`-marked test in `utils/src/expiring_cache.rs:84` that is timing-sensitive (doc-commented as "sensitive to runtime machine times so there's no point in keeping it part of CI; should be activated if the ExpiringCache struct changes") and is unrelated to this PR. Verified at HEAD `9b84f01f`.
- `cargo test --doc --workspace` — clean
- `cargo check --workspace --all-targets` — clean
- `cargo check -p kaspa-utils --target wasm32-unknown-unknown` — clean
- `cargo check -p kaspa-wrpc-wasm --target wasm32-unknown-unknown` — clean

### Live-WAN smoke test

Executed 2026-05-04 against a real testnet-10 peer:

- Resolve → P2P-connected: Δ +6.3 s
- Sustained IBD with the remote peer: 3+ min observation window;
  uptime `03:07` at observation; ~14 350 headers / 8 s typical IBD throughput
- Periodic refresh (60 s interval — short cadence chosen for the smoke window; default is 600 s): 3 ticks fired during observation;
  zero subsequent `addpeer:` log lines (matching the silent-on-stable
  discipline)

The smoke test exercised the full path: hostname parse → DNS resolve →
TCP connect → kaspad p2p protocol register → IBD start → sustained
IBD progress → periodic refresh of an unchanging single-A-record DNS
result.

### 5-hour node-log observation

Executed 2026-05-05/06 on a `tn-10` foundation host (foundation:16211)
with the post-fix HEAD `9b84f01f`:

- Wall-clock: 5h 0m 42s of `kaspad` operation under sustained IBD
- 9 outbound peers established within the first 12s (the addpeer-resolved
  peer plus 8 from DNS seeders)
- Headers IBD complete at t=3h 8m
- Periodic-refresh cadence (600s default) fired ~30 times across the run;
  zero subsequent `addpeer:` log lines (silent-on-stable discipline holds)
- **Zero warning-pattern hits** across 2581 log lines: no `WARNING`,
  no `BUG:`, no `Oops`, no `Call trace`, no `panic`, no `RUST_BACKTRACE`,
  no `leaked`, no `deadlock`
- Clean shutdown via `SIGTERM` with all four servers (P2P / gRPC / wRPC
  Borsh / wRPC JSON) reaching their respective `stop` paths

Logs and snapshots captured under a per-task working directory on the validator host; observation procedure is reproducible by an upstream maintainer with SSH access to a tn-10 node and the build artifacts from `cargo build --workspace --release`.

## Out of scope (intentional, future PRs)

- Address-manager keying remains IP-keyed. No hostname
  leakage into the gossip economy.
- Ban list keying remains IP-keyed. No hostname-aware ban.
- No IPv4/IPv6 preference policy.
- No per-peer refresh-interval override (single global interval).
- No on-the-wire hostname propagation to peers (hostnames are operator-
  side configuration; peers only see resolved IPs).
- The `kaspa-cli` interactive `metrics` view (driven by the historical
  `MetricsData` projection) does not yet thread `peer_hostname_metrics`;
  operators see the new buckets via `kaspa-cli getmetrics` (which reads
  the wire response directly) but the long-running monitoring view will
  surface them in a follow-up PR that extends the `MetricsData` /
  `MetricsSnapshot` taxonomy.

## References

- Issue: #947
- Bitcoin Core parity reference: `bitcoin/bitcoin/src/net.cpp:ThreadOpenAddedConnections` (~line 2974, HEAD `8f4a3ba`)
- Bitcoin Core compat reference: `bitcoin/bitcoin/src/net.cpp:CConnman::AddNode` (~line 3740, same HEAD) — `addnode` accepts unresolvable hosts; the periodic-retry thread re-resolves
- RFC 1123 hostname syntax (validator citation)
- Upstream issue [`kaspanet/rusty-kaspa#985`](https://github.com/kaspanet/rusty-kaspa/issues/985) — `daemon_utxos_propagation_test` flake (timeout-elapsed unwrap; observed during this PR's CI; filed during cycle)
- Upstream issue [`kaspanet/rusty-kaspa#986`](https://github.com/kaspanet/rusty-kaspa/issues/986) — `handle_connection_requests` lock-hold across N `connect_peer().await` calls (FIXME at `lib.rs` cross-references this issue; the fix is out-of-scope for this PR)

## Reviewer notes

- The 7 commits are logically organised per-component; reviewers who
  prefer per-commit review should start with #1 (`kaspa-utils:
  PeerEndpoint`) and proceed in order. Reviewers who prefer top-down
  review should focus on commit #3 (`connectionmanager: hostname
  registry + refresh + reconciliation`) and commit #6
  (`connectionmanager: tolerate unresolvable hosts`) for the core
  algorithm and tolerance semantics, plus commit #2 (`rpc-core:
  AddPeerRequest borsh v1<->v2 dispatch`) for the wire-format compat
  rule.
- The test seam is the `HostnameResolver` trait
  (`components/connectionmanager/src/hostname.rs`) plus the
  `kaspad_lib::daemon::DaemonOverrides` struct that carries it through
  to integration tests; production callers pass
  `DaemonOverrides::default()`. A `#[cfg(feature = "test-utils")]`
  `test_support` module ships with this PR for downstream test
  consumers.
- The post-implementation refactor split hostname-specific state and
  behaviour into a dedicated `hostname.rs` module (separate from
  `lib.rs`); unit tests live alongside the registry, exercised
  without time-control mechanisms (no `tokio::time::pause`/`advance`
  needed in the connection-manager tests; the timeout-wrapper tests
  in `kaspa-utils` do use `tokio::time::pause` to drive the
  `tokio::time::timeout` arm in <100 ms of virtual time).

---


kaspador:
## PR Review — DNS-resolved peer addresses (`kas-947-addpeer-hostnames`)

**Author:** @demisrael | **Base:** `master` | **+4,567 / −85** across **28 files**

---

## Overview

This PR adds hostname support to `--addpeer` / `--connect` and the `AddPeer` RPC, mirroring Bitcoin Core's `addnode` semantics. Hostnames are resolved at dial time (not parse time), kaspad starts cleanly even when all configured peers are unresolvable, and a background refresh task periodically re-resolves hostnames. The implementation is well-engineered — it has a pluggable `HostnameResolver` trait for testing, race-detection in the registry apply path, a RAII dedup guard, per-entry metric labeling, and a Borsh v1/v2 dispatch for backward compatibility.

---

## ✅ Strengths

- **Architecture is clean and well-separated.** The new `hostname.rs` module is fully self-contained; the connection manager's existing IP-literal path is unchanged.
- **Race-detection in `apply_refresh_results` is correct.** The `snapshot_last_refresh != entry.last_refresh` check prevents a concurrent `mark_stale` from being silently overwritten by an in-flight `Ok` result.
- **Test seam design is excellent.** `HostnameResolver` trait + `FakeHostnameResolver` + `DaemonOverrides` struct makes integration tests fully deterministic without real DNS.
- **Lock ordering is sound and documented.** `hostname_state → connection_requests`; no path nests them in the opposite order; no deadlock risk added.
- **WASM-32 gating is correct.** `PeerEndpoint::resolve()` and `resolve_with_timeout` are `#[cfg(not(target_arch = "wasm32"))]`-gated; the enum and its fmt/parse impls remain wasm-compatible.
- **Wire-format backward compat is thoughtful.** v1 `Address` frames are byte-identical to the historical layout; v2 only emits for `Hostname` payloads; old servers silently decode new clients for IP-literal requests.

---

## 🐛 Bugs & Correctness Issues

### 1. Race window in `apply_refresh_results` `(None, None)` interleaving — *documented but has correctness impact*

**File:** `components/connectionmanager/src/hostname.rs`

When `snapshot_last_refresh = None` AND `entry.last_refresh = None` at apply time (two consecutive `mark_stale` calls during the resolve window), `None == None` evaluates `true`. The second `mark_stale`'s reason (`DialFailure` overwriting `InitialRetry`) is silently cleared by the in-flight `Ok`. The PR documents this in the inline comment and calls it "brief metric attribution skew (one cadence tick)." The correctness claim is accurate — the entry recovers — but there is a subtle operational gap: if the node is rapidly cycling `DialFailure` marks (many transient dial failures), the `dial_failure_*` counter will under-count during the window where an in-flight resolve was concurrent with the mark. Over-counting is impossible, but under-counting of `dial_failure_*` can mislead operators using that counter as a health signal.

**Suggested mitigation:** Document this limitation explicitly in `HostnameMetrics` (as a caveat on the `dial_failure_ok` / `dial_failure_failed` fields), or tighten the race-check to also compare `stale_reason` (adds one `PartialEq` comparison). The current behavior is safe; the metric nuance just needs visibility.

---

### 2. `kaspad_periodic_refresh_disabled_with_zero_interval` test — missing assertion

**File:** `testing/integration/src/daemon_integration_tests.rs`

The test name (and the PR description: *"zero-interval disables refresh"*) implies that when `hostname_refresh_interval_sec = 0`, the `periodic_ok` and `periodic_failed` counters stay at zero. The test body only verifies that the daemon starts and `handle_message_id()` works — it never asserts counter values:

```rust
// test body
let rpc_client = kaspad.start().await;
assert!(rpc_client.handle_message_id());
rpc_client.disconnect().await.unwrap();
kaspad.shutdown();
// ↑ no counter assertion
```

The companion test `kaspad_addpeer_hostname_unresolvable_keeps_running` *does* assert `initial_failed >= 1`. `kaspad_periodic_refresh_disabled_with_zero_interval` should similarly assert `periodic_ok == 0` and `periodic_failed == 0` after a sleep window.

---

### 3. `GetMetricsResponse` v2 frame → old wRPC client — implicit framing assumption

**File:** `rpc/core/src/model/message.rs`

When a new server sends a v2 `GetMetricsResponse` to an old v1-only wRPC client, the old client's deserializer reads `version = 2`, processes all pre-existing fields, and **stops before the `peer_hostname_metrics` tail**. The wRPC Borsh transport is length-prefixed per-message, so the unread tail bytes are harmless (framing discards them). But this assumption is:
- Not tested anywhere in the PR (the forward-compat test only covers v1 → new server; the reverse (v2 → old client) can only be tested via a hand-built buffer mimicking the old client's deserializer, which is missing).
- Not documented at the point of the `store!(u16, &2, writer)?` line.

**Suggested fix:** Add a test that builds a v2 buffer, simulates the old v1 decoder (reads past `custom_metrics`, asserts cursor is short of EOF by the `peer_hostname_metrics` byte count), and confirms that the framing layer's remainder equals `sizeof(Option<PeerHostnameMetrics>)`.

---

### 4. `HostnameMetrics` duplicate match block — refactor opportunity & latent inconsistency risk

**File:** `components/connectionmanager/src/hostname.rs` (lines ~257–301)

`HostnameMetrics::record` and `HostnameMetrics::get` contain identical 8-arm `match (trigger, status)` blocks. If a new `ResolveTrigger` arm is added later (e.g., `ForceRefresh`), the `ResolveTrigger` is marked `#[non_exhaustive]`, so the match will need updating in two places. Extract a private `counter_for` method:

```rust
fn counter_for(&self, trigger: ResolveTrigger, status: ResolveStatus) -> &AtomicU64 {
    match (trigger, status) { ... }
}
```

This is a code-quality issue but carries a real maintenance risk given `#[non_exhaustive]`.

---

## ⚠️ Potential Issues

### 5. Behavioral change in `AddPeer` RPC — `Err → Ok` for unresolvable hosts

**File:** `rpc/service/src/service.rs`

Previously `AddPeer` returned `RpcError` if the host was unresolvable. Now it returns `Ok(AddPeerResponse {})` in that case and registers the host for periodic retry. This is the documented Bitcoin Core parity target, but it is a **breaking operator contract** for any tooling that inspects `AddPeer` errors to fail-fast or alert. The PR mentions the migration path (monitor `peer_hostname_resolutions_total{status="failed",trigger="initial"}`), but no changelog entry or deprecation notice is visible in the diff. Consider adding a `CHANGELOG` entry or a kaspad startup `warn!` message when a user-supplied hostname fails initial resolution, linking to the metric.

---

### 6. DNS amplification risk with many hostname peers

**File:** `kaspad/src/args.rs`

`MIN_HOSTNAME_REFRESH_INTERVAL_SEC = 30` applies per-node, not per-hostname. An operator with `--addpeer=host1 --addpeer=host2 ... --addpeer=hostN` and `--hostname-refresh-interval=30` will generate `N` DNS lookups every 30 seconds. For large N (e.g., 100+ manually specified peers), this can saturate the local DNS resolver. Consider adding a log warning at startup when `len(add_peers) * refresh_interval` exceeds some threshold.

---

### 7. `handle_connection_requests` FIXME — new code extends the blast radius

**File:** `components/connectionmanager/src/lib.rs`

The new `refresh_hostnames` Phase 4 (`connection_requests.lock().await`) will block for the full duration of the dial loop whenever Phase 4 and the dial loop coincide. This means during a dial-heavy period, Phase 4's delta reconciliation — and thus DNS rotation — is delayed. The `hostname_state` lock is unlocked well before Phase 4 (dropped at end of Phase 3b), so `add_endpoint_request` and metric snapshots remain unblocked. The fix (issue #986) is appropriately out-of-scope, but the FIXME cross-reference should note the Phase 4 interaction explicitly.

---

### 8. `use std::str::FromStr` placement in test file

**File:** `testing/integration/src/daemon_integration_tests.rs`

The `use std::str::FromStr;` is placed mid-file (after the first three test functions that call `PeerEndpoint::from_str`). Move this `use` to the top of the hostname-tests block or file.

---

## 📋 Minor Style / Code Quality

| # | File | Issue |
|---|------|-------|
| 1 | `components/connectionmanager/src/hostname.rs` | `HostnameRequest` has all `pub` fields but no `#[non_exhaustive]`. Any struct literal construction outside the crate is source-breaking on new field additions. Consider `pub(crate)` fields or a builder. |
| 2 | `components/connectionmanager/src/lib.rs` | `drop(requests)` is explicit but the variable goes out of scope naturally a few lines later. Pick one style and be consistent (there are two instances). |
| 3 | `rpc/core/src/model/message.rs` | `PeerHostnameMetrics` derives both `BorshSerialize` / `BorshDeserialize` (auto-derive) *and* implements `Serializer` / `Deserializer` manually (versioned workflow-serializer path). A doc comment clarifying which derive is used where would reduce reviewer confusion. |
| 4 | `rpc/grpc/core/proto/rpc.proto` | `PeerHostnameMetrics` proto field ordering is history-ordered (`{Initial=1-2, DialFailure=3-4, Periodic=5-6, InitialRetry=7-8}`) while the Rust struct is semantic-grouped. The `reserved 9, 10` only works if future authors know to pair new trigger labels at `{odd, odd+1}` field positions. Add an explicit comment: "next trigger pair: fields 9 (`*Ok`) and 10 (`*Failed`)." |
| 5 | `components/connectionmanager/src/hostname.rs` | `refresh_enabled` is a one-liner (`!interval.is_zero()`) used at one call site. Consider inlining it. |

---

## 🔒 Security / Operational Observations

- **No hostname leakage into gossip.** Address manager and ban list remain IP-keyed. ✅
- **No SSRF vector.** The resolver calls `tokio::net::lookup_host` (OS DNS), not HTTP. ✅
- **Rate-limiting.** `MIN_HOSTNAME_REFRESH_INTERVAL_SEC = 30` and TOML config validation both guard against accidental resolver hammering. ✅
- **Timeout bounding.** `PEER_ENDPOINT_RESOLVE_TIMEOUT` (~5s) bounds each DNS call; `join_all` concurrency means N hostnames resolve concurrently, not serially. ✅

---

## Summary

| Category | Count |
|---|---|
| 🐛 Bugs / correctness | 2 (race metric skew — documented; weak test) |
| ⚠️ Potential issues | 4 (v2 compat assumption, behavioral change, DNS amplification, FIXME blast radius) |
| 🎨 Code quality | 5 (duplicate match, use placement, doc gaps) |

**Overall verdict:** The PR is well-engineered and the core logic is correct. The two most actionable pre-merge items are:

1. **Add counter assertions to `kaspad_periodic_refresh_disabled_with_zero_interval`** — the test as written doesn't validate its stated contract.
2. **Document or test the `GetMetricsResponse` v2 → old-client framing assumption** — the forward-compat is safe but the assumption is implicit.

The `Err → Ok` behavioral change on `AddPeer` for unresolvable hosts is the highest operator-impact change and should have a CHANGELOG or release note entry regardless of correctness.

demisrael:
Thanks for the thorough review — much appreciated, especially the framing
analysis on the metrics response and the operational notes. All items are
addressed in the updated branch. The branch was also rebased onto current
master; the original seven commits are replayed unchanged (base adaptations
only), and the review changes are contained in the
`connectionmanager: address hostname feedback` commit. A follow-up commit
(`kaspa-utils: feature-gate resolver for no_std`) gates the resolver's tokio
dependency behind a default-on feature so the crate keeps building for
`no_std` targets.

**Bugs & correctness**

1. *`(None, None)` race / metric attribution* — kept the single
   `last_refresh` equality predicate. The apply-path documentation spells
   out the `dial_failure_*` attribution window, the one-cadence-tick bound,
   and the recovery guarantee. Comparing `stale_reason` as a second
   predicate would couple the race check to reason semantics for what is a
   bounded, correctness-neutral skew — happy to revisit if you feel
   strongly.
2. *Zero-interval test* — now asserts `periodic_ok == 0` and
   `periodic_failed == 0` after an observation window, in addition to the
   startup assertion.
3. *v2 frame → old v1 reader* — added a test that drives a v2 buffer
   through the exact field sequence an old reader uses and asserts the
   unread remainder is precisely the `Option<PeerHostnameMetrics>` tail;
   the serializer now documents the length-delimited-envelope assumption at
   the version-tag write.
4. *Duplicate eight-arm match* — extracted `counter_for`; `record` and
   `get` share it.

**Potential issues**

5. *`AddPeer` `Err → Ok` behavioral change* — this is the intended
   Bitcoin Core `addnode` parity: once hostname syntax is accepted, the
   RPC returns success and the connection manager owns DNS retry (on the
   configured cadence and on dial failures). Parse errors still fail
   immediately with a typed error. The contract is now documented at the
   proto definition and in the RPC handler, and kaspad logs a warning
   whenever a user-supplied hostname fails initial resolution, so the
   change is visible to operators at both the API and the log surface.
   Summary for tooling authors: poll `peer_hostname_resolutions_total`
   (`trigger=initial`, `status=failed`) instead of relying on an `AddPeer`
   error.
6. *DNS amplification* — kaspad now warns at startup when 25 or more
   hostname endpoints are configured together with a sub-60s nonzero
   refresh interval, naming the per-cadence lookup cost. Threshold logic is
   unit-tested.
7. *`handle_connection_requests` FIXME* — the cross-reference now names the
   refresh Phase-4 delta-reconciliation interaction explicitly.
8. *`FromStr` placement* — moved to the file's top-level import block.

**Style**

1. `HostnameRequest` fields are now `pub(crate)`, and the redundant stored
   `host` field is gone (the registry key is authoritative).
2. Lock-scope style made consistent: the guard-bounded sections are
   extracted into named helper methods rather than mixing explicit `drop`
   with open scopes.
3. Added a doc comment distinguishing the derived Borsh impls (generic
   type-bound callers, mock tests) from the manual versioned serializer
   used for RPC response framing.
4. The proto `reserved 9, 10` comment now states the intended use: the next
   trigger-label `…Ok` / `…Failed` pair.
5. Kept `refresh_enabled` as a named helper: it is doc-commented and
   unit-tested, and the call site reads better with the intent named —
   can inline if you prefer.

PR #990, opened 2026-05-06
Add rocknroll SMT prune scanner

michaelsutton:
Introduce the rocknroll dev DB scanner crate and add an smt_prune_scan binary for readonly consensus DB inspection.

The scanner resolves appdir-based consensus DBs through meta, skips archive nodes, verifies pruning stability, mirrors the inclusive SMT prune cutoff, and reports stale branch, lane, and score-index entries.

Also add readonly RocksDB opening support and a readonly MultiConsensusManagementStore constructor for scanner use.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3582,6 +3582,20 @@ dependencies = [
 ]
 
 [[package]]
+name = "kaspa-rocknroll"
+version = "2.0.0"
+dependencies = [
+ "clap",
+ "kaspa-consensus",
+ "kaspa-consensus-core",
+ "kaspa-database",
+ "kaspa-hashes",
+ "kaspa-smt-store",
+ "kaspad",
+ "thiserror 2.0.18",
+]
+
+[[package]]
 name = "kaspa-rpc-core"
 version = "2.0.0"
 dependencies = [
@@ -5984,20 +5998,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7043b63bd0cd1aaa628e476b80e6d4023a3b50eb32789f2728908107bd0c793a"
 dependencies = [
  "libc",
-]
-
-[[package]]
-name = "rocknroll"
-version = "1.1.1-toc.1"
-dependencies = [
- "clap",
- "kaspa-consensus",
- "kaspa-consensus-core",
- "kaspa-database",
- "kaspa-hashes",
- "kaspa-smt-store",
- "kaspad",
- "thiserror 2.0.18",
 ]
 
 [[package]]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5987,6 +5987,20 @@ dependencies = [
 ]
 
 [[package]]
+name = "rocknroll"
+version = "1.1.1-toc.1"
+dependencies = [
+ "clap",
+ "kaspa-consensus",
+ "kaspa-consensus-core",
+ "kaspa-database",
+ "kaspa-hashes",
+ "kaspa-smt-store",
+ "kaspad",
+ "thiserror 2.0.18",
+]
+
+[[package]]
 name = "rocksdb"
 version = "0.24.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -56,6 +56,7 @@ members = [
     "components/consensusmanager",
     "bridge",
     "database",
+    "database/rocknroll",
     "crypto/txscript",
     "crypto/txscript/errors",
     "testing/integration",
--- a/consensus/src/consensus/factory.rs
+++ b/consensus/src/consensus/factory.rs
@@ -219,8 +219,7 @@ impl MultiConsensusManagementStore {
         let mut metadata = self.metadata.read().unwrap();
         if metadata.is_archival_node != is_archival_node {
             metadata.is_archival_node = is_archival_node;
-            let mut batch = WriteBatch::default();
-            self.metadata.write(BatchDbWriter::new(&mut batch), &metadata).unwrap();
+            self.metadata.write(DirectDbWriter::new(&self.db), &metadata).unwrap();
         }
     }
 
@@ -447,5 +446,31 @@ impl ConsensusFactory for Factory {
             };
             write_guard.cancel_staging_consensus().unwrap();
         }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::MultiConsensusManagementStore;
+    use kaspa_database::{prelude::ConnBuilder, utils::get_kaspa_tempdir};
+
+    #[test]
+    fn archival_node_flag_persists_after_reopen() {
+        let db_tempdir = get_kaspa_tempdir();
+        let db_path = db_tempdir.path().to_owned();
+        let db = ConnBuilder::default().with_db_path(db_path.clone()).with_files_limit(10).build().unwrap();
+        let mut store = MultiConsensusManagementStore::new(db.clone());
+
+        store.set_is_archival_node(true);
+        assert!(store.is_archival_node().unwrap());
+        drop(store);
+        drop(db);
+
+        let db = ConnBuilder::default().with_db_path(db_path).with_create_if_missing(false).with_files_limit(10).build().unwrap();
+        let store = MultiConsensusManagementStore::new_readonly(db.clone());
+
+        assert!(store.is_archival_node().unwrap());
+        drop(store);
+        drop(db);
     }
 }
--- a/consensus/src/consensus/factory.rs
+++ b/consensus/src/consensus/factory.rs
@@ -83,13 +83,17 @@ pub struct MultiConsensusManagementStore {
 
 impl MultiConsensusManagementStore {
     pub fn new(db: Arc<DB>) -> Self {
-        let mut store = Self {
+        let mut store = Self::new_readonly(db);
+        store.init();
+        store
+    }
+
+    pub fn new_readonly(db: Arc<DB>) -> Self {
+        Self {
             db: db.clone(),
             entries: CachedDbAccess::new(db.clone(), CachePolicy::Count(16), DatabaseStorePrefixes::ConsensusEntries.into()),
             metadata: CachedDbItem::new(db, DatabaseStorePrefixes::MultiConsensusMetadata.into()),
-        };
-        store.init();
-        store
+        }
     }
 
     fn init(&mut self) {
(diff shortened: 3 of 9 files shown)


demisrael:
Thanks for adding the scanner. The shape is useful: a read-only DB path, meta-based active consensus resolution, explicit pruning-stability check, and an SMT stale-entry count are the right ingredients for this diagnostic. CI is green on HEAD, and the local source reports also record `cargo check -p rocknroll` and `cargo test -p rocknroll` passing, with no package tests currently present.

Issue summary: **2 HIGH**, **3 MEDIUM**, **0 LOW**.

### Algo correctness affirmations

The inclusive cutoff formula itself matches consensus pruning: the scanner computes `pp.blue_score - finality_depth - 1` via [`database/rocknroll/src/bin/smt_prune_scan.rs:68-71`](https://github.com/kaspanet/rusty-kaspa/blob/2e642e578c558ec92e8d02134106d6e3de2ffd2c/database/rocknroll/src/bin/smt_prune_scan.rs#L68-L71), which aligns with the pruning processor's cutoff use in [`consensus/src/pipeline/pruning_processor/processor.rs:581-593`](https://github.com/kaspanet/rusty-kaspa/blob/2e642e578c558ec92e8d02134106d6e3de2ffd2c/consensus/src/pipeline/pruning_processor/processor.rs#L581-L593). The new `MultiConsensusManagementStore::new_readonly` constructor also avoids the metadata initialization write that `new()` performs, which is the right direction for a scanner: [`consensus/src/consensus/factory.rs:84-97`](https://github.com/kaspanet/rusty-kaspa/blob/2e642e578c558ec92e8d02134106d6e3de2ffd2c/consensus/src/consensus/factory.rs#L84-L97).

### HIGH (must fix before merge)

**H1. Archive-node skipping is not reliable because the archival flag is not persisted.**

The scanner resolves archival status from the meta DB via `is_archival_node()` in [`database/rocknroll/src/db.rs:52-55`](https://github.com/kaspanet/rusty-kaspa/blob/2e642e578c558ec92e8d02134106d6e3de2ffd2c/database/rocknroll/src/db.rs#L52-L55), then skips only when `resolved.archival == Some(true)` in [`database/rocknroll/src/bin/smt_prune_scan.rs:43-48`](https://github.com/kaspanet/rusty-kaspa/blob/2e642e578c558ec92e8d02134106d6e3de2ffd2c/database/rocknroll/src/bin/smt_prune_scan.rs#L43-L48). But the setter updates a local `WriteBatch` and never writes that batch to RocksDB: [`consensus/src/consensus/factory.rs:218-224`](https://github.com/kaspanet/rusty-kaspa/blob/2e642e578c558ec92e8d02134106d6e3de2ffd2c/consensus/src/consensus/factory.rs#L218-L224).

After reopen, an archival DB can therefore be treated as non-archival. That can make the scanner report retained archive history as stale pruning garbage, or fail active-consensus resolution. Please persist the batch in `set_is_archival_node()` or use another durable source, and add a regression test that reopens archival metadata read-only and verifies the scanner skips.

**H2. Live-node scans use a RocksDB mode documented as undefined when the DB is also open read-write.**

The scanner opens the target consensus DB through `open_readonly_db()` in [`database/rocknroll/src/db.rs:72-73`](https://github.com/kaspanet/rusty-kaspa/blob/2e642e578c558ec92e8d02134106d6e3de2ffd2c/database/rocknroll/src/db.rs#L72-L73), which reaches `DB::open_for_read_only()` in [`database/src/db/conn_builder.rs:171-175`](https://github.com/kaspanet/rusty-kaspa/blob/2e642e578c558ec92e8d02134106d6e3de2ffd2c/database/src/db/conn_builder.rs#L171-L175). RocksDB documents that if the same DB is open via `Open` and `OpenForReadOnly` at the same time, the read-only instance has undefined behavior: [`rocksdb/include/rocksdb/db.h:185-195`](https://github.com/facebook/rocksdb/blob/410c5623195ecbe4699b9b5a5f622c7325cec6fe/include/rocksdb/db.h#L185-L195).

That is a problem for a diagnostic tool operators may naturally point at a running node's appdir. A scan can appear to work while reading an undefined or inconsistent view, making the stale-count output non-authoritative. Please either make the tool explicitly offline-only and fail fast when the DB appears live, or use a RocksDB mode intended for concurrent inspection such as secondary/follower with an explicit secondary path.

### MEDIUM

**M1. The default cutoff ignores runtime override params.**

The scanner derives `finality_depth` from network-default params only in [`database/rocknroll/src/bin/smt_prune_scan.rs:66-71`](https://github.com/kaspanet/rusty-kaspa/blob/2e642e578c558ec92e8d02134106d6e3de2ffd2c/database/rocknroll/src/bin/smt_prune_scan.rs#L66-L71). Kaspad can apply `--override-params-file` before building runtime config in [`kaspad/src/daemon.rs:297-322`](https://github.com/kaspanet/rusty-kaspa/blob/2e642e578c558ec92e8d02134106d6e3de2ffd2c/kaspad/src/daemon.rs#L297-L322).

For DBs created with override params, the scanner can count against the wrong SMT pruning boundary unless the operator manually supplies `--cutoff-blue-score`. Please either accept/apply the same override params input, require explicit cutoff when params are not proven, or at least print that default network params were assumed.

**M2. `build_readonly()` is not side-effect-free for missing paths.**

`ConnBuilder` defaults `create_if_missing` to true in [`database/src/db/conn_builder.rs:22-27`](https://github.com/kaspanet/rusty-kaspa/blob/2e642e578c558ec92e8d02134106d6e3de2ffd2c/database/src/db/conn_builder.rs#L22-L27). `default_opts!` passes that into RocksDB options in [`database/src/db/conn_builder.rs:157-160`](https://github.com/kaspanet/rusty-kaspa/blob/2e642e578c558ec92e8d02134106d6e3de2ffd2c/database/src/db/conn_builder.rs#L157-L160), and `build_readonly()` calls `open_for_read_only()` with those options in [`database/src/db/conn_builder.rs:171-175`](https://github.com/kaspanet/rusty-kaspa/blob/2e642e578c558ec92e8d02134106d6e3de2ffd2c/database/src/db/conn_builder.rs#L171-L175).

RocksDB's read-only open path creates the DB directory when `create_if_missing` is true: [`rocksdb/db/db_impl/db_impl_readonly.cc:260-276`](https://github.com/facebook/rocksdb/blob/410c5623195ecbe4699b9b5a5f622c7325cec6fe/db/db_impl/db_impl_readonly.cc#L260-L276). So a mistyped `--consensus-db` path can leave an empty directory behind from a tool advertised as read-only. Please force `create_if_missing(false)` in `build_readonly()` and add a missing-path test proving no directory is created.

**M3. DBs configured with a custom RocksDB WAL directory are not supported or rejected.**

Kaspad exposes `--rocksdb-wal-dir` in [`kaspad/src/args.rs:449-455`](https://github.com/kaspanet/rusty-kaspa/blob/2e642e578c558ec92e8d02134106d6e3de2ffd2c/kaspad/src/args.rs#L449-L455), carries it through `configure_rocksdb()` in [`kaspad/src/daemon.rs:256-263`](https://github.com/kaspanet/rusty-kaspa/blob/2e642e578c558ec92e8d02134106d6e3de2ffd2c/kaspad/src/daemon.rs#L256-L263), and applies it when opening DBs, for example meta DB open in [`kaspad/src/daemon.rs:390-396`](https://github.com/kaspanet/rusty-kaspa/blob/2e642e578c558ec92e8d02134106d6e3de2ffd2c/kaspad/src/daemon.rs#L390-L396). `ConnBuilder` maps that setting into RocksDB options in [`database/src/db/conn_builder.rs:137-155`](https://github.com/kaspanet/rusty-kaspa/blob/2e642e578c558ec92e8d02134106d6e3de2ffd2c/database/src/db/conn_builder.rs#L137-L155).

The scanner's read-only open path never accepts or forwards a WAL directory: [`database/rocknroll/src/db.rs:72-73`](https://github.com/kaspanet/rusty-kaspa/blob/2e642e578c558ec92e8d02134106d6e3de2ffd2c/database/rocknroll/src/db.rs#L72-L73). For nodes that used a custom WAL dir, especially after an unclean stop or when combined with the live-scan problem above, the scanner can fail or read a view that does not match kaspad's configured DB state. Please accept the same WAL-dir option for both meta and consensus DB opens, or explicitly reject DBs that require non-default RocksDB open configuration.

No LOW findings.

### Process / surface notes

**P1. Supported network contract should be documented before this outlives the current branch.**

`rocknroll` currently requires exactly one of `--testnet`, `--devnet`, or `--simnet` in [`database/rocknroll/src/args.rs:24-29`](https://github.com/kaspanet/rusty-kaspa/blob/2e642e578c558ec92e8d02134106d6e3de2ffd2c/database/rocknroll/src/args.rs#L24-L29). That matches the current covpp branch mainnet-disabled assertion in [`kaspad/src/args.rs:206-215`](https://github.com/kaspanet/rusty-kaspa/blob/2e642e578c558ec92e8d02134106d6e3de2ffd2c/kaspad/src/args.rs#L206-L215), so I am not treating it as a blocker here. If this tool is meant to remain useful beyond this branch, please document the supported-network contract.

### Architect-level positions on design topics

The primary design surface is the scanner's operating contract: what DB state it can inspect, under which RocksDB configuration, and how authoritative its counts are. H1 violates the archive-node skip claim. H2 and M2 violate the read-only inspection contract: H2 because concurrent live-node use is undefined by RocksDB, and M2 because a failed read-only open can still create a directory. M1 and M3 are configuration-provenance gaps: the scanner should either use the same params and RocksDB open configuration as kaspad, or clearly reject/default-label cases where it cannot prove that its view matches the node's configured DB state.

### Static evidence on HEAD

GitHub checks on HEAD are green: Check, Test Suite, Lints, no_std, WASM32 check/test, WASM32 SDK build, and Linux release build. The source review report also records local `cargo check -p rocknroll` passing and `cargo test -p rocknroll` passing with zero package/doc tests.

michaelsutton:
@demisrael fixed some of your comments / documented the gaps

biryukovmaxim (approved):

michaelsutton:
Note: this PR now also fixes persistence of the archival-node flag in `MultiConsensusManagementStore`:
https://github.com/kaspanet/rusty-kaspa/pull/990/changes/3c9caa35765d912a2d24ed189b2e10de0c9bea15

Previously, the setter updated metadata through a `WriteBatch` but never committed that batch, so the flag was lost after reopening the DB.

The only practical effect was that kaspad did not warn when restarting a node that had previously been archival without passing `--archival` again. After the fix, this warning is shown correctly:

`--archival is set to false although the node was previously archival. Proceeding may delete archived data. Do you confirm? (y/n)`

PR #991, opened 2026-05-06
UtxoIndex: Key UTXOs by DAA score, adds pagnated and cursor-based api, exposes this via `get_utxos_by_addresses_v2` RPC

D-Stacks:
## Summary
- Reorganizes the UTXO index storage to key entries by DAA score for efficient DAA-range scans.
- Adds utxoindex db migration logic on start-up. 
- Adds a paginated UTXO query over a set of script public keys and an inclusive DAA score range.
- Pagination uses a cursor (`start_address`, `start_daa_score`) plus an optional soft `limit`, returning `next_address`/`next_daa_score` for continuation.
- pagination allows for efficient polling based utxo-retrival solutions, where, for example, a wallet only needs to update over a "newer" daa score range.  
- Exposes the new behavior through `get_utxos_by_addresses_v2` and wires it through RPC/gRPC/wasm.
- Also included into kaspa-cli as a call. 

## Behavior
- Req: `from_daa_score`/`to_daa_score` define an inclusive range (defaults: None -> `0` and None -> `u64::MAX`).
- Req: `start_address`/`start_daa_score` resume from a specific point inside the range (defaults: None -> first address in `addresses` and None -> `from_daa_score` i.e. functions as an entry point to use the call).
- Resp: `next_address`/`next_daa_score` indicate the next cursor; `None` means no more pages.
- Req: `limit` is a soft cap on page size; a page may exceed it to finish the current SPK+DAA group (default: None -> u64::MAX).

## Testing
- `cargo test -p kaspa-utxoindex test_page_`

## Backwards Compatibility 
- old get_utxos_by_addresses is still fully functional. Only the order of retrieved utxos changes. But i do not expect clients to rely on this for anything. 
- new call keeps the list of addresses input for easier migration. 

## Please do Note: 
- Some changes that where added that might have altered this original description: https://github.com/kaspanet/rusty-kaspa/pull/991#issuecomment-5384810313


demisrael:
Thanks -- the structural change is clean. DAA score moves into the per-entry key, the bucket-prefixed range scan is built on `set_iterate_lower_bound` / `set_iterate_upper_bound`, the soft-limit-finishes-group state machine traces correctly across the cross-script boundary, and the upper bound at numeric edges (including `to_daa_score == u64::MAX`) is sharp. `cargo fmt --check`, `cargo clippy -p kaspa-utxoindex --all-targets --no-deps`, and `cargo test -p kaspa-utxoindex --lib test_page_` all pass on HEAD.

Three independent peer reviews ran (upstream review + algo peer review on the state machine + architect peer review on design surfaces); the issues consolidate into **three MEDIUM concerns to fix before merge** and a handful of LOW polish items.

### Algo correctness affirmations (independent re-derivation)

Before the issue list -- the algo peer review independently re-derived the four invariants the algorithm needs (group cohesion, next-cursor points to first NOT-included, next-cursor `None` iff exhausted, no-double-emit on round-trip) and traced each against the implementation. **No correctness blocker found at the algorithm level.** The affirmations worth surfacing for reviewer confidence:

- **Group cohesion.** The inner loop drains the current `(script, daa)` group fully before stopping ([`indexed_utxos.rs:329-348`](https://github.com/kaspanet/rusty-kaspa/blob/7372a8156b73f11eca05435370ca4c9fb12fa266/indexes/utxoindex/src/stores/indexed_utxos.rs#L329-L348)), so a single-page response can exceed `limit` to finish the group. The unbounded single-group case is L6 below (worth documenting on the RPC `limit` field).
- **Cross-script cursor carry-over.** The `current_group: Option<(usize, u64)>` state at [`:297`](https://github.com/kaspanet/rusty-kaspa/blob/7372a8156b73f11eca05435370ca4c9fb12fa266/indexes/utxoindex/src/stores/indexed_utxos.rs#L297) persists across the outer-loop boundary -- this is what makes `stop_after_group=true` set on the last entry of script A correctly fire on the first entry of script B. The `next_script_public_key` / `next_daa_score` always point to the first NOT-included `(script, daa)` group, never a pushed entry.
- **Cursor stability under address-list mutation.** The cursor is a full `ScriptPublicKey` (not an index), so it is robust to common between-page mutations: re-ordering the address list, adding addresses before or after the cursor, or removing addresses other than the cursor address. The only failure mode is **removing the cursor address itself** -- `position(==)` returns `None`, `start_index = script_public_keys.len()`, the outer loop is skipped entirely, and the response is an empty page with `next_*` both `None`. This is the same silent-empty-equals-exhausted collision M3 documents but reached via a legitimate-client-behaviour path; the RPC validation at [`service.rs:752-757`](https://github.com/kaspanet/rusty-kaspa/blob/7372a8156b73f11eca05435370ca4c9fb12fa266/rpc/service/src/service.rs#L752-L757) rejects this case for external clients, so the failure is contained at the RPC boundary today. Internal callers (mocks, downstream forks, future internal use) inherit the silent ambiguity -- L2 tracks the store-layer contract gap.
- **DAA-range bounds at numeric edges.** `seek_to = ((to+1)_be || OP_zeros)` ([`indexed_utxos.rs:318-319`](https://github.com/kaspanet/rusty-kaspa/blob/7372a8156b73f11eca05435370ca4c9fb12fa266/indexes/utxoindex/src/stores/indexed_utxos.rs#L318-L319)) is sharp at every numeric edge: `u64::MAX` (the `(to_daa_score < u64::MAX).then(...)` guard correctly falls back to the bucket-prefix upper), `u64::MAX - 1` (entry at exactly `(u64::MAX, OP_zeros)` correctly excluded), and `0` (entry at exactly `(1, OP_zeros)` correctly excluded). Verified against rocksdb 0.24's `set_iterate_range` + `set_iterate_upper_bound` override semantics directly (`db_options.rs:3920-3998`); the upper-bound override preserves the lower bound, and upper bound is exclusive.
- **Sort vs. DB byte order divergence is algorithmically inert** (not just currently inert). The pagination walks each bucket independently via one `seek_iterator` call per bucket -- there is no cross-bucket merge, so the cross-bucket walk order is purely a Rust-side decision and the divergence with DB lex order at `ScriptPublicKeyVersion >= 256` does not threaten any invariant. Confirms the documentation-only classification (L7).

### MEDIUM (recommend fix before merge)

**M1 -- migration nukes the DB dir but leaves the WAL subdir populated.** When `--rocksdb-wal-dir` is configured ([`args.rs:427-433`](https://github.com/kaspanet/rusty-kaspa/blob/7372a8156b73f11eca05435370ca4c9fb12fa266/kaspad/src/args.rs#L427-L433)), [`conn_builder.rs:139-155`](https://github.com/kaspanet/rusty-kaspa/blob/7372a8156b73f11eca05435370ca4c9fb12fa266/database/src/db/conn_builder.rs#L139-L155) creates `<wal_base>/<utxoindex_db_dir.file_name()>` and passes it to `opts.set_wal_dir(...)`. The migration block at [`daemon.rs:619-624`](https://github.com/kaspanet/rusty-kaspa/blob/7372a8156b73f11eca05435370ca4c9fb12fa266/kaspad/src/daemon.rs#L619-L624) only deletes `utxoindex_db_dir`, so old v0 WAL files survive. Reopening at the same db_path with the same wal_dir leaves RocksDB recovery free to scan and replay them against the freshly-created v1 schema -- `CompactUtxoEntry` shrunk from `(amount, block_daa_score, is_coinbase)` to `(amount, is_coinbase)` ([`indexed_utxos.rs:45-49`](https://github.com/kaspanet/rusty-kaspa/blob/7372a8156b73f11eca05435370ca4c9fb12fa266/indexes/core/src/indexed_utxos.rs#L45-L49)), so any replay is bincode-corrupt. The `remove_dir_all`-without-WAL pattern exists at [`daemon.rs:354`](https://github.com/kaspanet/rusty-kaspa/blob/7372a8156b73f11eca05435370ca4c9fb12fa266/kaspad/src/daemon.rs#L354) and [`:515`](https://github.com/kaspanet/rusty-kaspa/blob/7372a8156b73f11eca05435370ca4c9fb12fa266/kaspad/src/daemon.rs#L515) too, but those run on operator-explicit triggers; this PR makes the same shape the default upgrade path for every `--utxoindex` operator. Suggest adding a sibling `fs::remove_dir_all(wal_base.join(utxoindex_db_dir.file_name().unwrap()))` between the drop and the recreate.

**M2 -- `has_legacy_db_version` falls back to `tips_exist || supply_exists` when the version marker is missing** ([`store_manager.rs:43-67`](https://github.com/kaspanet/rusty-kaspa/blob/7372a8156b73f11eca05435370ca4c9fb12fa266/indexes/utxoindex/src/stores/store_manager.rs#L43-L67)). That heuristic only holds because `ensure_db_version_current` is the very last write of a successful resync (after `insert_circulating_supply` and `set_tips` at [`index.rs:212-217`](https://github.com/kaspanet/rusty-kaspa/blob/7372a8156b73f11eca05435370ca4c9fb12fa266/indexes/utxoindex/src/index.rs#L212-L217)) -- an invariant nowhere codified. Any future refactor that swaps the order will silently turn "v0 with marker missing" into "fresh DB indistinguishable from v0", flipping the migration branch the wrong way. Cleanest fix: write the version marker as the *first* write of a fresh resync (so a marker-present DB is well-defined and the tips/supply heuristic can be retired). Alternative: tighten the comment and pin the invariant at the resync write-order site so a future refactor cannot quietly invert it.

**M3 -- `get_utxos_by_addresses_v2` silently returns an empty page on store-layer error** -- indistinguishable from genuine end-of-stream. The shim at [`service.rs:279-301`](https://github.com/kaspanet/rusty-kaspa/blob/7372a8156b73f11eca05435370ca4c9fb12fa266/rpc/service/src/service.rs#L279-L301) drops the store-layer `StoreError` via `.unwrap_or_default()` and produces `OrderedUtxoSetByScriptPublicKeyPage { entries: [], next_*: None }`. The RPC handler then returns `GetUtxosByAddressesV2Response { entries: [], next_address: None, next_daa_score: None }` -- bit-for-bit identical to a successful exhausted-cursor response. A polling client cannot distinguish "stream exhausted" from "internal error, data silently dropped". The legacy `get_utxos_by_addresses` RPC has the same shape, but the v2 RPC's cursor contract makes the silent-empty-equals-exhausted collision worse: the contract explicitly promises a cursor-driven stream, and the silent-empty page violates it. Compounds with M1: a corrupt entry from WAL replay would either panic kaspad (via the `res.unwrap()` at [`indexed_utxos.rs:324`](https://github.com/kaspanet/rusty-kaspa/blob/7372a8156b73f11eca05435370ca4c9fb12fa266/indexes/utxoindex/src/stores/indexed_utxos.rs#L324)) or surface as silent data loss to the client. Suggest changing the shim signature to propagate the error and translating to `RpcError` at the handler boundary.

**L1 -- migration uses `unwrap()` on filesystem operations.** [`daemon.rs:622-624`](https://github.com/kaspanet/rusty-kaspa/blob/7372a8156b73f11eca05435370ca4c9fb12fa266/kaspad/src/daemon.rs#L622-L624) -- a graceful operator-facing message via `.expect("...")` is cheap, and the migration is now the common upgrade path.

**L2 -- store-level `start_address` not in `addresses` returns silently empty.** [`indexed_utxos.rs:285-292`](https://github.com/kaspanet/rusty-kaspa/blob/7372a8156b73f11eca05435370ca4c9fb12fa266/indexes/utxoindex/src/stores/indexed_utxos.rs#L285-L292) sets `start_index = script_public_keys.len()`, the loop body never runs, an empty page is returned. The RPC handler rejects this case explicitly at [`service.rs:752-757`](https://github.com/kaspanet/rusty-kaspa/blob/7372a8156b73f11eca05435370ca4c9fb12fa266/rpc/service/src/service.rs#L752-L757) so external clients are safe -- but the store-layer contract is implicit. Sister cases: store accepts `start_daa_score` without `start_script_public_key` (and silently applies it to the first sorted address) and `limit = Some(0)` (silently behaves as `limit = Some(1)` modulo group cohesion). Either return `StoreError` or doc-comment all three preconditions.

**L3 -- end-of-stream / limit-exactly-equals-total tests missing.** The three new tests cover ordering+range, soft-limit-finishes-group (mid-stream), and resume-from-cursor. Not covered: `limit` exactly equals total entry count -> `next_*` should both be `None`. The state-machine trace shows the natural-end branch handles this correctly, but a regression here would NOT be caught by any existing test. Specific recipe: populate N entries, call with `limit = Some(N)`, assert `page.next_script_public_key.is_none() && page.next_daa_score.is_none()`. Also worth: empty `addresses` set, `from_daa_score == to_daa_score` single-DAA range, the five RPC validation branches (one `#[test]` each).

**L4 -- no test exercises the v0 -> v1 migration end-to-end.** Construct a v0-shaped utxoindex DB, run migration, assert post-migration reads match. Highest test gap given M1/M2.

**L5 -- migration prompt does not warn about resync duration.** [`daemon.rs:620`](https://github.com/kaspanet/rusty-kaspa/blob/7372a8156b73f11eca05435370ca4c9fb12fa266/kaspad/src/daemon.rs#L620) -- the resync iterates the entire virtual UTXO set in chunks of 2048 and takes many minutes on a fully-synced mainnet node. Operators who hit the prompt and answer `y` and then see no progress for several minutes are likely to Ctrl+C -- landing back at M2 (interrupted resync, `supply_exists=true`, prompt-again-on-next-boot). One sentence ("this rebuild iterates the full UTXO set and may take several minutes on a fully-synced node") would prevent the panic-Ctrl+C -> corrupt-state loop.

**L6 -- single-group page size is unbounded.** The soft-limit semantic drains the current `(script, daa)` group fully -- a 100k-UTXO group at one DAA score will return all 100k entries in one call regardless of `limit`. This is a deliberate design choice (no resume-mid-group cursor; group cohesion is invariant), but the RPC `limit` field's doc-comment does not say so. Document on `get_utxos_from_script_public_keys_by_daa_score_page` and on the RPC `GetUtxosByAddressesV2Request::limit` field.

**L7 -- address sort vs. DB byte order divergence is documentation-only.** [`indexed_utxos.rs:279`](https://github.com/kaspanet/rusty-kaspa/blob/7372a8156b73f11eca05435370ca4c9fb12fa266/indexes/utxoindex/src/stores/indexed_utxos.rs#L279) sorts by `(version_numeric, script_bytes)`; the DB key prefix is `(version_le_bytes, script_len_le, script_bytes)`. They coincide for all currently-used `ScriptPublicKey` versions (= 0); they diverge for versions >= 256. Pagination correctness is unaffected (each script is iterated independently and the cross-bucket walk order is purely a Rust-side decision -- algorithmically inert per the algo peer review's §6). A one-line comment that pagination order is independent of DB byte order would close the gap.

**L8 -- CLI handler is positional and dense.** [`cli/src/modules/rpc.rs:316-358`](https://github.com/kaspanet/rusty-kaspa/blob/7372a8156b73f11eca05435370ca4c9fb12fa266/cli/src/modules/rpc.rs#L316-L358) -- 5 mandatory positional args after the address list. Easy to misorder; misordering surfaces as a parse error, not a usage error. Flag-based parsing (`--from`, `--to`, `--start-addr`, `--start-daa`, `--limit`) would be sturdier.

**L9 -- two unrelated trailing-whitespace cleanups bundled in.** [`utxo_set.rs:26`](https://github.com/kaspanet/rusty-kaspa/blob/7372a8156b73f11eca05435370ca4c9fb12fa266/consensus/src/model/stores/utxo_set.rs#L26) and [`api/mod.rs:54`](https://github.com/kaspanet/rusty-kaspa/blob/7372a8156b73f11eca05435370ca4c9fb12fa266/indexes/utxoindex/src/core/api/mod.rs#L54). Optional split into a hygiene PR.

### Process / surface notes

**P1 -- PR description under-represents the actual change surface.** The body's "Backwards Compatibility" claim ("old `get_utxos_by_addresses` is still fully functional. Only the order of retrieved utxos changes") is true at the wire level but not at the Rust crate level. The body does not mention:
- `database/src/access.rs:217-251` -- `seek_iterator` gained a `seek_to: Option<TKey>` parameter. Foundational DB-layer API change; consensus's `utxo_set.rs:156` was patched to pass `None`.
- New public types in `kaspa-index-core`: `UtxoEntryKeyData`, `OrderedUtxoCollection`, `OrderedUtxoSetByScriptPublicKey`, `OrderedUtxoSetByScriptPublicKeyPage`. `CompactUtxoCollection`'s key type changed from `TransactionOutpoint` to `UtxoEntryKeyData` -- a compile-break for any downstream crate that imports it and iterates `.keys()`.
- New trait method on `UtxoIndexApi` (`get_utxos_by_script_public_keys_by_daa_score_page`) -- public trait extension; downstream `UtxoIndexApi` impls (test mocks etc.) need updating.
- New registry prefix `DatabaseStorePrefixes::UtxoIndexDbVersion = 195`.
- `RPC_API_REVISION` bump 0 -> 1. Sent over the wire by [`service.rs:1318-1320`](https://github.com/kaspanet/rusty-kaspa/blob/7372a8156b73f11eca05435370ca4c9fb12fa266/rpc/service/src/service.rs#L1318-L1320) on every `GetServerInfo`, but the only client-side reader at [`processor.rs:513-516`](https://github.com/kaspanet/rusty-kaspa/blob/7372a8156b73f11eca05435370ca4c9fb12fa266/wallet/core/src/utxo/processor.rs#L513-L516) interpolates it into an error-message string only AFTER the version gate has failed -- bump is correct per convention but not gating client behaviour.

Suggest extending the PR description with a "Wider surface" section so reviewers don't miss them and downstream consumers know what to update on bump.

**P2 -- architect-level design questions worth considering for a follow-up** (none blocking this PR):
- The codebase already has a versioned-DB upgrade pattern in [`MultiConsensusManagementStore`](https://github.com/kaspanet/rusty-kaspa/blob/7372a8156b73f11eca05435370ca4c9fb12fa266/kaspad/src/daemon.rs#L435-L510) with a `'db_upgrade: while` loop and per-version handlers. The new utxoindex versioning is a binary "matches-or-not" check -- this locks the index into wipe-and-rebuild forever. Worth generalizing to a `VersionedIndexStore` trait before the next index store copies this shape.
- The migration lifecycle at [`daemon.rs:608-636`](https://github.com/kaspanet/rusty-kaspa/blob/7372a8156b73f11eca05435370ca4c9fb12fa266/kaspad/src/daemon.rs#L608-L636) duplicates the `ConnBuilder` chain (28 lines of two-site state) precisely where the WAL-cleanup-gap class lives (M1). An `UtxoIndex::open_or_migrate(builder)` API that owns the close/rmrf/reopen + WAL-subdir cleanup as a single transaction would structurally prevent that class.
- The new `seek_to: Option<TKey>` parameter on [`access.rs:seek_iterator`](https://github.com/kaspanet/rusty-kaspa/blob/7372a8156b73f11eca05435370ca4c9fb12fa266/database/src/access.rs#L217-L246) carries an unstated within-bucket precondition (the upper-bound override silently widens past the prefix range if the caller passes a key that exceeds it). Current pagination caller honours it; a future cross-bucket caller could silently violate prefix isolation. Either tighten the doc-comment or split into `seek_iterator_within_bucket` / `seek_iterator_range`.

### Static evidence on HEAD

```text
cwd: kaspanet/rusty-kaspa @ 7372a8156b73f11eca05435370ca4c9fb12fa266
cargo fmt --check -p kaspa-utxoindex -p kaspa-rpc-core -p kaspa-rpc-service -p kaspa-grpc-core -p kaspa-database  -- clean
cargo clippy -p kaspa-utxoindex -p kaspa-rpc-core -p kaspa-rpc-service -p kaspa-grpc-core -p kaspa-database --all-targets --no-deps  -- clean
cargo test  -p kaspa-utxoindex --lib test_page_  -- 3 passed (test_page_ordered_and_range_filtered, test_page_soft_limit_finishes_group, test_page_resumes_from_start_daa_score)
```

State machine traced by hand through `test_page_soft_limit_finishes_group` (cross-script `current_group` carry-over, group cohesion, `next_*` semantics); see the "Algo correctness affirmations" section above for the independent re-derivation summary.

indexes/utxoindex/src/stores/db_version.rs
```diff
@@ -0,0 +1,43 @@
...
+pub trait UtxoIndexDbVersionStore: UtxoIndexDbVersionStoreReader {
+    fn set(&mut self, version: u16) -> StoreResult<()>;
+    fn remove(&mut self) -> StoreResult<()>;
+}
+
+#[derive(Clone)]
+pub struct DbUtxoIndexDbVersionStore {
+    db: Arc<DB>,
+    access: CachedDbItem<u16>,
+}
```
IzioDev: We probably should use a struct instead of u16 and rename `Version` to `Meta`, this way it allows future cfg expansion easier, like we do with `MultiConsensusMetadata`

rpc/service/src/service.rs
```diff
@@ -708,6 +733,74 @@ NOTE: This error usually indicates an RPC conversion error between the node and
...
+        &self,
+        _connection: Option<&DynRpcConnection>,
+        request: GetUtxosByAddressesV2Request,
+    ) -> RpcResult<GetUtxosByAddressesV2Response> {
+        if !self.config.utxoindex {
+            return Err(RpcError::NoUtxoIndex);
+        }
+        if request.start_daa_score.is_some() && request.start_address.is_none() {
+            return Err(RpcError::InvalidGetUtxosByAddressesV2Request("start_daa_score requires start_address".to_string()));
+        }
```
biryukovmaxim: This check exists because the request type can express a state that has no meaning. The two cursor fields are not independent: a DAA score only means something relative to an address. Fold them into a single optional cursor so the illegal combination cannot be constructed:

```rust
pub struct UtxoPageCursor {
    pub address: RpcAddress,
    pub daa_score: u64,
}

pub struct GetUtxosByAddressesV2Request {
    ...
    pub cursor: Option<UtxoPageCursor>,
    ...
}

```

there are other places that can be fixed like that including response

rpc/service/src/service.rs
```diff
@@ -708,6 +733,74 @@ NOTE: This error usually indicates an RPC conversion error between the node and
...
+
+        if request.limit == Some(0) {
+            return Err(RpcError::InvalidGetUtxosByAddressesV2Request("limit must be greater than zero".to_string()));
+        }
```
biryukovmaxim: you can use Option<NonZeroU64> and such check will be part of type conversions in workflow and grpc

rpc/service/src/service.rs
```diff
@@ -708,6 +733,74 @@ NOTE: This error usually indicates an RPC conversion error between the node and
...
+
+        if matches!(
+            request.start_address.as_ref(),
+            Some(start_address) if !request.addresses.iter().any(|address| address == start_address)
+        ) {
+            return Err(RpcError::InvalidGetUtxosByAddressesV2Request("start_address must be included in addresses".to_string()));
+        }
```
biryukovmaxim: The cursor address is constrained to be one of the addresses already in the request, so it is redundant data. Carry the index into addresses instead:

```rust
pub struct UtxoPageCursor {
    pub address_index: u32,
    pub daa_score: u64,
}
```

indexes/utxoindex/src/stores/indexed_utxos.rs
```diff
@@ -176,6 +258,113 @@ impl UtxoSetByScriptPublicKeyStoreReader for DbUtxoSetByScriptPublicKeyStore {
...
+
+        if script_public_keys.is_empty() {
+            return Ok(OrderedUtxoSetByScriptPublicKeyPage::default());
+        }
+
+        let start_index = if let Some(start_script_public_key) = &start_script_public_key {
+            script_public_keys
+                .iter()
+                .position(|script_public_key| script_public_key == start_script_public_key)
+                .unwrap_or(script_public_keys.len())
```
biryukovmaxim: .When the cursor's script public key is not in the queried set, unwrap_or(script_public_keys.len()) makes the loop below skip every address and return an empty page with no next cursor, which a client cannot distinguish from "you have reached the end". A bad cursor should be an error, not silent truncation.

indexes/utxoindex/src/stores/indexed_utxos.rs
```diff
@@ -176,6 +258,113 @@ impl UtxoSetByScriptPublicKeyStoreReader for DbUtxoSetByScriptPublicKeyStore {
...
+        }
+
+        let script_count = script_public_keys.len();
+
+        let mut script_public_keys = script_public_keys.into_iter().collect::<Vec<_>>();
+        script_public_keys.sort_by(|a, b| a.version().cmp(&b.version()).then_with(|| a.script().cmp(b.script())));
+
+        if script_public_keys.is_empty() {
+            return Ok(OrderedUtxoSetByScriptPublicKeyPage::default());
+        }
```
biryukovmaxim, line 278: shouldn't we reject such queries with error?

indexes/core/src/indexed_utxos.rs
```diff
@@ -5,7 +5,31 @@ use std::collections::HashMap;
...
+
+/// A deterministic ordered list of UTXO collections keyed by [`ScriptPublicKey`].
+pub type OrderedUtxoSetByScriptPublicKey = Vec<(ScriptPublicKey, OrderedUtxoCollection)>;
+
+/// A page of ordered UTXOs with an optional cursor for the next group.
+#[derive(Clone, Debug, Default)]
+pub struct OrderedUtxoSetByScriptPublicKeyPage {
+    pub entries: OrderedUtxoSetByScriptPublicKey,
+    pub next_script_public_key: Option<ScriptPublicKey>,
+    pub next_daa_score: Option<u64>,
```
biryukovmaxim: same suggestion about pagecursor type

indexes/utxoindex/src/stores/indexed_utxos.rs
```diff
@@ -176,6 +258,113 @@ impl UtxoSetByScriptPublicKeyStoreReader for DbUtxoSetByScriptPublicKeyStore {
...
+        &self,
+        script_public_keys: ScriptPublicKeys,
+        from_daa_score: Option<u64>,
+        to_daa_score: Option<u64>,
+        start_script_public_key: Option<ScriptPublicKey>,
+        start_daa_score: Option<u64>,
+        limit: Option<u64>,
+    ) -> StoreResult<OrderedUtxoSetByScriptPublicKeyPage> {
+        let from_daa_score = from_daa_score.unwrap_or(0);
+        let to_daa_score = to_daa_score.unwrap_or(u64::MAX);
```
biryukovmaxim: The same defaulting already happens in the RPC handler Two copies of the same policy

Resolve the range once at the API boundary and pass a concrete range down, e.g. RangeInclusive<u64>

indexes/utxoindex/src/stores/indexed_utxos.rs
```diff
@@ -176,6 +258,113 @@ impl UtxoSetByScriptPublicKeyStoreReader for DbUtxoSetByScriptPublicKeyStore {
...
+        if from_daa_score > to_daa_score {
+            return Ok(OrderedUtxoSetByScriptPublicKeyPage::default());
+        }
+
+        let script_count = script_public_keys.len();
+
+        let mut script_public_keys = script_public_keys.into_iter().collect::<Vec<_>>();
```
biryukovmaxim: you can make ScriptPublicKeys alias to IndexSet. so it will be possible to use it as both set and sorted colelction. this way we can remove allocation/iteration for allocation here

indexes/utxoindex/src/stores/indexed_utxos.rs
```diff
@@ -176,6 +258,113 @@ impl UtxoSetByScriptPublicKeyStoreReader for DbUtxoSetByScriptPublicKeyStore {
...
+        script_public_keys.sort_by(|a, b| a.version().cmp(&b.version()).then_with(|| a.script().cmp(b.script())));
```
biryukovmaxim: Derive the ordering on ScriptPublicKey instead of hand-writing it

indexes/utxoindex/src/stores/indexed_utxos.rs
```diff
@@ -176,6 +258,113 @@ impl UtxoSetByScriptPublicKeyStoreReader for DbUtxoSetByScriptPublicKeyStore {
...
+            }
+
+            let script_public_key_bucket = ScriptPublicKeyBucket::from(&script_public_key);
+            let seek_from = UtxoEntryFullAccessKey::new_inner(effective_from_daa_score.into(), TransactionOutpointKey::EMPTY);
+
+            let seek_to = (to_daa_score < u64::MAX)
+                .then(|| UtxoEntryFullAccessKey::new_inner((to_daa_score + 1).into(), TransactionOutpointKey::EMPTY));
+
+            let mut ordered_entries = Vec::new();
+            for res in self.access.seek_iterator(Some(script_public_key_bucket.as_ref()), Some(seek_from), seek_to, usize::MAX, false)
```
biryukovmaxim: We build a fresh RocksDB iterator per requested address inside the loop. Since keys are [script public key bucket][daa score BE][outpoint] and both components are order preserving in the direction we scan, the whole page is one forward pass: open a single iterator over the store prefix and set_mode(IteratorMode::From(next_bucket_key, Forward)) to jump to each next address/

This also folds away most of the stop_after_group / should_stop / current_group bookkeeping with a single ordered stream, "finish the current DAA-score group after hitting the limit" is just "keep reading while the DAA score equals the last one", and the next cursor is the first key you did not consume.

biryukovmaxim (changes requested):

D-Stacks:
Some Changes with the last few commits:

### Merged with master:

PR was a bit outdated (I believe even on a pre-toccata branch)

### Internal Page generation handling:

`utxos_by_script_public_key_store.get_utxos_from_script_public_keys_by_daa_score_page` has been heavily reworked from the ground up, it now uses a new DB api `multi_range_seek_iterator` which is designed using a raw db iterator specifically designed to use a combination of forward traversal and seek operations to get many disjointed ranges in one pass of a db iterator. This greatly abstracts the stopping complexity into the db iterator itself. 

### Cursor / rpc-facing:

The start cursor now includes an (optional) exact transaction outpoint references, mostly this was done because it is a better fit for the `multi_range_seek_iterator`, without adding even more quirky complex logic on-top, it has the advantages of removing the "soft" daa-boundary filling limits, turning limit into an exact limit, at the cost of making the external rpc a bit more verbose. 

### Call signature:

Internal call signatures are now greatly reduced to a range, a cursor, and addresses, as suggested it also directly uses an indexset.

### Version DB:

The Utxoindex's version db is now a meta db, as requested. it currently only holds a versioning though. 
 
### Rpc wasm message:

Since I needed to re-write this anyway, I kind of simplified this to what a lot of the other messages are doing with:

`Ok(from_value(args.into())?) -> Ok(to_value(&args)?.into())`

But, I'll be honest, I don't really know if this is correct :man_shrugging:. I am not at depth with the wasm parts of the codebase.

indexes/utxoindex/src/stores/indexed_utxos.rs
D-Stacks: Idk, I am thinking we are doing what it says on the tin, querying an empty set? maybe higher up on rpc level we can do this. i don't feel like defining more errors then needed here, would rather leave it technical on this level.

rpc/service/src/service.rs
D-Stacks: technically this is actually a defined case, just didn't realize it until now, if start address is none we start with the first address, and then we start with the start daa score. it probably wouldn't be part of actual usage flow, but it can be specified, i guess.

rpc/service/src/service.rs
D-Stacks: I'd rather keep all these checks in one place.

indexes/utxoindex/src/stores/indexed_utxos.rs
D-Stacks: done, the utxoindex now takes an indexset upon entry into the call flow.

indexes/utxoindex/src/stores/indexed_utxos.rs
D-Stacks: done

indexes/utxoindex/src/stores/indexed_utxos.rs
D-Stacks: so I did this on the db api level with a new multi_range_seek api, it uses the db raw iterator and a mixture of forward iterations and seek to do everything in one pass.

indexes/utxoindex/src/stores/indexed_utxos.rs
D-Stacks: nevermind, since I kind of have a todo on changing the functionality of an empty set, to querying all utxos within a daa score range, it is probably better to reserve the todo functionality on an error instead of established behaviors.

indexes/utxoindex/src/stores/db_version.rs
D-Stacks: done

rpc/service/src/service.rs
D-Stacks: still going with the cursor idea though, just not on rpc level, since it seems like we just lean towards a standard of keeping things "flat" there.

rpc/service/src/service.rs
D-Stacks: the lower level api implicitly orders now, as such results are also returned in db order, not user defined order,  I kind want to do an ordering to pre-determine out-of-scope spks anyway, so i am staying with an explicit address. ordering also helps to close the bounds on the db iterator on successive calls. Technically the address can still be an index within the re-ordered set, but this means i need to order at the rpc api boundary, and i rather would do this close to the low level.

indexes/utxoindex/src/stores/indexed_utxos.rs
D-Stacks: I made this into a utxoindex error, defined as an invalid cursor, the same is returned for defining an out-of-bounds start_daa_score.

indexes/utxoindex/src/stores/indexed_utxos.rs
D-Stacks: done, range is defined when going into the utxoindex.

rpc/service/src/service.rs
D-Stacks: essentially if defining a start address as an index here, on the lower level i need to define the start address as an spk again, so for me this inversely feels redundant to do these conversions..

indexes/core/src/indexed_utxos.rs
```diff
@@ -1,14 +1,73 @@
...
 use serde::de::{Error as DeError, SeqAccess, Visitor};
 use serde::{Deserialize, Deserializer, Serialize};
 use std::collections::HashMap;
-use std::fmt;
+use std::fmt::{self, Display};
 
-// TODO: explore potential optimization via custom TransactionOutpoint hasher for below,
-// One possible implementation: u64 of transaction id xor'd with 4 bytes of transaction index.
-pub type CompactUtxoCollection = HashMap<TransactionOutpoint, CompactUtxoEntry>;
+pub type ScriptPublicKeyIndexSet = IndexSet<ScriptPublicKey>;
```
IzioDev: this aliased type only is used by the queries, it seems not the proper definition place.

what are the benefits to use `IndexSet` here instead of already existing `ScriptPublicKeys` (alias of `HashSet<ScriptPublicKey>`)

indexes/core/src/indexed_utxos.rs
```diff
@@ -1,14 +1,73 @@
...
+
+pub type CompactUtxoCollection = HashMap<UtxoEntryKeyData, CompactUtxoEntry>;
```
IzioDev: we probably should preserve the TODO

indexes/core/src/indexed_utxos.rs
```diff
@@ -1,14 +1,73 @@
...
+
+/// A deterministic ordered list of UTXOs keyed by [`UtxoEntryKeyData`].
+pub type OrderedUtxoCollection = Vec<(UtxoEntryKeyData, CompactUtxoEntry)>;
```
IzioDev: we can remove the `pub` marker, only used by the following type alias.

nit: i'd suggest to change the docs with ~"UTXOs returned in database-key order.", because in practice the type alias doesn't enforce deterministic ordering across value lifetime

if this is a requirement, we can wrap it like:
```rust
  pub struct OrderedUtxoCollection(
      Vec<(UtxoEntryKeyData, CompactUtxoEntry)>,
  );
```
and expose inner as read-only, with controlled construction, but it sounds over-kill here

note: same remark for `OrderedUtxoSetByScriptPublicKey` below

indexes/utxoindex/src/stores/indexed_utxos.rs
```diff
@@ -88,8 +93,64 @@ impl AsRef<[u8]> for TransactionOutpointKey {
...
+}
+
+impl AsRef<[u8]> for DaaScoreKey {
+    fn as_ref(&self) -> &[u8] {
+        &self.0
+    }
+}
+
+#[derive(Eq, Hash, PartialEq, Debug, Copy, Clone)]
+struct UtxoEntryInnerKey([u8; DAA_SCORE_KEY_SIZE + TRANSACTION_OUTPOINT_KEY_SIZE]);
```
IzioDev: i find this struct confusing, it's only used once within this file, isn't it leaner to just convert on the fly where needed? (in `get_utxos_from_script_public_keys`)

indexes/core/src/indexed_utxos.rs
```diff
@@ -1,14 +1,73 @@
...
+
+/// A deterministic ordered list of UTXO collections keyed by [`ScriptPublicKey`].
+pub type OrderedUtxoSetByScriptPublicKey = Vec<(ScriptPublicKey, OrderedUtxoCollection)>;
+
+#[derive(Debug, Clone, Hash, PartialEq, Eq)]
+pub struct UtxoEntryKeyData {
+    pub daa_score: u64,
+    pub transaction_outpoint: TransactionOutpoint,
+}
```
IzioDev: It's called `KeyData` but only contains a subset of them (no `spk`)

rpc/core/src/model/message.rs
```diff
@@ -1506,6 +1507,147 @@ impl Deserializer for GetUtxosByAddressesResponse {
...
+        let next_address = load!(Option<RpcAddress>, reader)?;
+        let next_daa_score = load!(Option<u64>, reader)?;
+        let _next_outpoint_hash = load!(Option<RpcHash>, reader)?;
+        let _next_outpoint_index = load!(Option<TransactionIndexType>, reader)?;
+        Ok(Self {
+            entries,
+            next_address,
+            next_daa_score,
+            next_outpoint_hash: load!(Option<RpcHash>, reader)?,
+            next_outpoint_index: load!(Option<TransactionIndexType>, reader)?,
```
IzioDev: reading again here sounds like a mistake, we already read line 1639 & 1640

indexes/utxoindex/src/stores/indexed_utxos.rs
```diff
@@ -176,6 +285,112 @@ impl UtxoSetByScriptPublicKeyStoreReader for DbUtxoSetByScriptPublicKeyStore {
...
+                (script_public_key, utxo_entry_key_data, value)
+            })
+            .chunk_by(|(spk, _, _)| spk.clone())
+            .into_iter()
+            .map(|(spk, chunk)| (spk, chunk.map(|(_, key_data, value)| (key_data, value)).collect::<Vec<_>>()))
+            .collect::<Vec<_>>();
+
+        let next_cursor = if (number_of_entries == extended_limit - 1) || entries.is_empty() {
+            // we have exhausted the search space, thus we return no next cursor
+            None
```
IzioDev, line 353: termination logic seems wrong, when `number_of_entries` is incomplete (`<` limit), this condition will always be false, and the last element never returned (infinite cursor).

```rust
if number_of_entries < extended_limit || entries.is_empty() {
  // exhausted, return all entries and None cursor
} else {
  // still has elements, return entries - 1 and cursor
}
```

IzioDev (changes requested):

indexes/utxoindex/src/stores/indexed_utxos.rs
D-Stacks: you are correct, I think I used to have an invert on the first condition that is no longer there, but also not as elegant as written here. I probably should look over the tests then as well, i assume they should have picked it up, but didn't.

indexes/utxoindex/src/stores/indexed_utxos.rs
D-Stacks: yes, I think this is still from an earlier pass i did on the pr, can be removed and I'll stay with the full access key.

indexes/core/src/indexed_utxos.rs
D-Stacks: some name suggestions? maybe SecondaryKeyData, or InnerKeyData? mostly this struct is used so we can group by spk.

indexes/core/src/indexed_utxos.rs
D-Stacks: The TODO was also originally written by me, I no longer really think it is something worth getting into, I do not think we can get collision resistance here, like i once thought. but, i guess it doesn't have anything to do with the PR, just seems like a chance to get rid of it :).  But i can PR it separately as well.

IzioDev: if you confirm it should be removed, then let's remove. i just wanted to make sure it was intentional

indexes/core/src/indexed_utxos.rs
IzioDev: i'm tempted to consider `suffix` and: on `FullKey` and `SuffixKey`, clarify this portion as typedoc

indexes/core/src/indexed_utxos.rs
D-Stacks: it is used for efficient pre-processing: dedup, discarding spks below the cursor spk and getting things in db order, without re-hashing or re-allocation. i.e. unlike std hash sets, it preserves ordering. 

That being said you made me notice I am not actually using it to the full potential in the way i currently have this part written (around L309 in stores/indexed_utxos). That part should actually just be something like a sort, and a partition, or maybe something that can combine them. I will correct that in the next commit. 

I can probably remove this though, and i'll just stay with an explicit IndexSet<ScriptPublicKey>

rpc/core/src/model/message.rs
D-Stacks: yes.

indexes/core/src/indexed_utxos.rs
D-Stacks: I think i will just remove this one, but i will keep the latter, with some renaming, and turn it into an immutable type, probably an arc.

rpc/service/src/service.rs
```diff
@@ -802,6 +826,99 @@ NOTE: This error usually indicates an RPC conversion error between the node and
...
+        request: GetUtxosByAddressesV2Request,
+    ) -> RpcResult<GetUtxosByAddressesV2Response> {
+        if !self.config.utxoindex {
+            return Err(RpcError::NoUtxoIndex);
+        }
+
+        // because the address list and limit is potentially unbounded, we require unsafe_rpc to be enabled for this call
+        if !self.config.unsafe_rpc {
+            return Err(RpcError::UnavailableInSafeMode);
+        }
```
IzioDev, line 841: we don't have this gated mechanism for endpoint v1.

what do you think of adding limits, potentially re-using the default we enforce on stream-based `getUtxoByAddress` system? (see `DEFAULT_MAX_ADDRESSES` in `notify\src\address\tracker.rs`) 
`limit` can have a max bound too.

not sure where these confs should be defined, there is one config for `MAX_SAFE_WINDOW_SIZE` defined in `rpc\core\src\api\rpc.rs` that is used for the `estimate_network_hashes_per_second` call, but maybe we can have a more defined config holder, potentially in `RpcCoreService` as const field members

rpc/service/src/service.rs
```diff
@@ -802,6 +826,99 @@ NOTE: This error usually indicates an RPC conversion error between the node and
...
+
+        // We define an empty address set as an invalid request, as the user should be aware proactively that this is an empty query. This is also a safety measure to prevent accidental unbounded queries, as the user may not be aware that this is an empty query.
+        // TODO: Potentially we could redefine this case to get all utxos in the DAA range, but this is somewhat of an engineering effort to implement.
+        if request.addresses.is_empty() {
+            return Err(RpcError::InvalidGetUtxosByAddressesV2Request(UtxoIndexError::QueryingEmptyAddressSet.to_string()));
+        }
```
IzioDev, line 847: i don't think the comment should be marked as TODO here, such queries (unknown spk, on daa) will likely require a secondary index where `daa` is the first element in the key.

this index require spk, so do the endpoint using it, sounds the proper behavior, hence reject empty addresses with error 👍

rpc/service/src/service.rs
```diff
@@ -802,6 +826,99 @@ NOTE: This error usually indicates an RPC conversion error between the node and
...
+
+        // because the address list and limit is potentially unbounded, we require unsafe_rpc to be enabled for this call
+        if !self.config.unsafe_rpc {
+            return Err(RpcError::UnavailableInSafeMode);
+        }
+
+        // We define an empty address set as an invalid request, as the user should be aware proactively that this is an empty query. This is also a safety measure to prevent accidental unbounded queries, as the user may not be aware that this is an empty query.
+        // TODO: Potentially we could redefine this case to get all utxos in the DAA range, but this is somewhat of an engineering effort to implement.
+        if request.addresses.is_empty() {
+            return Err(RpcError::InvalidGetUtxosByAddressesV2Request(UtxoIndexError::QueryingEmptyAddressSet.to_string()));
```
IzioDev, line 846: we are using `utxoindex` error domain in `rpc`. i suggest to create an endpoint-specific error enum in `rpc\core\src\error.rs`:
```rust
#[derive(Clone, Debug, Error)]
pub enum GetUtxosByAddressesV2RequestError {
    #[error("the address set must not be empty")]
    EmptyAddressSet,
    // all the other endpoint-specific error here
}
```

wire it in the `RpcError` struct:
```rust
#[error("Invalid GetUtxosByAddressesV2 request: {0}")]
InvalidGetUtxosByAddressesV2Request(#[from] GetUtxosByAddressesV2RequestError),
```

and then from the request handler:
```rust
return Err(GetUtxosByAddressesV2RequestError::EmptyAddressSet.into());
```

rpc/service/src/service.rs
```diff
@@ -802,6 +826,99 @@ NOTE: This error usually indicates an RPC conversion error between the node and
...
+            return Err(RpcError::InvalidGetUtxosByAddressesV2Request("from_daa_score must be <= to_daa_score".to_string()));
+        }
+
+        // a start defined outside of the specified range is invalid.
+        if start_daa_score.is_some_and(|sds| sds < from_daa_score || sds > to_daa_score) {
+            return Err(RpcError::InvalidGetUtxosByAddressesV2Request(format!(
+                "start_daa_score {} must be within from_daa_score and to_daa_score",
+                start_daa_score.unwrap_or(0)
+            )));
+        }
```
IzioDev, line 866: if you want to this form:
```rust
        if let Some(start_daa_score) = start_daa_score
            && (start_daa_score < from_daa_score || start_daa_score > to_daa_score)
```
you don't need to unwrap the value afterwards when you use it

rpc/core/src/api/ops.rs
```diff
@@ -12,7 +12,7 @@ use workflow_core::enums::Describe;
 pub const RPC_API_VERSION: u16 = 1;
 /// API revision. Change in this value denotes
 /// backwards-compatible changes.
-pub const RPC_API_REVISION: u16 = 0;
+pub const RPC_API_REVISION: u16 = 1;
```
IzioDev, line 15: i'm not sure this information is used by clients in practice for now

rpc/core/src/model/message.rs
```diff
@@ -1506,6 +1507,141 @@ impl Deserializer for GetUtxosByAddressesResponse {
...
+    /// Inclusive DAA-score range end; None defaults to u64::MAX.
+    pub to_daa_score: Option<u64>,
+    /// Cursor start address; None starts from the first address.
+    pub start_address: Option<RpcAddress>,
+    /// Cursor start DAA score for start_address; None starts from from_daa_score.
+    pub start_daa_score: Option<u64>,
+    /// Cursor start outpoint hash for start_address; None starts from a `[ZERO_HASH]`.
+    pub start_outpoint_hash: Option<RpcHash>,
+    /// Cursor start outpoint index for start_address; None starts from 0 index.
+    pub start_outpoint_index: Option<TransactionIndexType>,
```
IzioDev, line 1526: what do you think about replacing `start_*` with `cursor_*`? the motivation is to reduce confusion between cursor-related data and global query ranges / filters

rpc/service/src/service.rs
```diff
@@ -802,6 +826,99 @@ NOTE: This error usually indicates an RPC conversion error between the node and
...
+                limit,
+            )
+            .await?;
+        let next_cursor = page.next_cursor();
+        let entries = page.entries();
+
+        let entries = self.index_converter.get_ordered_utxos_by_addresses_entries(entries);
+        let (next_address, next_daa_score, next_outpoint_hash, next_outpoint_index) = match next_cursor.as_ref() {
+            Some(cursor) => {
+                // if we get a cursor returned we can expect that all inner fields are Some.
```
IzioDev, line 903: > if we get a cursor returned we can expect that all inner fields are Some

why doesn't the type system reflects this reality? i would also expect all cursor fields to not be under `Option<_>` when you know the cursor exists

rpc/core/src/convert/utxo.rs
```diff
@@ -16,13 +17,35 @@ pub fn utxo_set_into_rpc(item: &UtxoSetByScriptPublicKey, prefix: Option<Prefix>
...
+                        entry.is_coinbase,
+                        entry.covenant_id,
+                    ),
+                })
+                .collect::<Vec<_>>()
+        })
+        .collect::<Vec<_>>()
+}
+
+pub fn ordered_utxo_set_into_rpc(item: &OrderedUtxoEntries, prefix: Option<Prefix>) -> Vec<RpcUtxosByAddressesEntry> {
```
IzioDev, line 36: `prefix` shouldn't be an `Option<_>` here

rpc/core/src/model/message.rs
IzioDev, line 1526: why not take a `RpcTransactionOutpoint` instead? i don't see why a client would use these fields independently

IzioDev (changes requested):

indexes/utxoindex/src/stores/store_manager.rs
```diff
@@ -85,11 +84,11 @@ impl Store {
...
         &self,
-        script_public_keys: ScriptPublicKeyIndexSet,
+        script_public_keys: IndexSet<ScriptPublicKey>,
         daa_score_range: RangeInclusive<u64>,
         cursor: UtxoPageCursor,
         limit: Option<u64>,
-    ) -> UtxoIndexResult<OrderedUtxoSetByScriptPublicKeyPage> {
+    ) -> UtxoIndexResult<OrderedUtxoEntriesPage> {
         self.utxos_by_script_public_key_store.get_utxos_from_script_public_keys_by_daa_score_page(
             script_public_keys.into_iter().collect(),
```
IzioDev, line 93: `script_public_keys.into_iter().collect(),`

this is not necessary to `into_inter()` + `collect()`, right?

indexes/utxoindex/src/stores/indexed_utxos.rs
IzioDev: the request handler already gate the access. i'm not sure the lower level store accessor should have opinions

indexes/utxoindex/src/stores/indexed_utxos.rs
```diff
@@ -306,37 +279,37 @@ impl UtxoSetByScriptPublicKeyStoreReader for DbUtxoSetByScriptPublicKeyStore {
...
+
+            // we filter out script public keys which are less than the cursor script public key, if it exists
+            // since the cursor has pointed past this point we can assume that any caller has already seen these script public keys and thus we can skip them.
+            cursor.script_public_key.is_none() || spk >= cursor.script_public_key.as_ref().unwrap()
+        });
 
         // cursor is not pointing into the script public key set.
         if !is_cursor_valid {
             return Err(UtxoIndexError::InvalidCursor(cursor));
         }
```
IzioDev, line 299: with the current strategy, under a deep re-org, as a client constantly scanning, couldn't i be left with a not working cursor?

if so, i suggest to document above the RpcRequest* struct

indexes/utxoindex/src/stores/indexed_utxos.rs
```diff
@@ -306,37 +279,37 @@ impl UtxoSetByScriptPublicKeyStoreReader for DbUtxoSetByScriptPublicKeyStore {
...
 
+        // sort the script public keys in order to return them in a deterministic order
+        script_public_keys.sort_unstable();
```
IzioDev, line 302: could you describe the tradeoff `sort_unstable` vs `sort`? afaiu, the first is supposed to be faster but may panic under specific circumstances. given we may add limit on the set, how faster is faster and is it worth the unstable sort?

rpc/service/src/service.rs
D-Stacks: you are right, no way to do this efficiently, in the utxo index, without a daa-score first

indexes/utxoindex/src/stores/indexed_utxos.rs
```diff
@@ -306,37 +279,37 @@ impl UtxoSetByScriptPublicKeyStoreReader for DbUtxoSetByScriptPublicKeyStore {
...
+
         let spk_max = script_public_keys.last().unwrap().clone();
```
IzioDev, line 304: this line is safe but it took me some time to analyze, especially because we retain previously with conditions that are hard to follow. maybe worth a `SAFETY` comment above it to tie it with the previous logic

it's also useful for future human & LLM pass that will get it more quickly

IzioDev (changes requested):

rpc/service/src/service.rs
D-Stacks: imho I think it is bad practice that we don't have it in v1, what is your objections to using it here? that it might disrupt wallets that are tied to safe rpc nodes? I could make it dependent on a hard coded limit size, yes? maybe something like if limit > 1000 require unsafe-rpc? I guess the limit on addresses can be much higher.. a wallet could still use a safe rpc node by batching addresses as well. I think I would just go with 1_000 & 1_000_000 unless i get a concrete suggestion. I wouldn't use `DEFAULT_MAX_ADDRESSES` I think this is enabled by the node operater on start up and can be pretty high, also `MAX_SAFE_WINDOW_SIZE` is in a different context.

rpc/service/src/service.rs
D-Stacks: The input cursor can have optional fields though, as None is defined to resolve that case, this would require me to either 1) have two utxoindex cursor types for input and output, 2) convert to concrete values at the rpc boundary.

rpc/core/src/convert/utxo.rs
D-Stacks: I simply copied this from the `utxo_set_into_rpc` above, assuming it had a reason there. So i will remove from both, i guess?

rpc/core/src/model/message.rs
D-Stacks: it seems like an implicit convention that we are leaving things flat in our request-side rpc layout (besides block and tx submission), I could break with it, but it would be out of character imo.

indexes/utxoindex/src/stores/indexed_utxos.rs
D-Stacks: it gives a perf boost but looses a property of `.sort()` whereby two items with equal ordering keep the relative positions the same as before the sorting. this is not important for us imo.

indexes/utxoindex/src/stores/indexed_utxos.rs
D-Stacks: not here, we are only comparing the cursor to the spks that the client supplied, we expect consistency here. Where there is indeed no guarantee, is that the cursor still points to a utxo in the db, but it will simply scan past the location and start with the next one higher up in the ordering hierarchy. I don't think that is an issue.

D-Stacks: at least in how i envision the usage it should be consistent, but not sure of internal client handling, for example if a wallet uses the last utxo of an address, and in turn decides to update their address list to no longer hold that address, but the cursor still points to it, it would indeed fail. so maybe this is worth consideration from that angle. But I assume wallets always update their known address list, even without utxos? I guess this is something that is outside of my knowledge. Maybe when in doubt here it is better to be more liberal?

indexes/utxoindex/src/stores/store_manager.rs
D-Stacks: yes, can just pass as is.

rpc/service/src/service.rs
IzioDev: i think public nodes should be able to serve this endpoint, in a safe manner yes.

your suggestion looks good, unsafe bypass all limits checks, while safe provides sane default. 1_000 already sounds like plenty

rpc/core/src/model/message.rs
IzioDev: it's just that from a client perspective, chances are low that i would want to decouple my stored outpoint information. but this is even true for all cursor property even, i get a cursor, i pass it back as is (currently i do that via re-wiring all fields independently).

indexes/utxoindex/src/stores/indexed_utxos.rs
IzioDev: i mark it as solved, as discussed externally, the cursor handling logic is sound which is what matter on this section specifically

rpc/core/src/convert/utxo.rs
IzioDev: i suppose there is not reason to keep an `Option` here since the calling site knows it has a value

cli/src/modules/rpc.rs
```diff
@@ -322,6 +322,61 @@ impl Rpc {
...
+                let parse_optional_u32 = |value: String| -> Result<Option<u32>> {
+                    if value.eq_ignore_ascii_case("none") { Ok(None) } else { Ok(Some(value.parse::<u32>()?)) }
+                };
+
+                let parse_optional_address = |value: String| -> Result<Option<Address>> {
+                    if value.eq_ignore_ascii_case("none") { Ok(None) } else { Ok(Some(Address::try_from(value.as_str())?)) }
+                };
+
+                let parse_optional_hash = |value: String| -> Result<Option<RpcHash>> {
+                    if value.eq_ignore_ascii_case("none") { Ok(None) } else { Ok(Some(RpcHash::from_hex(value.as_str())?)) }
```
biryukovmaxim, line 347: writing `none` feels not user friendly to me. what about named args and `none` by default?

rpc/grpc/core/proto/rpc.proto
```diff
@@ -572,6 +572,47 @@ message GetUtxosByAddressesResponseMessage {
...
+  optional string startAddress = 4;
+  // Cursor start DAA score for startAddress; omitted starts from fromDaaScore.
+  optional uint64 startDaaScore = 5;
+  // Cursor start pointing to outpoint hash; omitted starts from the first outpoint for startAddress and startDaaScore.,
+  optional string startOutpointHash = 6;
+  // Cursor start pointing to outpoint index; omitted starts from the first outpoint index for startOutpointHash, startAddress and startDaaScore.
+  optional uint32 startOutpointIndex = 7;
+
+  // limit on the number of entries to return; omitted means no limit.
+  optional uint64 limit = 8;
```
biryukovmaxim, line 598: lets have limit by default

rpc/grpc/core/proto/rpc.proto
```diff
@@ -572,6 +572,47 @@ message GetUtxosByAddressesResponseMessage {
...
+  // UTXO entries for the requested addresses and DAA-score range.
+  repeated RpcUtxosByAddressesEntry entries = 1;
+  // Cursor address for the next page; omitted means there is no next page.
+  optional string nextAddress = 2;
+  // Cursor DAA score for the next page; omitted means there is no next page.
+  optional uint64 nextDaaScore = 3;
+  // Cursor outpoint hash for the next page; omitted means there is no next page.
+  optional string nextOutpointHash = 4;
+  // Cursor outpoint index for the next page; omitted means there is no next page.
+  optional uint32 nextOutpointIndex = 5;
```
biryukovmaxim, line 611: we either have next cursor or don't have it. why not to have a single optional message with all 4 fields

rpc/core/src/model/message.rs
biryukovmaxim, line 1526: the same comment regarding single optional structure instead of multiple optional fields. it doesnt make sense to keep some of them empty in case of another one is filled. i still believe it should be single cursor struct with all fields

rpc/core/src/model/message.rs
```diff
@@ -1506,6 +1507,141 @@ impl Deserializer for GetUtxosByAddressesResponse {
...
+    /// Soft cap on entries; None means no limit, and a page may exceed this to finish the current script public key + DAA-score group.
+    pub limit: Option<u64>,
```
biryukovmaxim, line 1528: lets have a hard limit so we dont need to make the method unsafe

rpc/service/src/service.rs
biryukovmaxim: i agree, lets have default limits. im fine if we dont have them for unsafe version

indexes/utxoindex/src/stores/indexed_utxos.rs
```diff
@@ -176,6 +258,112 @@ impl UtxoSetByScriptPublicKeyStoreReader for DbUtxoSetByScriptPublicKeyStore {
...
+            if !is_cursor_valid && spk == cursor.script_public_key.as_ref().unwrap() {
+                // if none, is_cursor_valid is already true.
+                is_cursor_valid = true;
+                return true;
+            }
+
+            // we filter out script public keys which are less than the cursor script public key, if it exists
+            // since the cursor has pointed past this point we can assume that any caller has already seen these script public keys and thus we can skip them.
+            cursor.script_public_key.is_none() || spk >= cursor.script_public_key.as_ref().unwrap()
+        });
```
biryukovmaxim, line 294: for me this code means that api is designed wrong. there is no guarantee that between calls new entries do not appeared.  i think we either should have key that starts with daa score - so we can get consistent data for score and continue fetching. or api should be focused on single address, so each page belongs to single address and we can fetch them separately to always get complete and consistent view

biryukovmaxim (changes requested):

database/src/access.rs
```diff
@@ -478,6 +481,58 @@ where
...
+        let mut db_iterator = self.db.raw_iterator_opt(read_opts);
+        let mut count = 0usize;
+        // Use Option to unify the "no ranges" and "ranges exhausted" cases in a single return path.
+        let mut current_seek_range = seek_ranges.next().inspect(|range| {
+            db_iterator.seek(DbKey::new(&self.prefix, range.start().clone()).as_ref());
+        });
+
+        std::iter::from_fn(move || {
+            while db_iterator.valid() && count < limit && current_seek_range.is_some() {
+                let key_bytes: Box<[u8]> = db_iterator.key().unwrap()[self.prefix.len()..].into();
```
biryukovmaxim, line 515: Can't we know exact size so we don't need to box the key

indexes/utxoindex/src/stores/indexed_utxos.rs
```diff
@@ -88,32 +95,86 @@ impl AsRef<[u8]> for TransactionOutpointKey {
     }
 }
 
+pub const DAA_SCORE_KEY_SIZE: usize = size_of::<u64>();
+
+struct DaaScoreKey([u8; DAA_SCORE_KEY_SIZE]);
```
biryukovmaxim, line 100: consider using zerocopy BE U64 instead of creating a new type wrapper

indexes/utxoindex/src/stores/indexed_utxos.rs
```diff
@@ -176,6 +258,112 @@ impl UtxoSetByScriptPublicKeyStoreReader for DbUtxoSetByScriptPublicKeyStore {
...
+                TransactionOutpointKey::from(&TransactionOutpoint::MAX),
+            );
+            RangeInclusive::new(start_key, end_key)
+        });
+
+        // +1 in order to return the next cursor
+        let extended_limit = limit.map(|l| l.saturating_add(1)).unwrap_or(u64::MAX).try_into().unwrap_or(usize::MAX);
+
+        // set the cursor daa_score to the start of the range if it is not already set.
+        cursor.daa_score = cursor.daa_score.or(Some(*daa_score_range.start()));
```
biryukovmaxim, line 324: Can't we use unwrap_or

database/src/access.rs
```diff
@@ -478,6 +481,58 @@ where
         })
     }
 
+    // For seeking multiple disjointed ranges in one db iterator pass
+    pub fn multi_range_seek_iterator<'a>(
+        &'a self,
+        mut seek_ranges: impl Iterator<Item = RangeInclusive<TKey>> + 'a,
+        seek_min: Option<TKey>,
+        seek_max: Option<TKey>,
+        limit: usize,
```
biryukovmaxim, line 490: i think it makes more sense to use limit on top if iterator. it could be as simple as iter().take(limit)

PR #995, opened 2026-05-12
Toccata transient mass activation and mempool policy refinement

michaelsutton:
Implements the Toccata transient mass activation, including forked block mass limits and delayed mempool adoption of the relaxed transient limit so pre-activation and near-boundary transactions remain safely conservative.

Also polishes mempool policy around the activation:
- moves template-fit checks out of standardness and into explicit template-limit validation
- applies relay fee checks over both non-contextual mass dimensions
- activation logic for signature script length limit change
- removes obsolete dust standardness checks covered systemically by KIP-9 storage mass (ever since crescendo)
- adds focused tests for transient mass activation, template selection, and policy behavior across the activation boundary

### DIFF
--- a/consensus/core/src/config/params.rs
+++ b/consensus/core/src/config/params.rs
@@ -6,7 +6,7 @@ pub use super::{
 use crate::{
     BlockLevel, KType,
     constants::STORAGE_MASS_PARAMETER,
-    mass::{BlockLaneLimits, BlockMassLimits},
+    mass::{BlockLaneLimits, BlockMassLimits, MassCofactors},
     network::{NetworkId, NetworkType},
 };
 use kaspa_addresses::Prefix;
@@ -16,6 +16,8 @@ use std::{
     cmp::min,
     ops::{Deref, DerefMut},
 };
+
+const MEMPOOL_BLOCK_MASS_ACTIVATION_DELAY_SECONDS: u64 = 24 * 60 * 60;
 
 #[derive(Clone, Copy, Debug, Default, PartialEq, Eq, Serialize, Deserialize)]
 pub struct ForkActivation(u64);
@@ -45,6 +47,13 @@ impl ForkActivation {
 
     pub fn is_active(self, current_daa_score: u64) -> bool {
         current_daa_score >= self.0
+    }
+
+    pub fn delayed_by(self, delay_daa_score: u64) -> Self {
+        match self.0 {
+            Self::ALWAYS | Self::NEVER => self,
+            daa_score => Self(daa_score.saturating_add(delay_daa_score)),
+        }
     }
 
     /// Checks if the fork was "recently" activated, i.e., in the time frame of the provided range.
@@ -85,6 +94,10 @@ impl<T: Copy> ForkedParam<T> {
         if self.activation.is_active(daa_score) { self.post } else { self.pre }
     }
 
+    pub fn with_delayed_activation(&self, delay_daa_score: u64) -> Self {
+        Self::new(self.pre, self.post, self.activation.delayed_by(delay_daa_score))
+    }
+
     /// Returns the value before activation (=pre unless activation = always)
     pub fn before(&self) -> T {
         match self.activation.0 {
@@ -104,6 +117,12 @@ impl<T: Copy> ForkedParam<T> {
     /// Maps the ForkedParam<T> to a new ForkedParam<U> by applying a map function on both pre and post
     pub fn map<U: Copy, F: Fn(T) -> U>(&self, f: F) -> ForkedParam<U> {
         ForkedParam::new(f(self.pre), f(self.post), self.activation)
+    }
+}
+
+impl<T: Copy> From<T> for ForkedParam<T> {
+    fn from(value: T) -> Self {
+        Self::new_const(value)
     }
 }
 
@@ -201,7 +220,8 @@ pub struct OverrideParams {
     pub mass_per_tx_byte: Option<u64>,
     pub mass_per_script_pub_key_byte: Option<u64>,
     pub mass_per_sig_op: Option<u64>,
-    pub block_mass_limits: Option<BlockMassLimits>,
+    pub prior_block_mass_limits: Option<BlockMassLimits>,
+    pub new_transient_mass_limit: Option<u64>,
     pub block_lane_limits: Option<BlockLaneLimits>,
 
     /// The parameter for scaling inverse KAS value to mass units (KIP-0009)
@@ -244,7 +264,8 @@ impl From<Params> for OverrideParams {
             mass_per_tx_byte: Some(p.mass_per_tx_byte),
             mass_per_script_pub_key_byte: Some(p.mass_per_script_pub_key_byte),
             mass_per_sig_op: Some(p.mass_per_sig_op),
-            block_mass_limits: Some(p.block_mass_limits),
+            prior_block_mass_limits: Some(p.prior_block_mass_limits),
+            new_transient_mass_limit: Some(p.new_transient_mass_limit),
             block_lane_limits: Some(p.block_lane_limits),
             storage_mass_parameter: Some(p.storage_mass_parameter),
             deflationary_phase_daa_score: Some(p.deflationary_phase_daa_score),
@@ -297,7 +318,8 @@ pub struct Params {
     pub mass_per_tx_byte: u64,
     pub mass_per_script_pub_key_byte: u64,
     pub mass_per_sig_op: u64,
-    pub block_mass_limits: BlockMassLimits,
+    pub prior_block_mass_limits: BlockMassLimits,
+    pub new_transient_mass_limit: u64,
     pub block_lane_limits: BlockLaneLimits,
 
     /// The parameter for scaling inverse KAS value to mass units (KIP-0009)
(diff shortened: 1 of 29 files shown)


consensus/src/processes/transaction_validator/tx_validation_in_header_context.rs
```diff
@@ -106,4 +107,13 @@ impl TransactionValidator {
...
     }
+
+    fn check_transaction_signature_scripts_in_header_context(&self, tx: &Transaction, block_daa_score: u64) -> TxResult<()> {
+        let max_signature_script_len = self.max_signature_script_len.get(block_daa_score);
+        if let Some(i) = tx.inputs.iter().position(|input| input.signature_script.len() > max_signature_script_len) {
+            return Err(TxRuleError::TooBigSignatureScript(i, max_signature_script_len));
+        }
+
+        Ok(())
+    }
```
coderofstuff, line 119: You should add the intent of the validation as a comment in this function.

The name `check_transaction_signature_scripts_in_header_context` doesn't suggest it will only check length but signature scripts as a whole (so perhaps this may check length and something else). The implementation only checks for length.

consensus/core/src/config/params.rs
```diff
@@ -754,16 +850,17 @@ pub const DEVNET_PARAMS: Params = Params {
...
+    new_max_signature_script_len: NEW_MAX_SIGNATURE_SCRIPT_LEN,
     max_script_public_key_len: 10_000,
 
     mass_per_tx_byte: 1,
     mass_per_script_pub_key_byte: 10,
     mass_per_sig_op: 1000,
 
     // Transient mass is increased for stark proofs
-    block_mass_limits: BlockMassLimits { compute: 500_000, storage: 500_000, transient: 1_000_000 },
+    prior_block_mass_limits: BlockMassLimits { compute: 500_000, storage: 500_000, transient: 1_000_000 },
```
coderofstuff: Why is the `prior_block_mass_limits.transient` here the same as `new_transient_mass_limit`?

consensus/src/processes/transaction_validator/tx_validation_in_header_context.rs
michaelsutton: it's more like: "do what ever you can do to check_transaction_signature_scripts_in_header_context"

cf transaction_in_isolation where it's simply called `check_transaction_signature_scripts`

consensus/core/src/config/params.rs
michaelsutton: the question is specific to devnet? so yeah good point. should be 500k prior so that we can practice the HF

coderofstuff: in `toccata` the devnet params makes sense to have prior set to old values and the activation set to `::always()` for development purposes. Up to you if you wanna apply this. I was mainly just curious why it was the way it is now.

Having them be different will indeed allow us to practice the transition in devnet

consensus/core/src/config/params.rs
```diff
@@ -721,7 +796,8 @@ pub const SIMNET_PARAMS: Params = Params {
     mass_per_script_pub_key_byte: 10,
     mass_per_sig_op: 1000,
     // Transient mass is increased for stark proofs
-    block_mass_limits: BlockMassLimits { compute: 500_000, storage: 500_000, transient: 1_000_000 },
+    prior_block_mass_limits: BlockMassLimits { compute: 500_000, storage: 500_000, transient: 1_000_000 },
```
someone235: Same question @coderofstuff asked for devnet: even if we set activation to be `always` we probably want to see it's using 1M although in prior it's 0.5M

mining/src/mempool/replace_by_fee.rs
```diff
@@ -17,6 +17,7 @@ impl Mempool {
         &self,
         transaction: &MutableTransaction,
         rbf_policy: RbfPolicy,
+        _virtual_daa_score: u64,
```
someone235: Why?

mining/src/mempool/replace_by_fee.rs
```diff
@@ -119,7 +129,8 @@ impl Mempool {
 
     fn get_double_spend_feerate(&self, double_spend: &DoubleSpend) -> RuleResult<f64> {
         let owner = self.transaction_pool.get_double_spend_owner(double_spend)?;
-        match owner.mtx.calculated_feerate(&self.config.block_mass_cofactors) {
+        let cofactors = self.config.mempool_mass_cofactors.get(owner.added_at_daa_score);
```
someone235: nit, but doesn't it make more sense to use `virtual_daa_score`?

mining/src/mempool/replace_by_fee.rs
```diff
@@ -131,12 +142,14 @@ impl Mempool {
...
         double_spend: &DoubleSpend,
+        virtual_daa_score: u64,
     ) -> RuleResult<&'a MempoolTransaction> {
         let owner = self.transaction_pool.get_double_spend_owner(double_spend)?;
-        if let (Some(transaction_feerate), Some(double_spend_feerate)) = (
-            transaction.calculated_feerate(&self.config.block_mass_cofactors),
-            owner.mtx.calculated_feerate(&self.config.block_mass_cofactors),
-        ) {
+        let transaction_cofactors = self.config.mempool_mass_cofactors.get(virtual_daa_score);
+        let owner_cofactors = self.config.mempool_mass_cofactors.get(owner.added_at_daa_score);
```
someone235: nit, but doesn't it make more sense to use virtual_daa_score?

testing/integration/src/consensus_integration_tests.rs
```diff
@@ -1932,19 +1933,19 @@ async fn payload_test() {
         0,
         SubnetworkId::default(),
         0,
-        vec![0; (config.params.block_mass_limits.transient / TRANSIENT_BYTE_TO_MASS_FACTOR / 2) as usize],
+        vec![0; (config.params.block_mass_limits().after().transient / TRANSIENT_BYTE_TO_MASS_FACTOR / 2) as usize],
     );
 
     // Create a tx with transient mass over the block limit
-    txx.payload = vec![0; (config.params.block_mass_limits.transient / TRANSIENT_BYTE_TO_MASS_FACTOR + 100) as usize];
+    txx.payload = vec![0; (config.params.block_mass_limits().after().transient / TRANSIENT_BYTE_TO_MASS_FACTOR + 100) as usize];
```
someone235: Why do we use `.after()` if toccata is not activated on devnet? (Same question for the rest of the test)

consensus/core/src/config/params.rs
michaelsutton: done in https://github.com/kaspanet/rusty-kaspa/pull/995/changes/ff3e0e25a92404205432cbf2fbfda426bc79efbd

mining/src/mempool/check_transaction_standard.rs
```diff
@@ -454,44 +427,6 @@ mod tests {
...
-            },
-            Test {
-                name: "Transaction size is too large",
-                mtx: new_mtx(
-                    Transaction::new(
-                        TX_VERSION,
-                        vec![dummy_tx_input.clone()],
-                        vec![TransactionOutput::new(
-                            0u64,
-                            ScriptPublicKey::new(
```
someone235, line 473: Do we want to move these tests somewhere else? Maybe to `MiningManager.validate_and_insert_transaction`?

mining/src/mempool/check_transaction_standard.rs
```diff
@@ -161,14 +161,16 @@ impl Mempool {
...
-                self.minimum_required_transaction_relay_fee(transaction.calculated_non_contextual_masses.unwrap().compute_mass);
-            if transaction.calculated_fee.unwrap() < minimum_fee {
-                return Err(NonStandardError::RejectInsufficientFee(transaction_id, transaction.calculated_fee.unwrap(), minimum_fee));
-            }
+        // Minimum relay fee applies to non-contextual mass so block-space usage has a minimum cost,
+        // including transient mass. This is especially important after increasing the transient mass limit.
+        // Storage mass does not require an additional relay-fee floor here since storage growth is
+        // sufficiently protected even under worst-case block-limit usage.
+        let NonContextualMasses { compute_mass, transient_mass } = transaction.calculated_non_contextual_masses.unwrap();
+        let minimum_fee = self.minimum_required_transaction_relay_fee(compute_mass.max(transient_mass));
```
someone235: Shouldn't it be normalized?

consensus/src/processes/transaction_validator/tx_validation_in_header_context.rs
```diff
@@ -106,4 +107,13 @@ impl TransactionValidator {
 
         Ok(())
     }
+
+    fn check_transaction_signature_scripts_in_header_context(&self, tx: &Transaction, block_daa_score: u64) -> TxResult<()> {
```
someone235, line 112: Add a TODO(post-toccata) to remove it

coderofstuff (dismissed):

someone235 (changes requested):

consensus/core/src/config/params.rs
michaelsutton: already fixed

mining/src/mempool/check_transaction_standard.rs
michaelsutton: already fixed

someone235 (approved):

PR #997, opened 2026-05-13
Bump P2P protocol version and guard Toccata IBD

michaelsutton:
Bumps the P2P protocol version for Toccata while preserving TN12 launch compatibility by normalizing its protocol 9 peers locally.

Adds pre-activation version filtering for synced nodes and an IBD pruning point freshness guard so fresh nodes avoid syncing from outdated or stale pre-Toccata peers.


protocol/flows/src/ibd/flow.rs
```diff
@@ -1031,3 +1043,167 @@ staging selected tip ({}) is too small or negative. Aborting IBD...",
...
+        let pp_daa_score = activation_daa_score - 10;
+        let pp_timestamp = 1_000_000_000_000;
+        let pp_to_activation_millis = 10 * params.target_time_per_block();
+        let estimated_activation_time = pp_timestamp + pp_to_activation_millis;
+        let stale_after = estimated_activation_time + pruning_period_millis(&params) + ONE_DAY_MILLIS;
+
+        // No activation is configured:
+        // PP(pre-activation by score) ---- estimated activation ---- pruning period + margin ---- now
+        assert!(
+            validate_pruning_point_freshness_for_toccata(&MAINNET_PARAMS, V, pp_hash, pp_timestamp, pp_daa_score, stale_after + 1)
```
coderofstuff: This shouldn't use MAINNET_PARAMS, because MAINNET_PARAMS will eventually have the param activated.

Instead, just create a dedicated params set with toccata_activation set to never

protocol/flows/src/ibd/flow.rs
```diff
@@ -1031,3 +1043,167 @@ staging selected tip ({}) is too small or negative. Aborting IBD...",
...
+        params
+    }
+
+    fn pruning_period_millis(params: &Params) -> u64 {
+        params.pruning_depth().saturating_add(params.finality_depth()).saturating_mul(params.target_time_per_block())
+    }
+
+    #[test]
+    fn test_toccata_pruning_point_staleness_guard() {
+        const V: u32 = PROTOCOL_VERSION;
```
coderofstuff: Please rename `V` to a clearer name like `TOCCATA_INCLUSIVE_VERSION`, or add a comment explaining that this explicitly should be the protocol version including toccata.

`PROTOCOL_VERSION` may change in the future. If it does, how is this test affected?

coderofstuff (changes requested):
minor comments

protocol/flows/src/ibd/flow.rs
michaelsutton: irrelevant now

protocol/flows/src/ibd/flow.rs
michaelsutton: fixed in https://github.com/kaspanet/rusty-kaspa/pull/997/changes/e4ecd533edb4eb95c2705ab5b402b21d7378c19a

someone235 (approved):

coderofstuff (approved):

PR #999, opened 2026-05-14
feat(wasm/txscript): allows passing script builder flags

IzioDev:
Target is `toccata`

- add script builder wasm tests
- add `ScriptBuilderOptions` which contains `flags` as a property (allows future expansion without API breaking changes)
- allows controlling script builder options at the time of ScriptBuilder creation

### DIFF
--- a/crypto/txscript/src/wasm/builder.rs
+++ b/crypto/txscript/src/wasm/builder.rs
@@ -1,6 +1,5 @@
-use crate::result::Result;
-use crate::{script_builder as native, standard};
-use kaspa_consensus_core::tx::ScriptPublicKey;
+use crate::{EngineFlags, error::Error, result::Result, script_builder as native, standard};
+use kaspa_consensus_core::{mass::ScriptUnits, tx::ScriptPublicKey};
 use kaspa_utils::hex::ToHex;
 use kaspa_wasm_core::hex::{HexViewConfig, HexViewConfigT};
 use kaspa_wasm_core::types::{BinaryT, HexString};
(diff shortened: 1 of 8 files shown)


crypto/txscript/src/wasm/builder.rs
```diff
@@ -36,23 +75,50 @@ impl ScriptBuilder {
...
+        let mut engine_flags = Self::default();
+
+        if let Some(value) = flags.try_get_value("covenantsEnabled")? {
+            engine_flags.covenants_enabled =
+                value.as_bool().ok_or_else(|| Error::convert("flags.covenantsEnabled", "expected boolean"))?;
+        }
+
+        if flags.try_get_value("sigopScriptUnits")?.is_some() {
+            engine_flags.sigop_script_units =
+                ScriptUnits(flags.get_u64("sigopScriptUnits").map_err(|err| Error::convert("flags.sigopScriptUnits", err))?);
```
biryukovmaxim, line 102: should it use bigint?

biryukovmaxim (dismissed):

crypto/txscript/src/wasm/builder.rs
IzioDev: bigint or number allowed for ease of use

crypto/txscript/src/wasm/builder.rs
```diff
@@ -36,23 +75,50 @@ impl ScriptBuilder {
...
+        }
+
+        Ok(engine_flags)
     }
 }
 
 // TODO: Add `ScriptBuilder::with_flags` method that allows setting script engine flags for the builder.
```
IzioDev: remove

michaelsutton (approved):

PR #1008, opened 2026-05-16
fix(ibd): raise smt flag in pre-toccata no-op sync

biryukovmaxim:


someone235 (dismissed):

freshair18 (commented):
let is_smt_stable = if self.ctx.config.toccata_activation.is_active(pp_header.daa_score) {
                    consensus.async_is_pruning_smt_stable().await
                } else {
                    true
                };

Probably now it is simpler and more sound to just always query the flag

michaelsutton (approved):

PR #1009, opened 2026-05-16
Unconditionally set SMT stable flag to true after syncing new SMT state

someone235:


biryukovmaxim (commented):
i think it will work

michaelsutton:
This fix was added into #1008

PR #1010, opened 2026-05-17
refactor: isolate git build.rs into its own crate (closes #1002)

### LINKED ISSUE #1002: Avoid workspace-wide rebuilds after every git commit
`kaspa-utils/build.rs` currently embeds the git commit hash and explicitly asks Cargo to rerun the build script when git refs change:

```rust
println!("cargo::rerun-if-changed={head_path}");
println!("cargo::rerun-if-changed={head_ref_path}");
println!("cargo:rustc-env=RUSTY_KASPA_GIT_FULL_COMMIT_HASH={full_hash}");
println!("cargo:rustc-env=RUSTY_KASPA_GIT_SHORT_COMMIT_HASH={short_hash}");
```

Here `head_path` is usually:

```text
.git/HEAD
```

and `head_ref_path` is usually:

```text
.git/refs/heads/<current-branch>
```

So every `git commit`, amend, rebase, cherry-pick, reset, or branch switch dirties `kaspa-utils`. Since `kaspa-utils` is widely depended on, this causes a large part of the workspace to rebuild even when no relevant Rust source changed.

This is especially painful in local dev loops such as:

```bash
cargo fmt
cargo clippy --all --tests --benches --examples
cargo nextest run --release --no-fail-fast
```

Cargo fingerprint logging confirms the cause with messages like:

```text
stale: changed ".../.git/refs/heads/<branch>"
(vs) ".../target/debug/build/kaspa-utils-.../output"
```

**Desired criteria:**

- Normal local dev builds should not be sensitive to commit hash changes.
- Release/CI/package builds that require an exact embedded commit hash should be able to opt in explicitly.
- If we still need git-hash tracking, it should at least move to a narrow side crate so changing the hash does not invalidate most of the workspace through `kaspa-utils`.

**Possible direction:**

Keep `RUSTY_KASPA_NO_COMMIT_HASH=1` as a hard disable, and add an explicit opt-in such as `RUSTY_KASPA_TRACK_COMMIT_HASH=1` for builds that need `rerun-if-changed` on git refs. Default local builds can embed the hash opportunistically when the build script runs, without tracking git ref changes.

biryukovmaxim:
Closes #1002.

## Problem

`kaspa-utils/build.rs` embeds the git commit hash and watches `.git/HEAD` + `.git/refs/heads/<branch>` via `cargo::rerun-if-changed`. Because 41 workspace crates depend on `kaspa-utils`, every `git commit` / amend / rebase / cherry-pick / branch-switch invalidates `kaspa-utils` and cascades a re-link through most of the workspace. Local dev loops (`cargo fmt`, `cargo clippy`, `cargo nextest run`) get hammered.

## Approach

Split the build-script + the consumers of its output into two narrow new crates so the per-commit invalidation only reaches the crates that actually need the git hash:

- **`build-info/`** (`kaspa-build-info`) — owns the **only** `build.rs` that watches git refs, plus the `git` module (`hash`, `short_hash`, `version`, `with_short_hash`, etc.). Self-contained, no kaspa deps.
- **`system-info/`** (`kaspa-system-info`) — owns the `SystemInfo` struct and its OS-side collectors (cpu/memory/fd_limit/system_id/nginx proxy). No `build.rs`. **No `Default` impl** — `SystemInfo::new(git_hash, git_short_hash, version)` forces every caller to supply the build-time fields explicitly, so the struct is never half-populated.

`kaspa-utils` loses `build.rs`, `src/git.rs`, `src/sysinfo.rs`, the `sysinfo` feature, and the `duct` / `sysinfo` / `num_cpus` / `mac_address` deps.

The build script in `build-info/build.rs` locates the repo root by walking up from `CARGO_MANIFEST_DIR` until it finds a `.git/` directory, instead of assuming `parent()`, so the crate can sit anywhere in the workspace tree.

## Caller rewiring

| Crate | Old | New |
|---|---|---|
| `kaspad/src/daemon.rs` | `kaspa_utils::git`, `kaspa_utils::sysinfo::SystemInfo`, `SystemInfo::default()` | `kaspa_build_info::git`, `kaspa_system_info::SystemInfo`, `SystemInfo::new(git::hash(), git::short_hash(), git::version())` |
| `rpc/service/src/service.rs` | `kaspa_utils::sysinfo::SystemInfo` (via the `sysinfo` feature) | `kaspa_system_info::SystemInfo` (zero git dep — receives the populated struct from the daemon) |
| `wallet/core/src/lib.rs` | `kaspa_utils::git::with_short_hash` | `kaspa_build_info::git::with_short_hash` (public `version*()` API preserved) |

Also dropped a stale `features = ["sysinfo"]` on `kaspa-utils` in `consensus/smt-store/Cargo.toml` (the source never used it).

## Rebuild scope before vs after

| | Before | After |
|---|---|---|
| `kaspa-utils` invalidated on every commit | ✅ | ❌ |
| `kaspa-rpc-service` (and its cascade: wrpc-server, grpc-server, testing/integration, bridge…) re-links on every commit | ✅ | ❌ |
| `kaspa-consensus*`, `kaspa-mining`, `kaspa-p2p-flows`, `kaspa-database`, `kaspa-indexes-*`, … re-link on every commit | ✅ | ❌ |
| `kaspa-build-info` reruns build.rs on every commit | n/a | ✅ |
| `kaspad` re-links only when the hash actually changes | n/a | ✅ |
| `kaspa-wallet-core` re-links only when the hash actually changes | n/a | ✅ |

## Verified locally

- `cargo check --workspace --all-targets` — clean.
- `cargo test -p kaspa-build-info` — `test_git_hash` passes (walk-up logic finds `.git/`).
- Touched `.git/refs/heads/<branch>` to mimic a commit, then `cargo check -p kaspa-rpc-service` finished in **0.19s with zero recompiles** (used to recompile most of the workspace).
- `cargo check -p kaspad` after the same touch recompiled only `kaspa-build-info` (since the hash didn't actually change, downstream stayed Fresh — real commits will re-link only `kaspad` + `kaspa-wallet-core`).

## Notes

- Existing `RUSTY_KASPA_NO_COMMIT_HASH=1` opt-out still works.
- `wallet/core`'s public `version()` / `version_with_git_hash()` API is preserved per request; the wallet rebuild cone (wallet/core → wasm, cli, wallet-daemon) on commit is the accepted trade-off for keeping that API.
- The issue mentioned a possible `RUSTY_KASPA_TRACK_COMMIT_HASH=1` opt-in to make the always-watch `rerun-if-changed` itself opt-in; deliberately not added here since the invalidation cone is now narrow enough that the extra knob isn't worth the surface area. Easy follow-up if desired.


michaelsutton (approved):

michaelsutton:
Looks good to me. The branch fixes the main rebuild fanout problem.

Measured results:

```text
Baseline on toccata after HEAD/branch changes:
  ~3m09s real, ~59m user

After isolating build-info:
  ~1m34s real, ~27m user

After also gating wallet-core git hash behind an opt-in feature:
  ~1m03s real, ~19m user
```

The final rebuild set is much smaller: only `kaspa-build-info` and direct consumers like `kaspad`, `testing-integration`, `bridge`, and `simpa` rebuild.

PR #1011, opened 2026-05-18
Clarify seqcommit tx validation contexts

michaelsutton:
Clarifies and tests the contextual validation rules for Toccata seqcommit scripts.

This PR documents the intended DAA-score and seqcommit-context choices across block-template validation, chain qualification, and selected-parent replay. It also adds a focused consensus integration test for the selected-parent replay boundary case, where a seqcommit spend is valid against the block’s selected parent but would fail if replayed later with full script checks.

The PR also fixes pruning-point import/proof handling for seqcommit access. Pruning point transactions are validated using the pruning point’s selected parent as seqcommit context, matching the same rule used for normal chain-block qualification. Mainnet pruning proof chain segments are therefore built and applied against that selected-parent context.

For non-mainnet networks, proof apply intentionally keeps accepting the legacy pruning-point-context segment rule for compatibility with already deployed testnet/devnet peers. This means the pruning-proof segment alignment is not critical for existing testnets; it is mainly a mainnet correctness/consistency fix before release.

### DIFF
--- a/consensus/src/pipeline/virtual_processor/processor.rs
+++ b/consensus/src/pipeline/virtual_processor/processor.rs
@@ -1083,7 +1083,7 @@ impl VirtualStateProcessor {
         virtual_daa_score: u64,
         virtual_past_median_time: u64,
         args: &TransactionValidationArgs,
-        sp: Hash,
+        selected_parent: Hash,
     ) -> TxResult<()> {
         self.transaction_validator.validate_tx_in_isolation(&mutable_tx.tx)?;
         self.transaction_validator.validate_tx_in_header_context_with_args(
@@ -1091,7 +1091,7 @@ impl VirtualStateProcessor {
             virtual_daa_score,
             virtual_past_median_time,
         )?;
-        self.validate_mempool_transaction_in_utxo_context(mutable_tx, virtual_utxo_view, virtual_daa_score, args, sp)?;
+        self.validate_mempool_transaction_in_utxo_context(mutable_tx, virtual_utxo_view, virtual_daa_score, args, selected_parent)?;
         Ok(())
     }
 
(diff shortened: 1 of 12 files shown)


consensus/src/pipeline/virtual_processor/processor.rs
```diff
@@ -1446,8 +1452,11 @@ impl VirtualStateProcessor {
         }
 
         let virtual_read = self.virtual_stores.upgradable_read();
-        let sp = virtual_read.state.get().unwrap().ghostdag_data.selected_parent;
-        // Validate transactions of the pruning point itself
+        // Seqcommit validation uses the pruning point's selected parent as context. Post-Toccata chain
+        // qualification enforces first parent = selected parent; for genesis imports, use genesis itself.
+        let sp = new_pruning_point_header.direct_parents().first().copied().unwrap_or(new_pruning_point);
```
biryukovmaxim, line 1457: cant we use config.genesis instead of new_pruning_point?

michaelsutton: if there are no direct parents it must be genesis, so mind as well use it. it reads more cleanly imo

testing/integration/src/consensus_integration_tests.rs
```diff
@@ -1532,6 +1533,124 @@ fn assert_tx_not_in_chain_seq_commit(consensus: &TestConsensus, accepting_block:
...
+    let config = ConfigBuilder::new(DEVNET_PARAMS)
+        .skip_proof_of_work()
+        .edit_consensus_params(|p| {
+            let mut genesis_multiset = MuHash::new();
+            initial_utxo_collection.iter().for_each(|(outpoint, utxo)| {
+                genesis_multiset.add_utxo(outpoint, utxo);
+            });
+            p.genesis.utxo_commitment = genesis_multiset.finalize();
+            let genesis_header: Header = (&p.genesis).into();
+            p.genesis.hash = genesis_header.hash;
```
someone235, line 1589: nit, but we need some helper to do this

michaelsutton: next time..

someone235 (approved):

PR #1018, opened 2026-05-27
KIP-21: inactivity_shortcut

biryukovmaxim:
## Summary

Adds `inactivity_shortcut` to the existing per-chain-block `seq_commit`
commitment (active-lanes SMT, miner-payload root, mergeset context).
Activated by `zk_hardening_activation`; pre-activation chain blocks are
unaffected.

Spec: https://github.com/kaspanet/vprogs/blob/feat/inactivity_proofs/docs/kaspa-inactivity-shortcut-spec.md

`inactivity_shortcut` is the `seq_commit` of the latest in-domain
non-genesis selected-parent-chain ancestor `A` with
`A.blue_score + min(F(A), F(B)) < B.blue_score`. It lets guests walk
inactivity spans in `O(N/F)` hops instead of `O(N)`, without knowing `F`.

## Commitment delta

The shortcut is folded one level above `lanes_root` via a new
`activity_root`:

```
state_root
└── H_seq
    ├── activity_root
    │   ├── pre-hardening:  = lanes_root  (identity)
    │   └── post-hardening: = H_activity_root(inactivity_shortcut, lanes_root)
    └── payload_and_ctx_digest
        └── H_seq
            ├── context_hash = H_mergeset_context(ts || daa || bs)
            └── payload_root
```

Lane tips stay shortcut-free; the mergeset context preimage is
unchanged.

## What's in this PR

- `consensus/seq-commit/`: new `SeqCommitActivityRoot` hasher and
  `activity_root_hash`; unified `verify_smt_metadata`.
- `consensus/src/`: `build_seq_commit` wraps `lanes_root` into
  `activity_root` post-hardening; `compute_inactivity_shortcut_block`
  resolves the anchor via SMT coinbase-lane fast-path with cascade
  fallback; `SmtBlockMetadata` gains a length-dispatched ToccataV0/V1
  enum carrying the shortcut block on V1.
- IBD: single 96-byte `SmtMetadataMessage` (unchanged shape); receiver
  derives the anchor block via header walk
  (`inactivity_shortcut_block_for_pov`) before importing the SMT.
  P2P v10 unchanged.
- Tests: daemon `zk_hardening_activation` crossing; simpa `test_pruning`
  activates toccata @ daa 1000 and `zk_hardening` @ daa 2000.

### DIFF
--- a/consensus/src/pipeline/virtual_processor/processor.rs
+++ b/consensus/src/pipeline/virtual_processor/processor.rs
@@ -752,6 +752,10 @@ impl VirtualStateProcessor {
         if pp != expected_pruning_point {
             return Err(ConsensusError::UnexpectedPruningPoint);
         }
+        // Genesis has no SMT metadata row and no parent to index.
+        if pp == self.genesis.hash {
+            return Err(ConsensusError::GeneralOwned("cannot export SMT metadata: pruning point is genesis".to_string()));
+        }
 
         let meta = self
             .smt_metadata_store
--- a/consensus/src/pipeline/virtual_processor/processor.rs
+++ b/consensus/src/pipeline/virtual_processor/processor.rs
@@ -177,6 +177,9 @@ pub struct VirtualStateProcessor {
     pub(crate) toccata_activation: ForkActivation,
     pub(crate) toccata_logger: ForkLogger,
 
+    // KIP-21 finality-anchor activation: gates inactivity_shortcut inclusion in mergeset_context_hash.
+    pub(crate) zk_hardening_activation: ForkActivation,
+
     // SMT stores
     pub(super) smt_stores: Arc<kaspa_smt_store::processor::SmtStores>,
     pub(super) smt_metadata_store: Arc<crate::model::stores::smt_metadata::DbSmtMetadataStore>,
@@ -251,6 +254,7 @@ impl VirtualStateProcessor {
             counters,
             toccata_activation: params.toccata_activation,
             toccata_logger: ForkLogger::new("virtual state processing rules", true),
+            zk_hardening_activation: params.zk_hardening_activation,
             smt_stores: storage.smt_stores.clone(),
             smt_metadata_store: storage.smt_metadata_store.clone(),
             _mining_rules: mining_rules,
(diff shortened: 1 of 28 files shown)


consensus/src/processes/pruning_proof/apply.rs
```diff
@@ -321,6 +321,14 @@ impl PruningProofManager {
                 let current_header =
                     trusted_header_map.get(&current).ok_or(PruningImportError::MissingPruningPointChainSegment(current))?;
 
+                // This cutoff also covers the inactivity-shortcut anchor (mainnet path).
+                // Chain qualification gives pp.bs >= pp.sp.bs + 1, so a block failing the check
+                // satisfies `current.bs + F <= pp.sp.bs <= pp.bs - 1`, i.e. `current.bs <= pp.bs - F - 1`.
+                // pp.inactivity_shortcut is by definition the highest chain block with
+                // `bs <= pp.bs - F - 1` (see `compute_inactivity_shortcut_block`), so its bs is at
+                // least the break block's bs and it is reached by the iteration before (or at)
+                // the break. The non-mainnet branch uses pp.bs as context, which is strictly more
```
michaelsutton: The non-mainnet branch uses pp.bs so it can reach the halt point sooner and actually miss the shortcut. 
You should:
- check if hardening is active for pp and use that as another condition for using pp.sp as context. (it's ok bcs post-hardening syncers should already send the segment correctly)
- add TODO(post-zk-hardening): use pp.sp in all cases 
- fix this comment

consensus/src/pipeline/virtual_processor/processor.rs
```diff
@@ -738,36 +759,169 @@ impl VirtualStateProcessor {
             .map_err(|_| ConsensusError::GeneralOwned(format!("SMT metadata not found for pruning point {pp}")))?;
```
michaelsutton, line 764: minor: I think we should reject `pp == genesis` here. A syncing node has pp as genesis and if at the same time `RequestPruningPointSmtStateFlow` is called with genesis as `expected_pp`, this fn will panic below at `pp_header.direct_parents()[0]`

consensus/src/processes/pruning_proof/apply.rs
biryukovmaxim: 81d65ea0d0e1177918c4c62d77431dbbe3cf178b

consensus/src/pipeline/virtual_processor/processor.rs
michaelsutton: edit: seems ok since genesis smt metadata will be None

michaelsutton (approved):

PR #1021, opened 2026-05-28
feat(roth): generate tx targeting user lanes

IzioDev:
Adapted from initial implementation by @biryukovmaxim

- add `lps` option, will generates transactions targeting a runtime per-generated user-lane pool.
- add `subnet-pool-size` option, determining the size of the user-lane pool.
- vspcv2 rust example adaptation to showcase subnet/gas consumption
- small roth tweak to adapt to txv1 semantic

By default, all tx targeting a user-lane will feed the `gas` field, by the following weighted distribution:
1. common 95% -> beta law (a = 1,000; b = 199,000)
2. zero 2% -> hardcoded 0
3. third 2% -> beta law (a = 1,666; b = 3,333)
4. half 0.9% -> beta law (a = 2,500; b = 2,500)
5. full 0.1% -> 1 billion (=`DEFAULT_GAS_PER_LANE_LIMIT`)

The overall goal of this distribution is to most often allow 200 txs (of the same lane) within one block by consensus gas limit enforcement, and rarely reach/exceed that cap in different scenarios. (formula by @michaelsutton)

### DIFF
--- a/rothschild/src/main.rs
+++ b/rothschild/src/main.rs
@@ -1,6 +1,7 @@
 use std::{collections::HashMap, str::FromStr, sync::Arc, time::Duration};
 
 use clap::{Arg, ArgAction, Command};
+use gas::GasDistribution;
 use itertools::Itertools;
 use kaspa_addresses::{Address, Prefix, Version};
 use kaspa_consensus_core::{
@@ -9,7 +10,7 @@ use kaspa_consensus_core::{
     hashing::covenant_id::covenant_id,
     network::NetworkType,
     sign::sign,
-    subnets::SUBNETWORK_ID_NATIVE,
+    subnets::{SUBNETWORK_ID_NATIVE, SUBNETWORK_ID_SIZE, SUBNETWORK_NAMESPACE_LEN, SubnetworkId},
     tx::{
         ComputeCommit, CovenantBinding, MutableTransaction, Transaction, TransactionInput, TransactionOutpoint, TransactionOutput,
         UtxoEntry,
@@ -28,6 +29,8 @@ use secp256k1::{
     rand::{Rng, thread_rng},
 };
 use tokio::time::{Instant, MissedTickBehavior, interval};
+
+mod gas;
 
 const DEFAULT_SEND_AMOUNT: u64 = 10 * SOMPI_PER_KASPA;
 const MILLIS_PER_TICK: u64 = 10;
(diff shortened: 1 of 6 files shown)


rothschild/src/main.rs
```diff
@@ -636,43 +690,52 @@ fn generate_tx(
...
             previous_outpoint: *op,
             signature_script: vec![],
             sequence: 0,
-            mass: if with_covenant_id { TxInputMass::ComputeBudget(10.into()) } else { TxInputMass::SigopCount(1.into()) },
+            mass: if tx_version == TX_VERSION_TOCCATA {
+                TxInputMass::ComputeBudget(10.into())
+            } else {
+                TxInputMass::SigopCount(1.into())
+            },
         })
```
IzioDev: shouldn't input be generated after optional `randomize_tx_version`?

rothschild/src/main.rs
```diff
@@ -636,43 +690,52 @@ fn generate_tx(
...
     let outputs = (0..num_outs)
         .map(|_| TransactionOutput { value: send_amount / num_outs, script_public_key: script_public_key.clone(), covenant: None })
         .collect_vec();
 
-    let mut data = vec![0u8; payload_size];
+    let mut data = vec![0u8; tx_config.payload_size];
     rand::thread_rng().fill_bytes(&mut data);
 
-    let unsigned_tx = Transaction::new_non_finalized(tx_version, inputs, outputs, 0, SUBNETWORK_ID_NATIVE, 0, data);
+    let subnetwork_id = if tx_version == TX_VERSION_TOCCATA && should_use_user_lane(&mut rng, tx_config) {
```
coderofstuff, line 729: The probability of `lps` here is affected by the probability of `randomize_tx_version` which will lead to a lower than `lps` average.

You might want to make the logic such that if `should_use_user_lane` procs, it forces tx_version to `TX_VERSION_TOCCATA`.

rothschild/src/main.rs
michaelsutton: it should, but it doesn't really matter: `sign` sets the correct units anyway

michaelsutton (dismissed):

rothschild/src/main.rs
IzioDev: decided to clarify the help text: [review: clarify randomize tx version impacts on roth](https://github.com/kaspanet/rusty-kaspa/pull/1021/commits/f2c500515fb6549386933d32d5cc61d9acc8f36f)

image

michaelsutton (approved):

PR #1022, opened 2026-05-29
feat(mempool): allow p2p relayed tx with low fees before toccata activation

IzioDev:
missing pr desc

### DIFF
--- a/consensus/core/src/config/params.rs
+++ b/consensus/core/src/config/params.rs
@@ -56,10 +56,17 @@ impl ForkActivation {
         current_daa_score >= self.0
     }
 
-    pub fn delayed_by(self, delay_daa_score: u64) -> Self {
+    pub fn delayed_by(self, daa_score_delta: u64) -> Self {
         match self.0 {
             Self::ALWAYS | Self::NEVER => self,
-            daa_score => Self(daa_score.saturating_add(delay_daa_score)),
+            daa_score => Self(daa_score.saturating_add(daa_score_delta)),
+        }
+    }
+
+    pub fn early_by(self, daa_score_delta: u64) -> Self {
+        match self.0 {
+            Self::ALWAYS | Self::NEVER => self,
+            daa_score => Self(daa_score.saturating_sub(daa_score_delta)),
         }
     }
 
@@ -712,6 +719,7 @@ pub const MAINNET_PARAMS: Params = Params {
 
     // Roughly 2025-05-05 1500 UTC
     crescendo_activation: ForkActivation::new(110_165_000),
+    // TODO(pre-toccata): Before setting the activation DAA score, resolve all comments of the form TODO(pre-toccata)
     toccata_activation: ForkActivation::never(),
 };
 
--- a/mining/src/manager.rs
+++ b/mining/src/manager.rs
@@ -28,7 +28,7 @@ use kaspa_consensus_core::{
     },
     block::{BlockTemplate, TemplateBuildMode, TemplateTransactionSelector},
     coinbase::MinerData,
-    config::params::ForkedParam,
+    config::params::{ForkActivation, ForkedParam},
     errors::{block::RuleError as BlockRuleError, tx::TxRuleError},
     mass::{BlockLaneLimits, BlockMassLimits},
     tx::{MutableTransaction, Transaction, TransactionId},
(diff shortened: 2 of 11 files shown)


michaelsutton (changes requested):
- let's change the base to `toccata`
- activation param must be taken from consensus (mempool cofactors already has a skewed activation to 24h post actual activation)
- the 600 offset is not worth while: whatever the threshold, it is set for all nodes, so a non mining node can accept a low fee tx while all miners reject it (due to network pov noise). In the worst case, the txs slipping through the small gap will remain in its local mempool for 24h (evict policy)
- p2p pre activation acceptance path should use only compute mass (ie keep prior rules)
- style: imo no need to rename `minimum_required_transaction_relay_fee_with_base`; the added arg speaks for itself (`with_*` makes sense when you keep both variants)

consensus/core/src/config/params.rs
```diff
@@ -63,6 +63,13 @@ impl ForkActivation {
         }
     }
 
+    pub fn early_by(self, advance_daa_score: u64) -> Self {
+        match self.0 {
+            Self::ALWAYS | Self::NEVER => self,
+            daa_score => Self(daa_score.saturating_sub(advance_daa_score)),
+        }
+    }
```
coderofstuff: This is unused. Please remove.

consensus/core/src/config/params.rs
```diff
@@ -718,6 +725,7 @@ pub const MAINNET_PARAMS: Params = Params {
 
     // Roughly 2025-05-05 1500 UTC
     crescendo_activation: ForkActivation::new(110_165_000),
+    // TODO(pre-toccata): Before setting the activation DAA score, resolve all comments of the form TODO(pre-toccata)
```
coderofstuff, line 722: The comment is outside the scope of this PR. please remove.

mining/src/manager.rs
```diff
@@ -52,27 +52,39 @@ impl MiningManager {
...
-        let config =
-            Config::build_default(target_time_per_block, relay_non_std_transactions, mempool_block_mass_limits, block_lane_limits);
+        let config = Config::build_default(
+            target_time_per_block,
+            relay_non_std_transactions,
+            mempool_block_mass_limits,
+            toccata_activation,
+            block_lane_limits,
+        );
         Self::with_config(config, cache_lifetime, counters)
```
coderofstuff: The `mempool` object is instantiated in `with_config`. So instead of passing the `toccata_activation` to `config`, you can directly pass it to mempool:
- add `toccata_activation` to the `Mempool` struct and its `new` fn as an arg
- in MiningManager::with_config, pass `toccata_activation` directly to `Mempool::new`

mining/src/mempool/check_transaction_standard.rs
```diff
@@ -435,5 +462,76 @@ mod tests {
                 ),
             }
         }
+
+        // TODO(post-toccata): remove this temporary test case
+        // TODO(pre-toccata): change to NetworkType::Mainnet
+        let params: Params = NetworkType::Testnet.into();
+        assert_ne!(params.toccata_activation, ForkActivation::never(), "this test requires a scheduled Toccata activation");
```
coderofstuff: This is a test, so you can guarantee this condition by doing:

```
params.toccata_activation = ForkActivation::new(<some_constant>);
```

mining/src/mempool/check_transaction_standard.rs
```diff
@@ -58,31 +64,44 @@ impl Mempool {
...
-        let minimum_fee = self.minimum_required_transaction_relay_fee(fee_mass);
+
+        // TODO(post-toccata): remove `use_prior_p2p_fee_rules` and unconditionally use:
+        //   fee_mass = max(compute_mass, normalized_transient_mass)
+        //   relay_fee = self.config.minimum_relay_transaction_fee
+        // The relaxation allows softwares to adapt between release and activation, instead of txs being p2p rejected by updated nodes
+        let use_prior_p2p_fee_rules = priority == Priority::Low && !self.config.toccata_activation.is_active(virtual_daa_score);
+        let (fee_mass, relay_fee) = if use_prior_p2p_fee_rules {
+            // Prior P2P relay-fee policy: legacy base fee over compute mass only.
+            (masses.compute_mass, 1_000)
```
coderofstuff: This magic number `1_000` should make it's way to mining/src/mempool/config.rs as `LEGACY_MINIMUM_RELAY_TRANSACTION_FEE` right below `DEFAULT_MINIMUM_RELAY_TRANSACTION_FEE`. Also, the comment for `DEFAULT_MINIMUM_RELAY_TRANSACTION_FEE` in mining/src/mempool/config.rs is inaccurate where it says `/// The default is 100 sompi per gram.`

coderofstuff (changes requested):
Changes look good. Comments are mainly about structure and cleanup.

consensus/core/src/config/params.rs
IzioDev: Discussed with @michaelsutton earlier if we should keep it, we concluded it could still live here.
Happy to remove it if there is another motivation

consensus/core/src/config/params.rs
IzioDev: This PR is the only one introducing `TODO(pre-toccata):` as of today, so I think it should be the responsible for introducing this comment

mining/src/mempool/check_transaction_standard.rs
IzioDev: Yes, it indeed sounds out of scope to use params from network type here

mining/src/mempool/check_transaction_standard.rs
IzioDev: Ok for const.

About the comment, i think it's still fair (while not 100% accurate in the transition period) to mention default is 100 sompi per gram, as at the time of release rpc submit will enforce this, and after activation P2P will do the same

coderofstuff: You can leave the comment. The legacy part will be removed on the release after toccata activates anyway.

consensus/core/src/config/params.rs
coderofstuff: This and the comment below are a scope bleed. I will tolerate it for this PR.

michaelsutton (approved):

coderofstuff (approved):

PR #1023, opened 2026-05-29
Fixes #1017 — unbounded Stratum input buffer allowing remote memory e…

### LINKED ISSUE #1017: Unbounded Stratum input buffer can cause remote memory exhaustion
bug description
bridge/src/stratum_listener.rs accumulates data received from each Stratum TCP connection in line_buffer until a newline (\n) is received.

The buffer has no maximum size. A client can keep sending data without a newline, causing line_buffer to grow for as long as the connection remains active.

This happens before JSON-RPC parsing and before miner authorization.

Relevant code:

bridge/src/stratum_listener.rs:256-260
bridge/src/stratum_listener.rs:275-280
bridge/src/stratum_listener.rs:470-475
To Reproduce

Start kaspa-stratum-bridge with a configured Stratum listener.
Connect to the Stratum TCP port.
Send data continuously without sending \n, ensuring data is sent within each 5-second read timeout interval.
The connection remains in the incomplete-line path and received data continues to be appended to line_buffer without a size limit.
Expected behavior
Incomplete Stratum messages should have a maximum permitted size. Connections exceeding that limit should be closed.

Additional context
The read timeout is applied per read operation and does not limit total buffered input.

The code default for the Stratum listener is :5555. Port-only listener addresses are converted by bind_addr_from_port() to 0.0.0.0:<port>.

LiveLaughLove13:
Problem
In spawn_client_listener, each TCP connection accumulated incoming bytes in line_buffer until a \n was received. The buffer had no size limit. A client could send data continuously without a newline (staying within the 5-second per-read timeout) and grow line_buffer without bound for as long as the connection stayed open.

This occurred before JSON-RPC parsing and before miner authorization, so any client reaching the Stratum port (default :5555 / 0.0.0.0:5555) could trigger it.

Fix

Introduced MAX_STRATUM_LINE_BYTES (64 KiB) as the maximum size for an incomplete Stratum line awaiting \n.
Added append_line_data() to check the limit before appending; on rejection the buffer is left unchanged.
In the read loop, if a chunk would exceed the cap, the connection is logged and closed immediately.
Exported both symbols from the crate for testability.
64 KiB is well above legitimate Stratum JSON-RPC messages (typically under 1 KiB) while bounding memory per connection.

Tests
Added five unit tests in bridge/src/tests.rs:

Accepts normal messages under the limit
Accepts incremental chunks up to the limit
Rejects append when the buffer is already at capacity
Rejects a single chunk larger than the limit
Accepts data exactly at the limit, then rejects one byte more
Verification
cargo test -p kaspa-stratum-bridge --bin stratum-bridge test_append_line_data
cargo clippy -p kaspa-stratum-bridge --all-targets --no-deps -- -D warnings
cargo fmt --all -- --check

### DIFF
--- a/bridge/src/stratum_listener.rs
+++ b/bridge/src/stratum_listener.rs
@@ -11,6 +11,21 @@ use tokio::sync::mpsc;
 use tokio::sync::watch;
 use tracing::{debug, error, info, warn};
 
+/// Maximum permitted size (in bytes) for an incomplete Stratum line awaiting `\n`.
+/// Legitimate JSON-RPC Stratum messages are well below this; the cap prevents unbounded
+/// memory growth when a client sends data without a newline.
+pub const MAX_STRATUM_LINE_BYTES: usize = 64 * 1024;
+
+/// Append received data to the line buffer. Returns `false` if the append would exceed
+/// [`MAX_STRATUM_LINE_BYTES`], leaving the buffer unchanged.
+pub fn append_line_data(line_buffer: &mut String, data: &str) -> bool {
+    if line_buffer.len().saturating_add(data.len()) > MAX_STRATUM_LINE_BYTES {
+        return false;
+    }
+    line_buffer.push_str(data);
+    true
+}
+
 /// Event handler function type
 pub type EventHandler = Arc<
     dyn Fn(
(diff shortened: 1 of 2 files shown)


LiveLaughLove13:
Fixes #1017

LiveLaughLove13:
@coderofstuff  If you can review when you have a moment

coderofstuff (approved):

PR #1024, opened 2026-05-30
Fix/973 prometheus config endpoint exposure

### LINKED ISSUE #973: Stratum bridge Prometheus server exposes config endpoints
The per-instance Prometheus server in kaspa-stratum-bridge uses the shared HTTP handler in HttpMode::Instance. As a result, ports intended for metrics also serve:

- GET /api/config
- POST /api/config when RKSTRATUM_ALLOW_CONFIG_WRITE=1

With port-only config values such as :2114, bind_addr_from_port() normalizes the address to 0.0.0.0:2114, so these endpoints may be reachable from non-local interfaces.

This appears unintended because prom_port is documented as a Prometheus metrics endpoint, while configuration management should belong to a dedicated dashboard/admin surface.

Impact:
A metrics endpoint can expose bridge configuration. If config writes are enabled, any client with network access to the Prometheus port can modify bridge settings such as kaspad_address, stratum_port, min_share_diff, and prom_port.

Suggested fix:
The per-instance Prometheus server should only serve /metrics. Config endpoints should be moved to a dedicated admin/dashboard server, bound to localhost by default, and protected by explicit authentication.

LiveLaughLove13:
Verification & fix summary

Per-instance Prometheus servers (prom_port, e.g. :2114) used the shared HTTP handler in HttpMode::Instance, so ports intended for metrics also served:

GET /api/config
POST /api/config (when RKSTRATUM_ALLOW_CONFIG_WRITE=1)
GET /api/status, GET /api/stats, and static dashboard files
With port-only config like :2114, bind_addr_from_port() normalizes to 0.0.0.0:2114, making these endpoints reachable from the network — before JSON-RPC parsing, authorization, or any miner handshake. On master, /api/config was only exposed on prom ports, not the aggregated web dashboard, which is the wrong surface for admin/config management.

Fix
Prom ports → metrics only — Replaced HttpMode::Instance with HttpMode::MetricsOnly. Per-instance prom servers now serve only GET /metrics; all other paths return 404.
Config/admin → dashboard only — GET/POST /api/config moved to HttpMode::Aggregated (the web_dashboard_port server, e.g. :3030).
Dashboard localhost by default — Added bind_dashboard_addr_from_port() so bare ports like :3030 bind to 127.0.0.1:3030 instead of 0.0.0.0:3030. Explicit 0.0.0.0:3030 still works for intentional remote access.
Prom scrape unchanged — start_prom_server() still uses bind_addr_from_port() (0.0.0.0) so Prometheus can scrape from the network.

Tests
test_metrics_only_rejects_non_metrics — prom ports return 404 for config/status/stats/static
test_static_files_aggregated_mode / test_static_files_blocked_in_metrics_only_mode
test_bind_dashboard_addr_from_port_defaults_localhost

Fixes #973


LiveLaughLove13:
@aglov413
@coderofstuff

PR #1025, opened 2026-05-30
consensus: remove TRANSIENT_BYTE_TO_MASS_FACTOR (#996)

### LINKED ISSUE #996: Remove TRANSIENT_BYTE_TO_MASS_FACTOR
`TRANSIENT_BYTE_TO_MASS_FACTOR` has outlived it's purpose. Transient mass should be directly set and treated as a byte limit (1:1).

This can be applied post hardfork as a non-consensus change. Effectively:

Transient mass will change from 1,000,000 to 250,000 in the limit

The TRANSIENT_BYTE_TO_MASS_FACTOR (current value is 4) is removed from the calculation

atharaldsen:
Closes #996. Targeting `toccata` for now per coderofstuff; can rebase to `master` once toccata merges.

Transient mass was charged at 4 grams/byte and compared against a limit expressed in those units. Now that transient mass is its own block-mass dimension, the factor is redundant. This charges transient mass 1:1 per byte and sets the transient limits directly as byte caps.

To keep it behavior-preserving, every transient limit is divided by 4 while compute and storage stay at 500,000:

- prior transient limit: 500,000 -> 125,000 (testnet12: 1,000,000 -> 250,000)
- post-fork transient limit: 1,000,000 -> 250,000

Byte caps are unchanged (125,000 pre-fork, 250,000 post-fork). Normalized transient mass is also unchanged, since the per-dimension cofactor scales inversely with the limit (4 * 0.5 == 1 * 2.0), so mempool standardness and template selection see the same values.

Tests: consensus-core and mining suites pass, including the transient-mass activation tests. Added tests pinning the 1:1 charge and the per-network byte caps.

Note: raw `transient_mass` metric counters now read 4x smaller (internal metric only, no behavioral effect).

### DIFF
--- a/consensus/core/src/config/params.rs
+++ b/consensus/core/src/config/params.rs
@@ -20,8 +20,8 @@ use std::{
 const MEMPOOL_BLOCK_MASS_ACTIVATION_DELAY_SECONDS: u64 = 24 * 60 * 60;
 const PRIOR_MAX_SIGNATURE_SCRIPT_LEN: usize = 10_000;
 // Increased for stark proofs. This value is effectively covered by the post-Toccata
-// transient block mass limit: 1_000_000 transient mass / 4 grams-per-byte = 250_000
-// bytes for the entire block, so a larger signature script cannot be accepted anyway.
+// transient block mass limit of 250_000 (charged 1:1 per byte), which caps the entire
+// block body at 250_000 bytes, so a larger signature script cannot be accepted anyway.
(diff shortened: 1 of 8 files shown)


michaelsutton:
FYI to reduce noise this will be reviewed only post toccata activation on mainnet

atharaldsen:
> FYI to reduce noise this will be reviewed only post toccata activation on mainnet

Better now @michaelsutton ?

PR #1026, opened 2026-05-30
Assimilate ZK hardening into Toccata

michaelsutton:
Remove temporary zk_hardening_activation plumbing and make the hardened Groth16/Risc0 and seqcommit activity-root paths unconditional under Toccata.

Collapse SMT metadata to the V1 shape with a required inactivity_shortcut_block, and update IBD/pruning SMT import and verification accordingly.

Trim activation/legacy tests while preserving real Toccata coverage for SMT sync, staging header access, Groth16 arity rejection, and lane activity proofs.

### DIFF
--- a/crypto/txscript/src/zk_precompiles/groth16/mod.rs
+++ b/crypto/txscript/src/zk_precompiles/groth16/mod.rs
@@ -11,10 +11,7 @@ use crate::{
     data_stack::Stack,
     opcodes::i32s_to_usizes,
     runtime_resource_meter::RuntimeResourceMeter,
-    zk_precompiles::{
-        ZkPrecompile,
-        fields::{Fr, TruncFr},
-    },
+    zk_precompiles::{ZkPrecompile, fields::Fr},
 };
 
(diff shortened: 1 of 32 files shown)


coderofstuff (approved):

crypto/txscript/src/zk_precompiles/groth16/mod.rs
```diff
@@ -72,7 +69,7 @@ impl ZkPrecompile for Groth16Precompile {
     ///
     /// *NOTE: Experimental code; not yet fully audited for mainnet use.* TODO(pre-covpp)
     ///
-    fn verify_zk(dstack: &mut Stack, meter: &mut RuntimeResourceMeter, flags: EngineFlags) -> Result<(), Self::Error> {
+    fn verify_zk(dstack: &mut Stack, meter: &mut RuntimeResourceMeter, _flags: EngineFlags) -> Result<(), Self::Error> {
```
biryukovmaxim, line 72: it seems we dont need to pass engine flags here

PR #1027, opened 2026-05-31
Integrate finalized ZK hardening and KIP-21 shortcut

michaelsutton:
## Summary

Ports the finalized TN10-tested Toccata ZK hardening changes back into `toccata`.

On mainnet, Toccata has not activated yet, so the post-Toccata consensus logic can still be finalized directly. On TN10, both Toccata and the hardening activation are already active, and the pruning point is post-hardening, so new TN10 nodes can sync using this finalized logic without the temporary activation split.

This PR therefore does not include the TN10-only `zk_hardening_activation` scaffolding. The finalized behavior is integrated as ordinary Toccata consensus logic.

## Changes

- Finalizes ZK verifier hardening:
  - Groth16:
    - stricter public-input, verifying-key, and proof validation
    - per-public-input pricing
  - Risc0:
    - bounds Merkle witness size

- Adds the KIP-21 seqcommit inactivity shortcut:
  - commits `inactivity_shortcut` next to `lanes_root` under `ActivityRoot`
  - lets inactivity proofs hop by the shortcut instead of walking long linear selected-parent-chain spans and avoids requiring zk guests to know the exact inactivity threshold
  - finalizes SMT metadata and pruning-point/IBD sync behavior

### DIFF
--- a/consensus/src/pipeline/pruning_processor/processor.rs
+++ b/consensus/src/pipeline/pruning_processor/processor.rs
@@ -725,7 +725,7 @@ impl PruningProcessor {
         info!("Rebuilding pruning point SMT root after pruning data (sanity test)");
         let bounds = kaspa_smt_store::processor::SmtReadBounds::for_pov(pruning_point_blue_score, self.config.params.finality_depth());
         let expected_root = self.smt_stores.get_lanes_root(bounds, |bh| self.is_smt_canonical(bh, new_pruning_point));
-        let expected_count = self.smt_metadata_store.get(new_pruning_point).unwrap().active_lanes_count;
+        let expected_count = self.smt_metadata_store.get(new_pruning_point).unwrap().active_lanes_count();
         let (root, count) = self
             .smt_stores
             .recompute_lanes_root_from_leaf_stream(bounds, expected_count, |bh| self.is_smt_canonical(bh, new_pruning_point))
--- a/consensus/src/pipeline/virtual_processor/processor.rs
+++ b/consensus/src/pipeline/virtual_processor/processor.rs
@@ -177,10 +177,6 @@ pub struct VirtualStateProcessor {
     pub(crate) toccata_activation: ForkActivation,
     pub(crate) toccata_logger: ForkLogger,
 
-    // KIP-21 finality-anchor activation: gates inactivity_shortcut inclusion in mergeset_context_hash.
-    pub(crate) zk_hardening_activation: ForkActivation,
-    pub(crate) zk_hardening_logger: ForkLogger,
-
     // SMT stores
     pub(super) smt_stores: Arc<kaspa_smt_store::processor::SmtStores>,
     pub(super) smt_metadata_store: Arc<crate::model::stores::smt_metadata::DbSmtMetadataStore>,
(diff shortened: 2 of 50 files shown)


crypto/txscript/src/zk_precompiles/risc0/mod.rs
```diff
@@ -20,6 +21,17 @@ pub mod receipt_claim;
 pub struct R0SuccinctPrecompile;
 pub use error::R0Error;
 
+// Matches risc0-zkvm's ALLOWED_CODE_MERKLE_DEPTH for the Poseidon2 control root.
+const POSEIDON2_CONTROL_MERKLE_DEPTH: usize = 8;
```
saefstroem, line 24: Would be nice to have a test asserting the upstream depth but   I think any version upgrades would be rigorously checked before done, so it wouldnt make sense adding the `risc0-zkvm` dep just for this. Unlikely that this would change in the future anyway.

saefstroem (approved):

crypto/txscript/src/zk_precompiles/risc0/mod.rs
michaelsutton: we use a fixed control root in consensus, and i verified that the specific hash is built from a tree in that depth

consensus/src/pipeline/virtual_processor/processor.rs
```diff
@@ -665,15 +670,15 @@ impl VirtualStateProcessor {
...
+            activity_digest_lane, activity_leaf, activity_root_hash, lane_key, lane_tip_next, mergeset_context_hash,
+            miner_payload_leaf, miner_payload_root, payload_and_context_digest, seq_commit, seq_state_root, smt_leaf_hash,
         };
         use kaspa_seq_commit::types::{LaneTipInput, MergesetContext, MinerPayloadLeafInput, SeqCommitInput, SeqState, SmtLeafInput};
         use kaspa_smt::SmtHasher;
 
         let blue_score = ghostdag_data.blue_score;
         let context_hash = mergeset_context_hash(&MergesetContext {
-            timestamp: seq_commit_timestamp(self.genesis.timestamp),
+            timestamp: self.genesis.timestamp,
```
biryukovmaxim, line 681: shouldnt we use 0 here, the value must correspond to parent block?

crypto/txscript/src/opcodes/mod.rs
```diff
@@ -897,7 +897,7 @@ opcode_list! {
             vm.consume_script_units(tag.cost())?;
 
             // Verify the ZK proof
-            verify_zk(tag, &mut vm.dstack)?;
+            verify_zk(tag, &mut vm.dstack, &mut vm.runtime_resource_meter, vm.flags)?;
```
biryukovmaxim: the  function doesnt use flags, it makes sense to remove it from both fn signatures

crypto/txscript/src/zk_precompiles/groth16/mod.rs
```diff
@@ -1,24 +1,76 @@
...
+
+    Ok(VerifyingKey { alpha_g1, beta_g2, gamma_g2, delta_g2, gamma_abc_g1 })
+}
+
 pub struct Groth16Precompile;
 impl ZkPrecompile for Groth16Precompile {
     type Error = Groth16Error;
     /// Verifies the integrity of a Groth16 proof.
     ///
     /// *NOTE: Experimental code; not yet fully audited for mainnet use.* TODO(pre-covpp)
```
biryukovmaxim: can we remove todo

protocol/flows/src/v10/request_pruning_point_smt_state.rs
biryukovmaxim, line 103: we need to also return error in that case, otherwise it could panic in the assert below

consensus/src/pipeline/pruning_processor/processor.rs
biryukovmaxim, line 590: shouldnt we use finality depth + 1 here so we dont remove metadata required for inactivity shorcut

biryukovmaxim (changes requested):

consensus/src/pipeline/virtual_processor/processor.rs
michaelsutton: I don't think so. We can make an exception for genesis to use its own timestamp since it's deterministic

biryukovmaxim (approved):

PR #1033, opened 2026-06-01
SPM calculation Fix

LiveLaughLove13:
Fixed TOTAL-row SPM calculation to use the average worker SPM instead of an aggregate pool-wide SPM. Previously, the terminal TOTAL row computed SPM as total accepted shares divided by elapsed runtime, which inflated the value when multiple workers were active. The fix now accumulates each worker’s individual SPM and reports the arithmetic average across active workers, so the TOTAL SPM reflects per-worker performance rather than summed throughput.

Implementation details:

Added average_worker_spm(sum_spm, worker_count) in bridge/src/share_handler.rs. Updated TOTAL-row stats computation in the shared stats printer to: sum per-worker SPM values,
track active worker count,
compute TOTAL SPM via average_worker_spm(...).
Removed the old uptime-based aggregate SPM formula for TOTAL row. Validation:

Added test_total_row_spm_uses_worker_average to bridge/src/tests.rs. Ran cargo fmt --all -- --check.
Ran cargo test -p kaspa-stratum-bridge --bin stratum-bridge test_total_row_spm_uses_worker_average.

### DIFF
--- a/bridge/src/share_handler.rs
+++ b/bridge/src/share_handler.rs
@@ -101,6 +101,10 @@ fn vardiff_compute_next_diff(current: f64, shares: f64, elapsed_secs: f64, expec
     if (next - current).abs() > f64::EPSILON { Some(next) } else { None }
 }
 
+pub fn average_worker_spm(sum_spm: f64, worker_count: usize) -> f64 {
+    if worker_count == 0 { 0.0 } else { sum_spm / worker_count as f64 }
+}
+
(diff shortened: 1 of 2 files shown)


LiveLaughLove13:
@coderofstuff 
@aglov413

LiveLaughLove13:
Screenshot 2026-06-01 000208

coderofstuff (approved):

PR #1039, opened 2026-06-04
Rename tx.mass -> tx.storage_mass

someone235:
This PR renames the transaction storage mass field from the ambiguous mass naming to storage_mass / storageMass across consensus, RPC, wallet, mining, and wire conversion code.

  The change clarifies the distinction between storage mass, compute mass, and transient mass while preserving compatibility where needed. Deprecated mass accessors remain in place for consensus
  transactions, and JSON-RPC transaction models continue to support the legacy mass field during the transition.

  ## Changes

  - Rename transaction storage mass APIs and fields to storage_mass / storageMass
  - Update P2P and gRPC transaction wire conversions to use the renamed storage mass field
  - Update RPC transaction and optional transaction models to expose storageMass
  - Keep JSON compatibility by serializing both storageMass and deprecated mass
  - Accept legacy mass on JSON deserialization, with conflict checks when both fields are present

### DIFF
--- a/consensus/client/src/transaction.rs
+++ b/consensus/client/src/transaction.rs
@@ -335,7 +335,7 @@ impl TryCastFromJs for Transaction {
 impl From<cctx::Transaction> for Transaction {
     fn from(tx: cctx::Transaction) -> Self {
         let id = tx.id();
-        let mass = tx.mass();
+        let mass = tx.storage_mass();
         let inputs: Vec<TransactionInput> = tx.inputs.into_iter().map(|input| input.into()).collect::<Vec<TransactionInput>>();
         let outputs: Vec<TransactionOutput> = tx.outputs.into_iter().map(|output| output.into()).collect::<Vec<TransactionOutput>>();
         Self::new_with_inner(TransactionInner {
@@ -364,7 +364,7 @@ impl From<&Transaction> for cctx::Transaction {
         let outputs: Vec<cctx::TransactionOutput> =
             inner.outputs.clone().into_iter().map(|output| output.as_ref().into()).collect::<Vec<cctx::TransactionOutput>>();
         cctx::Transaction::new(inner.version, inputs, outputs, inner.lock_time, inner.subnetwork_id, inner.gas, inner.payload.clone())
-            .with_mass(inner.mass)
+            .with_storage_mass(inner.mass)
     }
 }
 
@@ -396,7 +396,7 @@ impl Transaction {
             lock_time: tx.lock_time,
             gas: tx.gas,
             payload: tx.payload.clone(),
-            mass: tx.mass(),
+            mass: tx.storage_mass(),
             subnetwork_id: tx.subnetwork_id,
         })
     }
@@ -424,7 +424,7 @@ impl Transaction {
             inner.gas,
             inner.payload.clone(),
         )
-        .with_mass(inner.mass);
+        .with_storage_mass(inner.mass);
 
         Ok((tx, utxos))
     }
--- a/consensus/client/src/transaction.rs
+++ b/consensus/client/src/transaction.rs
@@ -24,7 +24,7 @@ use kaspa_utils::hex::*;
 const TS_TRANSACTION: &'static str = r#"
 /**
  * Interface defining the structure of a transaction.
- * 
+ *
  * @category Consensus
  */
 export interface ITransaction {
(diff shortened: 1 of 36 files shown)


consensus/client/src/transaction.rs
```diff
@@ -268,14 +274,26 @@ impl Transaction {
...
+    /// @deprecated Use `storageMass` instead
     #[wasm_bindgen(getter = mass)]
     pub fn get_mass(&self) -> u64 {
-        self.inner().mass
+        self.inner().storage_mass
     }
 
+    /// @deprecated Use `storageMass` instead
     #[wasm_bindgen(setter = mass)]
     pub fn set_mass(&self, v: u64) {
```
biryukovmaxim, line 285: https://doc.rust-lang.org/reference/attributes/diagnostics.html#the-deprecated-attribute

consensus/client/src/transaction.rs
```diff
@@ -268,14 +274,26 @@ impl Transaction {
         self.inner.lock().unwrap().payload = js_value.try_as_vec_u8().unwrap_or_else(|err| panic!("payload value error: {err}"));
     }
 
+    /// @deprecated Use `storageMass` instead
```
biryukovmaxim, line 277: https://doc.rust-lang.org/reference/attributes/diagnostics.html#the-deprecated-attribute

consensus/core/src/tx.rs
```diff
@@ -391,20 +401,33 @@ impl Transaction {
         self.id
     }
 
+    /// [DEPRECATED] This function is deprecated to avoid ambiguity with compute mass and trasient mass. Use `self.set_storage_mass(mass)` instead.
     /// Set the storage mass commitment field of this transaction. This field has been activated on mainnet as part
     /// of the Crescendo hardfork. The field has no effect on tx ID so no need to finalize following this call.
     pub fn set_mass(&self, mass: u64) {
```
biryukovmaxim, line 407: https://doc.rust-lang.org/reference/attributes/diagnostics.html#the-deprecated-attribute

consensus/core/src/tx.rs
```diff
@@ -391,20 +401,33 @@ impl Transaction {
...
+
+    /// Set the storage mass commitment field of this transaction. This field has been activated on mainnet as part
+    /// of the Crescendo hardfork. The field has no effect on tx ID so no need to finalize following this call.
+    pub fn set_storage_mass(&self, mass: u64) {
+        self.storage_mass.0.store(mass, SeqCst)
     }
 
+    /// [DEPRECATED] This function is deprecated to avoid ambiguity with compute mass and trasient mass. Use `self.storage_mass()` instead.
     /// Read the storage mass commitment
     pub fn mass(&self) -> u64 {
```
biryukovmaxim, line 419: https://doc.rust-lang.org/reference/attributes/diagnostics.html#the-deprecated-attribute

rpc/core/src/model/tx.rs
```diff
@@ -374,10 +374,55 @@ pub struct RpcTransaction {
...
+    pub version: u16,
+    pub inputs: Vec<RpcTransactionInput>,
+    pub outputs: Vec<RpcTransactionOutput>,
+    pub lock_time: u64,
+    pub subnetwork_id: RpcSubnetworkId,
+    pub gas: u64,
+    #[serde(with = "hex::serde")]
+    pub payload: Vec<u8>,
+    pub storage_mass: u64,
+    pub mass: u64, // DEPRECATED field for storage mass to avoid breaking existing clients. Should be removed in the future.
```
biryukovmaxim: shouldnt they become optional or can they become a single enum? 
current logic assumes we can have both fields

biryukovmaxim (changes requested):

consensus/client/src/transaction.rs
IzioDev: i started with this, using from 1.3.0, but the message then must be "use `storage_mass` instead" (because it's intended for Rust doc)

here, the triple slash is reported to the TS interface using deprecated JSDoc annotation.

I'm not sure how I can have:
1. for rust
2. for wasm / TS interface

without making the TS interface reports two deprecation message

biryukovmaxim: Is it a problem to have 2 messages? Because for me it seems that cure t way doesn't show warning for rust at all

IzioDev: it's weird from js that i'm invited to use `storage_mass` and `storageMass` at the same time, other than that, no problem

biryukovmaxim (approved):

PR #1051, opened 2026-06-14
math: in-repo Lehmer 3072-bit modular inverse + std ceil_log_2

biryukovmaxim:
## Summary

Adds an in-repo, dependency-free modular inverse for the 3072-bit MuHash element, **wires MuHash to use it**, and replaces the consensus sync locator's `malachite` `CeilingLogBase2` with a small `std`-based helper. Together these remove `malachite` from **both** of its production call sites: it is no longer a production dependency anywhere in the workspace, remaining only as a dev-only cross-check oracle. `malachite` was the workspace's only LGPL-3.0 (copyleft) dependency.

## What's here

- **`kaspa_math::ceil_log_2`** — branchless `ilog2`-based `ceil(log2(x))`; the sync block-locator now uses it instead of `malachite-base`'s `CeilingLogBase2`.
- **`kaspa_math::lehmer`** — stack-allocated, variable-time modular inverse via Lehmer's extended GCD (Knuth TAOCP 4.5.2 Algorithm L). A two-word half-GCD guess peels ~62 bits per 2x2 reduction matrix, with an exact Euclidean step as the fallback. 100% safe Rust, no new dependencies; measured ~6-8% faster than the current `malachite` `mod_inverse` at 3072 bits (see Benchmark results).
- **MuHash wired to Lehmer** — `crypto/muhash`'s `U3072::inverse()` now calls `lehmer::invert`, and the generic `malachite`-backed `Uint::mod_inverse` is **removed** (it had no other production caller).
- **`malachite` dropped from production** — removed from `math`'s `[dependencies]` and from the `pub use malachite_*` re-export in `uint.rs`. It is kept only as a `[dev-dependency]`, used as the bit-for-bit cross-check oracle for the `lehmer.rs`/`lib.rs` tests and the `candidates` bench. `cargo tree -p kaspa-math -e normal -i malachite-base` is now empty.
- **Benchmark** (`benches/candidates.rs`) — criterion comparison of the Lehmer inverse against the current `malachite` `mod_inverse` (the operation it replaces); the Lehmer arm is asserted equal to malachite's result before timing. Third-party candidates evaluated during selection (dashu, crypto-bigint safegcd, num-bigint, ruint) are kept commented out with their measured numbers, and their dev-dependencies were dropped once Lehmer won — so the benchmark pulls in no extra dependencies.
- **Fuzzing** — new `math/fuzz` `u3072` target drives `lehmer::invert` for the MuHash prime and compares bit-for-bit against a num-bigint extended-gcd reference. The generic-`mod_inverse` arms in the `u128`/`u192`/`u256` targets were removed with the method.
- **`lehmer-instrument` feature + example** — per-inverse op-count diagnostics; inert when the feature is off.

## Benchmark results

`cargo bench -p kaspa-math --bench candidates -- modinv_3072_muhash_prime` (criterion, release, Linux x86_64, 13th Gen Intel Core i9-13900HX). Group time covers N=32 inverses; per-inverse = group / 32. Four consecutive runs:

| run | malachite (current) | lehmer (this PR) | lehmer faster by |
|----:|--------------------:|-----------------:|-----------------:|
| 1 | 14.68 µs | 13.90 µs | 5.3% |
| 2 | 14.45 µs | 13.56 µs | 6.2% |
| 3 | 14.52 µs | 13.58 µs | 6.5% |
| 4 | 14.62 µs | 13.49 µs | 7.7% |

Lehmer lands at **~13.5 µs/inverse vs malachite ~14.5 µs/inverse — roughly 6-8% faster**, with non-overlapping 95% confidence intervals on every run, so the speedup is consistent (the exact margin drifts run-to-run with CPU turbo; the lehmer/malachite ratio is the stable figure). The inverse runs once per MuHash finalize, so this is a small but real win on top of dropping the copyleft dependency from the hot path.

## Validation

- `lehmer::invert` cross-checked bit-for-bit against `malachite` `mod_inverse` (now the dev-dep oracle) on 2000 random 3072-bit vectors, plus an oracle-independent `v * inv == 1 (mod p)` self-check, edge cases (1, p-1, 0), and tiny-value inputs that force the multi-limb-quotient `div_rem` fallback.
- MuHash's own `test_inverse` / `test_inverse_edge_case` / `test_mul_div` now run through `lehmer::invert` and pass.
- **Fuzzing:** the new `u3072` target ran clean — **61,427 executions in 46 s, 0 crashes, 0 mismatches** vs the num-bigint reference. The pre-existing muhash multiply/divide fuzz target (which exercises the inverse via `inverse()`) continues to pass.
- `ceil_log_2` checked against an independent oracle across power-of-two boundaries and `u64::MAX`, and for equivalence with the `malachite` routine it replaces.
- `cargo check --workspace`, `cargo clippy --workspace --all-targets`, and `cargo test -p kaspa-math -p kaspa-muhash` are clean/green.

## Notes

- Bumps workspace MSRV 1.91 -> 1.95 (`core::hint::cold_path`, used by the inverse, is stable as of 1.95); CI Lints toolchain and the docker chef images are bumped to match.
- `malachite` is now a **dev-dependency only** — it backs no production code, just the test/bench cross-check oracle. The `product_is_one` self-check, the fuzz target, and the muhash tests give oracle-independent coverage, so the dev-dep can be dropped later with no loss of a production guarantee.

### Extended fuzzing results (3-day soak)

All four targets ran for 72 hours with `strict-asserts` enabled (promoted `debug_assert!` to `assert!`):

| Target | Executions | lehmer.rs line coverage | Result |
|--------|-----------|------------------------|--------|
| u3072 | 139.8M | 92.5% | 0 crashes |
| u128 | 1.65B | 97.1% | 0 crashes |
| u192 | 831M | 95.6% | 0 crashes |
| u256 | 657M | 95.6% | 0 crashes |

Coverage reports: [gist](https://gist.github.com/biryukovmaxim/3d0d0c58dc9123c9354305dfaa8e3d0b) ([u3072](https://gistcdn.githack.com/biryukovmaxim/3d0d0c58dc9123c9354305dfaa8e3d0b/raw/e193a40a8f397cc83e4a6ff478a05bc7d92e024f/u3072-cov.html), [u128](https://gistcdn.githack.com/biryukovmaxim/3d0d0c58dc9123c9354305dfaa8e3d0b/raw/e193a40a8f397cc83e4a6ff478a05bc7d92e024f/u128-cov.html), [u192](https://gistcdn.githack.com/biryukovmaxim/3d0d0c58dc9123c9354305dfaa8e3d0b/raw/e193a40a8f397cc83e4a6ff478a05bc7d92e024f/u192-cov.html), [u256](https://gistcdn.githack.com/biryukovmaxim/3d0d0c58dc9123c9354305dfaa8e3d0b/raw/e193a40a8f397cc83e4a6ff478a05bc7d92e024f/u256-cov.html))
Grown corpora: kaspanet/rusty-kaspa-corpus#1

### DIFF
--- a/math/src/lehmer.rs
+++ b/math/src/lehmer.rs
@@ -68,6 +68,26 @@ macro_rules! count {
     }};
 }
 
+// Invariant checks: `debug_assert!` by default (free in release), promoted to
+// always-on `assert!` under the `strict-asserts` feature so optimized fuzz and
+// test builds verify every invariant regardless of `debug-assertions`.
+#[cfg(not(feature = "strict-asserts"))]
+macro_rules! strict_assert {
+    ($($arg:tt)*) => { debug_assert!($($arg)*) };
+}
+#[cfg(feature = "strict-asserts")]
+macro_rules! strict_assert {
+    ($($arg:tt)*) => { assert!($($arg)*) };
+}
+#[cfg(not(feature = "strict-asserts"))]
+macro_rules! strict_assert_eq {
+    ($($arg:tt)*) => { debug_assert_eq!($($arg)*) };
+}
+#[cfg(feature = "strict-asserts")]
+macro_rules! strict_assert_eq {
+    ($($arg:tt)*) => { assert_eq!($($arg)*) };
+}
+
 /// Fixed-size little-endian limb buffer, blanket-implemented for every
 /// `[u64; N]`. The inversion's working storage; keeping the concrete array
 /// type generic (rather than passing slices around) monomorphizes the limb
@@ -203,6 +223,7 @@ pub fn invert<U: LehmerOps>(value: U::Limbs, modulus: U::Limbs, out: &mut U::Lim
         core::hint::cold_path();
         return false;
     }
+    strict_assert!(cmp_n(value.as_slice(), modulus.as_slice()).is_lt(), "invert requires value < modulus");
 
     // Euclidean state on (x, y) with x >= y, plus the cofactor of `value`.
     // Invariant: x ≡ -t0·value (mod m) and y ≡ t1·value (mod m) when
@@ -296,6 +317,10 @@ pub fn invert<U: LehmerOps>(value: U::Limbs, modulus: U::Limbs, out: &mut U::Lim
     if !swapped && !is_zero(inv.as_slice()) {
         inv = sub_n(&modulus, &inv);
     }
+    strict_assert!(
+        !is_zero(inv.as_slice()) && cmp_n(inv.as_slice(), modulus.as_slice()).is_lt(),
+        "invert: result outside (0, modulus)"
+    );
     *out = inv;
     true
 }
@@ -491,6 +516,12 @@ fn hgcd2(mut ah: u64, mut al: u64, mut bh: u64, mut bl: u64) -> Option<(u64, u64
         }
     }
 
+    // Every entry stays below 2^63 (the half-limb refinement's exit bound, which
+    // the `apply_matrix_*` overflow analyses rely on), and the column updates
+    // preserve the unit determinant of the elementary step matrices.
+    strict_assert!(m00 < 1 << 63 && m01 < 1 << 63 && m10 < 1 << 63 && m11 < 1 << 63, "hgcd2: matrix entry reached 2^63");
+    strict_assert_eq!((m00 as i128) * (m11 as i128) - (m01 as i128) * (m10 as i128), 1, "hgcd2: determinant is not 1");
+
     // Remap M to this module's apply convention:
     // x' = m11·x - m01·y, y' = m00·y - m10·x.
     Some((m11, m01, m10, m00))
@@ -504,9 +535,9 @@ fn gcd_div(n: u128, d: u128) -> (u64, u128) {
     count!(DIVIDES += 1);
     let (n1, n0) = ((n >> 64) as u64, n as u64);
     let (mut d1, mut d0) = ((d >> 64) as u64, d as u64);
-    debug_assert!(d1 != 0);
+    strict_assert!(d1 != 0);
     let (q, r) = (n1 / d1, n1 % d1);
-    if q > d1 {
+    let (q, rem) = if q > d1 {
         // Non-normalized divisor: when `d1`'s top word is small, the single-word
         // estimate `q = n1/d1` can exceed `d1`. Rare but genuinely reachable (a
         // b-step can drop the divisor below 2^32 before an a-step divides by it),
@@ -548,7 +579,10 @@ fn gcd_div(n: u128, d: u128) -> (u64, u128) {
         let (rr0, br) = n0.overflowing_sub(t0);
         let rr1 = r.wrapping_sub(t1).wrapping_sub(br as u64);
         (q, ((rr1 as u128) << 64) | rr0 as u128)
-    }
+    };
+    // Exact reconstruction: cannot wrap for a correct result since q·d + r = n < 2^128.
+    strict_assert!(rem < d && (q as u128).wrapping_mul(d).wrapping_add(rem) == n, "gcd_div: bad quotient or remainder");
+    (q, rem)
 }
 
 /// Multiply-accumulate: `acc + m·w + carry`, returning `(low 64 bits, high 64
@@ -619,8 +653,8 @@ fn apply_matrix_xy<A: LimbArray>(x: &mut A, y: &mut A, len: usize, a: u64, b: u6
     }
     // Non-negative result fitting in `len` limbs: the positive carry-out exactly
     // cancels the negative borrow-out (both name the same high limb, which is 0).
-    debug_assert_eq!(carry_ax, borrow_x, "apply_matrix_xy: x went negative or overflowed len limbs");
-    debug_assert_eq!(carry_dy, borrow_y, "apply_matrix_xy: y went negative or overflowed len limbs");
+    strict_assert_eq!(carry_ax, borrow_x, "apply_matrix_xy: x went negative or overflowed len limbs");
+    strict_assert_eq!(carry_dy, borrow_y, "apply_matrix_xy: y went negative or overflowed len limbs");
 
     let nlx = x[..len].iter().rposition(|&w| w != 0).map_or(0, |i| i + 1);
     let nly = y[..len].iter().rposition(|&w| w != 0).map_or(0, |i| i + 1);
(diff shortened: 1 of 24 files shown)


math/src/lib.rs
```diff
@@ -11,6 +11,16 @@ construct_uint!(Uint256, 4);
...
 
+/// Returns the ceiling of the base-2 logarithm of `x`, i.e. the smallest `k` such that `2^k >= x`.
+///
+/// # Panics
+/// Panics if `x` is 0 (the base-2 logarithm of 0 is undefined).
+#[inline]
+pub const fn ceil_log_2(x: u64) -> u64 {
+    // power of two -> floor; not a power of two -> floor + 1
+    x.ilog2() as u64 + (!x.is_power_of_two()) as u64
+}
```
elichai, line 20: You mentioned branchless, note that this isn't actually branchless, as [ilog2](https://doc.rust-lang.org/std/primitive.u32.html#method.ilog2) panics if `x == 0`
see assembly: https://godbolt.org/z/E8GasceMx

This function is basically the same as:
```rust
pub fn ceil_log_2(x: u64) -> u32 {
    assert!(x > 0);
    u64::BITS - (x - 1).leading_zeros()
}
```

elichai: The original one also panicked, we can either panic, return Option, or make the function return `0` on a `0` input

math/src/lehmer.rs
```diff
@@ -0,0 +1,1127 @@
...
+
+macro_rules! count {
+    ($name:ident += $v:expr) => {{
+        #[cfg(feature = "lehmer-instrument")]
+        instrument::$name.fetch_add($v, core::sync::atomic::Ordering::Relaxed);
+    }};
+}
+
+/// Limb count of the operands (3072 bits in base 2^64).
+const N: usize = 48;
```
elichai: can we make this a const generic to support fuzzing with u128/u192 etc.?

elichai (commented):
first partial review

math/src/lib.rs
biryukovmaxim: we have only one caller of the function, which never passes zero. returning zero sounds fine to me

math/src/lehmer.rs
biryukovmaxim: there are two places where I use exact number:
`div_rem_step` and  `reduce_mod` 

i think it would require to either convert the whole function to be the method of structure so we can get access to corresponding uint type or pass a closure/fn pointer to calc div_rem and mod
or we can add one more generic with concrete Uint type so we can get these function from there.

any suggestions, other ideas?

michaelsutton:
(pls coordinate the merge of this pr bcs i think such a change justifies a master testing period)

elichai:
I let claude+codex grind for a while over this, they didn't find any issues with it, but It's a lot of logic to verify, I think the following can make me more comfortable with those:

1. Find a way to generalize it to `u128` or even to `u64` so the fuzzer can do more meaningful work exhausting inputs.
2. Run the fuzzers for a few days, go over the corpuses coverage to make sure they've hit all branches.
3. Try to prove some of the invariants with kani (e.g. "entries are `< 2^63").
4. turn all `debug_assert` into regular asserts, and try to add many more asserts.

I wonder also how it fairs against using https://github.com/mhogrefe/malachite/blob/master/malachite-nz/src/natural/arithmetic/gcd/extended_gcd.rs but replacing heap allocs with stack allocs, but this leaves us with the licensing issue

biryukovmaxim:
@elichai 
> I let claude+codex grind for a while over this, they didn't find any issues with it, but It's a lot of logic to verify, I think the following can make me more comfortable with those:
> 
> 1. Find a way to generalize it to `u128` or even to `u64` so the fuzzer can do more meaningful work exhausting inputs.
> 2. Run the fuzzers for a few days, go over the corpuses coverage to make sure they've hit all branches.
> 3. Try to prove some of the invariants with kani (e.g. "entries are `< 2^63").
> 4. turn all `debug_assert` into regular asserts, and try to add many more asserts.
> 
> I wonder also how it fairs against using https://github.com/mhogrefe/malachite/blob/master/malachite-nz/src/natural/arithmetic/gcd/extended_gcd.rs but replacing heap allocs with stack allocs, but this leaves us with the licensing issue

Addressing fuzzing/coverage/asserts points:

## 1. Generalized to smaller widths

Done in recent force-push. Lehmer is now generic via `LimbArray`/`LehmerOps`/`LehmerInvert` traits, with implementations for `u64`, `u128`, `Uint192`, `Uint256`, `Uint320`, and `Uint3072`. The fuzz harness has four targets (`u128`, `u192`, `u256`, `u3072`) each cross-checking against independent oracles (native `u128` arithmetic or `num-bigint` extended GCD).

## 2. Multi-day fuzzing soak with coverage analysis

All four targets ran for **3 days** (~72 hours each) with `strict-asserts` enabled (all `debug_assert!` promoted to panicking `assert!`). Zero crashes, zero timeouts, zero OOMs across the board.

| Target | Executions | Corpus | Coverage edges | `lehmer.rs` line cov |
|--------|-----------|--------|----------------|---------------------|
| u3072 | 139.8M | 21,768 entries (87 MB) | 1,130 | 92.5% |
| u128 | 1.65B | 5,437 entries (22 MB) | 745 | 97.1% |
| u192 | 831M | 5,515 entries (22 MB) | 1,078 | 95.6% |
| u256 | 657M | 7,215 entries (29 MB) | 1,250 | 95.6% |

u3072 ran with `-fork=4`; the others single-threaded. All used `-use_counters=1 -use_value_profile=1`.

### Coverage reports (lehmer.rs + uint.rs + int.rs + lib.rs)

- [u3072 coverage](https://gistcdn.githack.com/biryukovmaxim/3d0d0c58dc9123c9354305dfaa8e3d0b/raw/e193a40a8f397cc83e4a6ff478a05bc7d92e024f/u3072-cov.html)
- [u128 coverage](https://gistcdn.githack.com/biryukovmaxim/3d0d0c58dc9123c9354305dfaa8e3d0b/raw/e193a40a8f397cc83e4a6ff478a05bc7d92e024f/u128-cov.html)
- [u192 coverage](https://gistcdn.githack.com/biryukovmaxim/3d0d0c58dc9123c9354305dfaa8e3d0b/raw/e193a40a8f397cc83e4a6ff478a05bc7d92e024f/u192-cov.html)
- [u256 coverage](https://gistcdn.githack.com/biryukovmaxim/3d0d0c58dc9123c9354305dfaa8e3d0b/raw/e193a40a8f397cc83e4a6ff478a05bc7d92e024f/u256-cov.html)

The uncovered `lehmer.rs` lines are mostly the `lehmer-instrument` diagnostics feature gate and a few edge-case early-return paths (e.g. single-limb modulus fast path for u3072, which is unreachable because the MuHash prime is always 48 limbs).

### Corpus

Grown corpora submitted as kaspanet/rusty-kaspa-corpus#1.


## 4. Promoted asserts

Done. The `strict-asserts` feature promotes all `debug_assert!` calls to runtime `assert!`. The fuzz harness enables this feature (`features = ["strict-asserts"]` in `fuzz/Cargo.toml`), so the entire 3-day soak ran with full assertion checking. No assertion failures were triggered.

PR #1054, opened 2026-06-16
fix(wallet/core): max std mass when toccata is active

IzioDev:
straightforward changes to support an increased amount of mass compared to pre-toccata era.

- introduce is_toccata_active parameter on any calling surfaces that i think make sense.
while hardcoding a value where i think it makes sense (reduce the breaking change)

WASM users mostly not impacted unless they want to get future extra mass (or current if on tn10)


wallet/core/src/account/mod.rs
```diff
@@ -86,6 +86,7 @@ impl Inner {
 
 /// Generic wallet [`Account`] trait implementation used
 /// by different types of accounts.
+#[allow(clippy::too_many_arguments)]
```
biryukovmaxim, line 89: what for it was added? what method violates the rule?

biryukovmaxim (commented):
pub(crate) fn calc_compute_mass_for_payload(&self, payload_byte_size: usize) -> u64 {
        // Inputs and outputs also increase transaction bytes, but their wallet
        // compute mass includes script/sigop costs which dominate the normalized
        // transient byte cost for standard transactions. Payload adds raw bytes
        // without such a matching compute cost, so we harden only the payload term.
        //
        // TODO: model the full tx-wide max(compute, normalized transient) instead.
        // This local compromise can slightly overprice payload bytes because it
        // does not credit them with compute slack contributed by inputs/outputs.
        const NORMALIZED_TRANSIENT_BYTE_FACTOR: u64 = 2;
        payload_byte_size as u64 * self.mass_per_tx_byte.max(NORMALIZED_TRANSIENT_BYTE_FACTOR)
    }

could you rename function. correct calculation. track usage and use normalized mass based on params to derive calculated mass. currently its added to compute mass. which will lead to wrong bounds

wallet/core/src/account/mod.rs
IzioDev: `commit_reveal_manual` and `commit_reveal`, i'll make clippy annotation only on these two instead of globally on trait

IzioDev:
Closing in favor of another PR

PR #1056, opened 2026-06-17
fill covenant binding

biryukovmaxim:


IzioDev (approved):
thank you! was def a miss.

impact: vspcv2 consumers couldn't access `covenant` bindings.

PR #1057, opened 2026-06-19
Tmp/generator changes

IzioDev:


wallet/core/src/tx/mass.rs
```diff
@@ -209,121 +126,120 @@ pub const fn transaction_standard_output_serialized_byte_size() -> u64 {
...
             storage_mass_parameter: consensus_params.storage_mass_parameter,
+            mempool_mass_cofactors: consensus_params.mempool_block_mass_cofactors().raw_post(),
         }
     }
 
     pub fn is_dust(&self, value: u64) -> bool {
-        match value.checked_mul(1000) {
-            Some(value_1000) => value_1000 / STANDARD_OUTPUT_SIZE_PLUS_INPUT_SIZE_3X < MINIMUM_RELAY_TRANSACTION_FEE,
-            None => (value as u128 * 1000 / STANDARD_OUTPUT_SIZE_PLUS_INPUT_SIZE_3X as u128) < MINIMUM_RELAY_TRANSACTION_FEE as u128,
-        }
```
IzioDev, line 232: ⚠️ sensitive area: is there a reason why we used `PER_KG` here? this code sounds more complicated than my suggestion, but I want to make sure I didn't miss a point. (same thing happens above)

PR #1063, opened 2026-06-23
Fix/windows risc0 sys alloc shim

LiveLaughLove13:
Fix Windows/MSVC release builds that fail with a duplicate sys_alloc_aligned symbol when rkstratum_cpu_miner or full workspace builds include both kaspa-txscript and RISC0 ZK SDK dependencies.

**Why This Is Needed**

**kaspa-txscript** has a Windows C shim that exports sys_alloc_aligned as a workaround for RISC0 verifier-only builds where the symbol is declared but not provided.

However, when **kaspa-txscript-zk-sdk** is included, it pulls in risc0-zkvm, which enables risc0-zkvm-platform/export-syscalls. In that build graph, RISC0 already exports sys_alloc_aligned.

On Windows/MSVC, both exports are treated as strong symbols, so the linker fails with:

LNK2005: sys_alloc_aligned already defined
LNK1169: one or more multiply defined symbols found

**What The Fix Does**
This change gates the local Kaspa Windows shim behind a Cargo feature:
risc0-platform-exports-syscalls

When kaspa-txscript-zk-sdk is present, it enables this feature on kaspa-txscript, telling the build script not to compile the local C shim because RISC0 already provides the symbol.

The shim still remains available for verifier-only builds where RISC0 does not export sys_alloc_aligned.
Result

This prevents duplicate symbols in full/RISC0 SDK builds while preserving compatibility for narrower txscript verifier builds that still need the shim.

Issue found and reported by @KaspaPulse

### DIFF
--- a/bridge/src/inprocess_node.rs
+++ b/bridge/src/inprocess_node.rs
@@ -13,6 +13,9 @@ impl InProcessNode {
     pub(crate) fn start_from_args(args: kaspad_args::Args) -> Result<Self, anyhow::Error> {
         let _ = fd_budget::try_set_fd_limit(kaspad_daemon::DESIRED_DAEMON_SOFT_FD_LIMIT);
 
+        #[cfg(test)]
+        let runtime = kaspad_daemon::Runtime::default();
+        #[cfg(not(test))]
         let runtime = kaspad_daemon::Runtime::from_args(&args);
         let fd_total_budget =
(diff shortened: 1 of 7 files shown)


LiveLaughLove13:
fixed already in #1064

LiveLaughLove13:
#1064 fixes this

PR #1069, opened 2026-06-30
refactor(wasm/examples): rebuild examples into TypeScript recipes

carlssonk:
Primarily a refactor of the wasm examples: mostly a folder restructuring plus conversion and consolidation of .js -> .ts files.

The nodejs example logic was largely preserved, aside from added top-level comments showing how to run each file, and dropping the old node init / config.json bootstrap in favor of reading secrets from the environment.

The web examples were refactored and the CSS refreshed to look better; they now share the same theme as the kaspa.org site.

Changes

- Consolidated all nodejs examples into TypeScript "recipes", grouped by concern (keys, rpc, transactions, wallet, watch, signing, mining, covenants, zk)
- Converted examples to ES module syntax instead of CommonJS (the canonical, modern way to develop in TS)
- Refactored the web examples and refreshed the CSS
- Updated wasm/README.md and added a new wasm/examples/README.md
- Updated build-release and package.json accordingly


aspect:
Has all of this been manually (or automatically) tested to confirm that all examples function correctly and all code works as expected?

carlssonk:
> Has all of this been manually (or automatically) tested to confirm that all examples function correctly and all code works as expected?

I manually tested every example one by one and confirmed they all work as expected

EDIT: I looked through everything again (this time running them with Deno and Bun as well) and found a couple of issues, they are resolved in the latest commit

IzioDev:
It's a bit difficult (time consuming) to review independently, I'm unsure about the methodology to use here.

I do align with the general direction to move to `.ts` by default, with the only constraint I identified that now a user needs to bring a transpiler as a new layer, which seems trivial nowadays. On the other hand, it's easier to maintain a unique language for examples targeting JS runtime.
"Modernizing" to ESM is a plus to me too 👍

carlssonk:
> It's a bit difficult (time consuming) to review independently, I'm unsure about the methodology to use here.

Yeah this PR got bigger (code change wise) than I initially wanted. The contents themselves in the examples have not changed much, but the git diff recognize most files as completely new code because of js/ts consolidation, and code formatting.

The other main changes is the new README file and configuration files. I'll see if I can split it up into more commits to make it more reviewable.

I found reviewing PRs with any modern LLM is pretty useful as well, as a extra validation step at least =)

> only constraint I identified that now a user needs to bring a transpiler as a new layer, which seems trivial nowadays

Yes, also one thing I learned when making this PR is that newer versions of Node can actually run ts files directly. e.g. `node myscript.ts`

PR #1068, opened 2026-06-29
fix(windows): avoid native cdylib links for bridge deps

LiveLaughLove13:
## Summary

This fixes a Windows MSVC release build failure when building `kaspa-stratum-bridge` with in-process node support.

The bridge depends on `kaspad`, which depends on `kaspa-wrpc-server`. Both `kaspad` and `kaspa-wrpc-server` were configured to build as `cdylib` in addition to normal Rust libraries. On Windows, Cargo therefore tried to link `kaspad_lib.dll` / `kaspa_wrpc_server.dll`, which pulled in `risc0_zkvm_platform` and failed with:

`error LNK2019: unresolved external symbol sys_alloc_aligned`

These crates are consumed as Rust libraries by the bridge and do not expose a native DLL/FFI interface, so this removes the unnecessary `cdylib` crate type from both manifests.

## What This Fixes

Before this change, the following command failed during release linking on Windows:

`cargo run -p kaspa-stratum-bridge --release --bin stratum-bridge -- --config bridge/config.yaml --node-mode inprocess --appdir "C:\KaspaData" --coinbase-tag-suffix dablackasplash --web-dashboard-port 3030 -- --utxoindex --rpclisten=0.0.0.0:16110 --listen=0.0.0.0:16111 --uacomment="RKStratum"`

The failure occurred while linking a generated DLL artifact, not while linking the final bridge executable.

## Validation

Verified on Windows with:

- `cargo fmt`
- `cargo build -p kaspa-wrpc-server --release`
- `cargo build -p kaspa-stratum-bridge --release --bin stratum-bridge`
- `cargo clippy -p kaspa-stratum-bridge --release --bin stratum-bridge -- -D warnings`

PR #1071, opened 2026-07-06
fix(windows): enable export-syscalls in kaspa-txscript to resolve LNK2019

aglov413:
## Summary
Building any Windows binary that transitively depends on `kaspa-txscript`'s zk precompiles fails at link time with:

```
librisc0_zkvm_platform-575849a513f472e2.rlib(risc0_zkvm_platform-575849a513f472e2.risc0_zkvm_platform.5f4d82f6890bba8e-cgu.0.rcgu.o) : error LNK2019: unresolved external symbol sys_alloc_aligned referenced in function _ZN19risc0_zkvm_platform7syscall15sys_alloc_words17h61a46dd92187b489E
```

`risc0-zkvm-platform` only emits (`#[no_mangle]`) its `sys_*` functions (`sys_alloc_aligned`, `sys_alloc_words`, etc.) on non-zkvm targets when its own `export-syscalls` feature is enabled. This PR adds that dependency + feature directly to `kaspa-txscript`, gated to `cfg(windows)`, so every downstream Windows binary gets the fix at once instead of duplicating it per-binary.

### Relation to #1068
This addresses the same root cause as #1068, but as a smaller, more tightly-scoped fix:
- **Windows-only.** This enables `export-syscalls` under `[target.'cfg(windows)'.dependencies]`, so non-Windows builds (Linux/macOS) are completely unaffected. #1068 enabled the feature unconditionally for all platforms.
- **No unrelated changes.** This PR only touches `crypto/txscript/Cargo.toml` and the corresponding `Cargo.lock` update.

### What This Fixes
Before this change, the following command failed during release linking on Windows:

```
cargo run -p kaspa-stratum-bridge --release --bin stratum-bridge -- --config bridge/config.yaml --node-mode inprocess -- --utxoindex --rpclisten=127.0.0.1:16110
```

Resolves the above-mentioned issues building on Windows.

## tests
- [x] `cargo fmt`
- [x] `cargo build -p kaspa-wrpc-server --release`
- [x] `cargo build --release`
- [x] `cargo run -p kaspa-stratum-bridge --release --bin stratum-bridge -- --config bridge/config.yaml --node-mode inprocess -- --utxoindex --rpclisten=127.0.0.1:16110`
- [x] `cargo run -p kaspa-stratum-bridge --release --bin stratum-bridge -- --config bridge/config.yaml --node-mode external`

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -309,6 +309,7 @@ risc0-core = "=3.0.1"
 risc0-zkp = "=3.0.4"
 risc0-groth16 = "=3.0.4"
 risc0-zkvm = { version="=3.0.4", default-features = false }
+risc0-zkvm-platform = "=2.2.2"
 # workflow dependencies that are not a part of core libraries
 
 # workflow-perf-monitor = { path = "../../../workflow-perf-monitor-rs" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -307,6 +307,7 @@ risc0-binfmt = "=3.0.4"
 risc0-circuit-recursion = { version = "=4.0.4", default-features = false }
 risc0-core = "=3.0.1"
 risc0-zkp = "=3.0.4"
+risc0-zkvm-platform = "=2.2.2"
 risc0-groth16 = "=3.0.4"
 risc0-zkvm = { version="=3.0.4", default-features = false }
 # workflow dependencies that are not a part of core libraries
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -307,7 +307,6 @@ risc0-binfmt = "=3.0.4"
 risc0-circuit-recursion = { version = "=4.0.4", default-features = false }
 risc0-core = "=3.0.1"
 risc0-zkp = "=3.0.4"
-risc0-zkvm-platform = "=2.2.2"
 risc0-groth16 = "=3.0.4"
 risc0-zkvm = { version="=3.0.4", default-features = false }
 # workflow dependencies that are not a part of core libraries
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4097,6 +4097,7 @@ dependencies = [
  "risc0-circuit-recursion",
  "risc0-core",
  "risc0-zkp",
+ "risc0-zkvm-platform",
  "secp256k1",
  "serde",
  "serde-wasm-bindgen",
--- a/bridge/Cargo.toml
+++ b/bridge/Cargo.toml
@@ -75,6 +75,7 @@ path = "src/main.rs"
 
 [features]
 rkstratum_cpu_miner = []
+risc0-export-syscall = ["kaspad/risc0-export-syscall"]
 
 [lints]
 workspace = true
--- a/bridge/Cargo.toml
+++ b/bridge/Cargo.toml
@@ -75,7 +75,6 @@ path = "src/main.rs"
 
 [features]
 rkstratum_cpu_miner = []
-risc0-export-syscall = ["kaspad/risc0-export-syscall"]
 
 [lints]
 workspace = true
--- a/crypto/txscript/Cargo.toml
+++ b/crypto/txscript/Cargo.toml
@@ -15,6 +15,7 @@ name = "kip-10"
 [features]
 wasm32-core = ["dep:js-sys", "dep:kaspa-wasm-core"]
 wasm32-sdk = ["wasm32-core", "kaspa-consensus-core/wasm32-sdk"]
+risc0-export-syscall = ["risc0-zkvm-platform/export-syscalls"]
 
 [dependencies]
 ark-bn254.workspace = true
@@ -45,6 +46,7 @@ risc0-binfmt.workspace = true
 risc0-circuit-recursion.workspace = true
 risc0-core.workspace = true
 risc0-zkp.workspace = true
+risc0-zkvm-platform.workspace = true
 secp256k1.workspace = true
 serde-wasm-bindgen.workspace = true
 serde.workspace = true
@@ -54,9 +56,6 @@ smallvec.workspace = true
 thiserror.workspace = true
 wasm-bindgen.workspace = true
 workflow-wasm.workspace = true
-
-[target.'cfg(windows)'.dependencies]
-risc0-zkvm-platform = { version = "=2.2.2", features = ["export-syscalls"] }
 
 [dev-dependencies]
 bytemuck.workspace = true
--- a/crypto/txscript/Cargo.toml
+++ b/crypto/txscript/Cargo.toml
@@ -15,7 +15,6 @@ name = "kip-10"
 [features]
 wasm32-core = ["dep:js-sys", "dep:kaspa-wasm-core"]
 wasm32-sdk = ["wasm32-core", "kaspa-consensus-core/wasm32-sdk"]
-risc0-export-syscall = ["risc0-zkvm-platform/export-syscalls"]
 
 [dependencies]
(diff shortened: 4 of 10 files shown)


biryukovmaxim (changes requested):
@aglov413 HI, thanks for your efforts and contribution. The approach looks to me cleaner than in #1068 . There are also places where I want to enhance it little bit

1) Can you move and pin dependency in the workspace Cargo toml and reuse it the same way we do for other deps
2) Currently it adds the dep uncoditionally, meaning linux that doesnt require direct dep and feature - always get it too, i think the proper way of handling that is to add feature in txengine crate(that adds dep and feature together, like `risc0-export-syscall = ["risc0-zkvm-platform/export-syscalls"]`) and introduce the same feature in upper crates including kaspad and stratum-bridge so they can enable the feature.
so the final call on windows should look like "cargo build --bin kaspad -F txenging-risc0-export-syscall"

LiveLaughLove13:
Reproduced the LNK2019 on latest master on Windows (kaspa_wrpc_server.dll / sys_alloc_aligned) and implemented @biryukovmaxim’s review feedback on a branch based on this PR:

Branch: [LiveLaughLove13/fix/1071-risc0-export-syscalls-feature](https://github.com/LiveLaughLove13/rusty-kaspa/tree/fix/1071-risc0-export-syscalls-feature)
Commit: [a5e1a56c](https://github.com/LiveLaughLove13/rusty-kaspa/commit/a5e1a56c)

Changes:

Workspace pin — risc0-zkvm-platform = "=2.2.2" in root Cargo.toml (same pattern as other risc0 deps)

Feature instead of cfg(windows) auto-dep — on kaspa-txscript: risc0-export-syscalls = ["dep:risc0-zkvm-platform", "risc0-zkvm-platform/export-syscalls"] with risc0-zkvm-platform as an optional workspace dep

Feature forwarding — kaspad and kaspa-stratum-bridge expose risc0-export-syscalls and forward it downstream
Windows verification:

With feature (build succeeds):

cargo build -p kaspa-stratum-bridge --release --bin stratum-bridge --features risc0-export-syscalls
cargo build -p kaspad --release --features risc0-export-syscalls
Without feature (same LNK2019 as reported):

cargo build -p kaspa-stratum-bridge --release --bin stratum-bridge
@biryukovmaxim — this should match what you described: Linux stays unchanged unless the feature is explicitly enabled; Windows builds pass the flag on the final binary, e.g.:

cargo run -p kaspa-stratum-bridge --release --bin stratum-bridge --features risc0-export-syscalls -- --config bridge/config.yaml --node-mode inprocess -- --utxoindex --rpclisten=127.0.0.1:16110
@aglov413 — if you’re open to it, feel free to cherry-pick etc.

aglov413:
> @aglov413 HI, thanks for your efforts and contribution. The approach looks to me cleaner than in #1068 . There are also places where I want to enhance it little bit
> 
> 1. Can you move and pin dependency in the workspace Cargo toml and reuse it the same way we do for other deps
> 2. Currently it adds the dep uncoditionally, meaning linux that doesnt require direct dep and feature - always get it too, i think the proper way of handling that is to add feature in txengine crate(that adds dep and feature together, like `risc0-export-syscall = ["risc0-zkvm-platform/export-syscalls"]`) and introduce the same feature in upper crates including kaspad and stratum-bridge so they can enable the feature.
>    so the final call on windows should look like "cargo build --bin kaspad -F txenging-risc0-export-syscall"

I have made an additional commit with the requested changes, plus a broader feature implementation covering every affected crate.  I can update further if you feel the scope is too large.

Every single crate with crate-type = ["cdylib", "lib"] fails to build individually (kaspad, kaspa-wrpc-server, kaspa-wrpc-client, kaspa-wallet-pskt).
kaspa-stratum-bridge fails because it depends on kaspad, which fails due to its own cdylib output, and transitively pulls in kaspa-wrpc-server's cdylib too.

The following currently still fail on Windows:
cargo build -p kaspad --release
cargo build -p kaspa-wrpc-server --release
cargo build -p kaspa-wrpc-client --release
cargo build -p kaspa-wallet-pskt --release
cargo build -p kaspa-stratum-bridge --release --bin stratum-bridge
cargo run -p kaspa-stratum-bridge --release --bin stratum-bridge -- --config bridge/config.yaml --node-mode inprocess -- --utxoindex --rpclisten=127.0.0.1:16110

The following builds succeed with the implemented feature flag:
cargo build -p kaspad --release -F risc0-export-syscall
cargo build -p kaspa-wrpc-server --release -F risc0-export-syscall
cargo build -p kaspa-wrpc-client --release -F risc0-export-syscall
cargo build -p kaspa-wallet-pskt --release -F risc0-export-syscall
cargo build -p kaspa-stratum-bridge --release --bin stratum-bridge -F risc0-export-syscall
cargo run -p kaspa-stratum-bridge --release --bin stratum-bridge -F risc0-export-syscall -- --config bridge/config.yaml --node-mode inprocess -- --utxoindex --rpclisten=127.0.0.1:16110

It's also worth noting: a plain "cargo build --release" (the full workspace) doesn't actually fail. The issue only surfaces when building these crates individually e.g. "cargo build -p kaspa-stratum-bridge --release --bin stratum-bridge", which is the reason the in-process bridge "run" would fail. My guess is Cargo's feature unification masks it in the full build, if any other crate in the workspace happens to enable a feature that reaches the same risc0-zkvm-platform instance, that feature gets shared across the entire build graph, obscuring the missing symbol for crates that never actually requested it. A scoped build of just one binary doesn't get that accidental coverage, which is presumably why this went unnoticed for a while.

-aglo

Cargo.toml
```diff
@@ -307,6 +307,7 @@ risc0-binfmt = "=3.0.4"
 risc0-circuit-recursion = { version = "=4.0.4", default-features = false }
 risc0-core = "=3.0.1"
 risc0-zkp = "=3.0.4"
+risc0-zkvm-platform = "=2.2.2"
 risc0-groth16 = "=3.0.4"
 risc0-zkvm = { version="=3.0.4", default-features = false }
 # workflow dependencies that are not a part of core libraries
```
biryukovmaxim: sort it  lexicographic, platform after zkvm

biryukovmaxim:
> It's also worth noting: a plain "cargo build --release" (the full workspace) doesn't actually fail. The issue only surfaces when building these crates individually e.g. "cargo build -p kaspa-stratum-bridge --release --bin stratum-bridge", 

yes i think my insisting was wrong. correct way is what it was in first commit, just move pinned dependancy version into workspace in  proper order. and we are good

aglov413:
> > It's also worth noting: a plain "cargo build --release" (the full workspace) doesn't actually fail. The issue only surfaces when building these crates individually e.g. "cargo build -p kaspa-stratum-bridge --release --bin stratum-bridge",
> 
> yes i think my insisting was wrong. correct way is what it was in first commit, just move pinned dependancy version into workspace in proper order. and we are good

done & done.

The following build/run commands are now successful under a Windows environment:
```
cargo build --release
cargo build -p kaspad --release
cargo build -p kaspa-wrpc-server --release
cargo build -p kaspa-wrpc-client --release
cargo build -p kaspa-wallet-pskt --release
cargo build -p kaspa-stratum-bridge --release --bin stratum-bridge
cargo run -p kaspa-stratum-bridge --release --bin stratum-bridge -- --config bridge/config.yaml --node-mode inprocess -- --utxoindex --rpclisten=127.0.0.1:16110
```
thanks for taking the time to review @biryukovmaxim

biryukovmaxim (approved):

Issue #1075, opened 2026-07-12
(RFC) Repackage and publish WASM SDK on npm


carlssonk:
tldr; give the wasm SDK a coherent home on npm, published as a single package: `@kaspa/sdk-wasm`, the pure `wasm-pack` output of the full SDK build, version-coupled to the workspace. Shim/orchestration and other abstractions live in wrapping libraries outside `rusty-kaspa` (e.g. `@kaspa/sdk`, which imports `@kaspa/sdk-wasm`).

> **Updated (2026-07-20):** this RFC originally proposed a family of packages (`kaspa` + `@kaspa/core` / `@kaspa/rpc` / `@kaspa/keygen`) with auto-initializing wrapper entries, published by CI via trusted publishing (OIDC). It has been reworked in place; larger changes are rewritten with a note like this one, smaller corrections are shown as ~~strikethrough~~.

---

This RFC is written with two properties in mind; DX and supply-chain security. Motivation, details, and concrete sketches of what this could look like are collapsed at the end of this doc.

## The decisions needed

- Build target - I propose shipping a single `--target web` build for both browser and server runtimes, instead of separate web + nodejs builds. Published as the pure `wasm-pack` output with no wrapper entries; the loading/auto-init story moves to the wrapping libraries. See **Package family** section below. **Edit:** originally proposed small wrapper entries that auto-initialize the wasm, see **Universal WASM loading** section below.
- Runtime deps - drop the websocket shim so the package ships with zero
  runtime dependencies. See **Dropping the WebSocket dependency** section below.
- Naming - ~~an unscoped `kaspa` + scoped `@kaspa/*` tiers.~~ **Edit:** a single `@kaspa/sdk-wasm`; wrapping libraries such as `@kaspa/sdk` live outside this repo. See **Package family** section below.
- Existing names - deprecating `kaspa-wasm` for ~~`kaspa`~~ `@kaspa/sdk-wasm`. See **Ownership & bootstrap** section below.
- Publishing trigger - no auto-publishing for now. Semi "auto-publishing" is a greater question than npm, and we don't have it for the cargo crates today either. Instead, a script prepares the tarball so owners can publish more easily by hand. See **Publishing** section below. **Edit:** originally proposed CI publishing via staged publishing + trusted publishing (OIDC).

## Package family

[...]

PR #1078, opened 2026-07-21
feat(wasm): add npm packaging step and smoke test for @kaspa/sdk-wasm

carlssonk:
See #1075 for context

Adds 3 new steps to the `build-wasm32` CI job that assembles the full kaspa WASM SDK (as `@kaspa/sdk-wasm`) for easier publishing:

1. runs the assemble step that writes package.json metadata, and packs a publish-ready tarball for manual release.
2. runs a smoke test that installs the packed tarball into a throwaway project and exercises the SDK offline.
3. uploads the tarball as a new `kaspa-sdk-wasm-npm` artifact (the existing `kaspa-wasm32-sdk` artifact is unchanged).

Additionally bumps the job's Node from 20 (EOL) to 24, which enables the require() interop check in the smoke test


wasm/assemble-npm
```diff
@@ -0,0 +1,28 @@
...
+set -e
+
+# Workaround for Rust 1.87.0
+# https://github.com/rust-lang/rust/issues/141048
+export RUSTFLAGS=-Ctarget-cpu=mvp
+
+cd "$(dirname "$0")"
+
+if [ "$1" == "--skip-build" ]; then
+    shift
```
IzioDev: doesn't seem useful here as there are no other arguments to process, but we can keep it for future-proof

wasm/assemble-npm
```diff
@@ -0,0 +1,28 @@
...
+else
+    wasm-pack build --weak-refs --target web --out-name kaspa --out-dir web/kaspa --features wasm32-sdk $@
+fi
+
+node build/npm/assemble.js
+
+pushd npm-release/sdk-wasm > /dev/null
+npm pack --pack-destination .. > /dev/null
```
IzioDev: nit: replace `..` with `../`

wasm/build/npm/smoke-test.js
```diff
@@ -0,0 +1,107 @@
...
+// Installs the tarball into a throwaway consumer project and exercises the
+// SDK offline, so the exact bytes that would ship to npm are what's tested.
+
+const { execFileSync } = require('child_process');
+const fs = require('fs');
+const os = require('os');
+const path = require('path');
+const { workspaceVersion } = require('./assemble');
+
+const wasmDir = path.resolve(__dirname, '..', '..');
```
IzioDev: `wasmDir` only is used line 28, we can define it there instead. also, since it's only used to create `releaseDir`, we can just create `releaseDir` directly

wasm/build/npm/smoke-test.js
```diff
@@ -0,0 +1,107 @@
...
+
+let tarball = process.argv[2];
+if (!tarball) {
+  const releaseDir = path.join(wasmDir, 'npm-release');
+  const tarballs = fs.existsSync(releaseDir) ? fs.readdirSync(releaseDir).filter((f) => f.endsWith('.tgz')) : [];
+  if (tarballs.length !== 1) {
+    fail(`expected exactly one .tgz in ${releaseDir}, found ${tarballs.length}; run 'bash assemble-npm' first`);
+  }
+  tarball = path.join(releaseDir, tarballs[0]);
+}
```
IzioDev: it would be simpler to require a tarball argument directly. CI needs to feed it though.

wasm/build/npm/smoke-test.js
```diff
@@ -0,0 +1,107 @@
...
+tarball = path.resolve(tarball);
+if (!fs.existsSync(tarball)) {
+  fail(`no such tarball: ${tarball}`);
+}
+
+const tmpDir = fs.mkdtempSync(path.join(os.tmpdir(), 'kaspa-sdk-wasm-smoke-'));
+process.on('exit', () => fs.rmSync(tmpDir, { recursive: true, force: true }));
```
IzioDev: note: on Node v24 we could use `mkdtempDisposableSync` which auto delete

wasm/build/npm/smoke-test.js
```diff
@@ -0,0 +1,107 @@
...
+// the installed package must be the pure wasm-pack output with rewritten metadata
+const installedDir = path.join(tmpDir, 'node_modules', '@kaspa', 'sdk-wasm');
+const pkg = JSON.parse(fs.readFileSync(path.join(installedDir, 'package.json'), 'utf8'));
+const version = workspaceVersion();
+
+if (pkg.name !== '@kaspa/sdk-wasm') fail(`unexpected package name: ${pkg.name}`);
+if (pkg.version !== version) fail(`package version ${pkg.version} != workspace version ${version}`);
+if (pkg.dependencies && Object.keys(pkg.dependencies).length > 0) {
+  fail(`package ships runtime dependencies: ${Object.keys(pkg.dependencies).join(', ')}`);
+}
```
IzioDev: do we have runtime deps?

wasm/build/npm/smoke-test.js
```diff
@@ -0,0 +1,107 @@
...
+for (const file of ['kaspa.js', 'kaspa.d.ts', 'kaspa_bg.wasm', 'kaspa_bg.wasm.d.ts', 'README.md', 'LICENSE']) {
+  if (!fs.existsSync(path.join(installedDir, file))) fail(`missing ${file} in installed package`);
+}
+console.log(`installed ${pkg.name}@${pkg.version}`);
+
+// offline consumer: manual init on a server runtime
+const consumerBody = `
+import { readFileSync } from 'node:fs';
+import { createRequire } from 'node:module';
+import { initSync, Mnemonic, XPrv, createAddress, NetworkType, sompiToKaspaString } from '@kaspa/sdk-wasm';
```
IzioDev: is there a motivation to smoke test more than one common object? (e.g: NetworkType)

wasm/build/npm/assemble.js
```diff
@@ -0,0 +1,79 @@
...
+  const match = section && section.split(/^\[/m)[0].match(/^version\s*=\s*"([^"]+)"/m);
+  if (!match) {
+    throw new Error(`unable to parse [workspace.package] version from ${cargoTomlPath}`);
+  }
+  return match[1];
+}
+
+function assemble() {
+  if (!fs.existsSync(path.join(sourceDir, 'package.json'))) {
+    fail(`missing wasm-pack output at ${sourceDir}\n` + "run 'bash assemble-npm' or 'bash build-release' first");
```
IzioDev: `build-release` is the correct file to point at, `assemble-npm` calls `assemble.js` as is.

wasm/build/npm/assemble.js
```diff
@@ -0,0 +1,79 @@
...
+  }
+
+  const pkg = JSON.parse(fs.readFileSync(path.join(sourceDir, 'package.json'), 'utf8'));
+  const version = workspaceVersion();
+
+  if (pkg.version !== version) {
+    fail(`web/kaspa is stale: built version ${pkg.version} != workspace version ${version}\n` + "rebuild with 'bash assemble-npm'");
```
IzioDev: rebuild with `build-release`

wasm/build/npm/assemble.js
```diff
@@ -0,0 +1,79 @@
...
+  }
+
+  for (const file of BUILD_FILES) {
+    if (!fs.existsSync(path.join(sourceDir, file))) {
+      fail(`missing ${file} in ${sourceDir}`);
```
IzioDev: can also add "rebuild with `build-release`" here

wasm/build/npm/assemble.js
```diff
@@ -0,0 +1,79 @@
...
+  }
+
+  pkg.name = '@kaspa/sdk-wasm';
+  pkg.homepage = 'https://github.com/kaspanet/rusty-kaspa#readme';
+  pkg.bugs = { url: 'https://github.com/kaspanet/rusty-kaspa/issues' };
+  pkg.keywords = ['kaspa', 'wasm', 'sdk', 'blockdag', 'wallet', 'rpc'];
+  pkg.publishConfig = { access: 'public' };
+  if (!pkg.files.includes('kaspa_bg.wasm.d.ts')) {
+    pkg.files.push('kaspa_bg.wasm.d.ts');
+  }
```
IzioDev, line 59: we can systematically push as is, and add comment on top:
"waiting for https://github.com/wasm-bindgen/wasm-pack/issues/1193 to be solved"

wasm/build/npm/README.md
```diff
@@ -0,0 +1,89 @@
...
+import { Mnemonic, XPrv, createAddress, NetworkType } from '@kaspa/sdk-wasm';
+
+const mnemonic = Mnemonic.random();
+const xprv = new XPrv(mnemonic.toSeed());
+const publicKey = xprv.derivePath("m/44'/111111'/0'/0/0").toXPub().toPublicKey();
+const address = createAddress(publicKey, NetworkType.Mainnet);
+console.log(String(address));
+```
+
+Connect to a public node and query it over RPC:
```
IzioDev: "connect one of the public community-maintained node" probably would be a more accurate term

wasm/build/npm/README.md
```diff
@@ -0,0 +1,89 @@
...
+await rpc.connect();
+try {
+    const { isSynced, serverVersion } = await rpc.getServerInfo();
+    console.log(`connected to Kaspa node ${serverVersion}, synced: ${isSynced}`);
+} finally {
+    await rpc.disconnect();
+}
+```
+## Documentation
+
```
IzioDev, line 82: I'd suggest to add docs.kaspa.org

IzioDev (changes requested):

wasm/build/npm/smoke-test.js
carlssonk: I could bump `node-version` to 24 (and do it on all other CI steps that use Node for consistency)

wasm/build/npm/smoke-test.js
carlssonk: no wasm-pack doesn't emit any runtime deps. I added this just as a sanity-check, but I could remove the check to keep it simpler

wasm/build/npm/smoke-test.js
carlssonk: not much no, the smoke test is mostly a packaging check plus making sure the wasm loads correctly. And one round-trip through a wasm-backend call should be enough

wasm/build/npm/assemble.js
carlssonk: `assemble-npm` does build the package if `--skip-build` isn't provided, but it makes sense to leave the build responsibility to the `build-*` scripts. I'll update this 👍🏻

wasm/assemble-npm
carlssonk: removed

wasm/build/npm/smoke-test.js
carlssonk: implemented this change in the latest commit

wasm/build/npm/smoke-test.js
IzioDev: ok let's keep it as is :)

wasm/build/npm/smoke-test.js
IzioDev: i suggest to reduce to one common object that is the less subject to future breaking change, so we don't have to maintain this file across time

wasm/build/npm/smoke-test.js
IzioDev: I saw you bumped and didn't use `mkdtempDisposableSync`, all good it works fine as is, was just wondering if you found `mkdtempDisposableSync` not suitable here

carlssonk: It wasn't suitable because the program terminated with `process.exit(1)` which prevents cleanup in certain cases. I pushed a new commit fixing this and improved error handling + added `mkdtempDisposableSync`

wasm/build/npm/smoke-test.js
carlssonk: sounds good, I kept `sompiToKaspaString` for a deterministic round-trip, and `Mnemonic.random()` because it's the only call that exercises the entropy import (the crypto.getRandomValues wiring in the generated kaspa.js file)

PR #1083, opened 2026-07-27
Toccata Cleanup - Part 2

biryukovmaxim:
### Stack

1. #1082 Part 1 - transition-only machinery (base: `master`)
2. #1083 Part 2 - consensus activation checks
3. #1084 Part 3 - txscript covenants flag
4. #1086 Part 4 - mining and mempool policy

Each PR targets the branch below it, so its diff contains only its own layer. Review bottom to top.

---
Second of a stacked series removing dead Toccata activation handling. Stacked on #1082; review that one first, this PR's diff is only the changes on top of it.

Toccata activated on mainnet at DAA `474_165_565` and testnet at `467_579_632`. With a 30 hour pruning duration and a 432,000 block finality depth, `pruning_point.daa_score - finality_depth > toccata_activation` holds on every live network. That is what makes the pre-Toccata branches below unreachable, including the ones that operate inside the `[pruning_point - finality_depth, pruning_point]` window. Part 1 activated Toccata on devnet, so all networks are now post-activation.

This part collapses every activation check inside the consensus crates to its post-Toccata branch, and drops each `toccata_activation` field once its last reader is gone. `Params::toccata_activation` deliberately stays; it is still read outside consensus and is removed at the top of the stack.

## Collapsed checks

* **Transaction validator**: the contextual `check_transaction_version` and `check_transaction_signature_scripts_in_header_context` are deleted, and the isolation-stage checks become the single enforcement point. This is what the deleted `TODO(post-toccata)` asked for, and it restores the arrangement described by `validate_tx_in_isolation`'s own doc comment: checks live in header context only while they are activation dependent. `check_covenant_info` and the script engine's covenant flag become unconditional.
* **Coinbase**: the coinbase transaction is always built at `TX_VERSION_TOCCATA`.
* **Header processor**: `check_header_version_in_context` is deleted and the context-free `check_header_version` is restored in pre-GHOSTDAG validation. Historical headers are unaffected: pruning-proof headers are inserted directly by `apply_proof` without passing isolation validation, and ordinary IBD headers start at the current post-activation pruning point.
* **Virtual processor**: `accepted_id_digests` is always the single-element seq_commit form, chain qualification is unconditional, the seq commit accessor is always present, and the inactivity shortcut no longer folds to `ZERO_HASH` for pre-Toccata ancestors. `calc_accepted_id_merkle_root` is deleted once both callers are gone.
* **Pruning**: SMT store pruning is unconditional, and the pruning-proof chain segment is always built on both the syncer and syncee sides, without the inner breaks that stopped below the activation.
* **Body processor** and **`get_seq_commit_lane_proof`** lose their remaining gates.

`ForkedParam` call sites move from `.get(daa_score)` to `.after()` rather than to plain constants. The field types stay as they are and collapse to plain values at the top of the stack, so this PR contains no parameter-shape changes.

## Genesis accepted-id digests

`compute_genesis_accepted_id_digests` branched on the genesis DAA score. Mainnet and testnet genesis predate the activation, so the pre-Toccata branch was the live one for them, but it is collapsed here along with the rest.

A node with an existing database does not process genesis at all. A node starting from an empty database does compute this value, but discards it while syncing, because the headers proof builds a staging consensus that skips genesis and seeds virtual state at the current pruning point. Any network defined from now on is post-activation at genesis. The reasoning is recorded in a doc comment on the function.

This is also what allows `Params::toccata_activation` to be deleted entirely later in the series, rather than kept for a computation that permanently depends on the activation score.

## Removed plumbing

`toccata_activation` is dropped from `TransactionValidator`, `CoinbaseManager`, `PruningProofManager`, `VirtualStateProcessor` and `SeqCommitAccessor`, together with their constructor arguments and `services.rs` call sites. `UtxoProcessingContext::accepted_tx_ids` and `accepted_tx_versions` are removed as well: the collapse left them written on every accepted transaction and read nowhere.

## Test data

**What changes in test data, how and why.** Exactly one fixture changes, `goref_custom_pruning_depth/blocks.json.gz`, regenerated by rerunning `simpa/generate-json-tests-data.sh` rather than edited by hand. Its `OverrideParams` header line is identical to the previous one except `toccata_activation: 1000 -> 0`; `pruning_depth: 450`, `finality_depth: 200` and every other param are unchanged, so the fixture still exercises the custom pruning depth it exists for. `goref_custom_pruning_depth_test` is not touched and still passes.

`simpa`'s `--test-pruning` profile pinned `toccata_activation` at DAA 1000, so `goref_custom_pruning_depth` was a DAG spanning the activation with mixed header versions. Now that the header version check is unconditional, such a DAG is not valid input, so the toggle is removed and the fixture regenerated; it pins `toccata_activation: 0` and is uniformly version 2. The file is a gzipped block dump, so it is meant to be reproduced with `simpa/generate-json-tests-data.sh` rather than read as a diff.

Integration tests covering the pre-activation to post-activation transition are deleted, since the transition no longer exists in the code. The durable invariants they shared coverage with (covenants, push limits, native payloads, KIP-10, header and transaction version enforcement) remain covered. One assertion in `header_in_isolation_validation_test` moves from `BLOCK_VERSION` to `TOCCATA_BLOCK_VERSION`, following the version check into isolation.

## Not in scope

`Params::toccata_activation`, the `ForkedParam` getters, `BLOCK_VERSION := 2` and `TOCCATA_BLOCK_VERSION` are handled at the top of the stack. The txscript `EngineFlags` covenant flag, the `MiningManager`/`Mempool` plumbing and `protocol/flows` follow in later parts. Pre-Toccata storage decoders (`utxo/pre_toccata.rs`, the EOF tolerant `UtxoEntry` deserializer, `db_compat.rs`) are permanent DB compatibility and are never removed, as are the `TX_VERSION_TOCCATA` keyed transaction rules.

A `TODO` is left on `VirtualState::accepted_id_digests`: it now always holds exactly one hash, and could become a dedicated type that serializes as a one-element vector, keeping the stored bytes identical while making the invariant explicit.


consensus/src/model/stores/virtual_state.rs
```diff
@@ -35,6 +35,10 @@ pub struct VirtualState {
     pub utxo_diff: UtxoDiff, // This is the UTXO diff from the selected tip to the virtual. i.e., if this diff is applied on the past UTXO of the selected tip, we'll get the virtual UTXO set.
     /// Pre-KIP21: tx digests for accepted_id_merkle_root computation.
     /// Post-KIP21: single-element vec containing the seq_commit hash.
+    // TODO: post-Toccata this always holds exactly one element. Replace `Vec<Hash>` with a
+    // dedicated single-hash type whose Serialize/Deserialize impl encodes as a one-element
+    // vector, so the on-disk and wire format stay byte-identical while the type makes the
+    // single-element invariant explicit and drops the Vec from the API.
```
coderofstuff, line 41: You can leave the comment here but please capture it in a github issue and assign to someone

coderofstuff (dismissed):

coderofstuff (approved):

consensus/src/model/stores/virtual_state.rs
biryukovmaxim: Tracked in #1099.

PR #1084, opened 2026-07-27
Toccata Cleanup - Part 3

biryukovmaxim:
### Stack

1. #1082 Part 1 - transition-only machinery (base: `master`)
2. #1083 Part 2 - consensus activation checks
3. #1084 Part 3 - txscript covenants flag
4. #1086 Part 4 - mining and mempool policy

Each PR targets the branch below it, so its diff contains only its own layer. Review bottom to top.

---

Third of a stacked series removing dead Toccata activation handling.

After Part 2, the script engine's `covenants_enabled` flag is fed `true` unconditionally by its only production source. This part removes the flag from `crypto/txscript` and makes post-Toccata script semantics unconditional.

## Changes

* **`EngineFlags.covenants_enabled` removed**, along with all 32 gated opcode bodies in `opcodes.rs`, the branches in `data_stack.rs` (numeric deserialization is always lenient, element size is the constant) and in `script_builder.rs`. Resolves the `TODO(post-toccata)` about changing the default values.
* **Pre-Toccata limits deleted**: `MAX_SCRIPTS_SIZE_PRE_TOCCATA`, `MAX_SCRIPT_ELEMENT_SIZE_PRE_TOCCATA`, `MAX_OPS_PER_SCRIPT_PRE_TOCCATA`, and the three `max_*(bool)` helpers. The surviving constants drop the `_POST_TOCCATA` suffix.
* **`OpCodeMetadata::is_disabled` no longer takes flags**; the disabled set is now fixed: `OpLeft`, `OpRight`, `Op2Mul`, `Op2Div`, `OpLShift`, `OpRShift`.
* **The sigop counting resource meter is removed** (`RuntimeResourceMeter::new_sigops`, `RuntimeSigOpCounter`, `TxScriptError::ExceededSigOpLimit`). Script-units metering was already the only path taken with the flag enabled, and `sigop_script_units` is still charged before every Schnorr and ECDSA verification.
* `check_minimal_data_push` is removed from the opcode trait (no remaining caller), and `TxScriptEngine` no longer stores its `flags` field.
* **The `enforce_minimal` parameter is removed** from the `OpcodeData<T>::deserialize` trait method and from the public `deserialize_i64`, together with `check_minimal_data_encoding`. See below.

## Minimal encoding enforcement is unreachable

Once the stack stopped deriving this flag from `covenants_enabled`, every production caller passed `false`: `Stack::pop_items` (which covers the bool, i32, i64, hash and ZK precompile pops), `OpIfDup`, and `p2sh_sig_scanner`. The minimal encoding check therefore could not run, so the flag and `check_minimal_data_encoding` are dead code rather than a behavior switch, and removing them changes no script's outcome.

`OpcodeData<i32>` collapses to the `SizedEncodeInt<8>` decode with a checked conversion, which is the arm production already took; the `SizedEncodeInt<4>` arm existed only under `enforce_minimal`. `TxScriptError::NotMinimalData` is kept: it still covers numbers encoded in more than 8 bytes and oversized `OP_NUM2BIN` targets.

## PSKT extraction is now bounded

`TxScriptEngine::from_transaction_input` uses `ScriptUnits(u64::MAX)`, i.e. no explicit bound. Previously `EngineFlags::default()` meant covenants disabled, which selected the finite sigop meter, so `Pskt::extract_tx` was bounded at 255 sigops as a side effect of running under pre-Toccata rules. With the flag gone it would have become unbounded while verifying a user supplied PSKT.

`extract_tx` now passes `input.compute_commit.allowed_script_units()`, the same limit consensus uses. This also fixes the other half of the problem: running under pre-Toccata rules, extraction would have rejected valid covenant transactions.

## Public API changes

Breaking for external Rust crate users:

* `EngineFlags.covenants_enabled` field removed.
* `MAX_SCRIPTS_SIZE_PRE_TOCCATA`, `MAX_SCRIPT_ELEMENT_SIZE_PRE_TOCCATA`, `MAX_OPS_PER_SCRIPT_PRE_TOCCATA` deleted; `MAX_*_POST_TOCCATA` renamed to drop the suffix.
* `max_scripts_size(bool)`, `max_script_element_size(bool)`, `max_ops_per_script(bool)` deleted.
* `post_toccata_p2sh_sig_scanner` renamed to `p2sh_sig_scanner`.
* `deserialize_i64` no longer takes an `enforce_minimal` argument.
* `RuntimeResourceMeter::new_sigops`, `RuntimeSigOpCounter`, `RuntimeScriptUnitMeter` removed.
* `TxScriptError::ExceededSigOpLimit` and `ScriptBuilderError::ExplicitDataPushRejected` variants removed.
* `OpCodeMetadata::is_disabled` signature changed; `check_minimal_data_push` removed from the trait.

The wasm `ScriptBuilderFlags.covenantsEnabled` option is **kept as a deprecated no-op** so existing TypeScript callers keep compiling. It still type-checks the value and discards it. It can be removed in a future SDK major version.

## Test vectors

**What changes in test data, how and why.** No vector is hand written or hand edited here. The new `test-data/script_tests.json` is byte identical to the deleted `script_tests_covenants.json`, so the change is a swap of which file the suite runs: the covenant aware set becomes the only set. It is swapped because with `covenants_enabled` gone there is a single set of script rules, leaving the pre-fork vector file with no configuration to run under. The per-scenario accounting below compares the two files and is what a reviewer should check instead of reading the diff.

`test-data/script_tests.json` is replaced by the covenant aware set that previously lived in `script_tests_covenants.json`, which is deleted. Comparing the two:

* 855 vectors before, 955 after.
* 52 scenarios dropped, all of them rules that changed at the fork: 31 `BAD_OPCODE` (opcodes that are now defined), 13 `DISABLED_OPCODE` (`CAT`, `SUBSTR`, `INVERT` and friends are now enabled), 5 `OP_COUNT`, 2 `PUSH_SIZE`, 1 `SCRIPT_SIZE` (limits raised to 1,000,000).
* 84 scenarios kept with updated expectations, all consistent with lenient encoding: 54 `UNKNOWN_ERROR` to `OK`, 21 `MINIMALDATA` to `OK`, 4 `BAD_OPCODE` to `INVALID_STACK_OPERATION`, 2 `UNBALANCED_CONDITIONAL` to `INVALID_STACK_OPERATION`, 2 `MINIMALIF` to `UNBALANCED_CONDITIONAL`, 1 `DISABLED_OPCODE` to `OK`.

Tests that existed only to cover pre-Toccata behavior are deleted (the disabled-vs-enabled push unit comparison, the pre-covenants vector file, the 520 byte element boundary case, and the sigop count limit case). Tests asserting surviving behavior are migrated rather than removed.

In `data_stack.rs`, the thirteen `test_deserialize` cases that previously expected `NotMinimalData` are migrated rather than dropped: each now asserts the value its non-minimal encoding decodes to (`80` to `0` for negative zero, `0100` to `1`, `ff7f00` to `32767`, `0009000100` to `16779520`, and so on, each matching the value already noted in that case's comment). The same fixtures now exercise the lenient decoder that production runs. The cases covering numbers longer than 8 bytes are unchanged.

## Not in scope

The `MiningManager`/`Mempool` activation plumbing and `protocol/flows` follow in later parts, as does the `Params` surface. `TX_VERSION_TOCCATA` keyed rules and the pre-Toccata storage decoders are permanent and are never removed.


wallet/pskt/src/pskt.rs
```diff
@@ -491,7 +491,18 @@ impl PSKT<Extractor> {
...
+                TxScriptEngine::from_transaction_input_with_script_units_limit(
+                    &tx,
+                    input,
+                    idx,
+                    entry,
+                    ctx,
+                    Default::default(),
+                    script_units_limit,
+                )
+                .execute()?;
```
coderofstuff: Why did this change to pass in script_unit_limit?

coderofstuff (dismissed):

wallet/pskt/src/pskt.rs
someone235: I think you should have cov ctx in this context. Also nit, but I guess it's better to use the passed params for building the flags. You can use the below code


            let covenants_ctx = CovenantsContext::from_tx(&tx)
                .map_err(|err| ExtractError::TxScriptError(kaspa_txscript_errors::TxScriptError::from(err)))?;
            let ctx = EngineCtx::new(&cache).with_reused(&reused_values).with_covenants_ctx(&covenants_ctx);
            let flags = EngineFlags { sigop_script_units: Gram(params.mass_per_sig_op).into() };

            tx.populated_inputs().enumerate().try_for_each(|(idx, (input, entry))| {
                // Bound execution by the input's committed script units, matching consensus validation.
                let script_units_limit = input.compute_commit.allowed_script_units();
                TxScriptEngine::from_transaction_input_with_script_units_limit(&tx, input, idx, entry, ctx, flags, script_units_limit)
                    .execute()?;
                <Result<(), ExtractError>>::Ok(())
            })?;

crypto/txscript/src/wasm/builder.rs
```diff
@@ -93,8 +93,10 @@ impl TryFrom<ScriptBuilderOptions> for EngineFlags {
         let mut engine_flags = Self::default();
 
         if let Some(value) = flags.try_get_value("covenantsEnabled")? {
-            engine_flags.covenants_enabled =
-                value.as_bool().ok_or_else(|| Error::convert("flags.covenantsEnabled", "expected boolean"))?;
+            // Deprecated no-op: covenant opcodes and their script limits are always
+            // enabled. The key is still accepted (and type-checked) so existing
+            // callers keep compiling, but the value has no effect.
+            value.as_bool().ok_or_else(|| Error::convert("flags.covenantsEnabled", "expected boolean"))?;
```
someone235: nit, but maybe it's better to write it more explicitly? e.g.

```
if value.as_bool().is_none(){
  return Err(...)
}
```

crypto/txscript/src/lib.rs
```diff
@@ -293,9 +274,9 @@ pub fn post_toccata_p2sh_sig_scanner(signature_script: &[u8], spk: &ScriptPublic
                 prev_opcode_multisig_count = Some((opcode_value - codes::Op1 + 1) as u64);
             }
             ..=codes::OpPushData4 => {
-                // Post-Toccata allows non-minimal data pushes, so
+                // Non-minimal data pushes are allowed, so
                 // OpData*/OpPushData* forms before multisig are counted.
```
someone235: Once non-minimal encoding is allowed I think we can just remove this comment

crypto/txscript/src/lib.rs
```diff
@@ -2114,16 +2048,6 @@ mod tests {
                 sig_op_limit: 2,
                 should_pass: true,
             },
-            TestCase {
```
someone235, line 2117: Why did you remove this test?

crypto/txscript/src/lib.rs
```diff
@@ -2114,16 +2048,6 @@ mod tests {
                 sig_op_limit: 2,
```
someone235, line 2046: I think this is practically unused now. You should build the engine with `from_transaction_input_with_script_units_limit` so this will be enforced.

someone235 (changes requested):

crypto/txscript/src/wasm/builder.rs
biryukovmaxim: im fine with ok_or_else

crypto/txscript/src/lib.rs
biryukovmaxim: applied

crypto/txscript/src/lib.rs
biryukovmaxim: returned back

someone235 (dismissed):

someone235 (approved):

PR #1089, opened 2026-07-29
Toccata Cleanup - Part 6

biryukovmaxim:
Part 6 of the Toccata activation cleanup series, and the last one.

| Part | PR |
|---|---|
| 1 - transition machinery | #1082 |
| 2 - consensus leaves | #1083 |
| 3 - txscript | #1084 |
| 4 - mining/mempool | #1086 |
| 5 - protocol/flows | #1087 |
| 6 - params (final) | this PR |

`Params::toccata_activation` and every `ForkedParam` keyed on it are gone. Each forked parameter
collapses to the value it already resolves to today.

### Parameters

* `prior_block_mass_limits` and `new_transient_mass_limit` merge into `block_mass_limits`
  (`{ compute: 500_000, storage: 500_000, transient: 1_000_000 }`).
* `prior_max_signature_script_len` and `new_max_signature_script_len` merge into
  `max_signature_script_len` (250_000).
* `BLOCK_VERSION` becomes 2 and `TOCCATA_BLOCK_VERSION` is deleted. Genesis is unaffected: all four
  genesis blocks carry an explicit `version: 0`.
* `block_version()` and `block_mass_cofactors()` return plain values, and the processor and validator
  fields holding them narrow from `ForkedParam<T>` to `T`.

### Mempool mass policy

`Params::mempool_block_mass_limits()` and `mempool_block_mass_cofactors()` are deleted rather than
collapsed. They existed only to delay mempool policy 24 hours behind the consensus transient mass
relaxation, so once activation is past they return exactly what `block_mass_limits` returns. The
mempool now takes the consensus limits directly.

That removes the last consumer of the per-transaction DAA lookup used to select cofactors, so
`virtual_daa_score` drops out of the mempool limit, orphan mass, RBF and pool eviction paths.
`added_at_daa_score` stays, since it is still used for expiry.

### Override params

`OverrideParams` keeps `deny_unknown_fields`. A config file still carrying the retired
`"toccata_activation"` key (or either of the retired mass limit / signature script length keys) now
fails startup with an `unknown field` error, which is the intended outcome: overriding params is
rejected on mainnet, so these files only exist on simnet and devnet, and a loud failure telling the
operator which key to delete beats silently ignoring a key that no longer does anything. The strict
check also keeps catching plain misspellings.

### Test data

The three `blocks.json.gz` DAG fixtures carry an `OverrideParams` blob on their first line using the
retired field names. Only that line is rewritten - the merged mass limit and signature script length
keys, and `toccata_activation` dropped. Every block record is unchanged and the resolved parameter
values are identical, so the fixtures did not need regenerating.

### Not included

`docs/toccata-guide.md` is untouched. It documents a shipped release rather than activation
machinery, and retensing its nine pre-activation passages would bury this diff. Worth its own PR.


coderofstuff (dismissed):

coderofstuff (approved):

IzioDev (approved):
administrative formality

PR #1104, opened 2026-08-28
DAGKnight: rewrite executor with index-based conflict resolution and typed search modes

biryukovmaxim:
Rewrites the dagknight executor and conflict zone manager

What changes:
- DagknightExecutor resolves conflicts over indices into the parents slice (SmallVec grouping,
  sorted agreement groups) instead of hash maps and Arc<Vec<Hash>> buckets
- ConflictZoneManager encodes free vs committed search in the type system (CommittedSearch /
  FreeSearch markers) instead of a runtime flag; committed search is scoped to the subgroup's
  next chain ancestor, and find_last_known_tips walks only that chain subtree
- Rank search and tie-breaking reworked for the sorted-group structure, with survivors carried
  between k steps (passing UMC cascade is monotone in k)
- Stores and services are passed by reference instead of clones
- Unused cascade scaffolding removed

Pipeline call sites are unchanged: DagknightData keeps its public shape


consensus/src/consensus/services.rs
```diff
@@ -140,16 +139,15 @@ impl ConsensusServices {
             storage.topology_ghostdag_store.clone(),
         );
 
-        // TODO[DK]: Use a config or ForkActivation to gate this
```
coderofstuff, line 143: Please keep this TODO. I'm addressing it in a different PR.

consensus/src/processes/dagknight/umc_voting.rs
```diff
@@ -25,8 +25,9 @@ pub trait ColoringReader {
 pub struct UmcVotingContext<'a> {
     /// The latest common chain ancestor of the zone (conflict genesis).
     pub conflict_genesis: Hash,
-    /// Tips of the subgroup being evaluated (virtual block conditioned on these).
-    pub subgroup: &'a [Hash],
+    /// A member of the subgroup being evaluated; the subgroup shares one next chain
+    /// ancestor above the conflict genesis, so a single member identifies it.
+    pub subgroup_member: &'a Hash,
```
coderofstuff: Since we're changing this anyway, I think we should use the opportunity to change this to `next_chain_ancestor` (that's the only thing the subgroup was used for).

consensus/src/processes/dagknight/tie_breaking.rs
coderofstuff, line 1: Please move back the TieBreaking logic, structs and tests here. Protocol.rs should only construct the DagknightTieBreaker then use it.

consensus/src/processes/dagknight/protocol.rs
```diff
@@ -137,100 +171,121 @@ impl<
...
-               - optimize with index
-        */
-
-        let start = parents[0];
-
-        if start == self.genesis_hash {
-            return self.genesis_hash;
-        }
+    /// Tie-breaking rule in case of multiple winning subgroups with the same rank value
+    fn tie_breaking(
```
coderofstuff: please move this down to align with where `fn tie_breaking` used to be.

consensus/src/processes/dagknight/protocol.rs
```diff
@@ -137,100 +171,121 @@ impl<
...
+            }
         }
-        assert_eq!(1, curr_subgroup.len(), "Expected dagknight to have only a single parent at the end");
+    }
 
-        conflict_ordered_parents.reverse();
+    /// Follows the Calculate-Rank algorithm in the DK paper: returns the best
+    /// agreement groups (all surviving at the winning k) with their metadata,
+    /// each subgroup being a contiguous slice into the sorted agreement grouping
+    fn rank<'a>(&self, conflict_genesis: Hash, parents: &[Hash], agreement_grouping: &'a [Group]) -> SmallVec<[GroupMetadata<'a>; 4]> {
```
coderofstuff: please move this down to align with where `fn rank` used to be

consensus/src/processes/dagknight/protocol.rs
```diff
@@ -407,229 +427,249 @@ impl<
...
-
-    /// BTree entry
-    #[derive(Eq, Clone)]
-    pub struct CascadeTreeEntry {
-        pub hash: Hash,
-        pub floor: i64,
-    }
+    /// Finds the latest common chain ancestor of the given subgroup (indices into `parents`),
+    /// serving as the genesis of the conflict level the caller is about to resolve
+    fn common_chain_ancestor(&self, parents: &[Hash], subgroup: &[u16]) -> Hash {
```
coderofstuff: please move this up to align with where `fn common_chain_ancestor` used to be

consensus/src/processes/dagknight/protocol.rs
```diff
@@ -407,229 +427,249 @@ impl<
...
-        btree: BTreeSet<CascadeTreeEntry>,
-        rev_index: BlockHashMap<i64>,
-
-        // Exact counters
-        past_blues: BlockHashMap<u64>,
-        past_reds: BlockHashMap<u64>,
-        anticone_blues: BlockHashMap<u64>,
+/// DAGKnight tie-breaker implementing Algorithm 4 of the paper. Holds only the
+/// stores needed for tie-breaking, not the full executor.
+struct DagknightTieBreaker<
```
coderofstuff: in line with the other comment to move back the TieBreaking logic to `tie_breaker.rs`, please make sure to remove these.

consensus/src/processes/dagknight/protocol.rs
```diff
@@ -99,18 +124,12 @@ pub struct DagknightExecutor<
     pub genesis_hash: Hash,
     pub dagknight_store: Arc<C>,
     pub headers_store: Arc<O>,
-    pub relations_store: Arc<RwLock<D>>,
+    pub relations_store: D,
```
coderofstuff, line 100: Is it just `D` and not `Arc<D>`?

consensus/src/processes/dagknight/manager.rs
```diff
@@ -86,51 +89,138 @@ fn cleanup_k_colouring_locks() {
...
+/// Conflict zone manager with the search mode (free vs committed) encoded as a type
+/// parameter. Committed searches are scoped per call by the next chain ancestor
+/// passed to the search methods, so no NCA is held as struct state.
 pub struct ConflictZoneManager<
     C: DagknightStore + DagknightStoreReader,
     O: HeaderStoreReader,
     D: RelationsStoreReader,
-    R: ReachabilityStoreReader + Clone,
+    S: ReachabilityService,
+    M: SearchMode = CommittedSearch,
```
coderofstuff: Please use:

```suggestion
    R: ReachabilityService,
    S: SearchMode = CommittedSearch,
```

consensus/src/processes/dagknight/protocol.rs
```diff
@@ -240,136 +295,101 @@ impl<
...
+            self.headers_store.as_ref(),
             relations_service,
-            reachability_service.clone(),
+            &self.reachability_service,
         );
 
         // Calculate the subgroup's next chain ancestor above conflict_genesis
-        let subgroup_nca = self.reachability_service.get_next_chain_ancestor(subgroup[0], conflict_genesis);
-        conflict_zone_manager.fill_zone_data(subgroup, Some(subgroup_nca));
+        let subgroup_nca = self.reachability_service.get_next_chain_ancestor(first_subgroup_member, conflict_genesis);
```
coderofstuff: select_parent_from_k_colouring is always called for a particular subgroup, which means we already also know the NCA for this subgroup. We can take this opportunity to update the params of select_parent_from_k_colouring to pass in the NCA instead of having to recalculating it here all the time.

coderofstuff (changes requested):
Initial review first. Will continue reviewing once the comments are addressed. I generally agree with a majority of these changes. I still need to completely review the code after the requested changes are fixed.

consensus/src/processes/dagknight/protocol.rs
biryukovmaxim: consensus uses MTRelationsService which is Arc<RwLock<RelationsStoreReader>>. so i dont see the motication to wrap it in arc rwlock again

consensus/src/processes/dagknight/tie_breaking.rs
```diff
@@ -143,34 +117,39 @@ impl<
...
+
+        let mut group_tips = group_tips.into_iter();
+        let first_tip = group_tips.next().expect("group must be non-empty");
+        let group_tips = once(first_tip).chain(group_tips);
+        let all_tips = all_tips.into_iter();
 
-        // Committed (non-free-search) manager
-        let conflict_zone_manager = ConflictZoneManager::with_free_search(
+        // Calculate the subgroup's next chain ancestor above conflict_genesis
+        let subgroup_nca = self.reachability_service.get_next_chain_ancestor(*first_tip, conflict_genesis);
```
coderofstuff, line 140: We can get rid of this call by adding `next_chain_ancestor` to `GroupMetadata` (should be easy to wire it through from `Group`'s `common_ancestor`).

In general, since the `Group` is providing a nice way to keep the NCA info, we should try to get rid of extra downstream calls to get nca.

consensus/src/processes/dagknight/tie_breaking.rs
```diff
@@ -33,83 +32,58 @@ pub struct ReferenceCluster {
...
-    /// All tips of the DAG (parents of the virtual block).
-    pub all_tips: &'a [Hash],
-    /// The tied subgroups — each subgroup is a slice of tip hashes that agree on a common chain ancestor above the conflict genesis.
-    pub subgroups: &'a [GroupMetadata],
-    /// The mutual rank `k` shared by all tied subgroups.
-    pub k: KType,
-}
-
-/// Trait for tie-breaking logic, isolating it from the DAGKnight executor for testability.
-pub trait TieBreaker {
```
coderofstuff, line 49: This trait exists because there was a `SimpleTieBreaker` before and I was comparing results. The tie breaking area is still in flux and when exploring alternatives, we might end up with another DagknightTieBreaker (DagknightTieBreakerV2) that implements this trait.

I think we should keep it to maintain the expected contract from the pov of protocol.rs.

In general, the patterns for protocol.rs is to provide a context and expect a trait that expects that trait (TieBreakingContext, TieBreaker, UmcVoter, UmcVoterContext)

consensus/src/processes/dagknight/tie_breaking.rs
```diff
@@ -33,83 +32,58 @@ pub struct ReferenceCluster {
     pub chain_blocks: Vec<Hash>,
 }
 
-/// Input data for a tie-breaking call.
-pub struct TieBreakContext<'a> {
```
coderofstuff, line 37: In line with https://github.com/kaspanet/rusty-kaspa/pull/1104/changes#r3906135373 please construct this context and pass it to the trait.

consensus/src/processes/dagknight/protocol.rs
```diff
@@ -137,100 +151,89 @@ impl<
...
+                                .get_next_chain_ancestor(parents[parent as usize], conflict_genesis),
+                            parent,
+                        })
+                        .collect();
+                    agreement_grouping.sort_unstable();
+
+                    if agreement_grouping.iter().map(|group| &group.common_ancestor).all_equal() {
+                        // There is exactly one group, we don't rank; the loop head re-derives the
+                        // conflict genesis from this same subgroup to skip to the next level
+                        agreement_grouping.into_iter().map(|group| group.parent).collect()
```
coderofstuff, line 187: In the old impl, if there is exactly 1 group, we would try to find the top-most common ancestor (not just the next one). Is that behavior retained here?

Previous code was:

```
                let next_conflict_genesis = self.common_chain_ancestor(&curr_subgroup);
                assert_ne!(
                    next_conflict_genesis, conflict_genesis,
                    "Expected the conflict genesis to change after skipping a level of the conflict hierarchy but got {}",
                    conflict_genesis
                );
                conflict_genesis = next_conflict_genesis;
```

consensus/src/processes/dagknight/protocol.rs
```diff
@@ -119,7 +117,23 @@ impl<
...
+                .reverse_conflict_ordered_parents
+                .iter()
+                .rev()
+                .map(|&parent| parents[parent as usize])
+                .collect(),
+        }
+    }
+
+    // TODO[DK]: accept slice with unique values so we don't need to verify it again
+    fn dagknight_indices(&self, parents: &[Hash]) -> DagknightDataIndices {
```
coderofstuff, line 136: Could you elaborate why `dagknight_indices` (and also `DagknightDataIndices` is necessary?

consensus/src/processes/dagknight/protocol.rs
```diff
@@ -240,136 +243,160 @@ impl<
...
-        let evaluate = |k: KType| -> Option<()> {
-            let (filtered_groups_kv, best_groups): (HashMap<_, _>, Vec<GroupMetadata>) = group_map
-                .get_mut()
+    /// Follows the Calculate-Rank algorithm in the DK paper: returns the best
+    /// agreement groups (all surviving at the winning k) with their metadata,
+    /// each subgroup being a contiguous slice into the sorted agreement grouping
+    fn rank<'a>(&self, conflict_genesis: Hash, parents: &[Hash], agreement_grouping: &'a [Group]) -> SmallVec<[GroupMetadata<'a>; 4]> {
+        // Groups failing a k evaluation are dropped from further consideration
+        // (passing the UMC cascade is monotone in k)
+        let mut survivors: SmallVec<[&'a [Group]; 10]> =
```
coderofstuff, line 320: Thanks for cleaning this area up. The pattern here before was a mess. This is much easier to reason about.

consensus/src/processes/dagknight/protocol.rs
```diff
@@ -240,136 +243,160 @@ impl<
...
-                Some(())
+                survivors = next_survivors;
+                Some(best_groups)
             }
         };
 
-        let _search_result = RankSearcher::search(evaluate);
-        // let (best_k) = search_result.map(|r| (r.k, r.result)).unwrap();
-        best_groups_cell.take()
+        RankSearcher::search(evaluate).map(|result| result.result).unwrap_or_default()
```
coderofstuff, line 345: I think we can just `.unwrap()` or `.expect()` here. We don't expect a scenario where we can't find a `k` that satisfies UMC (because at most the entire conflict zone should be a K(k) cluster).

consensus/src/processes/dagknight/protocol.rs
```diff
@@ -137,100 +151,89 @@ impl<
...
+                    let mut agreement_grouping: SmallVec<[Group; 10]> = curr_subgroup
+                        .iter()
+                        .map(|&parent| Group {
+                            common_ancestor: self
+                                .reachability_service
+                                .get_next_chain_ancestor(parents[parent as usize], conflict_genesis),
+                            parent,
+                        })
+                        .collect();
+                    agreement_grouping.sort_unstable();
```
coderofstuff, line 182: Why do we need to sort groupings here?

coderofstuff (changes requested):
Second pass at review

consensus/src/processes/dagknight/protocol.rs
biryukovmaxim: Instead of hashmap[can]-> vec[parent]
We have after sorting 
[{ca1,idx1},{ca1, idx2},{ca2,idx3}]

So by using chunk_by we get the whole subgroup which just a slice of the vector of pairs(in code it's a struct named Group)
We also can use binary search for specific ancestor subgroup if we need to.

This sorting is basically grouping subgroups by common ancestor

consensus/src/processes/dagknight/protocol.rs
biryukovmaxim: I don't like to allocate new array for all hashes multiple times, where we basically can use slices or allocate indices for them, which basically leads to 1 or zero allocations. But I also don't want to change public interface right now, so I kept public input and output the same

consensus/src/processes/dagknight/tie_breaking.rs
biryukovmaxim: Fixed in https://github.com/kaspanet/rusty-kaspa/pull/1104/commits/a5888dab4554db1b835454599e51f6fc0e1e76c3

consensus/src/processes/dagknight/tie_breaking.rs
biryukovmaxim: Fixed in https://github.com/kaspanet/rusty-kaspa/pull/1104/commits/a5888dab4554db1b835454599e51f6fc0e1e76c3

consensus/src/processes/dagknight/tie_breaking.rs
biryukovmaxim: Fixed in https://github.com/kaspanet/rusty-kaspa/pull/1104/commits/a5888dab4554db1b835454599e51f6fc0e1e76c3

consensus/src/processes/dagknight/protocol.rs
biryukovmaxim: Fixed in https://github.com/kaspanet/rusty-kaspa/pull/1104/commits/a5888dab4554db1b835454599e51f6fc0e1e76c3

consensus/src/processes/dagknight/protocol.rs
biryukovmaxim: single group branch returns the subgroup unchanged and on next iteration we call `common_chain_ancestor`

consensus/src/processes/dagknight/protocol.rs
```diff
@@ -385,8 +410,8 @@ impl<
                     cascade_result.virtual_score,
                     baseline_result.accepted,
                     cascade_result.accepted,
-                    cascade_result.flips,
-                    cascade_result.voting_blocks
+                    baseline_result.flips,
+                    baseline_result.voting_blocks
```
coderofstuff, line 414: These should remain `cascade_result`. Cascade is what has flips (baseline has flips = 0 all the time) and the one we care to see the voting blocks for.

consensus/src/processes/dagknight/protocol.rs
```diff
@@ -240,136 +243,158 @@ impl<
...
+        subgroups: &[GroupMetadata],
+    ) -> usize {
         debug!("Winning groups had rank k = {}", subgroups[0].k);
-        let mutual_k = subgroups[0].k;
-
-        let winning_index = DagknightTieBreaker::new(
-            self.dagknight_store.clone(),
-            self.headers_store.clone(),
+        let k = subgroups[0].k;
+        let all_tips: SmallVec<[Hash; 20]> = agreement_grouping.iter().map(|group| parents[group.parent as usize]).collect();
```
coderofstuff, line 297: This needs to be the full tip set since the free search coloring must be done in the context of all tips in the current hierarchy (not just the tips of surviving groups). Please align to previous impl.

coderofstuff (changes requested):
Parts I agree with: 
- CZM and manager.rs - I agree with the changes here and this is a much better structure with the way you have it
- protocol.rs
  - rank - updated structure looks way better than it was before
- template types

Parts I'm on the fence about:
- protocol.rs
  - Usage of indices - seems like premature optimization. I haven't seen this become an issue. Did you take any measurements that drove you to do this work?
  - The original structure for `fn dagknight` was written in a way to align with how it looks like in the paper for Order-DAG. I wonder if the current structure is easy enough to reason about for other where it aligns with Algorithm 1
  - I'm worried about the sort potentially masking protocol inaccuracies (which would have to be fixed). In general the ordering of the parents should never affect the SP result. If we perform any sorting inside the algo though (like agreement_grouping.sort_unstable()) that might end up masking an issue that we have to fix.

protocol.rs is the only one I'm thinking about now mostly as the rest lgtm

consensus/src/processes/dagknight/protocol.rs
biryukovmaxim: agreement_grouping contains all competing groups of the conflict level. so the code is same

STP-KAS:
Outsider-safe property test for the `sort_unstable()` on agreement groups: `dagknight(parents)` vs shuffled parents must return the same selected parent. Opened as https://github.com/kaspanet/rusty-kaspa/pull/1132 against `dagknight` (not master). Does not restyle the executor. Please run it on this rewrite before merge.

consensus/src/model/stores/dagknight.rs
```diff
@@ -21,6 +21,34 @@ pub trait DagknightStoreReader {
     fn has(&self, dk_key: DagknightKey) -> Result<bool, StoreError>;
 }
 
+impl<T: DagknightStoreReader + ?Sized> DagknightStoreReader for &T {
```
D-Stacks, line 24: Some of these stores seem more verbose then they need to be can we not blanket the `impl` for both `Arc` and `&`?

something like:

```rs

// blanket impl for any smart pointer / reference wrapping a [`DagknightStoreReader`]
impl<P, T> DagknightStoreReader for P
where
    P: Deref<Target = T>,
    T: DagknightStoreReader + ?Sized, {
    fn get_selected_parent(&self, dk_key: DagknightKey) -> Result<Hash, StoreError> {
        self.deref().get_selected_parent(dk_key)
    }

    fn get_data(&self, dk_key: DagknightKey) -> Result<Arc<GhostdagData>, StoreError> {
        self.deref().get_data(dk_key)
    }

    fn has(&self, dk_key: DagknightKey) -> Result<bool, StoreError> {
        self.deref().has(dk_key)
    }
}
```
I guess it is the same for some other stores.

biryukovmaxim: ye, seems fine to me

STP-KAS: Grok Build heavy (Windows desk), running this on auto. Independent desk check, not an audit, not Kaspa-core.

Deref blanket is the wrong merge of these two impls.

1. It cannot sit next to the existing `&T` / `Arc<T>` impls. Both already `Deref`, so the blanket is E0119 overlap. It has to *replace* them.

2. House style in this crate is the narrow `&T` impl, not Deref. `RelationsStoreReader` is only:

```rust
impl<T: RelationsStoreReader + ?Sized> RelationsStoreReader for &T
```

(`consensus/src/model/stores/relations.rs`). MT wrappers (`MTRelationsService`) exist specifically so we do **not** trait-impl through `RwLockReadGuard` / `Ref` / `MutexGuard`. A `P: Deref<Target = T>` blanket would do exactly that.

3. This PR's actual need is passing refs instead of clones (`&T`). The executor still stores `Arc<C>` with `C: DagknightStoreReader` (`protocol.rs` ~L90–98). Method calls on that Arc autoderef to `C`; they do not need `Arc<T>: DagknightStoreReader`. Same for `DagknightStore`.

Ask: is there a call site that takes `S: DagknightStoreReader` *by value* as `Arc<C>` (not `&*arc`)? If yes, keep the explicit `Arc<T>` impl. If no, delete the Arc impl and keep `&T` only. Either way, skip the Deref blanket. Happy to re-check if you point at that call site.

Not a merge blocker. Remaining fence is still protocol.rs (`sort_unstable` on agreement groups vs paper Alg. 1; parent-order). Shuffle property is #1132, still open, still the thing to run on this rewrite before merge.

---
**Operator (human, [@StppStp](https://x.com/StppStp) / [STP-KAS](https://github.com/STP-KAS)):** Intentions are good; thought process is questionable. STP remains delusional. Si vis pacem, para bellum. Intern at https://sixpack.wtf/

consensus/src/processes/dagknight/protocol.rs
D-Stacks, line 182: I don't think we should use sort_unstable here, although deterministic, we are making ordering dependent on rust internals, which I do not think is good practice for potentially consensus critical code.  I think we should use something like tie-breaking sort positioning by block hashes?

STP-KAS: Grok Build heavy (Windows desk), running this on auto. Independent desk check, not an audit, not Kaspa-core.

Agree the order here is consensus-visible. Disagree that `sort_unstable` vs `sort` is the bug, and disagree that this should become the DK tie-breaker.

`Group` is (`consensus/src/processes/dagknight/mod.rs`):

```rust
#[derive(Debug, PartialEq, Eq, PartialOrd, Ord)]
pub struct Group {
    pub common_ancestor: Hash,
    pub parent: u16,
}
```

Derived `Ord` is `(common_ancestor, parent index)`. After `unique_by` on parent hashes, those keys are unique. For a unique total order, `sort_unstable` and `sort` produce the same permutation. Switching to `sort()` does not change selected parent. The pdqsort equal-key / rust-internals worry does not fire on this type.

The actual leak is the second key: **slice index**, not block hash. Shuffle `parents` and the same block gets a different `u16`, so within-group order changes. That is caller-layout, not protocol. Paper Alg. 2 does not sort. @coderofstuff already flagged this: a grouping sort can mask a parent-order bug.

If the sort stays (it is how `chunk_by` gets contiguous agreement groups), specify the comparator as hashes, not indexes:

```rust
agreement_grouping.sort_by_key(|g| (g.common_ancestor, parents[g.parent as usize]));
```

That is the hash tie-break you want, and it is still only a grouping key. Do not fold it into `tie_breaking()`. Rank + tie-break is a later protocol step (winner among groups), not how we cluster by next-chain-ancestor.

Selected parent must still be invariant under permutation of the parent slice. That is [#1132](https://github.com/kaspanet/rusty-kaspa/pull/1132). Existing `test_parent_ordering_stability` only does one `sort()` of a fixture, not shuffles. Please run the shuffle property on this rewrite before merge:

```
cargo test -p kaspa-consensus dagknight_selected_parent_invariant_under_parent_permutation -- --nocapture
```

Happy to re-check if you point at a downstream use of within-group `Group.parent` order that is supposed to be protocol (rank / UMC / conflict_ordered_parents).

---
**Operator (human, [@StppStp](https://x.com/StppStp) / [STP-KAS](https://github.com/STP-KAS)):** Intentions are good; thought process is questionable. STP remains delusional. Si vis pacem, para bellum. Intern at https://sixpack.wtf/

consensus/src/processes/dagknight/protocol.rs
```diff
@@ -137,100 +151,89 @@ impl<
...
-                if conflict_genesis != winning_conflict_genesis {
-                    conflict_ordered_parents.extend(subgroup.as_ref().iter().copied());
+                curr_subgroup => {
+                    // g = find the LCCA of the current subgroup -- the genesis of this conflict level
+                    let conflict_genesis = self.common_chain_ancestor(parents, curr_subgroup);
+                    debug!("conflict_genesis: {:#}", conflict_genesis);
+
+                    // Split the subgroup by the chain ancestor each parent follows above `conflict_genesis`;
+                    // parents within each group agree about the conflict zone induced by `conflict_genesis`
+                    let mut agreement_grouping: SmallVec<[Group; 10]> = curr_subgroup
```
D-Stacks, line 173: `SmallVec<[Group; 10]>`

why do we choose 10?

STP-KAS: Grok Build heavy (Windows desk), running this on auto. Independent desk check, not an audit, not Kaspa-core.

10 is not a protocol constant, and it is not `max_block_parents`.

This `SmallVec` is one `Group` per parent in `curr_subgroup` (then sorted so `chunk_by` can cluster). Length is `|parents|` after `unique_by`, not an estimated number of competing groups. Same function already uses 20 for that cardinality:

```rust
SmallVec<[u16; 20]>  // curr_subgroup, conflict_ordered_parents, all_tips
SmallVec<[Group; 10]> // agreement_grouping  <-- same length as curr_subgroup
SmallVec<[GroupMetadata<'a>; 4]> // rank survivors
```

Live parent cap is `Bps::max_block_parents()` = `clamp(k/2, 10, 16)` (`consensus/core/src/config/bps.rs`). 10 is the *floor* for 1 BPS. 10 BPS is k=124 → 16. So 10 spills on a legal 10-BPS block; 20 does not.

Spill is correct (not a consensus limit). The number is only a stack hint. If you keep a literal, use 20 with the sibling parent vecs, or `SmallVec::with_capacity(curr_subgroup.len())`. Do not treat 10 as Crescendo BPS.

Not a merge blocker. Remaining fence is still the grouping sort / parent-order invariant on L182.

---
**Operator (human, [@StppStp](https://x.com/StppStp) / [STP-KAS](https://github.com/STP-KAS)):** Intentions are good; thought process is questionable. STP remains delusional. Si vis pacem, para bellum. Intern at https://sixpack.wtf/

consensus/src/processes/dagknight/protocol.rs
```diff
@@ -137,100 +151,89 @@ impl<
...
-            // Shortcut condition to avoid doing unnecessary work
-            if agreement_grouping.len() == 1 {
-                // There is exactly one group, we don't rank anymore.
-                let (_, subgroup) = agreement_grouping.iter().next().unwrap();
-                curr_subgroup = subgroup.clone();
-                // Deduplication via `.unique()` may have reduced curr_subgroup to a single
-                // element (e.g., [T1, T1] -> [T1]). In that case we're done.
-                if curr_subgroup.len() <= 1 {
-                    break;
+        assert!(parents.len() <= u16::MAX as usize);
```
D-Stacks, line 154: not sure, but I think this can cause the network to crash if we actually end up with over u16:MAX tips.  Just something worth considering, especially when we start thinking about even higher bps settings. We really should verify u16::MAX is enough imo. or we need to pre-filter tips to ensure we get down to u16:MAX in ssav2. u16:MAX seems like plenty, but doing some rough maths, a 49% miner currently only needs about 4 hours of mining and withholding of tips and publishing them all at once to potentially trigger this assert.

STP-KAS: Grok Build heavy (Windows desk), running this on auto. Independent desk check, not an audit, not Kaspa-core.

The crash path is SSAV2, not block parents. Split the call sites.

Header processing already rejects `direct_parents().len() > max_block_parents` (`pre_ghostdag_validation.rs`, `TooManyParents`). 10 BPS that cap is 16 (`Bps::max_block_parents` = `clamp(k/2, 10, 16)`). `dagknight(&ctx.known_direct_parents)` and `dagknight(&virtual_parents)` never see `u16::MAX` on a valid block. Those paths cannot halt the network with this assert.

`sink_search_algorithm_v2` is different (`virtual_processor/processor.rs`):

```rust
let DagknightData { selected_parent: inner_candidate, conflict_ordered_parents } =
    self.dagknight_executor.dagknight(&sub_tips.iter().copied().collect_vec());
```

`sub_tips` starts as the full tip antichain, not `max_block_parents`. That is the slice this `u16` indexes. `assert!` is on in release. Hit it and the virtual thread panics.

Rough math at 10 BPS: 0.49 × 10 × 3600 × 3.7h ≈ 65536 withheld unmerged blocks, then dump. That is an antichain (not a chain). Honest merge only sees them after publish. Higher BPS shortens the clock; it does not need `u16::MAX` to be “not plenty.”

`u16` is a compactness choice for *one block’s parents*. SSAV2 is a different set. Sharing the width is the smell.

Do not keep a production `assert!` here. Dropping it without a check is worse: `0..parents.len() as u16` wraps. Options:
1. Pre-filter SSAV2 tips before DK (your ssav2 note; same spirit as `max_virtual_parent_candidates = max_block_parents * 3`).
2. Widen SSAV2 indices to `u32` and still bound work — DK on 65k tips is a CPU DoS even if it does not panic.
3. Header/virtual-parents can stay `u16` (or just `u8`; 16 fits).

Not a merge blocker for the executor rewrite vs paper Alg. 1, but it is a halt-the-node assert on a reachable virtual path. Please do not ship `assert!(len <= u16::MAX)` on SSAV2 input.

Happy to re-check if you point at a tip-set bound already applied before that `dagknight(&sub_tips)` call. I do not see one in this PR.

---
**Operator (human, [@StppStp](https://x.com/StppStp) / [STP-KAS](https://github.com/STP-KAS)):** Intentions are good; thought process is questionable. STP remains delusional. Si vis pacem, para bellum. Intern at https://sixpack.wtf/

consensus/src/processes/dagknight/protocol.rs
```diff
@@ -137,100 +151,89 @@ impl<
...
+        // Duplicate parents always land in one group and can never be split apart
+        let mut curr_subgroup: SmallVec<[u16; 20]> = (0..parents.len() as u16).unique_by(|&parent| parents[parent as usize]).collect();
```
D-Stacks, line 156: why 20?

STP-KAS: Grok Build heavy (Windows desk), running this on auto. Independent desk check, not an audit, not Kaspa-core.

Same answer as [`SmallVec<[Group; 10]>`](https://github.com/kaspanet/rusty-kaspa/pull/1104#discussion_r4054450193): 20 is not a protocol constant.

`curr_subgroup` is one `u16` per unique parent in this call. Header / virtual-parents are capped at `max_block_parents` = 16 at 10 BPS (`clamp(k/2, 10, 16)`). 20 is “16 plus some slack.” It is not `k`, not BPS, not `max_virtual_parent_candidates` (16×3 = 48).

SSAV2 calls this with the full tip antichain ([`dagknight(&sub_tips)`](https://github.com/kaspanet/rusty-kaspa/pull/1104#discussion_r4054484407)). 20 always spills there. The inline array is only a stack hint; spill is correct.

If this function is for block parents, use 16 (or a named `MAX_BLOCK_PARENTS` / `params.max_block_parents()`). If it also serves SSAV2, `SmallVec::with_capacity(parents.len())` and stop picking 10 vs 20 vs 4 in the same file.

---
**Operator (human, [@StppStp](https://x.com/StppStp) / [STP-KAS](https://github.com/STP-KAS)):** Intentions are good; thought process is questionable. STP remains delusional. Si vis pacem, para bellum. Intern at https://sixpack.wtf/

consensus/src/processes/dagknight/protocol.rs
D-Stacks: I don't think public facing should deal with indices, public facing side is working with actual tip and parent hashes I think the dagknight call should keep aligning with this for the caller, if just for ease of readability.

STP-KAS: Grok Build heavy (Windows desk), running this on auto. Independent desk check, not an audit, not Kaspa-core.

Agree. Drop the TODO on L121. Do not make indices public.

Current `pub fn dagknight(&self, parents: &[Hash]) -> DagknightData` already returns hashes. `dagknight_indices` / `DagknightDataIndices` are private. That is the right split: internal compactness, public hashes. @biryukovmaxim already said the public input/output stayed hashes so callers did not change.

Callers are hash-native:
- header coloring: `dagknight(&ctx.known_direct_parents)` → `incremental_coloring(..., dk_sp: Hash)`
- SSAV2: `pick_virtual_parents_v2(inner_candidate, conflict_ordered_parents)` then set-compares those hashes against the next `sub_tips`. Indices are meaningless once the tip slice is replaced.

Mapping `u16` → `Hash` at the public boundary is one collect. SSAV2 would do that collect anyway. Leaking `u16` into `DagknightData` saves nothing for those call sites and couples them to “index into the slice I just passed,” which is how the parent-order / `u16::MAX` issues leak out of the executor.

Keep `DagknightData` as hashes. Keep indices crate-private. If a later call site wants zero-copy, add a private method — do not change this signature.

---
**Operator (human, [@StppStp](https://x.com/StppStp) / [STP-KAS](https://github.com/STP-KAS)):** Intentions are good; thought process is questionable. STP remains delusional. Si vis pacem, para bellum. Intern at https://sixpack.wtf/

consensus/src/processes/dagknight/protocol.rs
```diff
@@ -137,100 +151,89 @@ impl<
...
+        let mut conflict_ordered_parents: SmallVec<[u16; 20]> = SmallVec::with_capacity(parents.len());
+        loop {
+            curr_subgroup = match core::mem::take(&mut curr_subgroup).deref() {
```
D-Stacks, line 159: This might just be me, but I found this loop very hard to follow. especially in regards to find out how the curr_subgroup is actually re-assigned, via the match, mem::take, pattern matching and lastly the if/else statement. Can't we just check for `len == 1` and reassign current_subgroup at the end of the loop, without so much nesting and the core call? something like:

```rs

loop {
    if curr_subgroup.len() == 1 {
        let sp = curr_subgroup[0];
        debug!("dk::sp: {} | reverse_conflict_ordered_parents: {:?}", sp, conflict_ordered_parents);
        return DagknightDataIndices {
            selected_parent: sp,
            reverse_conflict_ordered_parents: conflict_ordered_parents,
        };
    }

    // g = find the LCCA of the current subgroup -- the genesis of this conflict level
    let conflict_genesis = self.common_chain_ancestor(parents, &curr_subgroup);
    debug!("conflict_genesis: {:#}", conflict_genesis);

    // Split the subgroup by the chain ancestor each parent follows above `conflict_genesis`;
    // parents within each group agree about the conflict zone induced by `conflict_genesis`
    let mut agreement_grouping: SmallVec<[Group; 10]> = curr_subgroup
        .iter()
        .map(|&parent| Group {
            common_ancestor: self
                .reachability_service
                .get_next_chain_ancestor(parents[parent as usize], conflict_genesis),
            parent,
        })
        .collect();
    agreement_grouping.sort_unstable();

    // Determine the next subgroup
    curr_subgroup = if agreement_grouping.iter().map(|group| &group.common_ancestor).all_equal() {
        // There is exactly one group, we don't rank; the loop head re-derives the
        // top-most common chain ancestor of this same subgroup as the next conflict genesis
        agreement_grouping.into_iter().map(|group| group.parent).collect()
    } else {
        // Pick a "winner" among these subgroups
        let best_groups = self.rank(conflict_genesis, parents, &agreement_grouping);

        // Multiple best groups (same winning k) are resolved by a tie-breaking rule
        let winner: &GroupMetadata = if best_groups.len() > 1 {
            &best_groups[self.tie_breaking(conflict_genesis, parents, &agreement_grouping, &best_groups)]
        } else {
            best_groups.first().expect("rank should return at least one best group")
        };

        let winning_ancestor = winner.subgroup[0].common_ancestor;
        agreement_grouping
            .iter()
            .filter(|group| group.common_ancestor != winning_ancestor)
            .for_each(|group| conflict_ordered_parents.push(group.parent));

        winner.subgroup.iter().map(|group| group.parent).collect()
    };
}
```

as mentioned it might just be personal preference, or maybe there is something I am missing, but I find this to be much more readable rust.

STP-KAS: Grok Build heavy (Windows desk), running this on auto. Independent desk check, not an audit, not Kaspa-core.

Agree. `mem::take` + `match .deref()` is an ownership trick, not protocol. The inner `curr_subgroup =>` also **shadows** the outer `SmallVec` with a `&[u16]`. That is why the reassignment is hard to see.

Paper Alg. 1 / the pre-PR executor is `while |P| > 1`. @coderofstuff already asked to keep that shape. Your sketch is that loop with a `len == 1` return at the head. Same for `|P| >= 1`.

Prefer either:

```rust
while curr_subgroup.len() > 1 {
    let conflict_genesis = self.common_chain_ancestor(parents, &curr_subgroup);
    // ...
    curr_subgroup = /* next */;
}
let sp = curr_subgroup[0];
```

or the `if len == 1 { return }` form you wrote. Do not keep `mem::take` just to match `[sp]`.

Empty slice panics in `common_chain_ancestor` (`subgroup[0]`) in both versions. Header already rejects `NoParents`. Not introduced here.

Not a consensus change. This is the readability / paper-alignment fix. Remaining protocol fence is still the grouping `sort_unstable` (parent-order / #1132), not this loop sugar.

---
**Operator (human, [@StppStp](https://x.com/StppStp) / [STP-KAS](https://github.com/STP-KAS)):** Intentions are good; thought process is questionable. STP remains delusional. Si vis pacem, para bellum. Intern at https://sixpack.wtf/

consensus/src/processes/dagknight/protocol.rs
D-Stacks: From this exchange and reading the variable "all_tips" I feel like "all_tips" is not an accurate term here? when I read "all_tips" I assumed it to literally be all tips put into the original dagknight call. Furthermore I generally feel like the term tip here is probably conflicting with my idea of tips that I have from other parts of the codebase, I kind of read it as a global network variable, but in dagknight it is scoped, and doesn't have to have any relation with the term tip we use elsewhere. I think I would rather we rename this, possibly to conflict_tip, conflict_sink, or maybe something completely new, such as conflict_frontiers? this would be similar in fashion like we named the genesis of the conflict, not simply "genesis" but "conflict_genesis".

STP-KAS: Grok Build heavy (Windows desk), running this on auto. Independent desk check, not an audit, not Kaspa-core.

Agree. `all_tips` is the wrong word here.

This slice is `agreement_grouping` mapped through `parents[i]`: the **current conflict-level** competing parents (Alg. 1’s current `P` after unique). It is not:
- the original `dagknight(parents)` argument (SSAV2 can pass the full DAG-tip antichain there),
- the global tips store,
- only the rank survivors (`subgroups` / `group_tips` are that).

`TieBreakingContext` already documents it as “competing tips of the current conflict level” and still names the field `all_tips`. The `all_` is what made it look like the original call’s input. The `tips` is what collides with `stores/tips` and SSAV2 `sub_tips`.

@coderofstuff’s point still holds under a new name: free-search coloring is in the context of this whole current-level `P`, not only surviving groups. That is `agreement_grouping`, not `subgroups`.

Rename the local and `TieBreakingContext.all_tips` together. Prefer `conflict_frontier` (one antichain, parallel to `conflict_genesis`). Avoid `sink` (SSAV2 already owns that). `conflict_parents` is also fine if you want less poetry.

Not a consensus change. Remaining fence is still the grouping sort / parent-order, not this identifier.

---
**Operator (human, [@StppStp](https://x.com/StppStp) / [STP-KAS](https://github.com/STP-KAS)):** Intentions are good; thought process is questionable. STP remains delusional. Si vis pacem, para bellum. Intern at https://sixpack.wtf/

consensus/src/processes/dagknight/protocol.rs
```diff
@@ -240,136 +243,158 @@ impl<
...
+    ) -> usize {
         debug!("Winning groups had rank k = {}", subgroups[0].k);
-        let mutual_k = subgroups[0].k;
-
-        let winning_index = DagknightTieBreaker::new(
-            self.dagknight_store.clone(),
-            self.headers_store.clone(),
+        let k = subgroups[0].k;
+        let all_tips: SmallVec<[Hash; 20]> = agreement_grouping.iter().map(|group| parents[group.parent as usize]).collect();
+        let ctx = TieBreakingContext { conflict_genesis, all_tips: &all_tips, parents, subgroups, k };
```
D-Stacks, line 298: It seems weird to me that we pack this into a ctx, only to unpack it immediately after we call tie_break in tie_breaking.rs. I think we should either stay with a ctx, or just pass them as arguments, in current implementation this ctx just seems unnecessary, in the sense that we only use it to unpack it again..

STP-KAS: Grok Build heavy (Windows desk), running this on auto. Independent desk check, not an audit, not Kaspa-core.

Keep the ctx. Do not flatten the trait to five arguments.

@coderofstuff already required this shape: `protocol.rs` builds a context and a `TieBreaker` consumes it — same pair as `UmcVotingContext` / `UmcVoter`. Tests construct `TieBreakingContext` directly (`ctx_forward` / `ctx_reversed`) without going through the executor. That is the reason the struct exists, including a future `DagknightTieBreakerV2`.

The smell is only this hop in the impl:

```rust
fn tie_break(&self, ctx: &TieBreakingContext<'_>) -> usize {
    let TieBreakingContext { conflict_genesis, all_tips, parents, subgroups, k } = *ctx;
    self.tie_break_alg4(conflict_genesis, all_tips, parents, subgroups, k)
}
```

Pack in `protocol.rs`, unpack one function later, then five args again. That is unnecessary, not the ctx.

Stay with the ctx: `tie_break_alg4` should take `&TieBreakingContext` (or `tie_break` should just use `ctx.conflict_genesis` / `ctx.subgroups` and drop `tie_break_alg4`). One object through the impl.

`protocol.rs` packing is the right adapter: it turns `agreement_grouping` into the current-level parent hashes. That mapping belongs at the executor boundary, not inside Alg. 4.

Not a consensus change.

---
**Operator (human, [@StppStp](https://x.com/StppStp) / [STP-KAS](https://github.com/STP-KAS)):** Intentions are good; thought process is questionable. STP remains delusional. Si vis pacem, para bellum. Intern at https://sixpack.wtf/

consensus/src/processes/dagknight/protocol.rs
```diff
@@ -409,227 +434,11 @@ impl<
...
-    }
-
-    pub fn genesis(&self) -> u64 {
-        self.genesis
-    }
+struct DagknightDataIndices {
+    selected_parent: u16,
+    /// Indices into `tips`, ordered by conflict hierarchy with bottom-most conflicts first;
+    /// reverse-iterate to get parents from latest/topmost conflicts first (as `dagknight` returns)
+    reverse_conflict_ordered_parents: SmallVec<[u16; 20]>,
```
D-Stacks, line 441: maybe we can rename this to `descending_conflict_ordered_parents` or `dec_conflict_ordered_parents`? I also think we can then maybe get rid of the `ordered` part then as well, as descending implies order. This signals to me that it is bottom-up more then reverse_ordered on just a quick glance. Also thinking we can maybe maybe apply a similar renaming to `conflict_ordered_parents` with ascending / asc elsewhere. Currently conflict ordered doesn't mean much to me without significant context / comment reading, which granted, is well documented, but I still think the naming of conflict_ordered_parents could be more precise.

STP-KAS: Grok Build heavy (Windows desk), running this on auto. Independent desk check, not an audit, not Kaspa-core.

Agree `reverse_` and `conflict_ordered` are opaque. Disagree with `descending_` / `dec_` on **this** field.

Push order is bottom-most / deepest conflict first (first `LCCA` of the full set, losers pushed, then shallower levels). Public `DagknightData.conflict_ordered_parents` is that vec **reversed**: latest / topmost first (`iter().rev()` at L127–129). `reverse_` only means “not yet reversed for the caller.” It is not a protocol order.

`descending` without an axis is the same trap. If descending means from the tips, that is the **public** field, not this one. Putting `descending_` here would name the internal vec as the public order. `dec_` is worse than `reverse_`.

Drop `ordered` (and `reverse`). Name the axis. Rename the pair together so SSAV2 (`conflict_ordered_parents` → `pick_virtual_parents_v2`) stays aligned:

- internal (push order): `conflict_parents_deepest_first`
- public (after `.rev()`): `conflict_parents_latest_first`

`bottom_first` / `top_first` is the same split if you prefer DAG-direction words. The field comment also says “indices into `tips`”; it is indices into `parents`. Same `all_tips` collision.

Not a consensus change.

---
**Operator (human, [@StppStp](https://x.com/StppStp) / [STP-KAS](https://github.com/STP-KAS)):** Intentions are good; thought process is questionable. STP remains delusional. Si vis pacem, para bellum. Intern at https://sixpack.wtf/

consensus/src/processes/dagknight/manager.rs
```diff
@@ -86,51 +89,138 @@ fn cleanup_k_colouring_locks() {
...
-// be more in line with what the paper needs
-// Renamed from ghostdag_customized to k_colouring
+/// The search mode of a conflict zone colouring. Used as a type parameter of
+/// `ConflictZoneManager` so that free vs committed search differ in the manager
+/// type rather than in runtime options. The associated `Nca` type is the search
+/// bound the mode requires: free search carries none, committed search is scoped
+/// by the next chain ancestor passed per call.
+pub trait SearchMode {
+    /// Whether the search is unrestricted (free) or bounded to a subgroup region (committed)
+    const IS_FREE: bool;
```
D-Stacks, line 99: Maybe we can rename this to `const FREE_SEARCH_ACTIVATED: bool` (or IS_FREE_SEARCH_ACTIVE)? I find things like `!S::IS_FREE` really Hard to read further down the road without significant follow-ups and context on the associated items in the Conflict zone manger impl. This would help a little bit.

STP-KAS: Grok Build heavy (Windows desk), running this on auto. Independent desk check, not an audit, not Kaspa-core.

Agree `IS_FREE` is too short. `!S::IS_FREE` at L461 is “this is committed search” written as a double negative.

Don’t use `FREE_SEARCH_ACTIVATED` / `IS_FREE_SEARCH_ACTIVE`. This is a type parameter (`FreeSearch` vs `CommittedSearch`), not a runtime feature flag. “Activated” reads like a toggle you can flip on a live manager.

If you keep a bool, name it `IS_FREE_SEARCH` so it matches the paper and `DagknightKey.free_search`. For the committed-only assert, one polarity: `if S::IS_COMMITTED { ... }`, not `if !S::IS_FREE`. Don’t ship both consts (they will drift).

You can also drop some of these branches. `FreeSearch::is_agreeing_parent` is already `true`, so L578–582 (`if S::IS_FREE { all parents } else { filter agreeing }`) is the same as always filtering. The typed mode is the methods; the bool is mostly a store-key discriminator.

Not a consensus change.

---
**Operator (human, [@StppStp](https://x.com/StppStp) / [STP-KAS](https://github.com/STP-KAS)):** Intentions are good; thought process is questionable. STP remains delusional. Si vis pacem, para bellum. Intern at https://sixpack.wtf/

consensus/src/processes/dagknight/manager.rs
```diff
@@ -86,51 +89,138 @@ fn cleanup_k_colouring_locks() {
...
+/// parameter. Committed searches are scoped per call by the next chain ancestor
+/// passed to the search methods, so no NCA is held as struct state.
 pub struct ConflictZoneManager<
     C: DagknightStore + DagknightStoreReader,
     O: HeaderStoreReader,
     D: RelationsStoreReader,
-    R: ReachabilityStoreReader + Clone,
+    R: ReachabilityService,
+    S: SearchMode = CommittedSearch,
 > {
```
D-Stacks, line 178: So, I kind of want to address the elephant in the room here. I think we should name the type parameters in a meaningful, readable, and helpful way instead of trying to spell out C-O-D-E-R. Even if it is somewhat more verbose, I would prefer something like:

DKStore: DagknightStore + DagknightStoreReader,
HStoreR: HeaderStoreReader,
RelStoreR: RelationsStoreReader,
ReachStoreR: ReachabilityService,
SearchM: SearchMode = CommittedSearch,

Right now, C, O, D, E, R feels like it obfuscates code readability too much imho, and this is kind of across the board in the whole dagknight codebase.

STP-KAS: Grok Build heavy (Windows desk), running this on auto. Independent desk check, not an audit, not Kaspa-core.

Agree: stop spelling `CODER`. That is a joke, not a type system.

Don’t replace it with `DKStore` / `HStoreR` / `RelStoreR` / `ReachStoreR` / `SearchM`. That is a second abbreviation language. `HStoreR` still needs a legend, and `ReachStoreR` is wrong on this type: `R` here is `ReachabilityService`, not a store reader (this PR already changed that bound).

House style in this crate is `GhostdagManager<T, S, U, V>` — one letter, bound on the same line, no acronyms:

```rust
pub struct GhostdagManager<T: GhostdagStoreReader, S: RelationsStoreReader, U: ReachabilityService, V: HeaderStoreReader>
```

Match that:

```rust
pub struct ConflictZoneManager<
    T: DagknightStore + DagknightStoreReader,
    H: HeaderStoreReader,
    Rel: RelationsStoreReader,
    Rch: ReachabilityService,
    S: SearchMode = CommittedSearch,
>
```

Keep `S` for `SearchMode` (that one already means something). If a letter still collides, use the bound, not Hungarian (`StoreR`).

The actual `CODER` is `DagknightExecutor<C, O, D, E, R>` (`E` = UMC cascade store). Rename the pair in this PR so protocol.rs and manager.rs match. Don’t boil the rest of the DK crate in the same pass.

Not a consensus change.

---
**Operator (human, [@StppStp](https://x.com/StppStp) / [STP-KAS](https://github.com/STP-KAS)):** Intentions are good; thought process is questionable. STP remains delusional. Si vis pacem, para bellum. Intern at https://sixpack.wtf/

STP-KAS:
**Auto Grok** (Grok Build, Windows desk). Running on auto. Quality before speed. Independent desk check — not an audit, not Kaspa core, not a KIP.

**Desk vs this comment:** GitHub account is STP-KAS. The desk is delusional (Gramlane / project-delusional). The work here is Grok: trained on `pr-1104` head `a5888da`, current `dagknight` HEAD `ad45e24`, [#1132](https://github.com/kaspanet/rusty-kaspa/pull/1132), coderofstuff/dk-wiki (rank + UMC chapters still unreviewed), and the intern roundup. Unfinished business.

**Why / goal:** keep naming nits from crowding the merge gate. Remaining honest gate is parent-order invariance, not `IS_FREE`.

## Trained read of this rewrite

- Public `dagknight(&[Hash])` still returns hashes. Work is `u16` into the caller slice after `unique_by` (first occurrence).
- `Group { common_ancestor, parent: u16 }` derives `Ord`. `sort_unstable` + `chunk_by` is deterministic **given the indices**. The indices **are** input order. Shuffle the parent slice → different `u16`s → different within-group order. Rank walks that order.
- Current `dagknight` branch (not this PR) still HashMap-groups; loser order is HashMap iteration, with a TODO that it "should not matter." So this rewrite is stricter than live `dagknight` and still input-order-sensitive via indices.
- Rank = min `k` such that committed k-colouring of the subgroup VSP passes d-UMC with `g(k) = floor(sqrt(k))`. Practice uses **gray** (agreeing reds vote 0), not paper Alg. 3 representatives. That is a documented practical substitute (dk-wiki), not a line-for-line paper port.
- `SearchMode::IS_FREE` is a type-level const. Already replied on `r4056347613`. Not a runtime feature flag. No further naming from this desk unless a shuffle test fails.

**Please run [#1132](https://github.com/kaspanet/rusty-kaspa/pull/1132) on this head before merge.** Paper Alg. 1/2 does not sort parents. If SP is not invariant under permutation of the parent slice, that is the bug `sort_unstable` can mask.

Inline threads from 19–20 Sep are already answered. This desk will monitor replies. No more `CODER` / `IS_FREE` nits from here unless the author asks.

**Sources:** this PR `protocol.rs` / `mod.rs` `Group` / `rank_search.rs` / `manager.rs` `SearchMode`; [#1132](https://github.com/kaspanet/rusty-kaspa/pull/1132); [coderofstuff/dk-wiki](https://github.com/coderofstuff/dk-wiki); intern roundup https://x.com/kaspaunchained/status/2101676311244915028 ; catalog https://github.com/STP-KAS/kaspa-master-file/commit/ff29162

---

Standard disclaimer. This GitHub, not the topic above.

Intentions are good; thought process is questionable. STP remains delusional. Si vis pacem, para bellum.

Intern at https://sixpack.wtf/

PR #1105, opened 2026-08-31
Fix CI Lints, committed LKT ancestry, GHOSTDAG integration headers, goref replay, DK sink-search livelock, and test-template pruning samples

LiveLaughLove13:
# Fix CI Lints, committed LKT ancestry, GHOSTDAG integration headers, goref replay, DK sink-search livelock, and test-template pruning samples

## Why this is necessary

Six independent failures blocked `dagknight`:

1. **CI Lints** — `cargo clippy --workspace --tests --benches --examples -- -D warnings` (Rust 1.93.0) failed. `cargo fmt` was already clean.
2. **`test_find_last_known_tips_uses_correct_ancestry_type`** — committed LKT returned 3 tips instead of `[D, E]` because the NCA-scoping change dropped the conflict-genesis chain bound when NCA is `None`.
3. **`ghostdag_test`** — `UnexpectedHeaderBlueScore(11, 12)` because test headers still committed vanilla GHOSTDAG `blue_score`, while validation checks DK **coloring** `blue_score`.
4. **`goref_custom_pruning_depth_test`** — `UnexpectedDifficulty` (`505530428` vs `505533095`). Goref JSON fixtures are GHOSTDAG-recorded headers; always-on DK coloring can pick a different selected parent, so bits / DAA no longer match. The later `DbLifetime` / SIGABRT is destructor cleanup after that panic, not a second bug.
5. **`sink_search_algorithm_v2` livelock** — CI nextest sat at `SLOW` >10 minutes on `validate_body_in_context_test`, `basic_utxo_disqualified_test`, `double_search_disqualified_test`, and `mass_per_sig_op_does_not_change_block_capacity`. Same four hung locally (>300s). Not GitHub, not the browser: DK sink search retried a UTXO-invalid tip forever.
6. **Test-template sink search vs coloring SP** — after (5) unblocked the suite, nextest failed `selected_chain_test` (`KeyNotFound(PruningSamples/…15)`) and `sighash_type_commitment_test` (`StatusUTXOValid` vs `StatusUTXOPendingVerification`). The test block template helper still used GHOSTDAG v1 sink search while header pruning-point calc uses DK coloring.

No `#[allow(...)]`. Each change is the smallest equivalent rewrite, the missing bound, the documented DK gate, the missing GHOSTDAG pop, or aligning the test template helper with `resolve_virtual`. Selected-parent, k-colouring, UMC, and 100 BPS behavior are unchanged except for restoring the committed LKT chain filter that the ancestry test already documented. Live DAA / `blue_score` stay on coloring when DK is on.

---

## 1. Clippy (`3edd772d`)

CI treats every warn-by-default lint as an error. All five failures were in `kaspa-consensus`.

### Production

**`clippy::filter_next` — `consensus/src/processes/dagknight/manager.rs`**

The missing-coloring diagnostic still walks agreeing parents, keeps those with no coloring data, and takes the first that is a chain ancestor of any tip. Only the last step changed: `.filter(…).next()` → `.find(…)`. Same order, same predicate, same `Option`. This path only panics when coloring data is missing.

**`clippy::unnecessary_lazy_evaluations` — `consensus/src/processes/ghostdag/ordering.rs`**

`sort_blocks` still prefers `custom_topology_store` when present, otherwise `ghostdag_store` (DK coloring mergeset order).

`unwrap_or_else(|| &self.ghostdag_store)` → `unwrap_or(&self.ghostdag_store)` because the fallback is a field borrow, not work that needs to be deferred. Same store, same sort key.

### Tests only

**`clippy::type_complexity`** — JSON fixture rows are still `(id, parents, blue_work, bits, blue_score, daa_score, selected_parent)`. Named aliases only:

| Alias | Selected parent | Used by |
| --- | --- | --- |
| `ConflictLktTestBlock` | `Hash` | `test_czm_lkt_correctness` |
| `CapturedZoneTestBlock` | `Option<Hash>` | `test_czm_lkt_correctness_2` |

**`clippy::useless_vec`** — `fill_zone_data` already takes `&[Hash]`. `&vec![hash]` → `&[hash]`. Same one-element slice.

---

## 2. Committed LKT chain bound (`189d93ba`)

**File:** `consensus/src/processes/dagknight/manager.rs`  
**In:** `ConflictZoneManager::find_last_known_tips`

The NCA-scoping change replaced the committed check with `next_chain_ancestor.is_none_or(...)`. When the caller (and this unit test) passes `None`, `is_none_or` is always true, so the **conflict-genesis chain** filter never ran. Committed LKT then walked DAG parents like free search and accepted X (`left: 3, right: 2`).

Original committed rule (before NCA scoping): skip `curr` unless the conflict genesis is a chain ancestor of it. That is what the test is for: F→D,E and W→X,E, with X dropped because A is not a chain ancestor of X.

**Fix:** In committed search, bound the walk with NCA when it is present, otherwise with the conflict genesis:

```rust
is_chain_ancestor_of(next_chain_ancestor.unwrap_or(self.root), curr)
```

- `Some(nca)` — same NCA scoping as production zone fill
- `None` — restored genesis-chain filter (X skipped)
- Free search — still no skip (DAG walk → D, E, X)

Production `fill_zone_data` still passes `Some(nca)` in committed mode, so that path is unchanged.

---

## 3. Test header builder vs coloring validation (`1aaddf96`)

**File:** `consensus/src/consensus/test_consensus.rs`  
**In:** `TestConsensus::build_header_with_parents`

Header-processor validation already splits coloring and topology:

- `check_blue_score`, DAA, bits, past-median-time → **coloring** (DK selected parent + `incremental_coloring` when the executor is present)
- `check_blue_work` → **topology** (vanilla GHOSTDAG)

The test helper still set both `blue_score` and `blue_work` from vanilla GHOSTDAG. Insert then failed: header `blue_score = 12`, coloring `blue_score = 11`.

**Fix:** Build test headers the same way the processor validates them:

- `blue_score`, DAA, bits, PMT → coloring (DK selected parent when the executor is present)
- `blue_work` → topology GHOSTDAG

`ghostdag_test` still asserts on the **topology** store, so fixture selected parent / reds / blues / expected score are unchanged. No consensus algorithm change — only the test header builder.

---

## 4. Gate DAGKNIGHT for GHOSTDAG goref fixtures (`a64df8fd`)

**Files:**
- `consensus/core/src/config/mod.rs`
- `consensus/src/consensus/storage.rs`
- `consensus/src/consensus/services.rs`
- `testing/integration/src/consensus_integration_tests.rs`

`json_test` (`goref_custom_pruning_depth_test`, `goref_notx_*`, `goref_tx_small_*`) replays **GHOSTDAG-recorded** headers. DK was always constructed (`dagknight_store = Some(...)`), so header coloring used a DK selected parent. When that parent differs from GHOSTDAG, the DAA window and `bits` diverge from the fixture.

Rewriting live DAA / `blue_score` to topology would make goref pass by undoing DK header rules. The storage/services TODOs already asked for a config / `ForkActivation` gate.

**Fix:** Honor that gate. Do not change consensus DAA.

- `Config.enable_dagknight` defaults to **`true`** (production and DK tests stay on)
- `ConfigBuilder::disable_dagknight()`
- `ConsensusStorage` creates `dagknight_store` only when the flag is on; no store → no executor → coloring falls back to vanilla GHOSTDAG (`coloring_ghostdag_manager.ghostdag()`)
- `json_test` calls `disable_dagknight()` so goref fixtures validate as recorded

`ghostdag_test` and other DK paths keep the default. `TODO[DK]: replace with ForkActivation` stays on the field.

---

## 5. Drop rejected sink candidates (`be0a2f2a`)

**File:** `consensus/src/pipeline/virtual_processor/processor.rs`  
**In:** `VirtualStateProcessor::sink_search_algorithm_v2`

Header-only inserts finished. The first **body** insert hung `virtual_state_task`. Logs:

```
tips=1  prev_sink=genesis
candidate=block 3
calculate_utxo_state_relatively(genesis → block 3)
…repeat forever, same tip, same candidate
```

GHOSTDAG `sink_search_algorithm` **pops** a rejected candidate off the heap, then walks to its parents. v2 kept the rejected tip in `tip_set`. Genesis is a DAG ancestor of that tip, so

```rust
!is_dag_ancestor_of_any(parent, tip_set)
```

was false for genesis. Genesis was never added. The loop retried the same UTXO-invalid tip forever.

These tests insert bodies that fail UTXO (zero / wrong commitment). That is expected: `basic_utxo_disqualified_test` / `double_search_disqualified_test` exist to do that. With v1 they complete; with v2 they livelock. Same bug on a live node if the only body tip is UTXO-invalid.

**Fix:** Same pop as GHOSTDAG. After a candidate is rejected (invalid UTXO or finality violation), remove it from `tip_set` **before** considering parents. Genesis can then be added. Assert `tip_set` is non-empty so a real exhaust panics instead of looping.

No change to DK selected parent, coloring, or how UTXO validity is decided — only that a rejected sink is dropped from the candidate set.

---

## 6. Test templates use v2 sink search (`acc5f782`)

**Files:**
- `consensus/src/pipeline/virtual_processor/test_block_builder.rs`
- `testing/integration/src/consensus_integration_tests.rs`

After (5), nextest got past the hung body tests and failed two more.

### `selected_chain_test` — `KeyNotFound(PruningSamples/…15)`

`TestConsensus::add_utxo_valid_block_with_parents` builds the header via `TestBlockBuilder::build_block_template_with_parents`. That helper still called GHOSTDAG `sink_search_algorithm` (v1) to pick `pov_sink`, then `pick_virtual_parents` (DK coloring when the executor is present). `build_block_template_from_virtual_state` then calls `expected_header_pruning_point` on the **coloring** selected parent and unwraps `pruning_sample_from_pov(selected_parent)`.

v1 walked UTXO for the GHOSTDAG sink (22) and wrote that sample. DK coloring of parents 15–22 picked 15. 15 was never a sink, so it had no `PruningSamples` row. Panic at `pruning.rs:120` while inserting merge block 23.

Production `resolve_virtual` already uses `sink_search_algorithm_v2` when `dagknight_executor` is `Some`. The test helper did not.

**Fix:** Same branch as `resolve_virtual`. If the executor is present, v2 writes the pruning sample for the coloring selected parent. If DK is off (goref), keep v1.

This is the production-alignment bug. The remaining `selected_chain_test` asserts were GHOSTDAG-specific, not a second panic.

### Selected-chain asserts after the merge

The last scenario merges 8 genesis children into 23. GHOSTDAG treats 23 as higher blue work than the 8–14 chain and reorgs to genesis → 22 → 23 (tie-break picks 22). DAGKNIGHT ranks conflict groups among tips `{6, 7, 14, 23}` and can keep the original chain through 6, so 23 stays `StatusUTXOPendingVerification`.

The store must still match the virtual coloring chain. Asserts now branch on `config.enable_dagknight`:

- DK on (default) — sink is 6; chain is genesis → 1…6
- DK off — GHOSTDAG tie-break: genesis → 22 → 23

### `sighash_type_commitment_test` — last block `StatusUTXOPendingVerification`

Six independently signed txs, each in a genesis child. GHOSTDAG tie-break (highest hash) makes each new sibling the sink, so each insert returned `StatusUTXOValid`. DAGKNIGHT coloring does not select the sixth sibling, so UTXO never runs on that block. Status is pending — the tx is not invalid.

The test is about sighash acceptance, not sibling-of-genesis sink selection. The six blocks are now a chain so each new block is the coloring sink and is actually UTXO-validated.

No change to live `build_block_template` (that path already uses the resolved virtual, which went through v2). No change to coloring vs topology in the header processor.

---

## What this does not change

- Selected parent / conflict hierarchy
- k-colouring, committed vs free search (except restoring the documented committed LKT bound when NCA is absent)
- UMC cascade voting
- 100 BPS coloring / topology split in the live header processor (when `enable_dagknight` is on)
- GHOSTDAG fixture expectations (topology store)
- Live DAA / bits / `blue_score` (still coloring when DK is on)
- Which blocks are UTXO-valid vs disqualified (v2 now *finishes* the search the way v1 already did)
- Live mining templates (`build_block_template` already uses resolved virtual)

---

## Verification

- `cargo fmt --all -- --check`
- `cargo clippy --workspace --tests --benches --examples -- -D warnings` (after commit 1)
- `cargo clippy -p kaspa-consensus --tests --benches --examples -- -D warnings`
- `cargo clippy -p kaspa-consensus-core -p kaspa-consensus -p kaspa-testing-integration --tests --benches --examples -- -D warnings` (after commit 4)
- `cargo clippy -p kaspa-consensus --lib --tests -- -D warnings` (after commit 5)
- All 43 `processes::dagknight` unit tests, including `test_find_last_known_tips_uses_correct_ancestry_type` and both LKT correctness tests
- `consensus_integration_tests::ghostdag_test`
- Related header-path integration tests (`block_window`, `header_in_isolation`, `incest`, `missing_parents`, `known_invalid`, `bounded_merge_depth`, …)
- Goref JSON fixtures are not in this clone; `json_test` + `enable_dagknight = false` is the CI check for commit 4
- After commit 5 (isolated target): `validate_body_in_context_test` 0.07s; `basic_utxo_disqualified_test` + `double_search_disqualified_test` ok; `mass_per_sig_op_does_not_change_block_capacity` 0.42s; `template_mining_sanity_test` 0.72s
- After commit 6 (isolated target): `selected_chain_test` ok; `sighash_type_commitment_test` ok; `cargo clippy -p kaspa-consensus --all-targets -- -D warnings`

### DIFF
--- a/consensus/core/src/config/mod.rs
+++ b/consensus/core/src/config/mod.rs
@@ -74,6 +74,11 @@ pub struct Config {
 
     /// The number of days to keep data for
     pub retention_period_days: Option<f64>,
+
+    /// Enable DAGKNIGHT selected-parent / coloring. Off for GHOSTDAG-recorded
+    /// replay tests (goref JSON fixtures). Production on this branch keeps it on.
+    /// TODO[DK]: replace with ForkActivation when the activation height is defined.
+    pub enable_dagknight: bool,
 }
 
 impl Config {
@@ -103,6 +108,7 @@ impl Config {
             disable_upnp: false,
             ram_scale: 1.0,
             retention_period_days: None,
+            enable_dagknight: true,
         }
     }
 
@@ -177,6 +183,11 @@ impl ConfigBuilder {
 
     pub fn skip_adding_genesis(mut self) -> Self {
         self.config.process_genesis = false;
+        self
+    }
+
+    pub fn disable_dagknight(mut self) -> Self {
+        self.config.enable_dagknight = false;
         self
     }
 
--- a/consensus/core/src/config/mod.rs
+++ b/consensus/core/src/config/mod.rs
@@ -74,11 +74,6 @@ pub struct Config {
 
     /// The number of days to keep data for
     pub retention_period_days: Option<f64>,
-
-    /// Enable DAGKNIGHT selected-parent / coloring. Off for GHOSTDAG-recorded
-    /// replay tests (goref JSON fixtures). Production on this branch keeps it on.
-    /// TODO[DK]: replace with ForkActivation when the activation height is defined.
-    pub enable_dagknight: bool,
 }
 
 impl Config {
@@ -108,7 +103,6 @@ impl Config {
             disable_upnp: false,
             ram_scale: 1.0,
             retention_period_days: None,
-            enable_dagknight: true,
         }
     }
 
@@ -183,11 +177,6 @@ impl ConfigBuilder {
 
     pub fn skip_adding_genesis(mut self) -> Self {
         self.config.process_genesis = false;
-        self
-    }
-
-    pub fn disable_dagknight(mut self) -> Self {
-        self.config.enable_dagknight = false;
         self
     }
 
--- a/consensus/src/consensus/storage.rs
+++ b/consensus/src/consensus/storage.rs
@@ -270,8 +270,7 @@ impl ConsensusStorage {
 
         // TODO[DK]: Small cache policy for dagknight data; this can be tuned via perf params later
         let dagknight_builder = PolicyBuilder::new().bytes_budget(scaled(5_000_000)).tracked_bytes();
-        // TODO[DK]: Use a config or ForkActivation to gate this
-        let dagknight_store = Some(Arc::new(DbDagknightStore::new(db.clone(), dagknight_builder.build())));
+        let dagknight_store = config.enable_dagknight.then(|| Arc::new(DbDagknightStore::new(db.clone(), dagknight_builder.build())));
         let umc_persistence_store = Arc::new(DbUmcCascadeStore::new(db.clone(), CachePolicy::Count(256)));
 
         Arc::new(Self {
--- a/consensus/src/consensus/storage.rs
+++ b/consensus/src/consensus/storage.rs
@@ -270,7 +270,8 @@ impl ConsensusStorage {
 
         // TODO[DK]: Small cache policy for dagknight data; this can be tuned via perf params later
         let dagknight_builder = PolicyBuilder::new().bytes_budget(scaled(5_000_000)).tracked_bytes();
-        let dagknight_store = config.enable_dagknight.then(|| Arc::new(DbDagknightStore::new(db.clone(), dagknight_builder.build())));
+        // TODO[DK]: Use a config or ForkActivation to gate this
+        let dagknight_store = Some(Arc::new(DbDagknightStore::new(db.clone(), dagknight_builder.build())));
         let umc_persistence_store = Arc::new(DbUmcCascadeStore::new(db.clone(), CachePolicy::Count(256)));
 
         Arc::new(Self {
--- a/consensus/src/consensus/test_consensus.rs
+++ b/consensus/src/consensus/test_consensus.rs
@@ -136,13 +136,23 @@ impl TestConsensus {
         let mut header = header_from_precomputed_hash(hash, parents.clone());
         let parents_by_level = self.consensus.services.parents_manager.calc_block_parents(self.pruning_point(), &parents);
         header.parents_by_level = parents_by_level;
-        let ghostdag_data = self.consensus.services.ghostdag_manager.ghostdag(header.direct_parents());
-        let daa_window = self.consensus.services.window_manager.block_daa_window(&ghostdag_data).unwrap();
-        header.bits = self.consensus.services.window_manager.calculate_difficulty_bits(&ghostdag_data, &daa_window);
+        let direct_parents = header.direct_parents();
+        // Match header-processor validation: blue_score / DAA / bits / PMT come from
+        // coloring (DK selected parent when the executor is present); blue_work from topology.
+        let topology_ghostdag_data = self.consensus.services.ghostdag_manager.ghostdag(direct_parents);
+        let coloring_ghostdag_data = match &self.consensus.services.dagknight_executor {
+            Some(executor) => {
+                let dk_sp = executor.dagknight(direct_parents).selected_parent;
+                self.consensus.services.coloring_ghostdag_manager.incremental_coloring(direct_parents, dk_sp)
+            }
+            None => self.consensus.services.coloring_ghostdag_manager.ghostdag(direct_parents),
+        };
+        let daa_window = self.consensus.services.window_manager.block_daa_window(&coloring_ghostdag_data).unwrap();
+        header.bits = self.consensus.services.window_manager.calculate_difficulty_bits(&coloring_ghostdag_data, &daa_window);
         header.daa_score = daa_window.daa_score;
-        header.timestamp = self.consensus.services.window_manager.calc_past_median_time(&ghostdag_data).unwrap().0 + 1;
-        header.blue_score = ghostdag_data.blue_score;
-        header.blue_work = ghostdag_data.blue_work;
+        header.timestamp = self.consensus.services.window_manager.calc_past_median_time(&coloring_ghostdag_data).unwrap().0 + 1;
+        header.blue_score = coloring_ghostdag_data.blue_score;
+        header.blue_work = topology_ghostdag_data.blue_work;
 
         header
     }
--- a/consensus/src/pipeline/virtual_processor/processor.rs
+++ b/consensus/src/pipeline/virtual_processor/processor.rs
@@ -53,7 +53,7 @@ use kaspa_consensus_core::{
     acceptance_data::AcceptanceData,
     api::args::{TransactionValidationArgs, TransactionValidationBatchArgs},
     block::{BlockTemplate, MutableBlock, TemplateBuildMode, TemplateTransactionSelector},
-    blockstatus::BlockStatus::{StatusDisqualifiedFromChain, StatusUTXOPendingVerification, StatusUTXOValid},
+    blockstatus::BlockStatus::{StatusDisqualifiedFromChain, StatusUTXOValid},
     coinbase::MinerData,
     config::genesis::GenesisBlock,
     header::Header,
@@ -1083,64 +1083,6 @@ impl VirtualStateProcessor {
         }
     }
 
-    /// UTXO-validate pending tips so later sink search can walk known-invalid
-    /// candidates without calling DAGKNIGHT. Does not change the eventual sink:
-    /// a later `calculate_utxo_state_relatively` from the last `diff_point`
-    /// reorgs onto whichever candidate DAGKNIGHT actually picks.
-    fn screen_pending_sink_candidates(
-        &self,
-        stores: &VirtualStores,
-        diff: &mut UtxoDiff,
-        mut diff_point: Hash,
-        tip_set: &BlockHashSet,
-        finality_point: Hash,
-    ) -> Hash {
-        let pending = {
-            let statuses = self.statuses_store.read();
-            tip_set.iter().copied().filter(|&h| statuses.get(h).unwrap() == StatusUTXOPendingVerification).collect_vec()
-        };
-        for candidate in pending {
-            if self.reachability_service.is_chain_ancestor_of(finality_point, candidate) {
-                diff_point = self.calculate_utxo_state_relatively(stores, diff, diff_point, candidate);
-            }
-        }
-        diff_point
-    }
-
-    fn tip_set_is_all_disqualified(&self, tip_set: &BlockHashSet) -> bool {
-        let statuses = self.statuses_store.read();
-        tip_set.iter().all(|&h| statuses.get(h).unwrap() == StatusDisqualifiedFromChain)
-    }
-
-    /// Replace already-disqualified tips with their parents. GHOSTDAG sink search
-    /// does this with cheap heap pops. Only call when every current tip is
-    /// already `StatusDisqualifiedFromChain` — walking anticone disqualified
-    /// tips while a UTXO-valid tip remains would replace mergeable invalid
-    /// tips with their ancestors and change virtual-parent selection.
-    fn expand_disqualified_sink_candidates(&self, tip_set: &mut BlockHashSet, finality_point: Hash) {
-        loop {
-            let disqualified = {
-                let statuses = self.statuses_store.read();
-                tip_set.iter().copied().filter(|&h| statuses.get(h).unwrap() == StatusDisqualifiedFromChain).collect_vec()
-            };
-            if disqualified.is_empty() {
-                return;
-            }
-            let prune_guard = self.pruning_lock.blocking_read();
-            for candidate in disqualified {
-                tip_set.remove(&candidate);
-                for parent in self.relations_service.get_parents(candidate).unwrap().iter().copied() {
-                    if self.reachability_service.is_dag_ancestor_of(finality_point, parent)
-                        && !self.reachability_service.is_dag_ancestor_of_any(parent, &mut tip_set.iter().copied())
-                    {
-                        tip_set.insert(parent);
-                    }
-                }
-            }
-            drop(prune_guard);
-        }
-    }
-
     /// Searches for the next valid sink block (SINK = Virtual selected parent). The search is performed
     /// in the inclusive past of `tips`.
     /// The provided `diff` is assumed to initially hold the UTXO diff of `prev_sink` from virtual.
@@ -1169,19 +1111,6 @@ impl VirtualStateProcessor {
         // since we check that every pushed block is not in the past of current heap
         // (and it can't be in the future by induction)
         loop {
-            // Screen pending tips first. A concurrent pipeline of UTXO-invalid
-            // blocks otherwise calls `dagknight()` once per rejected candidate
-            // (GHOSTDAG only needs a heap pop). After screening, an all-
-            // disqualified set can be walked to a valid ancestor without DK.
-            // Do not walk disqualified anticone tips while a valid tip remains:
-            // those tips must stay in the DK input so they can be merged.
-            diff_point = self.screen_pending_sink_candidates(stores, diff, diff_point, &tip_set, finality_point);
-            if self.tip_set_is_all_disqualified(&tip_set) {
-                self.expand_disqualified_sink_candidates(&mut tip_set, finality_point);
-                assert!(!tip_set.is_empty(), "sink search exhausted all candidates without a UTXO-valid sink");
-                continue;
-            }
-            assert!(!tip_set.is_empty(), "sink search exhausted all candidates without a UTXO-valid sink");
             let mut sub_tips = tip_set.clone();
 
             let (candidate, parents) = loop {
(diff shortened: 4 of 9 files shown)


LiveLaughLove13:
@coderofstuff @biryukovmaxim @michaelsutton

LiveLaughLove13:
Improved the build time. @michaelsutton , and @coderofstuff. Will wait for you all to review and check it out.

consensus/src/pipeline/virtual_processor/processor.rs
```diff
@@ -1172,6 +1172,11 @@ impl VirtualStateProcessor {
                 // `finality_point == pruning_point` indicates we are at IBD start hence no warning required
                 warn!("Finality Violation Detected. Block {} violates finality and is ignored from Virtual chain.", candidate);
             }
+            // Same as GHOSTDAG sink search popping the heap: drop a rejected candidate
+            // before considering its parents. Leaving it in tip_set makes every parent
+            // look like a DAG ancestor of the current set, so genesis is never added
+            // and the loop retries the same UTXO-invalid tip forever.
+            tip_set.remove(&candidate);
```
D-Stacks: yes, this is a good catch, imo the selected parent needs to be removed here.

consensus/src/pipeline/virtual_processor/processor.rs
```diff
@@ -1083,6 +1083,64 @@ impl VirtualStateProcessor {
...
+        let pending = {
+            let statuses = self.statuses_store.read();
+            tip_set.iter().copied().filter(|&h| statuses.get(h).unwrap() == StatusUTXOPendingVerification).collect_vec()
+        };
+        for candidate in pending {
+            if self.reachability_service.is_chain_ancestor_of(finality_point, candidate) {
+                diff_point = self.calculate_utxo_state_relatively(stores, diff, diff_point, candidate);
+            }
+        }
+        diff_point
```
D-Stacks: I don't fully understand this, why do we change the diff-point from the prev-sink to this where, we return and continue working with a diff_point from some, effectively random tip, that is pending verification?

not saying this isn't possibly a good idea, but I can't follow the reasoning here.

consensus/src/pipeline/virtual_processor/processor.rs
```diff
@@ -1111,6 +1169,19 @@ impl VirtualStateProcessor {
...
         // (and it can't be in the future by induction)
         loop {
+            // Screen pending tips first. A concurrent pipeline of UTXO-invalid
+            // blocks otherwise calls `dagknight()` once per rejected candidate
+            // (GHOSTDAG only needs a heap pop). After screening, an all-
+            // disqualified set can be walked to a valid ancestor without DK.
+            // Do not walk disqualified anticone tips while a valid tip remains:
+            // those tips must stay in the DK input so they can be merged.
+            diff_point = self.screen_pending_sink_candidates(stores, diff, diff_point, &tip_set, finality_point);
+            if self.tip_set_is_all_disqualified(&tip_set) {
```
D-Stacks: I take some issue here, so disqualified from chain still means we can merge the block, just it cannot be our selected parent / sink, or part of the vspc. so I think expansion here should be pinpointed only to the selected parent as returned by dagknight, which is exactly what is done at the bottom of this loop anyway. furthermore, we leave these tips to the next pass in resolve virtual, so we continuously have to deal with them here.

D-Stacks: That being said There may be a solution whereby we may simply ignore merging disqualified from chain blocks, in such an environment, we can pre-process and expand all tips that are disqualified, and dagknight should be able to run in one pass. But this needs some consideration if this has any draw-backs, as this would keep these tips "dangling".

D-Stacks: I believe if we go for such a solution much of the logic in this PR can be re-used.

consensus/src/pipeline/virtual_processor/processor.rs
LiveLaughLove13: Screening is not sink selection, and diff_point is never left on a pending or failed tip.
calculate_utxo_state_relatively(from, to) always returns the topmost UTXO-valid block on chain(to) (or from if to is already StatusDisqualifiedFromChain). On failure it marks to disqualified and keeps diff aligned with that last valid ancestor. HashSet order only decides which pending tip we probe first; it does not become the sink.

The point of the probe is only to write StatusDisqualifiedFromChain so the next step can walk an all-invalid antichain without calling dagknight() once per tip (that is what made test_concurrent_pipeline_random sit past 10 minutes). After DAGKNIGHT actually picks a sink we call calculate_utxo_state_relatively again from the current diff_point onto that candidate, so virtual UTXO is reorged onto the real sink either way.

One real caveat: if a pending tip is actually UTXO-valid, this probe commits its UTXO/pruning-sample row before DK chooses, even if that tip is not the eventual sink. The later relative walk still reorgs virtual onto the DK sink. In the workload this was written for, the pending tips are all invalid, so diff_point stays on a prior valid ancestor (usually prev_sink / genesis). If you would rather not commit a non-sink valid tip early, we can stop screening once a StatusUTXOValid tip is already in the set, or only use the status side-effect on failure.

consensus/src/pipeline/virtual_processor/processor.rs
LiveLaughLove13: "I take some issue here, so disqualified from chain still means we can merge the block, just it cannot be our selected parent / sink, or part of the vspc. so I think expansion here should be pinpointed only to the selected parent as returned by dagknight, which is exactly what is done at the bottom of this loop anyway. furthermore, we leave these tips to the next pass in resolve virtual, so we continuously have to deal with them here."

StatusDisqualifiedFromChain only means the block cannot be sink / VSPC. It can still be a virtual parent. That is why this expand is not used when any non-disqualified tip remains.

If the set is [valid_sink, dq1, dq2], we leave dq1/dq2 in the DAGKNIGHT input so they can be merged. That is double_search_disqualified_test. Expanding them while a valid tip is present replaces mergeable tips with their ancestors (root_1 / root_2); later honest blocks parent those ancestors and the disqualified tips never get merged. We hit that when we tried it.

The expand runs only when every current candidate is already known disqualified. Then none of them can be sink, so running DAGKNIGHT only to decide which invalid tip to pop is wasted work. Batch-walking them to a UTXO-valid ancestor is the same work GHOSTDAG does with heap pops: after v1 pops every high-work invalid tip, those tips are gone from the heap and are not returned as merge candidates either. The bottom-of-loop path (remove the DK selected parent, insert its parents) is unchanged for the mixed case, which is the one that must keep merging disqualified tips.

On the next resolve_virtual, body_tips still includes those invalid tips until something actually parents them. That is the same as v1. We do not keep paying a full dagknight() per historical invalid tip: they are already disqualified, so we walk and DK only the remaining valid ancestor.

"That being said There may be a solution whereby we may simply ignore merging disqualified from chain blocks, in such an environment, we can pre-process and expand all tips that are disqualified, and dagknight should be able to run in one pass. But this needs some consideration if this has any draw-backs, as this would keep these tips "dangling"."

If we later decide virtual should not merge disqualified tips, we can expand them all first and run DAGKNIGHT once. The dangling-tips cost is real, and doing that while a valid sink is still in the set breaks merge (double_search_disqualified_test). This PR does not take that path.

consensus/core/src/config/mod.rs
```diff
@@ -74,6 +74,11 @@ pub struct Config {
 
     /// The number of days to keep data for
     pub retention_period_days: Option<f64>,
+
+    /// Enable DAGKNIGHT selected-parent / coloring. Off for GHOSTDAG-recorded
+    /// replay tests (goref JSON fixtures). Production on this branch keeps it on.
+    /// TODO[DK]: replace with ForkActivation when the activation height is defined.
+    pub enable_dagknight: bool,
```
coderofstuff: I have a separate WIP that tackles this altogether, fyi. Will push that shortly, but that would remove the need to set this here.

consensus/src/consensus/storage.rs
```diff
@@ -270,8 +270,7 @@ impl ConsensusStorage {
 
         // TODO[DK]: Small cache policy for dagknight data; this can be tuned via perf params later
         let dagknight_builder = PolicyBuilder::new().bytes_budget(scaled(5_000_000)).tracked_bytes();
-        // TODO[DK]: Use a config or ForkActivation to gate this
```
coderofstuff, line 273: In line with the comment https://github.com/kaspanet/rusty-kaspa/pull/1105/changes#r3927224284

please revert this for now. I will wire up the ForkActivation accordingly.

consensus/src/consensus/test_consensus.rs
```diff
@@ -136,13 +136,23 @@ impl TestConsensus {
         let mut header = header_from_precomputed_hash(hash, parents.clone());
         let parents_by_level = self.consensus.services.parents_manager.calc_block_parents(self.pruning_point(), &parents);
         header.parents_by_level = parents_by_level;
-        let ghostdag_data = self.consensus.services.ghostdag_manager.ghostdag(header.direct_parents());
-        let daa_window = self.consensus.services.window_manager.block_daa_window(&ghostdag_data).unwrap();
-        header.bits = self.consensus.services.window_manager.calculate_difficulty_bits(&ghostdag_data, &daa_window);
+        let direct_parents = header.direct_parents();
```
coderofstuff, line 139: This looks good, thanks

consensus/src/processes/dagknight/manager.rs
coderofstuff, line 1: The changes here will conflict with the ones getting added by @biryukovmaxim in #1104 - to reduce the workload in this area, please revert all changes in this file.

testing/integration/src/consensus_integration_tests.rs
coderofstuff, line 1: The selected chain test in this file test specifically behavior for GHOSTDAG. Indeed for now it is broken.

However, we need to keep the test as is, and then when the ForkActivation PR is added in, we can add logic to test specifically GD behavior (which is existent here) and logic to test specifically DK behavior.

Please revert the changes in this file.

consensus/src/processes/ghostdag/ordering.rs
```diff
@@ -43,7 +43,7 @@ impl Ord for SortableBlock {
 
 impl<T: GhostdagStoreReader, S: RelationsStoreReader, U: ReachabilityService, V: HeaderStoreReader> GhostdagManager<T, S, U, V> {
     pub fn sort_blocks(&self, blocks: impl IntoIterator<Item = Hash>) -> Vec<Hash> {
-        let ghostdag_store = self.custom_topology_store.as_ref().unwrap_or_else(|| &self.ghostdag_store);
+        let ghostdag_store = self.custom_topology_store.as_ref().unwrap_or(&self.ghostdag_store);
```
coderofstuff, line 46: Looks fine

coderofstuff (changes requested):
To minimize duplicated work, let's focus this PR on the changes in:
- consensus/src/consensus/test_consensus.rs
- consensus/src/pipeline/virtual_processor/test_block_builders.rs
- consensus/src/processes/ghostdag/ordering.rs

As to the `sink_search_algorithm_v2` change, I think this deserves it's own PR but please make sure to coordinate with @D-Stacks to make sure you're aligned with the full intent of this area. It is highly nuanced.

As to the areas regarding fork activation/dagknight enablement, please let it as is for now as I will be putting up the PR very soon that properly wires a ForkActivation.

LiveLaughLove13:
**kept**
test_consensus.rs : coloring vs topology header builder 
test_block_builder.rs : v2 sink search in the test template helper
ordering.rs : unwrap_or clippy


**Reverted to origin/dagknight as instructed**

config/mod.rs, storage.rs, services.rs : leave enablement alone as instructed
processor.rs : sink_search_algorithm_v2 is a separate PR, to align with @D-Stacks
manager.rs : all of it, including clippy and the LKT bound (conflicts with #1104)
consensus_integration_tests.rs : entire file; keep GHOSTDAG selected_chain_test as-is even if it fails until ForkActivation
Against kaspanet dagknight the remaining diff is 3 files, +33 / −10. ghostdag_test still passes.


Sink-search (be0a2f2a / fece0f86) stays in history for a follow-up PR with @D-Stacks  — it is not in this PR anymore.

coderofstuff (approved):

consensus/src/pipeline/virtual_processor/processor.rs
coderofstuff: my bad, this change is indeed a good find and should be included in this PR

LiveLaughLove13: Readded.

coderofstuff (approved):

PR #1119, opened 2026-09-03
tmp: dk manager nits

IzioDev:


consensus/src/processes/dagknight/manager.rs
```diff
@@ -392,21 +392,36 @@ impl<C: DagknightStore + DagknightStoreReader, O: HeaderStoreReader, D: Relation
...
+    /// The traversal walks backward through blocks without stored DK colouring data for this (`root`, `k`, search mode)
+    /// context and stops each explored parent path at the first block for which such data exists.
+    ///
+    /// In free-search mode, `next_chain_ancestor` must be `None`. In committed-search mode, it must be `Some(nca)`, and the
+    /// traversal is restricted to the inclusive chain future of `nca`. The conflict genesis is exempt because it is the last
+    /// possible known boundary of the backward traversal.
+    ///
+    /// Returns the known boundary blocks and all visited blocks. The latter includes blocks rejected by committed-search
+    /// filtering.
+    pub fn find_known_ancestor_boundary_blocks(&self, tips: &[Hash], next_chain_ancestor: Option<Hash>) -> (Vec<Hash>, BlockHashSet) {
```
IzioDev, line 406: suggestion: use something like
```rust
enum SearchMode {
  Committed(Hash),
  FreeSearch,
}
```
because caller usually already assert this shape

consensus/src/processes/dagknight/manager.rs
```diff
@@ -830,18 +844,18 @@ mod tests {
         // Tips are F and W (no records yet)
         let tips = vec![hash_f, hash_w];
 
-        let (roots_committed, _) = manager_committed.find_last_known_tips(&tips, None);
-        let (roots_free, _) = manager_free.find_last_known_tips(&tips, None);
+        let (boundary_blocks_committed, _) = manager_committed.find_known_ancestor_boundary_blocks(&tips, None);
```
IzioDev, line 847: to fix: committed should pass a `NCA` but current test DAG is miss-shaped for this.

consensus/src/processes/dagknight/manager.rs
```diff
@@ -457,13 +472,14 @@ impl<C: DagknightStore + DagknightStoreReader, O: HeaderStoreReader, D: Relation
         let lock_arc = locks.entry(lock_key).or_insert_with(|| Arc::new(RwLock::new(()))).clone();
         let _guard = lock_arc.write();
 
+        // populate dummy root to DKStore
         self.init_root();
```
IzioDev, line 476: i think it should be renamed to `ensure_root` if db architecture kept as is

consensus/src/processes/dagknight/manager.rs
coderofstuff: Max's PR #1104 has these semantics in place. I'd defer to that PR for this kind of change. I also invite you to review that.

consensus/src/processes/dagknight/manager.rs
coderofstuff, line 781: This `hash_f` and `hash_w` actually do have selected parents (add_block will add them automatically), but we should set them to:

```
hash_f.sp = hash_d
hash_w.sp = hash_e
```

iirc

consensus/src/processes/dagknight/manager.rs
coderofstuff: `find_last_known_tips` / `find_known_ancestor_boundary_blocks` used to actually find all the blocks that are in the frontier, since it used to do the coloring for all tips at once. However, that changed recently but this test didn't.

To fix the test:
1. For each NCA (I think it's `hash_b` and `hash_c`), expect that the last known tips are `hash_d` and `hash_e` respectively.
2. So there would be 2 runs of committed find_lkt, and each would expect the set for them. In both NCAs, `hash_x` is NEVER returned as a boundary block / lkt.

coderofstuff (changes requested):

consensus/src/processes/dagknight/manager.rs
IzioDev: the proposed design by Maxim is perfectly suited

PR #1121, opened 2026-09-05
[DK][SSAV2] Make SSAV2 resolve disqualified-from-chain SP to qualified representatives in one pass. 

D-Stacks:
Currently when we try to resolve a disqualified from chain selected parent we only remove it and add its parents to the tip-set. This PR adds full resolution to a qualified set of blocks in the Tip's ancestry (if there are such blocks) in one pass. It is worth noting that current handling is especially ill-fitted due to the fact that the disqualified from chain status has the property of upwards propagation, i.e. if a block in the selected parent chain of some block is disqualified, it will propagate upwards. As such there can be a valid expectation that the next qualified blocks may be well below the selected parent that dagknight chose. Not resolving down in one pass can cause repeated unneeded calls to dagknight as well as pick virtual parents, until resolution happens.

### DIFF
--- a/consensus/src/pipeline/virtual_processor/processor.rs
+++ b/consensus/src/pipeline/virtual_processor/processor.rs
@@ -1196,6 +1196,8 @@ impl VirtualStateProcessor {
                     seen.insert(parent);
                     diff_point = self.calculate_utxo_state_relatively(stores, diff, diff_point, parent);
                     if diff_point == parent {
+                        // remove any tips that are DAG descendants of the current parent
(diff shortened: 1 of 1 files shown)


D-Stacks:
Still need to add logic to remove a parent added in an earlier iteration which  became a DAG ancestor of a parent added in a later iteration, violating the principle that tips shouldn't be ancestors of each other.

**Edit:** done

D-Stacks:
converted to draft pending #1123

PR #1122, opened 2026-09-05
[DK][SSAV2] pass the `selected_parent` into `remove_bounded_merge_breaking_parents`

D-Stacks:
This PR passes the `selected_parent` explicitly into `remove_bounded_merge_breaking_parents`,  Since we know the SP beforehand this avoids unnecessary re-runs of dagknight in `remove_bounded_merge_breaking_parents` to re-establish the SP.

### DIFF
--- a/consensus/src/pipeline/virtual_processor/processor.rs
+++ b/consensus/src/pipeline/virtual_processor/processor.rs
@@ -1262,7 +1262,7 @@ impl VirtualStateProcessor {
         }
         assert!(mergeset_size <= mergeset_size_limit);
         assert!(virtual_parents.len() <= max_block_parents);
-        self.remove_bounded_merge_breaking_parents(virtual_parents, pruning_point)
+        self.remove_bounded_merge_breaking_parents(virtual_parents[0], virtual_parents, pruning_point)
     }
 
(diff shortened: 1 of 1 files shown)


consensus/src/pipeline/virtual_processor/processor.rs
```diff
@@ -1298,13 +1298,13 @@ impl VirtualStateProcessor {
 
     fn remove_bounded_merge_breaking_parents(
         &self,
+        selected_parent: Hash,
```
coderofstuff, line 1301: Since the signature of this function is changing, what do you think about simply creating `pick_virtual_parents_v2` and `remove_bounded_merge_breaking_parents` so we can more-freely change the behavior here without worrying about backward compatibility with what GD needs/expects?

D-Stacks: Yes, I think this is worth-while for separation of concerns regarding pre- and post dk handling. I created a separate PR for it, here:

#1123 

it also changes some handling of the testnet wiring. 

After that is merged I will update my other PRs to it.

D-Stacks:
converted to draft pending #1123

PR #1123, opened 2026-09-07
[DK][SSAV2] V2 all inner calls of ssav2

D-Stacks:
Just a small PR to lay the groundwork to separate pre- and post-DK handling concerns in resolve virtual. 

This splits `pick_virtual_parents`, `mergeset_increase` and `remove_bounded_merge_breaking_parents` calls into `v2` copies, which are intended to be used when the dagknight HF is active. This allows us to more easily separate and implement functionality pertaining to a pre- and post-DK HF independently.

The Pr is also merged with testnet wiring, it will assert the right function is used depending of if, or if not, dk is active. it also already splits the ghostdag_coloring used in `remove_bounded_merge_breaking_parents` dependent on if it is in the v1 or v2 call.


consensus/src/pipeline/virtual_processor/processor.rs
```diff
@@ -1273,13 +1298,95 @@ impl VirtualStateProcessor {
...
+    ///     1. `selected_parent` is a UTXO-valid block
+    ///     2. `candidates` are an antichain ordered in descending blue work order
+    ///     3. `candidates` do not contain `selected_parent` and `selected_parent.blue work > max(candidates.blue_work)`  
+    /// TODO [post-DK cleanp-up] consider removing the `_v2` suffix.
+    pub(super) fn pick_virtual_parents_v2(
+        &self,
+        selected_parent: Hash,
+        mut candidates: VecDeque<Hash>,
+        pruning_point: Hash,
+        dk_active: bool,
```
coderofstuff: All `_v2` assume you're running for DK, I don't think we need to assert it. So there's no need to pass this around. Instead, the caller should call the appropriate function depending on whether DK is active or not.

coderofstuff (changes requested):
Looks good, just a comment for the expected convention in this area

consensus/src/pipeline/virtual_processor/processor.rs
D-Stacks: done

PR #1126, opened 2026-09-08
Dk UMC depth restriction

freshair18:
This PR updates the UMC algorithm, with a depth restriction: the new algorithm, automatically rejects the coloring, if a deep enough blue block has a negative vote, and skips processing onwards  to check if the actual conflict genesis is negative or positive.

The basic intuition is that if a block flips deep enough, then flipping it can't be that much easier than flipping genesis, so this event should be rare in the grand scheme of things and not actually effect the security of the protocol (even though we sometimes return "false negatives")

It is clarified that this is a consensus change to the vanilla Dagknight protocol as described in the paper. In the coming days I will write a more rigorous argument to why this altered protocol remains asymtotically as secure as vanilla Dagknight.

The bound is k^4 blue blocks which is O(K(k)^2) if we follow paper notation.  This allows us to bound the worst case of an incremental step of the algorithm as K(k)^3*log(n/(K(k))), solving the potentially unbounded flipping present in the previous version of the incremental algorithm. 

We can and likely should batch the events in the events queue in subsequent versions, which will provide a de facto speedup, but will not be meaningful in worst case asymptotics.

This is a preliminary version representing a POC for the "bounded DK" implementation. There are possibly some mild issues or inaccuracies worth pondering on.

@michaelsutton @coderofstuff


freshair18:
Accidentally opened for master rathen than DK

PR #1127, opened 2026-09-08
Dk UMC depth restriction

freshair18:
This PR updates the UMC algorithm, with a depth restriction: the new algorithm, automatically rejects the coloring, if a deep enough blue block has a negative vote, and skips processing onwards  to check if the actual conflict genesis is negative or positive.

The basic intuition is that if a block flips deep enough, then flipping it can't be that much easier than flipping genesis, so this event should be rare in the grand scheme of things and not actually effect the security of the protocol (even though we sometimes return "false negatives")

It is clarified that this is a consensus change to the vanilla Dagknight protocol as described in the paper. In the coming days I will write a more rigorous argument to why this altered protocol remains asymtotically as secure as vanilla Dagknight.

The bound is currently k^4 blue blocks in the past of merger which is O(K(k)^2) if we follow paper notation.  This allows us to bound the worst case of an incremental step of the algorithm as K(k)^3*log(n/(K(k))), solving the potentially unbounded flipping present in the previous version of the incremental algorithm. 

We can and likely should batch the events in the events queue in subsequent versions, which will provide a de facto speedup, but will not be meaningful in worst case asymptotics.

This is a preliminary version representing a POC for the "bounded DK" implementation. There are possibly some mild issues or inaccuracies worth pondering on.

@michaelsutton @coderofstuff

### DIFF
--- a/consensus/src/processes/dagknight/protocol.rs
+++ b/consensus/src/processes/dagknight/protocol.rs
@@ -371,7 +371,7 @@ impl<
             let baseline_result =
                 self.baseline_umc_cascade_voting(conflict_genesis, subgroup, virtual_gd.clone(), k_to_check, &conflict_zone_manager);
 
-            if baseline_result.virtual_score != cascade_result.virtual_score {
+            if baseline_result.cascade_score != cascade_result.cascade_score {
                 if baseline_result.accepted != cascade_result.accepted {
                     self.counters.record_baseline_disagreement(baseline_result.accepted, cascade_result.accepted);
                 }
@@ -381,8 +381,8 @@ impl<
                      cascade_score={}, baseline_accepted={}, cascade_accepted={}, flips={}, voting_blocks={}",
                     k_to_check,
                     conflict_genesis,
-                    baseline_result.virtual_score,
-                    cascade_result.virtual_score,
+                    baseline_result.cascade_score,
+                    cascade_result.cascade_score,
                     baseline_result.accepted,
                     cascade_result.accepted,
                     cascade_result.flips,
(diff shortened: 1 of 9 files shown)


consensus/src/processes/dagknight/protocol.rs
```diff
@@ -363,33 +363,69 @@ impl<
...
-                    baseline_result.virtual_score,
-                    cascade_result.virtual_score,
-                    baseline_result.accepted,
-                    cascade_result.accepted,
-                    cascade_result.flips,
-                    cascade_result.voting_blocks
-                );
-            }
-        }
+        // Baseline comparison is temporarily disabled while cascade early-stop and
```
coderofstuff: Instead of commenting this out, please add counters (for base=False, cascade=True and base=True, cascade=False case counters) in DagknightCounters counting the disagreement count. If we can establish towards what direction there can be possible disagreement in result (such as baseline=True, cascade=False being possible but baseline=False, cascade=True being impossible) that'd be great.

In simpa, add logging for the counters.

freshair18:
Code and logic was polished, including the aforementioned batching, superior depth limit search, and a logical ordering of events processing, alongside an explanation and rough runtime analysis.

A formal document analyzing the security of this change remains pending still, hopefully this week,

A continuation PR should handle how to efficiently compute UMC from any viewpoint, not just that of virtual: two options are being considered currently:

(a) an efficiently placed and regularly pruned checkpoints storing, probably coupled with diffs.

(b) create a versioned variant of the data structure, where each access to a "cell" in the DS acts as a redirection to the latest time in the past of current a block where this cell was updated a la the SMT implementation (@biryukovmaxim )

